PR #8433, opened 2017-02-06
missing streamOpen() in Magento\Framework\Filesystem\Io\File

elioermini:
I think would be useful to have like in M1 the method `streamOpen()`.
There is no other way to initialise the field `$_streamHandler` of the class File.
Other methods in the class File rely on the initialisation of `$_streamHandler` and having it null I suppose that leaves the class just possible to use only extending it.


vrann:
@elioermini thank you for contribution.
The class \Magento\Framework\Filesystem\Io\File is not used anywhere in Magento and effectively deprecated in favor of \Magento\Framework\Filesystem\File\ReadInterface. We have the story in the backlog to remove \Magento\Framework\Filesystem\Io framework altogether.

elioermini:
Hi @vrann thanks for the clarification. Looks like \Magento\Framework\Filesystem\File\WriteInterface is what I need. Kind regards

vrann:
@elioermini great thanks for taking time and contributing your code.

PR #8445, opened 2017-02-07
Changed name of dispatched object in delete event

clementbeudot:
in catalog_category_delete_after_done event dispatch in delete method (:1020)
changed dispatched object name from "product" to "category"

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/Category.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/Category.php
@@ -1022,7 +1022,7 @@ class Category extends AbstractResource
         $this->getEntityManager()->delete($object);
         $this->_eventManager->dispatch(
             'catalog_category_delete_after_done',
-            ['category' => $object]
+            ['product' => $object, 'category' => $object] 
         );
         return $this;
     }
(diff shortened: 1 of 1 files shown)


ishakhsuvarov:
@clementbeudot Hi! Thank you for your contribution.
Unfortunately we can not accept this kind of change, since some extension or customization potentially may rely on that naming convention. In this case such rename may cause unexpected behavior or errors.

clementbeudot:
Ok thank you. I know that it could potentially break something. But it is really strange to obtain a "product" when expecting a "category".

Maybe something to change later.

ishakhsuvarov:
@clementbeudot 
Thats true, however too risky for now. Thanks for understanding.

fooman:
Maybe it could add the product additionally as that would not break existing and add a deprecation notice on the category?

clementbeudot:
hey @fooman, good idea as it would not break anything and people who are trying to use this event won't be stuck finding out why there is no "category" object passed in the event.

ishakhsuvarov (approved):

vrann:
@clementbeudot  Thank you for the contribution!

PR #8446, opened 2017-02-07
Fix "each()" function call on potentially invalid data

giacmir:
In some cases the `$attributesData` variable can be `false` at this point of execution. If this happens `count($attributesData)` returns 1 and then the code inside the `if` is evaluated. But then the call to `each` fails with error `Variable passed to each() is not an array or object in /var/www/html/vendor/magento/module-catalog/Model/ResourceModel/AbstractResource.php`

It is necessary to check that `$attributesData` is actually an array before attempt to `count` its content.

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
@@ -563,7 +563,7 @@ abstract class AbstractResource extends \Magento\Eav\Model\Entity\AbstractEntity
             }
         }
 
-        if (sizeof($attributesData) == 1) {
+        if (is_array($attributesData) && sizeof($attributesData) == 1) {
             $_data = each($attributesData);
             $attributesData = $_data[1];
         }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8446) <br/>All committers have signed the CLA.

ishakhsuvarov:
@giacmir Hi, could you please provide an example where `$attributesData` is going to be false?
Also, did you consider checking this way: `$attributesData !== false`?
Thanks!

giacmir:
@ishakhsuvarov It happens on an edge case if you call `getAttributeRawValue` with a nonexistent product id and a non-typed attribute (such as 'sku'). This is due to the fact that `fetchRow` ultimately calls `PDOStatement::fetch` that can return false when no results are found. Then, since the 'sku' is a non-typed attribute the following `if` statement is not executed and the `$attributesData` is `false` when it is checked with 'sizeof'.

`$attributesData !== false` could also work, and is probably faster, but since the following code uses `each` that is an array specific function checking with `is_array` seems safer to me.

ishakhsuvarov:
@giacmir 
Thank you. Sounds really reasonable to me.
Could you please sign CLA so I can continue with the merge?
Thank you.

giacmir:
@ishakhsuvarov I tried, and it says "Thank you for signing the CLA" and sends me back to github but nothing happens..

ishakhsuvarov:
@giacmir Could you please check, that name and e-mail on the commit is the same as in Github profile? Currently Github shows a "question mark" near the name on your [commit](https://github.com/magento/magento2/pull/8446/commits/f0f61b74aef9281bd0ff65d1064b9969dc872341).

ishakhsuvarov:
@giacmir Thank you

okorshenko:
@giacmir that you for this PR. Your PR merged to develop branch

PR #8454, opened 2017-02-07
Fix outlining invoice item price with the header label 'price'

joost-florijn-kega:
Fix the outlining of the invoice item price with the header label 'price' by having the same feed distance as used in the header.

In the header the label 'price' has a feed distance of 360:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Sales/Model/Order/Pdf/Invoice.php#L99

This fixes issue https://github.com/magento/magento2/issues/8453


okorshenko:
@joost-florijn-kega thank you for your contribution. Unfortunately, we are not accepting PRs to 2.1 branch for now. Delivery procedure to patch releases is more complicated. We are working on establishing this process for 2.1 and 2.0 branches. Looks like this issue exists in develop branch as well. If you would like to fix this issue in develop branch, please, create new pull request to develop. 
Closing this PR for now. We will consider backporting of this fix to 2.1.x release once it will be fixed in develop branch.

PR #8462, opened 2017-02-08
Fix product option files not copying to order dir.

evktalo:
Copy product option file to order directory from quote also when media storage is set to file system, not just database.

Also rename _rootDirectory variable to _mediaDirectory, since that's what it is.


app/code/Magento/Catalog/Model/Product/Option/Type/File.php
```diff
@@ -492,7 +489,15 @@ public function copyQuoteToOrder()
      * @param string $url
      * @return $this
      */
-    public function setCustomOptionDownloadUrl($url)
+    public function setCustomOp
```
okorshenko: this code looks broken. @evktalo could you please review this code and fix it, so that i can proceed with acceptance of this PR. Thank you

okorshenko (changes requested):

okorshenko:
Hi @evktalo could you please fix unit tests so that we can process this PR?

evktalo:
Yes, I will work on fixing the tests. Thank you!

okorshenko:
@evktalo I will the the tests on our side and will proceed with PR acceptance. Thank you!

magento-team:
@evktalo thank you for your contribution

Issue #8287, opened 2017-01-26
InputException while updating cart with Minimum order amount enabled


sahirzm:
### Preconditions
1. Magento CE 2.1.3 
2. PHP 7.0.13
3. OS: Ubuntu 16.10
4. Apache 2.4.18

### Steps to reproduce
1. In Admin, go to Stores->Configuration->Sales->Sales 
2. In Minimum Order Amount, Select Yes for Enable
3. Fill in Minimum Amount and Description Message.
4. Save Config.
5. In Frontend, Add product to cart.
6. Go to Cart, Change Qty and click Update Shopping Cart. 

### Expected result
Cart Should be Updated with new quantity.

### Actual result
Error 500:
[Wed Jan 25 20:23:35.168951 2017] [:error] [pid 9139] [client 127.0.0.1:55620] PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Magento\\Framework\\Exception\\InputException::__construct() must be an instance of Magento\\Framework\\Phrase, string given, called in /var/www/vendor/magento/module-quote/Model/ShippingAddressManagement.php on line 121 and defined in /var/www/vendor/magento/framework/Exception/InputException.php:51
Stack trace:
#0 /var/www/vendor/magento/module-quote/Model/ShippingAddressManagement.php(121): Magento\\Framework\\Exception\\InputException->__construct('Kindly place an...')
#1 /var/www/vendor/magento/module-quote/Model/Quote/ShippingAssignment/ShippingProcessor.php(66): Magento\\Quote\\Model\\ShippingAddressManagement->assign('3', Object(Magento\\Quote\\Model\\Quote\\Address))
#2 /var/www/vendor/magento/module-quote/Model/Quote/ShippingAssignment/ShippingAssignmentProcessor.php(76): Magento\\Quote\\Model\\Quote\\ShippingAssignment\\ShippingProcessor->s in /var/www/vendor/magento/framework/Exception/InputException.php on line 51, referer: https://localhost/checkout/cart/

miakusha:
Hi @sahirzm thanks for report
We've created internal ticket MAGETWO-64011 to address this issue.

okorshenko:
Closing this issue. It was fixed in https://github.com/magento/magento2/pull/8474 for develop branch

magento-team:
Internal ticket to track issue progress: MAGETWO-64399

DavaGordon:
[...]

PR #8474, opened 2017-02-08
Fix for https://github.com/magento/magento2/issues/8287

ericrisler:
Added missing __() wrap on execption. Without it a new exception is thrown: PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Magento\Framework\Exception\InputException::__construct() must be an instance of Magento\Framework\Phrase, string given

Fix for https://github.com/magento/magento2/issues/8287

### DIFF
--- a/app/code/Magento/Quote/Model/ShippingAddressManagement.php
+++ b/app/code/Magento/Quote/Model/ShippingAddressManagement.php
@@ -118,7 +118,7 @@ class ShippingAddressManagement implements \Magento\Quote\Model\ShippingAddressM
         $address->setCollectShippingRates(true);
 
         if (!$quote->validateMinimumAmount($quote->getIsMultiShipping())) {
-            throw new InputException($this->getMinimumAmountErrorMessage()->getMessage());
+            throw new InputException(__($this->getMinimumAmountErrorMessage()->getMessage()));
         }
 
         try {


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8474) <br/>All committers have signed the CLA.

okorshenko (changes requested):
@ericrisler thank you for your contribution. Your PR fixes the issue but I'd recommend alternative solution.

The issue itself goes from https://github.com/magento/magento2/blob/develop/app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php#L46

If error message is configured the method returns it as is. I would recommend return Phrase object from the method \Magento\Quote\Model\Quote\Validator\MinimumOrderAmount\ValidationMessage::getMessage() in all cases. In this way, we don't need to wrap method result second time. Plus, the method itself will have strict return type which is better from code design perspective. 

СС: @sidolov

okorshenko (approved):

okorshenko:
@ericrisler thank you for your contribution. We merged your PR with some adjustments

warperharper:
This doesn't resolve the issue for me. Add one item to cart OK. Second item fails if value is below minimum amount. If value over minimum amount is added to cart, you can add more items, but you cannot delete items if removing them brings the value below minimum. 2.1.6

mzenner1:
I'm having the same issue as @warperharper.  I can add one item just fine.  After that, I can't add anything else to my cart or even remove the existing item.  My minimum order amount is $60.  I added something to my cart that was $2.50.  Magento 2.1.6.

Please fix this guys.  My client needs the minimum order amount

ifeltsweet:
Still broken on 2.1.6. The change doesn't seem to be merged into the code.

ifeltsweet:
Still broken in 2.1.8... O_O

okorshenko:
Fix will be available in 2.2

Ctucker9233:
@okorshenko Any plans to backport seeing as how this was an issue with 2.1?

PR #8481, opened 2017-02-09
Remove zend json checkout

dmanners:
As Zend1 has been EOL we should consider removing Zend1 components in Magento2. In this pull request I am replacing the usage of Zend_Json with the Magento Framework lib's serializer json object.

### DIFF
--- a/app/code/Magento/Checkout/Block/Cart/Shipping.php
+++ b/app/code/Magento/Checkout/Block/Cart/Shipping.php
@@ -86,4 +86,12 @@ class Shipping extends \Magento\Checkout\Block\Cart\AbstractCart
     {
         return $this->_storeManager->getStore()->getBaseUrl();
     }
+
+    /**
+     * @return bool|string
+     */
+    public function getSerializedCheckoutConfig()
+    {
+        return $this->serializer->serialize($this->getCheckoutConfig());
+    }
 }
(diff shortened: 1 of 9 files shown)


app/code/Magento/Checkout/Block/Cart/Shipping.php
```diff
@@ -21,26 +21,35 @@ class Shipping extends \Magento\Checkout\Block\Cart\AbstractCart
...
-     * @codeCoverageIgnore
+     * @throws \RuntimeException
      */
     public function __construct(
         \Magento\Framework\View\Element\Template\Context $context,
         \Magento\Customer\Model\Session $customerSession,
         \Magento\Checkout\Model\Session $checkoutSession,
         \Magento\Checkout\Model\CompositeConfigProvider $configProvider,
         array $layoutProcessors = [],
+        \Magento\Framework\Serialize\Serializer\Json $serializer = null,
```
fooman: I think if you are adding a new optional param like this it should come last as you are trying to avoid breaking anything.

dmanners: Oh yeah thanks @fooman for catching that.

okorshenko:
@dmanners  thank you for your contribution!

PR #8485, opened 2017-02-09
Fix/imposible customize product collection

dvynograd:
Right now it's impossible to customize product collection on the catalog products list page.

As i can see, Magento supposed to provide this ability in \Magento\Catalog\Block\Product\ListProduct::_beforeToHtml 

`        $this->_eventManager->dispatch(
            'catalog_block_product_list_collection',
            ['collection' => $this->_getProductCollection()]
        );

        $this->_getProductCollection()->load();
`

but due to Magento has the observer CatalogBlockProductCollectionBeforeToHtmlObserver wich listen this event and loads collection. So it prevent to collection modifications in future customizations.

Following PR is fixing that issue.


app/code/Magento/Review/Observer/CatalogBlockProductCollectionLoadAfter.php
```diff
@@ -1,13 +1,13 @@
 <?php
 /**
- * Copyright © 2013-2017 Magento, Inc. All rights reserved.
+ * Copyright © 2016 Magento. All rights reserved.
```
maghamed: why did you change copyright ?

dvynograd: I don't know will put in back

dvynograd: Fixed

app/code/Magento/Review/etc/frontend/events.xml
```diff
@@ -9,7 +9,7 @@
     <event name="tag_tag_product_collection_load_after">
         <observer name="review" instance="Magento\Review\Observer\TagProductCollectionLoadAfterObserver" shared="false" />
     </event>
-    <event name="catalog_block_product_list_collection">
-        <observer name="review" instance="Magento\Review\Observer\CatalogBlockProductCollectionBeforeToHtmlObserver" shared="false" />
+    <event name="catalog_product_collection_load_after">
+        <observer name="review" instance="Magento\Review\Observer\CatalogBlockProductCollectionLoadAfter" shared="false" />
     </event>
```
maghamed, line 14: We can't accept such fix because it will lead to general performance degradation of Product Collection load time. 
Theoretically it's correctly to treat reviews as a part of Product Aggregation Root, but this behavior should be consistent with how Catalog APIs work (aka ProductRepositoryInterface::get and ProductRepositoryInterface::getList, whether these API calls fill ProductInterface with Reviews or not).
For now Review module doesn't provide Extension attribute for ProductInterface. 
So, to prevent inconsistency between ProductCollection::load and ProductRepository::getList as well as not introduce performance degradation we will not add product reviews for all product collection load calls.

Proposed fix for this issue, taking into account Backward Compatibility requirements.
1. Add Optional dependency (which by default is equal to NULL) to Magento\Catalog\Block\Product\ListProduct class. With name $collectionLoader
2. Class $collectionLoader  will have the only public method - load($collection). Which loads collection passed as input parameter.
3. Method load of this class could be pluginized and logic which exists now in CatalogBlockProductCollectionBeforeToHtmlObserver, could be added as a plugin for  $collectionLoader::load($productCollection)

maghamed: @dvynograd  are there any updates?

maghamed (changes requested):

ishakhsuvarov:
@dvynograd 
Would you like to continue with this PR? If so, please see review comments.
Thank you

ishakhsuvarov:
@dvynograd I am closing this PR for now due to lack of activity.
Please check the code review comments, update with the latest `develop` and reopen this ticket if you wish to continue.
Thanks!

Issue #8315, opened 2017-01-27
Error with StdoTest


dmanners:
### Preconditions

1. Magento: develop
2. PHP 7.0.15

### Steps to reproduce

1. Run `./vendor/bin/phpunit --debug -v -c dev/tests/unit/phpunit.xml.dist setup/src/Magento/Setup/Test/Unit/Module/I18n/Dictionary/Writer/Csv/StdoTest.php`

### Expected result

1. Assertion to show as all green

### Actual result

1. PHPUnit cannot output anything as it uses the STDOUT stream which has just been destroyed in the teadown of the test.

I am not even 100% if this testing is actually makes sense. But because of the `__destruct` method in `Magento\Setup\Module\I18n\Dictionary\Writer\Csv` the output will always be closed and then PHPUnit will fail when trying to write the test results to the screen.

dmanners:
This also causes the same issue while running the test `setup/src/Magento/Setup/Test/Unit/Module/I18n/FactoryTest.php`

dmanners:
If you run phpunit with the `--stderr` option then this problem will not occur. Is this the desired option here? I mean it works but to me it seems that there could be deeper issues with closing `STDOUT` that might not be obvious.

Actually that seems to cause issues with the session tests also.

dmanners:
Is this an obsolete class? I found it referenced in `dev/tests/static/testsuite/Magento/Test/Legacy/_files/obsolete_classes.php` if that is the case can we remove the class or tests?

schmengler:
"Is this an obsolete class?" I hope so. It can break everything by closing STDOUT...

okorshenko:
@dmanners thank you for reporting the issue. Internal ticket: MAGETWO-64853

okorshenko:
@dmanners @schmengler class `Magento\Setup\Module\I18n\Dictionary\Writer\Csv` is not obsolete. This class replaces removed class `Magento\Tools\I18n\Dictionary\Writer\Csv`.

The file `dev/tests/static/testsuite/Magento/Test/Legacy/_files/obsolete_classes.php` has the follwing format: [`Removed\Class\Name`, `Class\Name\To\Use`]

orlangur:
[...]

PR #8487, opened 2017-02-09
Mark the STDO test as skipped

dmanners:
This is related to https://github.com/magento/magento2/issues/8315.

Basically phpunit uses the STDO and so we should not close this. Either remove the test or skip the test, I dont mint either.


okorshenko:
@dmanners thank you for your contribution. We will take a look at this issue

PR #8498, opened 2017-02-09
Product->save() shouldn't be called directly

stansm:
Here is a problem with calling Product->save() directly in ProductWebsiteLinkRepository:

When product has custom options it should have 'has_options' and possibly 'required_options' set to true but calling Product->save() without $product->setCanSaveCustomOptions(true); will result in both 'has_options' and 'required_options' to be set to false.

Also calling $product->save() won't invalidate internal ProductRepository's cache (instances , instancesById) and this would produces other unexpected issues.

### DIFF
--- a/app/code/Magento/Catalog/Model/ProductWebsiteLinkRepository.php
+++ b/app/code/Magento/Catalog/Model/ProductWebsiteLinkRepository.php
@@ -37,7 +37,7 @@ class ProductWebsiteLinkRepository implements \Magento\Catalog\Api\ProductWebsit
         $product = $this->productRepository->get($productWebsiteLink->getSku());
         $product->setWebsiteIds(array_merge($product->getWebsiteIds(), [$productWebsiteLink->getWebsiteId()]));
         try {
-            $product->save();
+            $this->productRepository->save($product);
         } catch (\Exception $e) {
             throw new CouldNotSaveException(
                 __(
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8498) <br/>All committers have signed the CLA.

vrann:
@stansm Thanks you, change looks good.
Can you please write integration test which exposes the issue?

stansm:
@vrann ok, I'll do it this weekend

vrann:
@stansm please verify this bug report. Is that what you had in mind?

**Steps to Reproduce**
1. Create visible product in default store and assign it to category. SKU: "test".
2. Add custom option to the product, i.e. type text field, SKU test-test, proce $10
3. Navigate to the frontpage of default store to the category, click on the product. Observe field to enter value of the custom option for the proce $10
4. Create website#2 with the store and store view
5. Invoke Web API POST /V1/products/:sku/websites per this schema [http://devdocs.magento.com/swagger/](http://devdocs.magento.com/swagger/)
```
{
  "productWebsiteLink": {
    "sku": "test",
    "website_id": 2
  }
}
```
6. Navigate to the product page in either website

**Expected result:**
product is displayed with the custom option

**Actual result:**
custom option is missing on the product page

vrann:
@stansm please verify steps and I'll merge pull request upon manual verification. You are welcome to add tests as a separate pull request though!

stansm:
@vrann  Instead of 4 I've just had a code like:
```
$siteLink = $productWebsiteLink->create();
$siteLink->setSku($product->getSku());
$siteLink->setWebsiteId(2);
$productWebsiteLinkRepository->save($siteLink);`
```
and after it executed 'has_options' and 'required_options' were set to false on this product (I've traced it and that's how it should really be according to the logic in the Product::beforeSave). And even that 'has_options' and 'required_options' were gone, custom options were still showing on a product page but were missing when added to the cart. This is what I know for now..

vrann:
@stansm thank you, that is probably same the flow but from the code perspective. I reproduced the issue using the steps to reproduce.

PR #8505, opened 2017-02-10
Update the Adminhtml image tree JSON

dmanners:
In this PR I have updated the CMS Image JsonTree to use the core Magento Json serilizer lib rather than Zend1' json class.


okorshenko:
@dmanners thank you for your continuous contribution!

PR #8526, opened 2017-02-13
Update Save.php

arosiek:
Problem came when I added new simple product, then I added children product to it. Magento in this case "upgrade" product to configurable but in redirectBack goes to edit form with _current=true so in url 'type=simple' still. This makes problem when I have product attributes for simple type but not for configurable. Attribute is not saved but with type=simple in url it is still visible so js validation throws error.

This fix updates url 'type' param to be current product type.

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
@@ -117,6 +117,7 @@ class Save extends \Magento\Catalog\Controller\Adminhtml\Product
                 $productId = $product->getEntityId();
                 $productAttributeSetId = $product->getAttributeSetId();
                 $productTypeId = $product->getTypeId();
+                $this->getRequest()->setParam('type', $productTypeId);
 
                 $this->copyToStores($data, $productId);
 


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8526) <br/>All committers have signed the CLA.

ihor-sviziev:
It's bad idea to change current request object. We need to change redirect url instead

arosiek:
@ihor-sviziev We have this option or copy all params and create new url. Why shouldn't I modify this request param if magento guys modified variable value? Basicly this is the aim.  Adding new redirecting option would increase logical complexity which is already to high in this method.

orlangur:
> This makes problem when I have product attributes for simple type but not for configurable

This is just a misconfiguration. In Magento 2 when editing product you're not sticked to only one product type - URL parameter just defines the display of initial page, switching between simple/configurable/virtual/downloadable product types occurs automatically.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
```diff
@@ -117,6 +117,7 @@ public function execute()
                 $productId = $product->getEntityId();
                 $productAttributeSetId = $product->getAttributeSetId();
                 $productTypeId = $product->getTypeId();
+                $this->getRequest()->setParam('type', $productTypeId);
```
orlangur, line 120: Modifying a request is not a good idea. This could be added to `$resultRedirect->setPath` calls but seems like it is not needed if you configure product attributes correctly.

orlangur (changes requested):

app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
arosiek: Did you check the problem? problem is magento not getting the type from product if it is set in url.

orlangur: This is explained here: https://github.com/magento/magento2/pull/8526#issuecomment-279990671 You should not have attributes applicable to configurable and not applicable to simple products as they are interchangeable.

arosiek: and it appears only in the first save (add attribute for simple and not configurable product, set this attribute to be required. Then create new simple product, add children product, save).

arosiek:
@orlangur Did you check the problem? problem is magento not getting the type from product if it is set in url.

and it appears only in the first save (add attribute for simple and not configurable product, set this attribute to be required. Then create new simple product, add children product, save).

and this is not misconfiguration because:
1. I can do that in clean magento.
2. I need to do that because of the project.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
arosiek: Why modyfing request is not a good idea? adding new redirecting option would increase logical complexity which is already to high in this method.

orlangur:
> problem is magento not getting the type from product if it is set in url

It is not supposed to get type from URL for anything. `/type/simple/` and `/type/configurable/` should work the same way because type is switched dynamically depending on user actions.

"Apply To" field for product attributes was removed from UI specifically to avoid this kind of problems.

I remember such misconfiguration existed for `Color` attribute at some point. Which attributes in clean Magento are enabled for only one of simple/configurable types at the moment? This needs to be fixed and not the behavior of controller.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
orlangur: No other `setParam` called here. `type` parameter in proper place does not increase logical complexity while changing request object in some random place do increase.

arosiek: in did, but type is changed here so why incorrect param should still exists in request? this is not inceasing complexity but fixing the value in the request object as it is modified in product.

orlangur: I cannot add more on this :) Maybe some common principle, like CQRS, exists and it says that we should not modify what we received in request.

In general, the fewer mutations - the better. For example, in https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/Model/Product/TypeTransitionManager.php#L48 product type changing responsibility is clearly encapsulated. Here we just modify a request in arbitrary place.

PR #8562, opened 2017-02-15
Product name not shown in message when trying to add more products to cart then there are in stock

joost-florijn-kega:
This is a fix for issue https://github.com/magento/magento2/issues/8560

### DIFF
--- a/app/code/Magento/CatalogInventory/Api/StockStateInterface.php
+++ b/app/code/Magento/CatalogInventory/Api/StockStateInterface.php
@@ -27,13 +27,14 @@ interface StockStateInterface
 
     /**
      * @param int $productId
+     * @parem $product
      * @param float $itemQty
      * @param float $qtyToCheck
      * @param float $origQty
      * @param int $scopeId
      * @return int
      */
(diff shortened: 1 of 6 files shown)


app/code/Magento/CatalogInventory/Api/StockStateInterface.php
```diff
@@ -27,13 +27,14 @@ public function verifyNotification($productId, $scopeId = null);
...
      * @param int $productId
+     * @parem $product
      * @param float $itemQty
      * @param float $qtyToCheck
      * @param float $origQty
      * @param int $scopeId
      * @return int
      */
-    public function checkQuoteItemQty($productId, $itemQty, $qtyToCheck, $origQty, $scopeId = null);
+    public function checkQuoteItemQty($productId, $product, $itemQty, $qtyToCheck, $origQty, $scopeId = null);
```
orlangur, line 37: Fix is not correct - you cannot change Api methods like this.

Since you already have a `productId` within method, for a proper fix you need to find out why `name` is not loaded correctly in these circumstances.

orlangur (changes requested):

PR #8568, opened 2017-02-15
Fix quote's outdated shipping address overwriting PayPal Express shipping address

torreytsui:
# Issue

PayPal responded shipping address fail to save against quote on Braintree PayPal Express checkout

# Preconditions
Magento EE 2.1.3

# Steps to reproduce

1. Log in to an account
2. Add product to the basket
3. Go to *One Page Checkout*
4. Enter every shipping address details; select a shipping method; and continue to the payment step
5. Don't proceed to payment; Go back to the basket and click PayPal Express checkout
6. Enter paypal credential (or proceed with sandbox default address data); proceed

paypal-one-click-outdated-shipping-address

# Expected Result

On `braintree/paypal/review/`, delivery address should show the PayPal's shipping address.

For sandbox, it should be:

```
Mr John Doe
123 Division Street
Apt. #1
Chicago, Greater Manchester, 60618
United States
```

paypal-one-click-expected-address

# Actual Result

The delivery address shows the outdated shipping address which entered from One Page Checkout

### DIFF
--- a/app/code/Magento/Braintree/Model/Paypal/Helper/QuoteUpdater.php
+++ b/app/code/Magento/Braintree/Model/Paypal/Helper/QuoteUpdater.php
@@ -84,6 +84,14 @@ class QuoteUpdater extends AbstractHelper
 
         $quote->collectTotals();
 
+        /**
+         * Unset shipping assignment to prevent from saving / applying outdated data
+         * @see \Magento\Quote\Model\QuoteRepository\SaveHandler::processShippingAssignment
+         */
+        if ($quote->getExtensionAttributes()) {
+            $quote->getExtensionAttributes()->setShippingAssignments(null);
+        }
+
         $this->quoteRepository->save($quote);
     }
 
--- a/app/code/Magento/Braintree/Model/Paypal/Helper/QuoteUpdater.php
+++ b/app/code/Magento/Braintree/Model/Paypal/Helper/QuoteUpdater.php
@@ -130,11 +130,6 @@ class QuoteUpdater extends AbstractHelper
         $shippingAddress->setCollectShippingRates(true);
 
         $this->updateAddressData($shippingAddress, $details['shippingAddress']);
-
-        // PayPal's address supposes not saving against customer account
-        $shippingAddress->setSaveInAddressBook(false);
-        $shippingAddress->setSameAsBilling(false);
-        $shippingAddress->unsCustomerAddressId();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


vrann:
@torreytsui thank you for contribution, started verification

vrann:
@torreytsui repeated the steps to reproduce, on the latest develop branch. The result is different from what is reported: after step 6, it opens up PayPal pop-up with the address which I entered on step 4:
screen shot 2017-02-15 at 4 25 18 pm

PayPal Express does not allow me to enter any other address. Am I doing something wrong?

YevSent:
Hi, @torreytsui, I've few important remarks:

- the changes in https://github.com/magento/magento2/pull/8568/files#diff-ebdcbe8410d2ed8e92e1a63993c9a6cfR135 (with next lines) and https://github.com/magento/magento2/pull/8568/files#diff-ebdcbe8410d2ed8e92e1a63993c9a6cfR162 (with next lines) should be moved to `\Magento\Braintree\Model\Paypal\Helper\QuoteUpdater::updateAddressData`.

- There is no `unsCustomerAddressId` method, it works because the address entity extends `DataObject`, but you should avoid using magic methods the `setCustomerAddressId(null)` should be used instead.

 - The problem with `region` property is not solved, so addresses will contain the region from the previous shipping address. You need to get region title by region received from Braintree PayPal.

torreytsui:
@vrann thanks for your response.

The popup is alright. I used full screen mode on Mac so it opened on a new tab instead. If you click `change` on the popup, you should be able to change your address; but it does not matter with this issue.

I can see the url on your screenshot is different from what I have. You were being redirected to `checkoutnow/` while I were in `one-touch-login-sandbox/`. 

Please also note that only Braintree PayPal Express has this issue, PayPal Express is alright.

Despite of the endpoint differences, I don't think it matter. If you keep going, when you arrive in the review page, you should be able to see the out dated shipping address rendered, as long as you have ever set a shipping address against quote (e.g., through one page checkout)

torreytsui:
@joni-jones Thanks for the feedbacks. I have updated the PR according to your first two points.

I'm looking into the `region` issue now. Will update it soon.

vrann:
@torreytsui thank you for the update.
Yeah, I noticed later that you are using Braintree PayPal Express. Please finish the changes and I will re-test it again.

vrann:
@torreytsui Hi, just checking up is there something you can add to this PR any time soon?

PR #8589, opened 2017-02-17
Set correct primary keys for temporary tables in product flat indexer

jarnooravainen:
Set 'row_id' as primary key because joins are made on it. Without this change product flat reindex can get practically stuck while making insert-from-select query in Magento\Catalog\Model\Indexer\Product\Flat\FlatTableBuilder::_fillTemporaryFlatTable.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8589) <br/>All committers have signed the CLA.

jarnooravainen:
Fixed this PR to be compatible with both CE and EE, although seems that CE is not affected by this issue. Field to be set as primary key is retrieved from the metadata, so for CE field will be entity_id and for EE row_id.

maghamed (approved):
@jarnooravainen  That's a very good catch! Well Done!

vrann:
@jarnooravainen Thank you for the contribution!

PR #8593, opened 2017-02-17
Fix #7371

### LINKED ISSUE #7371: Wrong price upon selecting product options after switching currency
Encountered an issue when selecting product options using the swatches in product details page wherein the price of the item in a previously selected currency gets displayed instead of the corresponding price in the currently selected currency. Disabling Block HTML cache type seems to prevent this issue from happening.

### Preconditions
1. Magento 2.1.x
2. Store supports multiple currencies
3. All cache types are enabled

### Steps to reproduce
1. Select a currency (e.g., GBP) then view a product's details:
image
2. Select product options in the current product details page :
image
3. Switch to a different currency (e.g., JPY):
image
4. Select a product variant; the price reverts back to the previous currency's value but with the current currency's symbol:
image

### Expected result
Correct equivalent price in the current currency should be displayed

### Actual result
The price in the previously selected currency gets displayed

rossluk:
'catalog_rule_price' is added to products only before load of the Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Product\Collection and it is used only for detecting the minimal price here - Magento\Catalog\Pricing\Price\BasePrice. Due to it the price has never converted to the site currency.

### DIFF
--- a/app/code/Magento/CatalogRule/Pricing/Price/CatalogRulePrice.php
+++ b/app/code/Magento/CatalogRule/Pricing/Price/CatalogRulePrice.php
@@ -99,9 +99,9 @@ class CatalogRulePrice extends AbstractPrice implements BasePriceProviderInterfa
                         $this->product->getId()
                     );
                 $this->value = $this->value ? floatval($this->value) : false;
-                if ($this->value) {
-                    $this->value = $this->priceCurrency->convertAndRound($this->value);
-                }
+            }
+            if ($this->value) {
+                $this->value = $this->priceCurrency->convertAndRound($this->value);
             }
         }
 
--- a/app/code/Magento/CatalogRule/Test/Unit/Pricing/Price/CatalogRulePriceTest.php
+++ b/app/code/Magento/CatalogRule/Test/Unit/Pricing/Price/CatalogRulePriceTest.php
@@ -208,12 +208,20 @@ class CatalogRulePriceTest extends \PHPUnit_Framework_TestCase
 
     public function testGetValueFromData()
     {
+        $catalogRulePrice = 7.1;
+        $convertedPrice = 5.84;
+        
+        $this->priceCurrencyMock->expects($this->any())
+            ->method('convertAndRound')
+            ->with($catalogRulePrice)
+            ->will($this->returnValue($convertedPrice));
+        
         $this->saleableItemMock->expects($this->once())->method('hasData')
             ->with('catalog_rule_price')->willReturn(true);
         $this->saleableItemMock->expects($this->once())->method('getData')
-            ->with('catalog_rule_price')->willReturn('7.1');
+            ->with('catalog_rule_price')->willReturn($catalogRulePrice);
 
-        $this->assertEquals(7.1, $this->object->getValue());
+        $this->assertEquals($convertedPrice, $this->object->getValue());
     }
 
     public function testGetAmountNoBaseAmount()
--- a/app/code/Magento/CatalogRule/Test/Unit/Pricing/Price/CatalogRulePriceTest.php
+++ b/app/code/Magento/CatalogRule/Test/Unit/Pricing/Price/CatalogRulePriceTest.php
@@ -210,12 +210,12 @@ class CatalogRulePriceTest extends \PHPUnit_Framework_TestCase
     {
         $catalogRulePrice = 7.1;
         $convertedPrice = 5.84;
-        
+
         $this->priceCurrencyMock->expects($this->any())
             ->method('convertAndRound')
             ->with($catalogRulePrice)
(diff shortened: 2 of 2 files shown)


okorshenko:
@rossluk could you please provide clear test cases? what is the issue and how to reproduce it. And how we can test your solution.

Thank you

rossluk:
@okorshenko 
1. Install Magento;
2. Add new currency to your website (By default there is USD only, leave USD as base currency and add Euro, then update rates to see the difference between product prices in different currency);
3. Create configurable product with some modifications (more then one);
4. Create Catalog Rule with some discount (for instance 50% for Default Category);
5. Go to product page (which was created on step 3);
6. Change site currency from base to Euro;
7. You should see the normal base price in Euro and it's ok, but if you change product option you'll see product price changed. It will have Euro sign but price value will be in USD. Option prices are received from server (swatches json config) and always have Base and Final price in main site currency (these prices for product options are loaded from \Magento\CatalogRule\Pricing\Price\CatalogRulePrice through the price resolver and it's a problem);

orlangur:
@okorshenko, steps are described in #7371, you know :)

@rossluk, fix looks incorrect to me. Check the logic in https://github.com/magento/magento2/commit/e3cecafccd999ee3b9ab5df0094dec02c12fad8c#diff-ca5d201faa2fc6d81caa299b64589a3bR91
From what I can understand here, value stored in product must be already converted.

So, if the issue is relevant for `develop` branch, you should find a place where incorrect value is stored to product and make this place multi-currency-aware. Sorry, didn't play with the code, thus, I may oversee the whole picture.

app/code/Magento/CatalogRule/Pricing/Price/CatalogRulePrice.php
```diff
@@ -99,9 +99,9 @@ public function getValue()
                         $this->product->getId()
                     );
                 $this->value = $this->value ? floatval($this->value) : false;
-                if ($this->value) {
-                    $this->value = $this->priceCurrency->convertAndRound($this->value);
-                }
+            }
+            if ($this->value) {
```
orlangur, line 103: https://github.com/magento/magento2/pull/8593#issuecomment-281721542

orlangur (changes requested):

app/code/Magento/CatalogRule/Pricing/Price/CatalogRulePrice.php
rossluk: @orlangur Sorry, I can't understand what I should see in diff. Problem is 'catalog_rule_price' which is added to join in Magento\CatalogRuleConfigurable\Plugin\ConfigurableProduct\Model\ResourceModel\AddCatalogRulePrice. I didn't see any mention about current currency (if value from db already in current currency we should see something about it) or conversion of the value in this file, and I didn't see any places where 'catalog_rule_price' or 'rule_price' converts to current currency after. Maybe I miss something?

orlangur:
```
if ($this->product->hasData(self::PRICE_CODE)) {
    $this->value = floatval($this->product->getData(self::PRICE_CODE)) ?: false;
```

Where this value is stored? I believe currency conversion is missed there.

So, it's better to store totally prepared value than convert it after.

However, if the value stored in product is not store-aware by design, fix looks ok to me. To avoid regression it's better to create proper integration test, with catalog rule and multiple currencies, as change of started to fail unit test does not enforce anything (i.e. somebody doing next changes will just do the same - change hardcoded expectations - as test case is not sticked to any concrete scenario).

rossluk:
```php
if ($this->product->hasData(self::PRICE_CODE)) {
    $this->value = floatval($this->product->getData(self::PRICE_CODE)) ?: false;
```
This value is stored in the 'catalogrule_product_price' table. As I know all prices in Magento is converted after (from base currency to selected on site) and it's too much work to store prepared value because there is should be added a new dimension by currency per store. Value which is stored in product is website-aware, but currency is another dimension (you can have many currencies for one store view).

orlangur: I see... Thanks, that's exactly what I was asking.

Then
> fix looks ok to me

Integration test would be really valuable to not ever break this place again, but it's up to you, of course, whether invest time in it or not.

orlangur (approved):

okorshenko:
@rossluk thank you for your contribution! PR has been merged to develop branch!
@orlangur thank you for review and your feedback.
@sdzhepa great job on testing of this solution!

Issue #6151, opened 2016-08-13
Can't delete last item in cart if Minimum Order is Enable 


philipvandebriel:
### Preconditions

Update magento 2.0.8 to Magento 2.1 
Debian Jessie
Mysql 5.6
Php 5.6
1. Put item in cart  
2. Try delete item , after popup error
### Steps to reproduce

1. Go  to admin -> SALES -> sales
2. Enable -> Minimum Order Amount
3. Minimum Amount = 10 or something
### Expected result
1. After remove product from mini cart then the cart need to be empty 
### Actual result
1. Last product stay in the cart can't remove
   Try to remove the item then we get this error :

Recoverable Error: Argument 1 passed to Magento\Framework\Exception\InputException::__construct() must be an instance of Magento\Framework\Phrase, string given, called in /var/www/clients/client1/web1/web/vendor/magento/module-quote/Model/ShippingAddressManagement.php on line 121 and defined in /var/www/clients/client1/web1/web/vendor/magento/framework/Exception/InputException.php on line 51

If disable  Minimum Order Amount then everything is fine.

stripcurtains:
I'm having the exactly same issue with 2.1

stripcurtains:
@wert2all Is there a temporary fix for it?

AVoskoboinikov:
Hi, guys.

Thanks for reporting this, I've created ticket #MAGETWO-57354 in our internal Jira. This issue will be closed after fix.

csd-rhuang:
Hi, 

I experienced the same issue. And even worse, I will get a http 500 error when I add the 3rd item to the cart when the minimum order is enabled. 

It is Magento 2.1. Bitnami package on AWS.

Thanks

Rex

vijay-wagento:
For my ongoing project in Magento 2.1 I have done below things to come out from this problem,

1) Create Plugin

```
 <type name="Magento\Quote\Model\Quote">
        <plugin name="min-orderamount-skip-delete" type="Namespace\ModuleName\Model\QuoteAfter" sortOrder="10"/>
    </type>
```

2) in QuoteAfter class

```
public function afterValidateMinimumAmount()
    {
        $actionName = $this->request->getActionName();
        if($actionName === 'delete') {
            return true;
        }
    }

```

This solved my problem and I got success to delete item.

ivankaranjac:
magento 2.1.2
Problem with deleting, updating and adding to cart once an item is added to the cart

[...]

PR #8600, opened 2017-02-18
Fix: various problems with minimum order amount enabled

ericrisler:
See https://github.com/magento/magento2/issues/6151 for the basics.

When minimum order amount is enabled:
1. add 1 product to the cart; total is below min order amount -> no error.
2. Try to remove that product -> error about minimum amount.
3. Try to add more to the cart with total still below min order -> error about minimum amount
4. Add a product that will bring the total cart amount **over** the min order amount and no errors!

I will request the Magento team takes a look at this line. I don't understand why the minimum amount is being validated here but perhaps it has something to do with multi-shipping checkouts. However, it botches the min order amount logic badly in it's current form.


okorshenko:
@ericrisler thank you for your contribution. Unfortunately the issue exists, but it can't be fixed in proposed way. Closing this PR.

PR #8603, opened 2017-02-18
Updated parent inheritance

koenner01:
Fix for https://github.com/magento/magento2/issues/8587

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Attribute/Frontend/Image.php
+++ b/app/code/Magento/Catalog/Model/Product/Attribute/Frontend/Image.php
@@ -11,7 +11,7 @@
  */
 namespace Magento\Catalog\Model\Product\Attribute\Frontend;
 
-class Image extends \Magento\Eav\Model\Entity\Attribute\Frontend\AbstractFrontend
+class Image extends \Magento\Eav\Model\Entity\Attribute\Frontend\DefaultFrontend
 {
     /**
      * Store manager


maghamed (changes requested):
replied in the issue https://github.com/magento/magento2/issues/8587#issuecomment-283409665

@koenner01  Why do you need Image to be successor of DefaultFrontend?

> An example of where this is already done in the core:
> Magento\Eav\Model\Entity\Attribute\Frontend\Datetime

We don't have these neither in 2.0.*
https://github.com/magento/magento2/blob/2.0.13/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/Datetime.php#L11
nor in 2.1.*
https://github.com/magento/magento2/blob/2.1.4/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/Datetime.php#L11
nor in develop branch 
https://github.com/magento/magento2/blob/develop/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/Datetime.php#L11

Actually for now Magento\Eav\Model\Entity\Attribute\Frontend\DefaultFrontend  class doesn't have any successor.

Currently the only place where Default Frontend class is used is in the AbstractAttribute class
```
    /**
     * Retrieve frontend instance
     *
     * @return \Magento\Eav\Model\Entity\Attribute\Frontend\AbstractFrontend
     */
    public function getFrontend()
    {
        if (empty($this->_frontend)) {
            if (!$this->getFrontendModel()) {
                $this->setFrontendModel($this->_getDefaultFrontendModel());
            }
            $this->_frontend = $this->_universalFactory->create($this->getFrontendModel())->setAttribute($this);
        }

        return $this->_frontend;
    }
```

and this behavior looks correct for me.
What customization abilities you are losing, having such hierarchy of inheritance ? 

Magento discourages usage of inheritance. So, a deep inheritance tree is a "smell" from architectural point of view.

PR #8608, opened 2017-02-19
Replace Zend_Mail (ZF1) with Zend\Mail (ZF2)

schmengler:
This PR probably needs more work, at least more testing, I cannot believe that it is so easy.

Opening for review and acceptance tests. If it's working I plan to refactor the code a bit (see TODO comments) - feedback appreciated!


orlangur:
> needs more work, at least more testing, I cannot believe that it is so easy

True, true... One integration test is failing, another one caused fatal error :) Check [Travis CI](https://travis-ci.org/magento/magento2/builds/203273924).

Note that refactoring should be performed in such a way that all existing customizations relying on `Magento_Framework` and not `Zend_Mail` directly work as before (just like integration tests, people will not be happy getting such errors after upgrade).

Will share my thoughts on the code a bit later.

schmengler:
> True, true... One integration test is failing, another one caused fatal error :) Check Travis CI.

I've seen it, but could not look into the details yet. I tried to take care not to change the `Magento_Framework` behavior, and it's possible that tests are only failing because mocks have to be changed.

I'm waiting for your feedback on the code now!

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -7,14 +7,25 @@
  */
 namespace Magento\Framework\Mail;
 
-class Message extends \Zend_Mail implements MessageInterface
+use Zend\Mime\Mime;
+use Zend\Mime\Part;
+
+/**
+ * @todo composition instead of inheritance for better testability
```
orlangur: Do it now and expose only necessary public methods.

Otherwise people may rely on Zend implementation rather then on narrowed interface and we will continue to have problems like Fatal Error with absent  `getBodyHtml`.

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -7,14 +7,25 @@
...
+class Message extends \Zend\Mail\Message implements MessageInterface
 {
+
     /**
      * @param string $charset
      */
     public function __construct($charset = 'utf-8')
     {
-        parent::__construct($charset);
+        $this->encoding = $charset;
```
orlangur: Rename constructor argument accordingly and use `setEncoding` method of `\Zend\Mail\Message` instead (currently `$this->getHeaders()->setEncoding($encoding);` is missed).

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -25,35 +36,40 @@ public function __construct($charset = 'utf-8')
...
     /**
-     * Set message body
+     * Set message type
      *
-     * @param string $body
+     * @param string $type
      * @return $this
      */
-    public function setBody($body)
+    public function setMessageType($type)
```
orlangur: Please preserve order of methods to simplify checking of correspondence during review (getBody -> getBody, setBody -> setBody etc).

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -25,35 +36,40 @@ public function __construct($charset = 'utf-8')
...
-     * @return string
+     * @param null|object|string|\Zend\Mime\Message $body
+     * @return \Zend\Mail\Message
      */
-    public function getBody()
+    public function setBody($body)
     {
-        return $this->messageType == self::TYPE_TEXT ? $this->getBodyText() : $this->getBodyHtml();
+        if (is_string($body) && $this->messageType === MessageInterface::TYPE_HTML) {
+            $body = self::htmlMimeFromString($body);
```
orlangur: Method name must begin with verb, I suggest `createHtmlMimeFromString`.

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -25,35 +36,40 @@ public function __construct($charset = 'utf-8')
...
-     * Set message body
-     *
-     * @return string
+     * @param null|object|string|\Zend\Mime\Message $body
+     * @return \Zend\Mail\Message
      */
-    public function getBody()
+    public function setBody($body)
     {
-        return $this->messageType == self::TYPE_TEXT ? $this->getBodyText() : $this->getBodyHtml();
```
orlangur: Calls of `getBodyHtml` must be fixed all over the code, this is why we saw Fatal Error on Travis. Integration tests do not use mocks usually.

lib/internal/Magento/Framework/Mail/Test/Unit/MessageTest.php
```diff
@@ -14,6 +14,7 @@ class MessageTest extends \PHPUnit_Framework_TestCase
 
     protected function setUp()
     {
+        $this->markTestSkipped('obsolete ZF 1 test');
```
orlangur, line 17: Test must be reworked within this pull request.

lib/internal/Magento/Framework/Mail/Test/Unit/Template/TransportBuilderTest.php
```diff
@@ -53,6 +54,7 @@ protected function setUp()
         $objectManagerHelper = new \Magento\Framework\TestFramework\Unit\Helper\ObjectManager($this);
         $this->templateFactoryMock = $this->getMock(\Magento\Framework\Mail\Template\FactoryInterface::class);
         $this->messageMock = $this->getMock(\Magento\Framework\Mail\Message::class);
+        $this->messageMock->method('getHeaders')->willReturn(new Headers());
```
orlangur: `Headers` class must be mocked as well.

lib/internal/Magento/Framework/Mail/Test/Unit/TransportTest.php
```diff
@@ -5,45 +5,32 @@
...
-     * @expectedExceptionMessage No body specified
+     * @expectedExceptionMessage Invalid email; contains no "To" header
      */
     public function testSendMessageBrokenMessage()
     {
-        $this->_messageMock->expects($this->any())
-            ->method('getParts')
-            ->will($this->returnValue(['a', 'b']));
+        $transport = new \Magento\Framework\Mail\Transport(
+            new \Magento\Framework\Mail\Message()
```
orlangur, line 18: `Message` class must be mocked here.

lib/internal/Magento/Framework/Mail/Transport.php
```diff
@@ -6,7 +6,8 @@
  */
 namespace Magento\Framework\Mail;
 
-class Transport extends \Zend_Mail_Transport_Sendmail implements \Magento\Framework\Mail\TransportInterface
+
```
orlangur: Useless new line and `phpcs` does not enforce it yet.

lib/internal/Magento/Framework/Mail/Transport.php
```diff
@@ -6,7 +6,8 @@
...
+class Transport extends \Zend\Mail\Transport\Sendmail implements \Magento\Framework\Mail\TransportInterface
```
orlangur: Do aggregation instead of inheritance.

lib/internal/Magento/Framework/Mail/Transport.php
```diff
@@ -20,8 +21,8 @@ class Transport extends \Zend_Mail_Transport_Sendmail implements \Magento\Framew
      */
     public function __construct(\Magento\Framework\Mail\MessageInterface $message, $parameters = null)
     {
-        if (!$message instanceof \Zend_Mail) {
-            throw new \InvalidArgumentException('The message should be an instance of \Zend_Mail');
+        if (!$message instanceof \Zend\Mail\Message) {
```
orlangur: I don't think we should keep classes inherited from `Zend` classes in the system.

Just create necessary instance of `\Zend\Mail\Message` in `sendMessage` method instead.

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -7,14 +7,25 @@
  */
 namespace Magento\Framework\Mail;
 
-class Message extends \Zend_Mail implements MessageInterface
+use Zend\Mime\Mime;
+use Zend\Mime\Part;
+
+/**
+ * @todo composition instead of inheritance for better testability
+ * - add a ZendMailDecorator interface with getZendMail() method for usage in \Magento\Framework\Mail\Transport
```
orlangur: Please no. To me it looks like `\Zend\Mail\Message` has all necessary setters already.

lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -7,14 +7,25 @@
...
+ * @todo get rid of temporal coupling (setMessageType() + setBody())
+ * - deprecate setMessageType(), implement a HtmlMessage decorator instead
```
orlangur: I don't think such decorator would be really valuable. Do we have some needed cases not covered with current text/html types implementation?

orlangur (changes requested):

orlangur:
@schmengler, thanks for your efforts, implementation in general looks quite clean and minimalistic (just does what is really necessary).

Please make sure all builds are green before next review iteration.

okorshenko:
Hi @schmengler 
Are you going to proceed this with PR?

schmengler:
@okorshenko yes as soon as I find some time for it

okorshenko:
@schmengler thank you. Please let us know when you  will be ready. Thank you

lib/internal/Magento/Framework/Mail/Message.php
schmengler: Fortunately `getBodyHtml` is only used in integration tests. I'm trying to fix them

lib/internal/Magento/Framework/Mail/Message.php
schmengler: The problem with the current implementation is that you have to call `setMessageType()` before `setBody()`. I'm now considering changing the `MessageInterface` interface. As it should be used via `TransportBuilder`, it should not have serious impact.

- The current signatures `setBody(mixed $body)` and `mixed getBody()` are not meaningful. Either "mixed" must be replaced with "string" or we are leaking the zend abstraction.
- The interface does not allow multitype mails (this is the needed case not covered. Something that Magento always has been lacking of)

One idea is, replacing `setMessageType()`, `setBody()` and `getBody()` with:

- `setBodyHtml(string $html)`
- `setBodyText(string $text)`
- `string getBodyText()`
- `string getBodyHtml()`

This allows text mails, HTML mails, and multitype mails, does not require calling them in any specific order and does not reveal anything of the underlying Zend mail component.

schmengler: Another advantage is that this resembles the old zf1 implementation, so 3rd party code is actually less likely to break even though we have to change an `@api` interface

schmengler:
Because it's hidden now in the "outdated" review, here again my concerns about the message interface:

The problem with the current implementation is that you have to call `setMessageType()` before `setBody()`. I'm now considering changing the `MessageInterface` interface. As it should be used via `TransportBuilder`, it should not have serious impact.

- The current signatures `setBody(mixed $body)` and `mixed getBody()` are not meaningful. Either "mixed" must be replaced with "string" or we are leaking the zend abstraction.
- The interface does not allow multitype mails (this is the needed case not covered. Something that Magento always has been lacking of)

One idea is, replacing `setMessageType()`, `setBody()` and `getBody()` with:

- `setBodyHtml(string $html)`
- `setBodyText(string $text)`
- `string getBodyText()`
- `string getBodyHtml()`

This allows text mails, HTML mails, and multitype mails, does not require calling them in any specific order and does not reveal anything of the underlying Zend mail component.

Another advantage is that this resembles the old zf1 implementation, so 3rd party code is actually less likely to break even though we have to change an `@api` interface

I already added one method `getRawMessage()` because otherwise it would not be possible to have a Zend transport implementation by only using the message interface and not the underlying Zend message, so the PR will cause a BC break anyways. But IMHO it's a minor one and it is necessary to get the mail interfaces right.

schmengler:
And while we are at it, shouldn't `TransportInterface` look like this:

```
    function sendMessage(MessageInterface $message);
```

instead of 

```
    function sendMessage()
```

?

I don't understand, why the "Transport" implementation receives the message in the constructor. I'd rather have the transport instance reusable for different messages.

With the current implementation we have the `TransportBuilder` class, which actually builds a message. That would be much simpler and clearer if we did not couple the transport to the message like currently.

lib/internal/Magento/Framework/Mail/Test/Unit/TransportTest.php
schmengler: I disagree. It's part of the same component and not using external resources. To test handling of an incomplete message, we create a real message without "to" header. A stub just complicates things and a mock will make unnecessary assumptions about the implementation details.

okorshenko:
Hi @schmengler 
Just FYI: we stabilized this branch and working on the testing of new changes.

0m3r:
What about getHeaders and setHeaders for \Magento\Mail\Message? 

I want to set headers like a in-reply-to, message-id, references etc.

klein0r:
Is there a reason why this has been included into 2.2.8?

orlangur:
@klein0r yep, backward compatibility is the reason.

PR #7688, opened 2016-12-06
Add PHP 7.1 to Travis build

ktomk:
This change is for forward compatibility with PHP 7.1.

Add PHP 7.1 (currently PHP 7.1-RC6 on Travis) to the Travis build. Allow
failures with it as failures are expected.

To get it to work, for PHP 7.1 composer ignores platform requirements.

Refs:

- https://github.com/magento/magento2/issues/5701

PR #8609, opened 2017-02-20
[PHP 7.1 Compatibility] Void became a reserved word

orlangur:
Before this change `Creditmemo > Void` and `Invoice > Void` actions in Admin UI had no chance to work under PHP 7.1 as "void" became a reserved word.

Unit and static tests are now able to complete on Travis CI under PHP 7.1: https://travis-ci.org/orlangur/magento2/builds/203260140 (this temporary branch includes #7688 changes).

Corresponding PHPCS sniff is expanded with new reserved words and executed against full codebase.
Before: https://travis-ci.org/orlangur/magento2/jobs/203213045
After: https://travis-ci.org/orlangur/magento2/jobs/203230641

### DIFF
--- a/lib/internal/Magento/Framework/App/Router/ActionList.php
+++ b/lib/internal/Magento/Framework/App/Router/ActionList.php
@@ -34,7 +34,7 @@ class ActionList
         'for', 'foreach', 'function', 'global', 'goto', 'if', 'implements', 'include', 'instanceof',
         'insteadof','interface', 'isset', 'list', 'namespace', 'new', 'or', 'print', 'private', 'protected',
         'public', 'require', 'return', 'static', 'switch', 'throw', 'trait', 'try', 'unset', 'use', 'var',
-        'while', 'xor',
+        'while', 'xor', 'void',
     ];
 
     /**
--- a/app/code/Magento/Sales/Controller/Adminhtml/Order/Creditmemo/VoidAction.php
+++ b/app/code/Magento/Sales/Controller/Adminhtml/Order/Creditmemo/VoidAction.php
@@ -7,7 +7,7 @@ namespace Magento\Sales\Controller\Adminhtml\Order\Creditmemo;
 
 use Magento\Backend\App\Action;
 
-class Void extends \Magento\Backend\App\Action
+class VoidAction extends Action
 {
     /**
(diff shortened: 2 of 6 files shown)


lib/internal/Magento/Framework/App/Router/ActionList.php
```diff
@@ -34,7 +34,7 @@ class ActionList
         'for', 'foreach', 'function', 'global', 'goto', 'if', 'implements', 'include', 'instanceof',
         'insteadof','interface', 'isset', 'list', 'namespace', 'new', 'or', 'print', 'private', 'protected',
         'public', 'require', 'return', 'static', 'switch', 'throw', 'trait', 'try', 'unset', 'use', 'var',
-        'while', 'xor',
+        'while', 'xor', 'void',
```
orlangur, line 37: @antonkril why do we even need such check instead of simple fallback to a `VoidAction` in case `Void` class is not available?

cscd98:
@okobchenko when can we expect this in a release? Thank you

okorshenko:
@orlangur thank you for your contribution. 
@craigcarnell this fix will be a part of 2.2.0 release

lucasvm:
Any fix on this for  2.1.9.? Im getting the same problem with Void word.

orlangur:
@lucasvm `2.1.x` is not going to become PHP 7.1 compatible, please stick to PHP 7.0 until you have a chance to upgrade Magento.

bellodaniel:
Hello, I'm using magento 2.2.2, and still get this error when running php bin/magento setup:di:compile 

`app@b585e061527e:~/html$ bin/magento setup:di:compile

Compilation was started.

Repositories code generation... 1/7 [====>-----------------------]  14% 4 secs 46.0 MiB

Fatal error: Cannot use 'Void' as class name as it is reserved in /var/www/html/vendor/magento/module-sales/Controller/Adminhtml/Order/Creditmemo/Void.php on line 10`

orlangur:
@bellodaniel please check if contents of this file in your instance correspond to changes in this PR, probably upgrade to 2.2.2 was unsuccessful somehow in your case, clearing `vendor` folder and running `composer install` should help.

PR #8610, opened 2017-02-20
Fix #8315: test setup/src/Magento/Setup/Test/Unit/Module/I18n/Dictionary/Writer/Csv/StdoTest.php crashes in debug mode

### LINKED ISSUE #8315: Error with StdoTest

orlangur:
It was crashing when running PHPUnit with `--debug` flag due to `STDOUT` closing in destructor of object under test.

I don't think this test is enforcing any useful behavior thus it would be better to simply remove it.

### DIFF
--- a/setup/src/Magento/Setup/Test/Unit/Module/I18n/Dictionary/Writer/Csv/StdoTest.php
+++ b/setup/src/Magento/Setup/Test/Unit/Module/I18n/Dictionary/Writer/Csv/StdoTest.php
@@ -5,25 +5,15 @@
  */
 namespace Magento\Setup\Test\Unit\Module\I18n\Dictionary\Writer\Csv;
 
+use Magento\Setup\Module\I18n\Dictionary\Writer\Csv\Stdo;
+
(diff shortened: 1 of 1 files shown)


okorshenko:
Has a fix for: magento/magento2#8315

okorshenko:
@orlangur thank you for your contribution!

PR #8611, opened 2017-02-20
Fix #8308: test setup/src/Magento/Setup/Test/Unit/Model/Cron/JobSetCacheTest.php crashes in debug mode

### LINKED ISSUE #8308: Fatal error with testSetCache
### Preconditions

1. magento: develop
2. PHP 7.0.15

### Steps to reproduce

1. Run `./vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist`

### Expected result

1. Tests to complete

### Actual result

1. Test do not complete exit code is 2

When running phpunit in debug mode as `./vendor/bin/phpunit --debug -c dev/tests/unit/phpunit.xml.dist setup/src/Magento/Setup/Test/Unit/Model/Cron/JobSetCacheTest.php` you will see

```
Fatal error: Method Symfony\Component\Console\Input\ArrayInput::__toString() must not throw an exception, caught ErrorException: preg_match() expects parameter 2 to be string, array given in /var/www/html/mage2-github/vendor/phpunit/phpunit/src/Framework/TestCase.php on line 0
```

This seems to be caused by `JobSetCacheTest::setCacheDataProvider` and the multi-level array used in the `$cacheEnable = new ArrayInput(['command' => 'cache:enable', 'types' => ['cache1']]);`

orlangur:
It was crashing when running PHPUnit with `--debug` flag due to object passed via data provider (which is never a good idea - there were serious troubles with such data providers littering Bamboo in the past; cannot be solved via implementing listener preventing such kind of data provider AFAIR).

### DIFF
--- a/setup/src/Magento/Setup/Test/Unit/Model/Cron/JobSetCacheTest.php
+++ b/setup/src/Magento/Setup/Test/Unit/Model/Cron/JobSetCacheTest.php
@@ -10,17 +10,21 @@ use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputDefinition;
 use Symfony\Component\Console\Input\InputArgument;
 
+/**
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
 class JobSetCacheTest extends \PHPUnit_Framework_TestCase
 {
     /**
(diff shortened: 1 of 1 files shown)


okorshenko:
@orlangur thank you for your contribution.

PR #8617, opened 2017-02-20
Remove Zend_Json from Customer module

dmanners:
In this PR I have updated the Customer module's auth pop-up to use the core Magento Json serilizer lib rather than Zend1' json class.

I have also added a new method to be used in the template and a new test to cover this new method.


PieterCappelle:
Keep up the good work @dmanners! :-)

okorshenko:
@dmanners thank you for removing Zend_Json from Customer module! Greate job!

PR #8623, opened 2017-02-20
Fix check for boolean product attributes

TKlement:
Based on that Issue [#6634](https://github.com/magento/magento2/issues/6634) here is a fix to add also boolean product attributes to display in frontend.

Magento2 handles translations with the Phrase instance. A product attribut with type boolean uses Phrase instance but the validation verify for strings.
This fix also includes the instance Phrase for the verification.

### DIFF
--- a/app/code/Magento/Catalog/Block/Product/View/Attributes.php
+++ b/app/code/Magento/Catalog/Block/Product/View/Attributes.php
@@ -12,6 +12,7 @@
 namespace Magento\Catalog\Block\Product\View;
 
 use Magento\Catalog\Model\Product;
+use Magento\Framework\Phrase;
 use Magento\Framework\Pricing\PriceCurrencyInterface;
 
 class Attributes extends \Magento\Framework\View\Element\Template
@@ -85,8 +86,8 @@ class Attributes extends \Magento\Framework\View\Element\Template
                 } elseif ($attribute->getFrontendInput() == 'price' && is_string($value)) {
                     $value = $this->priceCurrency->convertAndFormat($value);
                 }
-
-                if (is_string($value) && strlen($value)) {
+    
+                if ($value instanceof Phrase || (is_string($value) && strlen($value))) {
                     $data[$attribute->getAttributeCode()] = [
                         'label' => __($attribute->getStoreLabel()),
                         'value' => $value,


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8623) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=8623) before we can accept your contribution.<br/><hr/>**Timo Klement** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

app/code/Magento/Catalog/Block/Product/View/Attributes.php
```diff
@@ -85,8 +86,8 @@ public function getAdditionalData(array $excludeAttr = [])
                 } elseif ($attribute->getFrontendInput() == 'price' && is_string($value)) {
                     $value = $this->priceCurrency->convertAndFormat($value);
                 }
-
-                if (is_string($value) && strlen($value)) {
+    
+                if ($value instanceof Phrase || (is_string($value) && strlen($value))) {
```
orlangur, line 90: Sorry, from second look I have a little suggestion :) As `Phrase` accepts empty text, we need to check its length too.

So, it would be nice to add something like 
```
$value = $value instanceof Phrase ? (string)$value : $value;
```
prior to `if` instead of additional condition.
___
Would be nice to add a unit test for this case, but it's up to you, I believe.
___
And please link commit to your GitHub account: https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/ ;)

orlangur (changes requested):

app/code/Magento/Catalog/Block/Product/View/Attributes.php
maghamed: I don't see a big difference between current solution and the one you proposed.
It's not mandatory to convert Phrase into string to pass data into template

maghamed (approved):

app/code/Magento/Catalog/Block/Product/View/Attributes.php
orlangur: @maghamed, right part of `(is_string($value) && strlen($value))` is not checked in case of `Phrase` currently.

I didn't think deeper if this is a problem or not (like, maybe there is no need to check length at all).

ishakhsuvarov:
@TKlement Fix has now been delivered to the `develop` branch. Thanks for contributing!

app/code/Magento/Catalog/Block/Product/View/Attributes.php
orlangur: @ishakhsuvarov @maghamed so check for `strlen` could be simply removed?

onepack:
I applied the fix and it shows the yes/ no values now with succes on the frontend.
However, the color attribute that is of type single select also shows value "No" now when no color is selected for a product.

ventouris:
I have the community version 2.1.6 and still have this issue.

mudithae:
In Related to community version 2.1.6 I think real reason for this kind of behaviour is lies in the logic of fetching attributes. 
vendor/magento/module-eav/Model/ResourceModel/ReadHandler.php

`foreach ($connection->fetchAll($unionSelect) as $attributeValue) {
                if (isset($attributesMap[$attributeValue['attribute_id']])) {
                    $entityData[$attributesMap[$attributeValue['attribute_id']]] = $attributeValue['value'];
                } else {
                    $this->getLogger()->warning(
                        "Attempt to load value of nonexistent EAV attribute '{$attributeValue['attribute_id']}'
                        for entity type '$entityType'."
                    );
                }
            }`

I think $attributesMap array should be the one that need to be loop. If the result from all $unionSelect doesn't have any value for the $attributesMap element, then it should fetch the default value for the attribute.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
southerncomputer: This failure to check for empty string results in empty attributes displaying as N/A in product view versus suppressing lines that have no value. Please fix to 2.1-develop !

southerncomputer:
This causes Empty EAV attributes to display as: N/A rather than line being suppressed on 2.1.7 - it also breaks my dynamically loaded attributes (via xml call) when used with && to include length check!

Had to back-out this change on 2.1.7 to function as normal!

PR #8638, opened 2017-02-21
Make Magento compatible with composer ~1.3.0

### LINKED ISSUE #8635: Incompatibility with composer ~1.3.0 and setup:di:compile
### Preconditions

1. Install composer 1.3.0 or higher
2. Check out Magento 2.0.10

### Steps to reproduce

1. run composer update
2. run bin/magento setup:di:compile-multi-tenant -vvv

### Expected result

1. compilation runs fine

### Actual result

  [Magento\Framework\Exception\LocalizedException]
  Source class "\Magento\Catalog\Model\Product\OptionInterface" for "\Magento\Catalog\Model\Product\OptionInterfacePersistor" genera
  tion does not exist.

I've added the solution to this in https://magento.stackexchange.com/questions/147219/optioninterfacepersistor-error-when-compiling-magento-in-production-mode/161010#161010

I will create a pull request for you to fix this.

AntonEvers:
Fixes #8635

### DIFF
--- a/app/code/Magento/Captcha/etc/di.xml
+++ b/app/code/Magento/Captcha/etc/di.xml
@@ -38,6 +38,6 @@
         </arguments>
     </type>
     <type name="Magento\Checkout\Block\Cart\Sidebar">
-        <plugin name="login_captcha" type="\Magento\Captcha\Model\Cart\ConfigPlugin" sortOrder="50" />
+        <plugin name="login_captcha" type="Magento\Captcha\Model\Cart\ConfigPlugin" sortOrder="50" />
     </type>
 </config>
--- a/app/code/Magento/Catalog/etc/di.xml
+++ b/app/code/Magento/Catalog/etc/di.xml
@@ -397,7 +397,7 @@
         </arguments>
     </type>
     <type name="Magento\CatalogInventory\Model\Config\Backend\ShowOutOfStock">
-        <plugin name="showOutOfStockValueChanged" type="\Magento\Catalog\Model\Plugin\ShowOutOfStockConfig"/>
+        <plugin name="showOutOfStockValueChanged" type="Magento\Catalog\Model\Plugin\ShowOutOfStockConfig"/>
     </type>
     <virtualType name="Magento\Catalog\Block\Category\Widget\Link" type="Magento\Catalog\Block\Widget\Link">
         <arguments>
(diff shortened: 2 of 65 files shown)


ktomk:
It would be great to see a regression test looking for these values. DI itself does ltrim from top of my head, scanning over the files w/ xpath and starts-with could do it. Should also help extension authors to check their XMLs.

Small POC: https://gist.github.com/ktomk/a7690cc3353298eec609fdc1d76ad295

orlangur:
@ktomk, I remember leading backslash started to be trimmed at some point as such situation  produced inconsistent behavior before that. As to me it would be better to be strict about that and simply disallow leading backslash.

As this is a BC break it could be done in two phases: first deprecate the ability to use leading backslash and then remove in some future release.

ktomk:
True, it makes sense to first deprecate this, I also thought about extension authors.

IMHO there is nothing wrong to fix the `di.xml`s within the repo first. Looks like this commit might miss the tests, my example counts more files (86) than in the changeset (65). The example I linked runs against Magento 2.1.2 Community so does not catch upstream changes since then.

ktomk:
And not to miss: Thanks for the initiative here, I had lost why it's not compatible with the stable composer version. ~~~That to say, a short explanation why this is necessary would be great to read~~~ /e: it's in the [corresponding Magento Stackexchange answer](http://magento.stackexchange.com/questions/147219/optioninterfacepersistor-error-when-compiling-magento-in-production-mode/161010#161010).

orlangur:
Fixing this repo is an easiest part, there are even some existing tests reading all active `di.xml`, like https://github.com/magento/magento2/blob/develop/dev/tests/static/testsuite/Magento/Test/Integrity/Di/CompilerTest.php#L398 (not sure why it is skipped though).

And also if we bump composer version in `composer.json` it will test the leading backslash absence by design for all builds using DI pre-compilation (at least functional tests, don't remember about integration tests).

maghamed:
looks like we need to introduce static test (intergrity) which will cover these changes and prevent us from  new occurrences of leading backslashes appear.
@ajpevers  could you  please accompany these changes with a static test?

ktomk:
@maghamed If I may suggest to do these static checks already when validating the XML files with the schema, see #8743.

maghamed:
@ktomk  yes, this proposal sounds reasonable

maghamed:
@ajpevers  thanks for your contribution, I'll close current PR and continue processing #8743 which @ktomk  referred to as it contains your changes and accompanies them with test

orlangur:
@maghamed will it become Merged after #8743 merged? :) If not, I propose to reopen this one so they both become Merged when changes delivered to mainline.

ktomk:
@orlangur Yes, due to your request, #8743 has 0b243b8 (as first commit, see https://github.com/magento/magento2/pull/8743/commits). I rebased it on this PR's branch. ;)

orlangur:
@ktomk yeah, I really appreciate this, I'm asking if #8638 will remain `Closed` or become `Merged`.

ktomk:
Ah, okay. I think Github marks it as merged but now that you ask ... ;)

PR #8420, opened 2017-02-04
Refactor contact module

schmengler:
This PR contains:

- removal of Zend Framework 1 dependencies
- usage of result factories in controller
- show meaningful error messages to user if possible
- separated `sendEmail` and `validatedParams` methods for easier rewriting

     I intended to move it out of the controller altogether but it seems to be against what the team had in mind with the abstract controller

Why?

- In Magento 1 the contact module was a popular one to take as an example for how a simple module should work. Let's keep it aligned with current best practices
- Having coding validation and email sending all in the `execute` method makes extending it hard (as stated above, I'd move it out of the controller but did not want to take the PR too far)

PR #8641, opened 2017-02-21
makes $layerResolver protected in Category view

BlackIkeEagle:
When you *need* to overwrite this View controller but want to inherit
since most of the functionality is exactly the same you have access to
everything but the $layerResolver which is kind of annoying. Its also
pointless to need to copy the constructor to your child class just
because of this. So make $layerResolver protected.

Signed-off-by: BlackEagle <ike.devolder@gmail.com>

### DIFF
--- a/app/code/Magento/Catalog/Controller/Category/View.php
+++ b/app/code/Magento/Catalog/Controller/Category/View.php
@@ -62,7 +62,7 @@ class View extends \Magento\Framework\App\Action\Action
      *
      * @var Resolver
      */
-    private $layerResolver;
+    protected $layerResolver;
 
     /**
      * @var CategoryRepositoryInterface


orlangur:
> you have access to everything but the $layerResolver which is kind of annoying

Protected properties are just a bad smelling legacy.

> It's also pointless to need to copy the constructor to your child class just because of this

It is the lesser evil. Takes some 0.5s in modern IDEs.

See https://github.com/magento/magento2/pull/8420#discussion_r99948607 for details.

maghamed:
@BlackIkeEagle  Magento discourages usage of Inheritance as a mean for extension/customization of entities. 
So, ideally there shouldn't be inheritance at all among the Magento entities. And all existing examples of inheritable you can see in our codebase could be considered as Technical Debt.
Thus, if there is no Inheritance -> you no need the protected property. 

What you can do in this specific case:

if you need to overwrite/extend layer object being instantiated 
`$this->layerResolver->create(Resolver::CATALOG_LAYER_CATEGORY);`
you could provide own version of Magento\Catalog\Model\Layer\Resolver object trough DI into \Magento\Catalog\Controller\Category\View

current implementation of \Magento\Catalog\Controller\Category\View couldn't be considered as a good example of Controller's code, because there is some logic which doesn't belong to controller (initCategory).

But nevertheless, I believe that you could just manipulate with external dependencies which are injected using DI, and substitute them if needed on own versions. 
If you need several entities (read versions) of  \Magento\Catalog\Controller\Category\View but configured differently - you could achieve this creating VirtualType(-s) based on  \Magento\Catalog\Controller\Category\View.

BlackIkeEagle:
The change was made so we don't have to copy all code from the current Category View controller to our overriding controller. But we could do it with DI and have some duplication.

PR #8642, opened 2017-02-21
Update Save.php

josefbehr:
If quantity is changed using mass action for attribute changes, but no stock item exists for the product, saving the newly created stockItem in StockItemRepository->save($stockItem) returns without saving the stockItem, since $stockItem->getProductId() is null, due to it being added to the array with key '0' (see line 165). To fix this, the array key has been changed to "product_id", such that StockItemRepository successfully creates the stockItem.
Also, in my opinion, StockItemRepository->save() should not just return without notice, especially as it already encapsulates the functionality in a try/catch block. Throwing an exception would actually tell the user that something went wrong.

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Action/Attribute/Save.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Action/Attribute/Save.php
@@ -162,7 +162,7 @@ class Save extends \Magento\Catalog\Controller\Adminhtml\Product\Action\Attribut
                         $this->attributeHelper->getStoreWebsiteId($storeId)
                     );
                     if (!$stockItemDo->getProductId()) {
-                        $inventoryData[] = $productId;
+                        $inventoryData['product_id'] = $productId;
                     }
 
                     $stockItemId = $stockItemDo->getId();


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8642) <br/>All committers have signed the CLA.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Action/Attribute/Save.php
```diff
@@ -162,7 +162,7 @@ public function execute()
                         $this->attributeHelper->getStoreWebsiteId($storeId)
                     );
                     if (!$stockItemDo->getProductId()) {
-                        $inventoryData[] = $productId;
+                        $inventoryData['product_id'] = $productId;
```
orlangur, line 165: Hmm, looks like it never worked: https://github.com/magento/magento2/blob/c8b7223139366e8030b0b8ae43406495e46ec95e/lib/internal/Magento/Framework/Service/Data/AbstractObjectBuilder.php#L100

orlangur (approved):

maghamed (approved):

okorshenko:
@josefbehr thank you for your contribution! 
@orlangur thank you for review!

PR #8668, opened 2017-02-24
Add correct return type in order service

cmuench:
The method return $this. I corrected the doc-block.


okorshenko:
@cmuench thank you for your contribution!

PR #8694, opened 2017-02-27
Fix #8691

### LINKED ISSUE #8691: Language pack inheritance order is incorrect
Direct quote from the Magento 2 documentation:
(http://devdocs.magento.com/guides/v2.0/config-guide/cli/config-cli-subcommands-i18n.html)

> If necessary, you can specify several parent packages. The parent packages are applied on a first listed, first used basis.

And 

> In the preceding example:
> 
> language_package_one descends from en_au_package and en_au_package descends from en_ie_package
> language_package_two descends from en_ca_package and en_ca_package descends from en_us_package
> If the Magento application cannot find word or phrase in the en_GB package, it looks in other packages in following sequence:
> 
> parent-package-one/language_package_one
> <vendorname>/en_au_package
> <vendorname>/en_ie_package
> parent-package-two/language_package_two
> <vendorname>/en_ca_package
> <vendorname>/en_us_package
> Specifying all inheritances between the language packages might result in creating circular inheritance chains. Use Magento\Test\Integrity\App\Language\CircularDependencyTest test to locate and fix such chains.
> 

The documentation clearly states that 'package_one' should be loaded before 'package_two'; However if your 'package_one' name comes after 'package_two' alphabetically the order will not be respected.

### Preconditions
1. Magento 2.4-develop
2. PHP7.2

### Steps to reproduce
1. Create 2 language packs with package names "pack_a" & "pack_z"
2. Create a project language pack which uses the previously created packs.
3. Make sure you add "pack_z" as the first package to use then "pack_a"

```
<?xml version="1.0"?>
<language xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:App/Language/package.xsd">
   ...
    <use vendor="myvendor" package="pack_z"/>
    <use vendor="myvendor" package="pack_a"/>
</language>
```

### Expected result
1. Translations should be firstly fetched from "pack_z" then "pack_a"

### Actual result
1. Translations are firstly fetched from "pack_a" then from "pack_z"

koenner01:
Fix for https://github.com/magento/magento2/issues/8691


lib/internal/Magento/Framework/App/Language/Config.php
```diff
@@ -141,4 +141,15 @@ public function getUses()
...
         return $this->_data['use'];
     }
+    
+    /**
+     * Set sort order
+     *
+     * @param string|int $value
+     * @return void
+     */
+    public function setSortOrder($value)
```
maghamed, line 151: we can't add setter method to the Config object which suppose to be immutable by design.
So, after initialization it shouldn't be modified

maghamed (changes requested):

ishakhsuvarov:
Hi @koenner01 
I am closing this PR due to inactivity. Please reopen if you wish to continue at any time.
Thank you.

PR #8706, opened 2017-02-28
Remove Zend_Json from the unit test in the CustomerImportExport module

dmanners:

This PR removes the usage of Zend_Json in the CustomerImportExport module.

### DIFF
--- a/app/code/Magento/CustomerImportExport/Test/Unit/Model/ResourceModel/Import/CustomerComposite/DataTest.php
+++ b/app/code/Magento/CustomerImportExport/Test/Unit/Model/ResourceModel/Import/CustomerComposite/DataTest.php
@@ -105,7 +105,7 @@ class DataTest extends \PHPUnit_Framework_TestCase
             ->getMock();
         $jsonDecoderMock->expects($this->once())
             ->method('decode')
-            ->willReturn(\Zend_Json::decode($bunchData));
+            ->willReturn(json_decode($bunchData, true));
         $jsonHelper = $helper->getObject(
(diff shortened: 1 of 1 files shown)


orlangur (changes requested):
Cool! 👍 

But, it can be removed from all tests at once, I observe 38 occurrences of `Zend_Json` in `*Test.php`.

Just to decrease the overhead of multiple pull requests processing ;)

dmanners:
Hi @orlangur I have been working through module by module simply so that I can encapsulate the changes. I dont just want to do a find an replace on the full code base and then try to fix all the problems. For me smaller changes like this is actually how you get a monolith moving in the right direction.

orlangur:
@dmanners, ok, you can go your way, no offense.

> just want to do a find an replace on the full code base and then try to fix all the problems

That's exactly how I like to do it :) Most likely there would be no problem with tests, otherwise some of them could be left untouched (like this: https://github.com/orlangur/magento2/commit/77a6b51dd619786911d2cce564f0399367f2f301). Fix as much problems as possible all over the code + leave some tricky cases for further atomic implementation. The thing which would make me sad is fixing tests manually which can be done automatically and only once.

orlangur (approved):

dmanners:
@orlangur dont worry no offence taken. I just started this by stripping out Zend_Json overall places and then had to scale it back step by step. These things take time.

vrann (approved):

ishakhsuvarov:
@dmanners We have now merged your PR to the `develop` branch. Thank you for your contribution!

PR #8707, opened 2017-02-28
Remove zend json from email

dmanners:

This PR removes the usage of Zend_Json in the Email module.

### DIFF
--- a/app/code/Magento/Email/Test/Unit/Model/TemplateTest.php
+++ b/app/code/Magento/Email/Test/Unit/Model/TemplateTest.php
@@ -84,6 +84,11 @@ class TemplateTest extends \PHPUnit_Framework_TestCase
      */
     private $templateFactory;
 
+    /**
+     * @var \Magento\Framework\Serialize\Serializer\Json|\PHPUnit_Framework_MockObject_MockObject
+     */
+    private $serilizerMock;
+
     protected function setUp()
     {
         $this->context = $this->getMockBuilder(\Magento\Framework\Model\Context::class)
@@ -138,6 +143,8 @@ class TemplateTest extends \PHPUnit_Framework_TestCase
             ->setMethods(['create'])
             ->disableOriginalConstructor()
             ->getMock();
+
+        $this->serilizerMock = $this->getMockBuilder(\Magento\Framework\Serialize\Serializer\Json::class)->getMock();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


orlangur (changes requested):
Typo "serilizer", besides that LGTM 👍

app/code/Magento/Email/Test/Unit/Model/TemplateTest.php
```diff
@@ -84,6 +84,11 @@ class TemplateTest extends \PHPUnit_Framework_TestCase
      */
     private $templateFactory;
 
+    /**
+     * @var \Magento\Framework\Serialize\Serializer\Json|\PHPUnit_Framework_MockObject_MockObject
+     */
+    private $serilizerMock;
```
maghamed: @dmanners  could you please fix this type

maghamed (approved):

orlangur (approved):

vrann:
@dmanners Thank you for the contribution!

Ctucker9233:
@okorshenko Why Closed and not Merged?

orlangur:
@Ctucker9233 nice catch, there is typo in mainline currently due to not merged last commit of this PR: https://github.com/magento/magento2/blob/develop/app/code/Magento/Email/Test/Unit/Model/TemplateTest.php#L90

For PRs which did not become Merged automatically would be nice to have an automated check that all changes came to mainline (so that PR can be safely closed when we have all changes but some commit hashes are different).

Ctucker9233:
@orlangur Agreed. I just thought I'd ask because all previous comments and updates made it look like it was good to go and then it's suddenly closed with no explanation.

okorshenko:
@Ctucker9233  this PR is shown closed on GitHub but not merged, because the contributor did rebase of the branch and commit hashes changed. If we missed some commits from this PR, please, create additional PR to resolve the issue. Thank you

orlangur:
@okorshenko in fact, not :) This was the case in another PR but not here.

In https://github.com/magento/magento2/commits/develop/app/code/Magento/Email/Test/Unit/Model/TemplateTest.php we can see https://github.com/magento/magento2/commit/ce261ee0f554306528cd4763902c5c89710f70c3 has the same hash as in this PR.

> For PRs which did not become Merged automatically would be nice to have an automated check that all changes came to mainline (so that PR can be safely closed when we have all changes but some commit hashes are different).

This could sound as an overkill, simpler check could be - check if first commit from PR has the same hash in mainline. If not - rebase occured, if yes - need to check if all commits were added to mainline.

Additional PR created.

PR #8710, opened 2017-02-28
Add afterLoad-method to ArrayBackend

josefbehr:
ArrayBackend was missing the afterLoad method that explodes the saved string into an array again.

### DIFF
--- a/app/code/Magento/Eav/Model/Entity/Attribute/Backend/ArrayBackend.php
+++ b/app/code/Magento/Eav/Model/Entity/Attribute/Backend/ArrayBackend.php
@@ -29,6 +29,25 @@ class ArrayBackend extends \Magento\Eav\Model\Entity\Attribute\Backend\AbstractB
 
         return parent::beforeSave($object);
     }
+    
+    /**
+     * After Load Attribute Process
+     *
+     * @param \Magento\Framework\DataObject $object
+     * @return \Magento\Eav\Model\Entity\Attribute\Backend\AbstractBackend
+     */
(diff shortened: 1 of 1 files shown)


vrann:
@josefbehr Thank you for the contribution!

This is a major change in the Magento APIs, as you can see ин Travis Integration tests, the first one is failing because expected result type differs from what it received:
```
1) Magento\CatalogImportExport\Model\Import\ProductTest::testProductWithWrappedAdditionalAttributes
Array (...) does not match expected type "string".
/home/travis/build/magento/magento2/dev/tests/integration/testsuite/Magento/CatalogImportExport/Model/Import/ProductTest.php:1404
```
Same would happen for other extensions which relied on string before.

I'm closing this pull request, however, if this fix is needed for some important bug, please create a new one and describe:
- the reason for the fix - some case which is failing for you
- steps to reproduce
- expected vs actual result
This will allow justifying the importance of the use case versus the broken backward compatibility.

josefbehr:
@vrann 
Thanks for your review and reply.
The problem which prompted the pull request was an attribute with "backend" set to the ArrayBackend and a multiselect in the frontend, which did save the selected values as expected, but  did not show (on load after save) the selected values as actually selected, since apparently the uiComponent expects an array (as it should). The frontend-field was added to category_form.xml. I guess either the component would need to be changed to work with a string, or the backend and all that depends on the current implementation would need fixing.
Should I open another pull request or issue regarding the described problem, then?

vrann:
@josefbehr yes, please create another PR

PR #8711, opened 2017-02-28
Remove Zend_Json from the persistent module remember me status observer

dmanners:

This PR removes the usage of Zend_Json in the persistent module.


ishakhsuvarov:
@dmanners PR has now been merged to the `develop` branch. Thank you for contributing!

PR #8712, opened 2017-02-28
Removed uncommon Comment-Syntax

DavidLambauer:
The Syntax of /**#@+ isn't used usually. Maybe it's a remnant of an old merge conflict?

### DIFF
--- a/lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php
+++ b/lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php
@@ -21,7 +21,7 @@ use Magento\Framework\Phrase;
  */
 class PhpCookieManager implements CookieManagerInterface
 {
-    /**#@+
+    /**
      * Constants for Cookie manager.
      * RFC 2109 - Page 15
      * http://www.ietf.org/rfc/rfc6265.txt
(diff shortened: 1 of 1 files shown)


vrann:
This is excerpt from the Magento Coding Standards:

## 2.6. DocBlock Templates
If there is declaration of multiple consecutive elements of same type, the same contents of docblock may be relevant to all of them. In this case individual docblocks for those elements they may be replaced by a docblock template.
Docblock template consists of 2 docblock comments:
Starting comment is before first element of the group, distinguished using "#@+" and formatted as follows:
```
/**#@+
 *
 */
```
Ending comment is after the last element of the group, distinguished using "#@-" and formatted as follows:
```
/**#@-*/
```
For example, declaration of multiple class constants or attributes:
 Collapse source
```
class Mage_Core_Model_Layout extends Varien_Simplexml_Config
{
    /**#@+
     * Supported layout directives
     */
    const TYPE_BLOCK = 'block';
    const TYPE_CONTAINER = 'container';
    /**#@-*/
 
    /**#@+
     * Scheduled structure elements operations
     *
     * @var array
     */
    protected $_scheduledMoves = array();
    protected $_scheduledRemoves = array();
    /**#@-*/
```

DavidLambauer:
I see, thanks for the description!

PR #8714, opened 2017-02-28
Quote values in IN() predicate to avoid cast issues

xi-ao:
Values in IN() predicate must be quoted as it may lead to cast issues otherwise.

Experienced: Two SKUs composed of numbers, one with prefix, one without.
- **123456789** (imported)
- **123456789-old** (already in database)

When importing the first one, the query returns both because of MySQL casting:
`SELECT sku FROM catalog_product_entity WHERE sku IN (123456789)` returns
123456789
123456789-old

while `SELECT sku FROM catalog_product_entity WHERE sku IN ("123456789")` returns the correct response
123456789


maghamed (approved):
@xi-ao  good catch!

okorshenko:
@xi-ao thank you for your contribution

PR #8716, opened 2017-02-28
Improvement: Added Finally Block to Error Handling

DavidLambauer:
The Error Handling is related to the Try/Catch Block above. Due to the finally block, it is more clear that the error handling belongs to the Try/Catch statement. 

This Commit is just an improvement and doesn't change the behavior of the method.

### DIFF
--- a/app/code/Magento/Store/Controller/Store/SwitchAction.php
+++ b/app/code/Magento/Store/Controller/Store/SwitchAction.php
@@ -82,12 +82,12 @@ class SwitchAction extends Action
             $error = __('Requested store is inactive');
         } catch (NoSuchEntityException $e) {
             $error = __('Requested store is not found');
-        }
-
-        if (isset($error)) {
-            $this->messageManager->addError($error);
-            $this->getResponse()->setRedirect($this->_redirect->getRedirectUrl());
-            return;
+        } finally {
+            if (isset($error)) {
+                $this->messageManager->addError($error);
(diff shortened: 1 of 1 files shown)


orlangur (changes requested):
This will not work under HHVM: https://github.com/facebook/hhvm/issues/5162

See http://devdocs.magento.com/guides/v2.1/architecture/tech-stack.html
> Magento is compatible with but not supported for:
> HHVM 3.9 PHP interpreter

app/code/Magento/Store/Controller/Store/SwitchAction.php
```diff
@@ -82,12 +82,12 @@ public function execute()
...
-
-        if (isset($error)) {
-            $this->messageManager->addError($error);
-            $this->getResponse()->setRedirect($this->_redirect->getRedirectUrl());
-            return;
+        } finally {
+            if (isset($error)) {
+                $this->messageManager->addError($error);
+                $this->getResponse()->setRedirect($this->_redirect->getRedirectUrl());
+                return;
```
maghamed, line 89: not supported by HHVM https://3v4l.org/LZ0Mq
that's why it's implemented as a flag in Magento codebase

maghamed (changes requested):

maghamed:
agree with @orlangur , current implementation is more reliable taking HHVM support into account

PR #8723, opened 2017-02-28
Add active class to search form wrapper for more theming flexibility

andrewkett:
Currently when the search form is focused I can only theme the search label (because it has an active class) and the input element itself (because it is focused). I cannot currently theme the input wrapping .control div as there are no wrapping elements identifying that the form has been focused. 

Adding an active class to the wrapping .minisearch form allows us to theme any children elements when the search form is focused.

### DIFF
--- a/app/code/Magento/Search/view/frontend/web/form-mini.js
+++ b/app/code/Magento/Search/view/frontend/web/form-mini.js
@@ -122,6 +122,7 @@ define([
          * @param {Boolean} isActive
          */
         setActiveState: function (isActive) {
+            this.searchForm.toggleClass('active', isActive);
             this.searchLabel.toggleClass('active', isActive);
 
             if (this.isExpandable) {


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8723) <br/>All committers have signed the CLA.

maghamed (approved):

ishakhsuvarov:
@andrewkett We have now merged the proposed fix into the `develop` branch. Thank you for contribution!

Flamestyle:
Magento 2 Сlicking on the search box does not work on iPad, when I set the width: 0.
Theme Luma, module Search. I want the field to pop out when I click on the icon.

Example without a bug:

Click on the label element and change the position of the input element (the input element shifted). The click works.

Example with a bug:

It is necessary that when the label is clicked, the width of the input element changes (I set the input to width: 0, padding: 0, when the width of the click is set to width: 250px.) Result: the width change on the ipad does not work.
I tried to add a line that you wrote, but it crashes after the first open search form. Please help. What need add else?

Flamestyle:
@andrewkett, what do you think this problem is related to this script?

**Сlicking on the search box does not work on iPad, when I set the `width: 0` for `input`.**

Theme Luma, module Search. I want the field to pop out when I click on the icon.

**Example without a bug:**

Click on the `label` element and change the `position` of the `input` element (the `input` element shifted). The click works.

**Example with a bug:**

It is necessary that when the `label` is clicked, the `width` of the `input` element changes (I set the input to `width: 0`, `padding: 0`, when the width of the click is set to `width: 250px`.) Result: the width change on the ipad does not work.

Ctucker9233:
@magento-team Can this be backported to 2.2?

Issue #8566, opened 2017-02-15
Setting 'show_out_of_stock' to 'No' has no effect


koenner01:
Setting the value of 'Display Out of Stock Products' in the backend to 'No' does not have the effect desired. If you set this to 'No' you would expect that only products that are in stock are shown.

### Preconditions
1. PHP7.0
2. Checked in both MG2.1.3 CE and MG2.1.4 CE

### Steps to reproduce
1. Create a configurable product and set itself and all associated simples to out of stock
2. Set the value for 'Display Out of Stock Products' to 'No' in the store configuration
3. Go to a product list containing your configurable product

### Expected result
1. The configurable product should not be shown

### Actual result
1. The configurable product is shown

We traced this issue back to Magento\CatalogInventory\Helper\Stock.
In the function addIsInStockFilterToCollection a flag check is done on 'require_stock_items'. Because this flag is never set or used in Magento (wtf!), it causes the second parameter in addStockDataToCollection to always be false. This has the undesired effect that out of stock products are shown...

```
public function addIsInStockFilterToCollection($collection)
{
    $stockFlag = 'has_stock_status_filter';
    if (!$collection->hasFlag($stockFlag)) {
        $isShowOutOfStock = $this->scopeConfig->getValue(
            \Magento\CatalogInventory\Model\Configuration::XML_PATH_SHOW_OUT_OF_STOCK,
            \Magento\Store\Model\ScopeInterface::SCOPE_STORE
        );
        $resource = $this->getStockStatusResource();
        $resource->addStockDataToCollection(
            $collection,
            !$isShowOutOfStock && $collection->getFlag('require_stock_items')
        );
        $collection->setFlag($stockFlag, true);
    }
}
```

It is our opinion that the '&& $collection->getFlag('require_stock_items')' should be removed as it is never set or referenced in the entire Magento project. Removing this part also fixes the fact the out-of-stock products are shown

orlangur:
`git blame` shows it was introduced in https://github.com/magento/magento2/commit/c966dc171b2a18cea3ff36ad545d8810441d3a83 and this flag probably exists in another edition only (thus should be treated as always `true` here).

Corefix:
I can verify this is still a problem, it should properly just change && to ||.

I created a PR that seems to fix this for me: https://github.com/magento/magento2/pull/8736

koenner01:
[...]

PR #8736, opened 2017-03-01
Update Stock.php

Corefix:
Relates to issue: https://github.com/magento/magento2/issues/8566

A flag set to null will always overwrite the isShowOutOfStock variable. So you can't set the default value from the admin configuration.


vrann:
@Corefix Thank you for the contribution!

PR #8760, opened 2017-03-02
Fix issue #8675

renatocason:


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8760) <br/>All committers have signed the CLA.

ishakhsuvarov (approved):

ishakhsuvarov:
Hi @renatocason 
We have analyzed this PR, and can not accept it at the moment due to the backwards compatibility policy. We are going to work on a solution which would allow us to fix it in the future releases.

Thank you for your contribution.

ishakhsuvarov (commented):
In this case changing interfaces parent causes more changes, which have to be done (e. g. add set/get methods for extension attributes, which is currently impossible due to backwards compatibility policy)

Issue #7103, opened 2016-10-20
Filters can't filter ranged from 0 


Corefix:
### Preconditions

1. Community Version 2.1.2
2. PHP 7.0.10
3. Some products already in Magento with products with quantity 0 and products without
### Steps to reproduce

1. Go to Magento admin
2. Go to Products > Catalog
3. Select Filters
4. Enter in Quantity from 0 to 0
5. Press apply filters
### Expected result

1. I would expect to see all products with quantity of 0 to be displayed
### Actual result

1. All products are displayed

### Comment

I have traced the problem back to the following class:
`\Magento\Ui\Component\Filters\Type\Range`

and the method:
`applyFilterByType`

I'm not sure why the condition in it is how it is, and thats why I didn't want to just do a PR fixing this.

miakusha:
Hi @Corefix , thanks for report
We've created internal ticket MAGETWO-60044 to address this issue.

okorshenko:
@Corefix Thank you for reporting the issue
@giacmir thank you for fixing this issue in PR: magento/magento2#8776

PR #8776, opened 2017-03-03
Range filter doesn't works with 0  values in admin #7103

giacmir:
This fixes issue #7103 and refactors test for the `prepare` method of `\Magento\Ui\Component\Filters\Type\Range` class.

The old test wasn't actually checking anything. The test was still "green" even if in the meantime some methods changed name and with invalid data was passed to the dataprovider.

Now the test fails when something is wrong and there are some more data cases added to the dataprovider.


okorshenko:
@giacmir thank you for your contribution! Great job

ishakhsuvarov:
@giacmir Your fix has been delivered to the `develop` branch. Thank you for contribution!

PR #8778, opened 2017-03-03
magento/magento2: #8765

cavalier79:
M2.1.4 : Magento\Catalog\Model\ResourceModel\AbstractResource::getAttributeRawValue() returns false negative

- fixed adding a strong check validation on return value.

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
@@ -568,6 +568,6 @@ abstract class AbstractResource extends \Magento\Eav\Model\Entity\AbstractEntity
             $attributesData = $_data[1];
         }
 
-        return $attributesData!==false ? $attributesData : false;
+        return $attributesData === false ? false: $attributesData;
     }
 }
--- a/app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
@@ -568,6 +568,6 @@ abstract class AbstractResource extends \Magento\Eav\Model\Entity\AbstractEntity
             $attributesData = $_data[1];
         }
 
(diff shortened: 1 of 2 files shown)


app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
```diff
@@ -568,6 +568,6 @@ public function getAttributeRawValue($entityId, $attribute, $store)
             $attributesData = $_data[1];
         }
 
-        return $attributesData ? $attributesData : false;
+        return $attributesData!==false ? $attributesData : false;
```
orlangur: Please add spaces around binary operator.

Also, as to me this statement would be less confusing if you invert condition (`=== false` and swap ternary operator parts).

orlangur (changes requested):

app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
```diff
@@ -568,6 +568,6 @@ public function getAttributeRawValue($entityId, $attribute, $store)
             $attributesData = $_data[1];
         }
 
-        return $attributesData ? $attributesData : false;
+        return $attributesData === false ? false: $attributesData;
```
orlangur: Thanks! Just one more space in `false:` left to add :)

It would be really nice if you squash all changes into one commit and force push into the same branch, but I do not insist on this.

orlangur (changes requested):

app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
cavalier79: Fixed.. but i've still have some problem with squash :D

orlangur (approved):

maghamed (approved):

vrann:
@cavalier79 Thank you, accepted.

PR #8782, opened 2017-03-03
Fix method named class being interpreted as a class

AydinHassan:
`PhpScanner` incorrectly interprets a method named `class` (valid PHP) as an actual class.

In the future this code should probably be refactored to make sure it *is* actually a class definition rather than the current method of blacklisting any other conflicting token streams.

### DIFF
--- a/setup/src/Magento/Setup/Module/Di/Code/Scanner/PhpScanner.php
+++ b/setup/src/Magento/Setup/Module/Di/Code/Scanner/PhpScanner.php
@@ -237,7 +237,7 @@ class PhpScanner implements ScannerInterface
             }
 
             if (($tokens[$tokenIterator][0] == T_CLASS || $tokens[$tokenIterator][0] == T_INTERFACE)
-                && $tokens[$tokenIterator - 1][0] != T_DOUBLE_COLON
+                && $tokens[$tokenIterator - 1][0] != T_DOUBLE_COLON && $tokens[$tokenIterator - 2][0] != T_FUNCTION
             ) {
                 $classes = array_merge($classes, $this->_fetchClasses($namespace, $tokenIterator, $count, $tokens));
             }
--- a/setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/ClassesScannerTest.php
+++ b/setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/ClassesScannerTest.php
@@ -22,6 +22,6 @@ class ClassesScannerTest extends \PHPUnit_Framework_TestCase
         $pathToScan = str_replace('\\', '/', realpath(__DIR__ . '/../../') . '/_files/app/code/Magento/SomeModule');
         $actual = $this->model->getList($pathToScan);
         $this->assertTrue(is_array($actual));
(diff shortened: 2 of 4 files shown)


setup/src/Magento/Setup/Module/Di/Code/Scanner/PhpScanner.php
```diff
@@ -237,7 +237,7 @@ protected function _getDeclaredClasses($file)
             }
 
             if (($tokens[$tokenIterator][0] == T_CLASS || $tokens[$tokenIterator][0] == T_INTERFACE)
-                && $tokens[$tokenIterator - 1][0] != T_DOUBLE_COLON
+                && $tokens[$tokenIterator - 1][0] != T_DOUBLE_COLON && $tokens[$tokenIterator - 2][0] != T_FUNCTION
```
AydinHassan, line 240: `- 2` as the immediate previous token is whitespace.

AydinHassan:
Ah I forgot Magento still runs on PHP 5. Should I just skip the unit test on anything less that 7?

Edit: Skipped the test on PHP < 7.

AydinHassan:
Hmm - looks like a few other things can't deal with a method named `class` either. I will take a closer look when I get some more time.

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/ClassesScannerTest.php
```diff
@@ -22,6 +22,6 @@ public function testGetList()
         $pathToScan = str_replace('\\', '/', realpath(__DIR__ . '/../../') . '/_files/app/code/Magento/SomeModule');
         $actual = $this->model->getList($pathToScan);
         $this->assertTrue(is_array($actual));
-        $this->assertCount(5, $actual);
+        $this->assertCount(6, $actual);
```
kschroeder, line 25: I ran into this same issue (also because of the ::class thing) on #8965.  If it *is* the same thing then this test should remain with a count of 5.

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Scanner/PhpScannerTest.php
```diff
@@ -66,4 +66,19 @@ public function testCollectEntities()
 
         $this->assertEquals([], $this->_model->collectEntities($this->_testFiles));
     }
+
+    public function testMethodNamedClassDoesNotGetRecognisedAsAClass()
+    {
+        if (version_compare(PHP_VERSION, '7.0.0', '<')) {
```
orlangur, line 72: PHP < 7 is not supported anymore, just sync with `develop` branch.

orlangur (changes requested):

ishakhsuvarov:
@AydinHassan Please sync with the latest `develop` and check the review comments if you wish to continue work on this Pull Request.
Thanks.

AydinHassan:
Hi @ishakhsuvarov i'm not interested in working on this anymore, so it can just be closed.

ishakhsuvarov:
@AydinHassan Closing per your comment.
Thank you for collaboration.

PR #8796, opened 2017-03-06
Update AbstractModel.php

redelschaap:
PHP's function `in_array` returns `TRUE` when evaluating a needle `0` against a list of strings, which is a known issue. In my case I got an exception because a small integer was used to construct a new `DateTime` object. Calling the function in strict mode fixes this. See also http://stackoverflow.com/questions/13846769/php-in-array-0-value.


maghamed (approved):

maghamed:
@redelschaap  could you please make a pull request into develop branch , because for a while we don't process PRs into other branches

maghamed (changes requested):
please make Pull Request into Develop branch

redelschaap:
I've created a new PR to the develop branch. See #8991.

PR #8808, opened 2017-03-06
Change link text from "Report a Bug" to "Report an Issue"

Zifius:
Sorry to rise this again, but after discussion on contribution to Magento 2 at Mage Unconference it came to my mind that "Bug" sounds too early 200x-ish as not each issue/problem can be considered a bug, some are feature requests or a typo fix which is not really a "Bug" per definition. Besides GitHub, which is the place to report M2 issues, has its tab named as "Issues"

Thanks for considering!

### DIFF
--- a/app/code/Magento/Backend/i18n/en_US.csv
+++ b/app/code/Magento/Backend/i18n/en_US.csv
@@ -250,7 +250,7 @@ Minute,Minute
 "JavaScript may be disabled in your browser.","JavaScript may be disabled in your browser."
 "To use this website you must first enable JavaScript in your browser.","To use this website you must first enable JavaScript in your browser."
 "This is only a demo store. You can browse and place orders, but nothing will be processed.","This is only a demo store. You can browse and place orders, but nothing will be processed."
-"Report a Bug","Report a Bug"
+"Report an Issue","Report an Issue"
 "Store View:","Store View:"
 "Stores Configuration","Stores Configuration"
 "Please confirm scope switching. All data that hasn\'t been saved will be lost.","Please confirm scope switching. All data that hasn\'t been saved will be lost."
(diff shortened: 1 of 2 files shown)


orlangur:
I read hundreds of official responses in this repository with remark "GitHub is for bug reports, not for improvement suggestions, questions or discussions" :)

Good to know if this changed afterall and there is no need, for instance, to use Magento Forums for feature requests.

Zifius:
@orlangur I believe this did not change and Forums stand for collecting feature requests and detailed discussion of proposed changes

orlangur:
> some are **feature requests** or a typo fix which is not really a "Bug" per definition

> Forums stand for collecting **feature requests**

Then I don't understand. Typo IS really a bug [per definition](https://www.utest.com/articles/yes-typos-are-bugs-too).

"Issues" section in this repository is narrowed to bugs only as clearly stated by [Issue reporting guidelines](https://github.com/magento/magento2/wiki/Issue-reporting-guidelines):

> If you are sure that the problem you are experiencing is caused by a bug, file a new issue in a Github issue tracker following the recommendations below.

I'm not a native speaker of English though and "Report an Issue" could sound more natural/less early 200x-ish. So, I'm not against this change, just don't want to increase confusion.

Zifius:
Well, anyhow there are guidelines linked when reporting a new issue https://github.com/magento/magento2/issues/new as you mentioned so I believe the link title change won't create big confusion or misunderstanding

This PR is more about consistency in wording Bug vs Issue in GitHub's context of naming things. I'd prefer hearing feedback from core team about what they think about this change. Any word @maksek @okorshenko?

okorshenko:
@Zifius thank you for this PR. We started internal discussion with TechWriters, DevDocs and UX Teams about this change

okorshenko:
@Zifius thank you for your contribution. Your improvement has need approved!

Zifius:
Once again my pleasure @okorshenko, thanks for approving!

PR #8812, opened 2017-03-07
Fix count SQL on products sold collection

jameshalsall:


avoelkl:
Hi @jameshalsall!
Thanks for the PR!
Actually I was trying to reproduce what your PR solves, can you tell me where/in which cases you had an issue?
It seems the Ordered Quantity works correctly?

jameshalsall:
If you use `Magento\Reports\Model\ResourceModel\Product\Sold\Collection` to fetch the most sold products anywhere in user land code, it breaks on the count SQL as the table alias is incorrect.

Example code from my project:

```php
<?php

use Magento\Reports\Model\ResourceModel\Product\Sold\CollectionFactory as ProductsSoldCollectionFactory;

class MostPurchasedProductSkuProvider
{
    private $collectionFactory;
    private $storeManager;

    public function __construct(ProductsSoldCollectionFactory $productsSoldCollectionFactory)
    {
        $this->collectionFactory = $productsSoldCollectionFactory;
    }

    public function getSkus(int $limit, \DateTimeInterface $from, \DateTimeInterface $to): array
    {
        if ($limit < 1) {
            return [];
        }

        $collection = $this->collectionFactory->create();

        $collection->addOrderedQty($from->format('Y-m-d H:i:s'), $to->format('Y-m-d H:i:s'));
        $collection->setOrder('ordered_qty');
        $collection->setPageSize($limit);

        $collection->getSelect()->columns('order_items.sku');

        $skus = [];
            $skus[] = $item['sku'];
        }

        return $skus;
    }
}

avoelkl:
Thanks for the update @jameshalsall!
Mind to add a test as well, like in `Magento\Reports\Test\Unit\Model\ResourceModel\Report\Quote\CollectionTest` for example? That would be supercool. Let me know if you need help.

jameshalsall:
No problem. I'll add one later

jameshalsall:
I've added a test to cover this functionality now.

vrann:
@jameshalsall Thank you for the contribution!

PR #8822, opened 2017-03-08
Upgrade PHP CS Fixer to v2

keradus:
Hi.
I have noticed that you updated PHP CS Fixer to be under FriendsOfPHP vendor, which is great !

Let us put it forward - v1 has reached end of life.
This is my proposal how to update it - with all the upgrade needed to use new version.

Commit logs are meaningful, I did changes one by one.

Keep in mind that v2 brings a lot changes in rules and detects more and more violations - even when using same ruleset (like more rules in PSR2 "level"/"set" or same rule now is more powerful to cover more cases). For that, it's also worth to re-apply the rules on the codebase. To try to keep this PR atomic and easy see the changes, I apply the rules in separated PR - #8823 .
Feel free to close any of the PRs.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8822) <br/>All committers have signed the CLA.

.gitignore
```diff
@@ -25,6 +25,8 @@ atlassian*
 /.grunt
 /Gruntfile.js
 /package.json
+/.php_cs
+/.php_cs.cache
```
orlangur, line 29: I see the same line in composer's `gitignore`... No way to specify other path, like `var/.php_cs.cache` or something, right?

.php_cs.dist
```diff
@@ -0,0 +1,50 @@
+<?php
```
orlangur, line 1: Do this file has to be in the root? No other options like `.github` or `.travis` folders?

Then please do changes similar to https://github.com/magento/magento2/blame/develop/.htaccess#L225 in both `.htaccess` and `.htaccess.sample`.

.php_cs.dist
```diff
@@ -0,0 +1,50 @@
...
+ * vendor/bin/static-review.php hook:install dev/tools/Magento/Tools/StaticReview/pre-commit .git/hooks/pre-commit
+ */
+$finder = PhpCsFixer\Finder::create()
+    ->name('*.phtml')
+    ->exclude('dev/tests/functional/generated')
+    ->exclude('dev/tests/functional/var')
+    ->exclude('dev/tests/functional/vendor')
+    ->exclude('dev/tests/integration/tmp')
+    ->exclude('dev/tests/integration/var')
+    ->exclude('lib/internal/Cm')
```
orlangur, line 18: Please exclude `generated` folder also (`var/generation` was moved to `generated`).

Some of these paths are irrelevant already but it will not spoil anything and better to be addressed in separate PR or within this PR processing by Magento folks.

.php_cs.dist
```diff
@@ -0,0 +1,50 @@
...
+    ->setFinder($finder)
+    ->setRules([
+        '@PSR2' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'concat_space' => ['spacing' => 'one'],
+        'include' => true,
+        'new_with_braces' => true,
+        'no_empty_statement' => true,
+        'no_extra_consecutive_blank_lines' => ['use'],
+        'no_extra_consecutive_blank_lines' => true,
```
orlangur: Quick snippet:
```
php -r 'var_dump(["no" => ["use"], "no" => true]);'
array(1) {
  ["no"]=>
  bool(true)
}
```

Is there some generator for https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/UPGRADE.md#renamed-rules ? :) Probably should be just `['use']`.

.php_cs.dist
```diff
@@ -0,0 +1,50 @@
...
+
+return PhpCsFixer\Config::create()
+    ->setFinder($finder)
+    ->setRules([
+        '@PSR2' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'concat_space' => ['spacing' => 'one'],
+        'include' => true,
+        'new_with_braces' => true,
+        'no_empty_statement' => true,
```
orlangur, line 35: `join_function` -> `no_alias_functions` is missing (or omitted due to changed behavior?)

.php_cs.dist
```diff
@@ -0,0 +1,50 @@
...
+    ->exclude('pub/static')
+    ->exclude('setup/vendor')
+    ->exclude('var');
+
+return PhpCsFixer\Config::create()
+    ->setFinder($finder)
+    ->setRules([
+        '@PSR2' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'concat_space' => ['spacing' => 'one'],
```
orlangur, line 32: `operators_spaces` -> `binary_operator_spaces` is missing

.php_cs.dist
```diff
@@ -0,0 +1,50 @@
...
+        'no_multiline_whitespace_before_semicolons' => true,
+        'no_singleline_whitespace_before_semicolons' => true,
+        'no_trailing_comma_in_singleline_array' => true,
+        'no_unused_imports' => true,
+        'no_whitespace_in_blank_line' => true,
+        'object_operator_without_whitespace' => true,
+        'ordered_imports' => true,
+        'standardize_not_equals' => true,
+        'ternary_operator_spaces' => true,
+    ]);
```
orlangur, line 49: Except mentioned ruleset is identical 👍

orlangur (changes requested):

.gitignore
keradus: deafult location of `.php_cs.cache` is where `.php_cs.dist` is.

.php_cs.dist
keradus: I got confused is it supposed to be `var/generation` or `generated`.
Anyway, indeed. Out of the scope of this PR. it's up to you which paths should be excluded. I only updated the namespaces and so on. Not changing any logic.

.php_cs.dist
keradus: I didn't change the way where you keep the file. If you want to move it to non-default location and point the location every time you run the PHP CS Fixer from CLI - it's possible, it's up to you. Also, I doubt that this file should be used only on Travis env.
Anyway, change of where you store the file is outside of scope of this PR.

I updated `.htaccess, thanks

.php_cs.dist
keradus: It has wider scope

.php_cs.dist
keradus: `join_function` - works only with `join` function.
new fixer works with dozens of them.

.php_cs.dist
keradus: I first translated rules - originally rules `extra_empty_lines` and `remove_lines_between_uses` were both translated into `no_extra_consecutive_blank_lines`, but with different config.
Then sort lines in array. Indeed, should be only once.

.php_cs.dist
orlangur: No, I'm asking whether `php-cs-fixer` checks any other location besides current folder by default. If not - no need to move configuration file.

.php_cs.dist
orlangur: This one: https://github.com/magento/magento2/tree/develop/generated

Ok, no objections from my side, could be updated with any other paths later.

.php_cs.dist
orlangur: Anything undesired/unexpected? It's not an issue if now it fixes more cases than before :)

.php_cs.dist
orlangur: Aha, I see, and it is not configurable, ok.

.php_cs.dist
keradus: some thinks it change too much operators. eg in this repo i noticed you like to have space around `!`. Don't know if this is intended or not.
I do not want to define the ruleset you want to use or change paths. Just pure v1->v2 upgrade. If you want to update ruleset, go ahead ;) but it's your decision.

keradus:
Funny there are random Travis failures in this repo...

.php_cs.dist
orlangur: Ok :) Here we like consistency, not "zero spaces for concat, one space otherwise".

orlangur:
>  0 => 'Your password reset link has expired.'

Aha, didn't see this error before, probably this test relies on system time.

.php_cs.dist
keradus: Good for you ;)
Again, my purpose was not to figure out which ruleset you want to follow, that's part of the community ;) I just took what was in file

.php_cs.dist
orlangur: @keradus please clarify this item and we are done.

keradus: How config path is determined:
- path from CLI argument
- path of what you are fixing (`php-cs-fixer fix foo/` -> `foo/.php_cs(.dist)`)
- current working dir (`php-cs-fixer fix foo/` -> `.php_cs(.dist)`)

to work the best when one may want to run tool providing only subfolders, like
`php-cs-fixer fix foo/`
and then
`php-cs-fixer fix bar/`
the best DX is when the config file is on `cwd` level, so root of the project

orlangur: Ok :) Too many details, what I needed is if `php-cs-fixer` can look into some `.php_cs/` folder also.

orlangur:
LGTM 🥇 

I don't think anybody could do `php-cs-fixer` upgrade better.

Now let's wait for Fabien come here and say we are using [outdated](http://symfony.com/doc/current/contributing/community/releases.html#schedule) Symfony components and then I could say I saw everything in my life :)

orlangur (approved):

.php_cs.dist
keradus: no (except when manually passing the config file as parameter)

keradus:
Thank you.
Having approval, I signed CLA.

keradus:
So... could this be merged then ?

okorshenko:
@keradus thank you for your contribution!
@orlangur thank you for review!

PR #8824, opened 2017-03-08
Popup-Modal not closing on Safari/Windows

HansSchouten:
This pull request is a fix for issue: [#8761](https://github.com/magento/magento2/issues/8761)

The first thing I noticed when looking into the css was that the `-webkit-transition-duration` was missing. Therefore, the transition event was never triggered in Safari. The modal directly popped up on clicking the *remove item* button in the shopping cart.
After adding the transition duration, Safari did no longer show the modal at all. This was because the visibility transition acts on the parent, while the transition is set on its child. I divided the animation into a visibility transition at the parent `.modal-slide, .modal-popup` and a transform transition at its child `.modal-inner-wrap`. As a result, Safari users can now remove items from cart without the need to refresh a greyed out page.

While debugging I replicated the issue in JSFiddle:
http://jsfiddle.net/muL7Lpnx/
And wrote the following fix:
http://jsfiddle.net/muL7Lpnx/1/

After updating the Magento Less source files, I successfully tested the fix in the following browsers:
Safari 5.1.7
Chrome 56
IE 11
Microsoft Edge 38
Firefox 50.1.0
Opera 43

### DIFF
--- a/lib/web/css/source/components/_modals.less
+++ b/lib/web/css/source/components/_modals.less
@@ -48,9 +48,15 @@
     right: 0;
     top: 0;
     visibility: hidden;
+    opacity: 0;
+    -webkit-transition: visibility 0s .3s, opacity .3s ease;
+            transition: visibility 0s .3s, opacity .3s ease;
 
     &._show {
         visibility: visible;
+        opacity: 1;
+        -webkit-transition: opacity .3s ease;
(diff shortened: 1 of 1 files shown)


magento-team:
@Hansschouten Thank you for the contribution

PR #8831, opened 2017-03-09
Expose final price in the API

JulienPradet:
Hi,

In order to use the API in a SPA, the final price is needed.  
Especially for configurable products that have a price at 0.

Should I add a test for this ? If yes, where should it be ? I didn't find anything specifically related to the API yet.

### DIFF
--- a/app/code/Magento/Catalog/Api/Data/ProductInterface.php
+++ b/app/code/Magento/Catalog/Api/Data/ProductInterface.php
@@ -105,6 +105,13 @@ interface ProductInterface extends \Magento\Framework\Api\CustomAttributesDataIn
     public function getPrice();
 
     /**
+     * Product final price
+     *
+     * @return float|null
+     */
+    public function getFinalPrice();
+
+    /**
      * Set product price
      *
      * @param float $price


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8831) <br/>All committers have signed the CLA.

orlangur:
Hm :) I believe there should be a way to obtain final price using existing API methods. Such case seems too crucial to be overlooked.

JulienPradet:
That's what I thought too but couldn't find any. ^^

There are the group prices exposed in a seperate repository but couldn't find another repository related to prices or any final_price in the ProductRepositories.

However, I'd be super glad to be wrong and to be pointed to the correct endpoint :)

And the file I changed seems to be the converging point in the API :
* https://github.com/magento/magento2/blob/f1dc91b8fe71eb5fdc27c49542fc3aeb6bef8ac2/app/code/Magento/Catalog/Api/ProductRepositoryInterface.php#L72
* https://github.com/magento/magento2/blob/f1dc91b8fe71eb5fdc27c49542fc3aeb6bef8ac2/app/code/Magento/Catalog/Api/Data/ProductSearchResultsInterface.php#L17

orlangur:
Thanks for sharing your findings 👍 I would suggest something concrete if I had it offhand.

I suppose final price is hidden by some abstraction layers intentionally for the upgradability/extensibility purposes. Probably available as extension attribute or something.

Let's wait for somebody familiar with this area of Magento :) I'm quite interested in elegant answer as well.

app/code/Magento/Catalog/Api/Data/ProductInterface.php
```diff
@@ -105,6 +105,13 @@ public function setAttributeSetId($attributeSetId);
     public function getPrice();
 
     /**
+     * Product final price
+     *
+     * @return float|null
+     */
+    public function getFinalPrice();
```
vrann, line 112: Such change is not be allowed per our backward compatibility policy: there might be implementations of this interface which will be broken.

Instead, you can implement Extension Attribute and populate it with the final price when it's loaded. Extension attribute is backward compatible change and will be allowed.

See this as a reference: app/code/Magento/Bundle/etc/extension_attributes.xml
Documentation: http://devdocs.magento.com/guides/v2.0/extension-dev-guide/attributes.html

vrann (changes requested):

app/code/Magento/Catalog/Api/Data/ProductInterface.php
JulienPradet: Would you consider it if I make a new PR for it, or should I create my own module ?

vrann:
@JulienPradet @orlangur Here is more background on why final_price is not part of the Data\ProductInterface at the first place.  The final price is coupled to context/scope of the customer who retrieved the product. There could be different prices per different customer groups, catalog rules etc. Without the context, we can't rely on the final price value.

ProductInterface is returned from the Product Repository, which does not have any information on the context of the request, its job is to provide invariant information about the Product. 
Final price is something called "calculated field" -- this is not information permanently stored in the database, it is calculated depending on the input parameters of the service. For such cases, we introduce another service which can tell the final price based on the product and context information.

Having that, we cannot accept this PR. I would recommend creating separate service for the final price in your extension. I would be glad to review that and add to the core.

real34:
Hi @vrann, I understand why you don't want this backwards incompatible update however it will need changes in the code.

We have found leaks of this dependency at several locations in the core that would be solved by updating the `ProductInterface` too... otherwise requiring the removal of injected `ProductRepositoryInterface` in their constructors:

* https://github.com/magento/magento2/blob/b14d70e55dc3691c1134d255489f65dfa620e548/app/code/Magento/Checkout/Helper/Data.php#L270 the product here is in fact a `ProductInterface`, because the docblock of the `getProduct()` method is incorrect: a `\Magento\Catalog\Model\Product` is documented even though the product is the result of a `AbstractItem::getProduct()` call on an injected `Magento\Catalog\Api\ProductRepositoryInterface` (see https://github.com/magento/magento2/blob/b14d70e55dc3691c1134d255489f65dfa620e548/app/code/Magento/Quote/Model/Quote/Item/AbstractItem.php#L141)
* https://github.com/magento/magento2/blob/b14d70e55dc3691c1134d255489f65dfa620e548/app/code/Magento/ProductAlert/Model/Observer.php#L235 the product here also is the result of a `\Magento\Catalog\Api\ProductRepositoryInterface::getById()` call and should be treated as a `ProductInterface`

There may be other locations impacted by these kind of issues (docblocks are often misleading since a `\Magento\Catalog\Model\Product` could have been documented instead of a `ProductInterface`) I have not looked deeper.

Since in the core all implementations of a `\Magento\Catalog\Api\Data\ProductInterface` seems to have the `getFinalPrice()` method implemented it may end up being a better idea to update the interface even though it would break backward compatibility (but may be considered as a bugfix!).

PS: the "context" is already available anyway since a customer session is injected and used in https://github.com/magento/magento2/blob/b14d70e55dc3691c1134d255489f65dfa620e548/app/code/Magento/Catalog/Model/Product/Type/Price.php#L445

orlangur:
> Since in the core all implementations of a `\Magento\Catalog\Api\Data\ProductInterface` seems to have the `getFinalPrice()` method implemented it may end up being a better idea to update the interface

That's exactly how implementation SHALL NOT evolve with time :)

Api with some elaborated interfaces were introduced to reduce coupling between objects and for better concerns separation. Keeping in mind we have a huge legacy codebase, in most cases under the hood of new interfaces there are still old fat models, like the `Product` model with 35 injected dependencies in constructor.

So, there are still a lot of places in code relying on existing implementation rather than on contracts defined by interfaces. It does not mean that we should extend contracts so that bad code does not violate them. What it really means is that we need to eliminate all such violations and, for instance, do not call `getFinalPrice()` where `ProductInterface` is passed.

Thereby we have more freedom in next Magento releases as only a reasonable subset of `Product` methods reflecting its responsibility needs to be preserved. `getFinalPrice` can be removed as it is not a part of public interface or even the whole `Product` model implementation can be replaced with another one implementing just `ProductInterface` and no more than that.

And, of course, the fact that code follows contracts shall be enforced with static tests.

PR #8832, opened 2017-03-09
Replace Zend_Json from the Magento Quote module

dmanners:
Since Zend 1 is past EOL we should migrate from Zend_Json to the `\Magento\Framework\Serialize\Serializer\Json` class.


okorshenko:
@dmanners thank you for your contribution!

PR #8835, opened 2017-03-09
Replace Zend_Json from the Magento Review module

dmanners:

This PR removes the usage of Zend_Json in the Review module.


okorshenko:
@dmanners thank you for your contribution!

PR #8839, opened 2017-03-10
Rename to DataObjectTest

mfdj:
**This PR suggests two things**

1. Rename test to `DataObjectTest` because it tests `DataObject` class
2. use PSR-2 property-name for test subject

**Rationale**

1. I wanted a reference for how `DataObject` behaves and couldn't find a test. I was half finished writing my own test case before discovering that what I needed was `ObjectTest`? I think renaming it could help others.
2. PSR-2 coding style dictates that protected/private properties should not be prefixed with an  underscore


orlangur:
LGTM 👍 I'm surprised `Go To > Test` in PhpStorm cannot handle this.

Regarding leading underscore removal - please don't do it manually here and there, it would be more effective as automated change.

orlangur (approved):

mfdj:
@orlangur maybe I need to setup my phpstorm?

This is what I see on "go to > test" it prompts me to create a test?

screen shot 2017-03-10 at 9 37 02 am

**Update** 

Ok weird, it worked after reloading PhpStorm — I think it just need to reindex or something. Thanks for the tip 👍

orlangur:
@mfdj I mean I tried on mainline without your change and it didn't pick up wrongly named test properly.

okorshenko:
@mfdj thank you for your contribution!
@orlangur thank you for review

PR #8864, opened 2017-03-13
allows $this->getProduct to be null in product view block

BlackIkeEagle:
From the following code we can find out there is a chance the result of
getProduct in \Magento\Catalog\Block\Product\View can be null:

~~~ php
/**
 * Retrieve current product model
 *
 * @return \Magento\Catalog\Model\Product | null
 */
public function getProduct()
{
    if (!$this->_coreRegistry->registry('product') && $this->getProductId()) {
        $product = $this->productRepository->getById($this->getProductId());
        $this->_coreRegistry->register('product', $product);
    }
    return $this->_coreRegistry->registry('product');
}
~~~

In the case there is no product in the registry and the getProductId
method returns something that gets interpreted as false, the product is
not fetched and also noet set in the registry. As a result we get the
product from the registry but its not there so the output of
getProduct() is null.

When we use the output of getProduct() further in the code without
checking we will eventually have breakage.

Signed-off-by: BlackEagle <ike.devolder@gmail.com>

1. magento/magetno2#<issue_number>: Issue title
2. ...

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Block/Product/View.php
+++ b/app/code/Magento/Catalog/Block/Product/View.php
@@ -7,6 +7,8 @@ namespace Magento\Catalog\Block\Product;
 
 use Magento\Catalog\Api\ProductRepositoryInterface;
 use Magento\Catalog\Model\Category;
+use Magento\Catalog\Model\ProductFactory;
+use Magento\Framework\App\ObjectManager;
 
 /**
  * Product View block
@@ -63,6 +65,11 @@ class View extends AbstractProduct implements \Magento\Framework\DataObject\Iden
     protected $productRepository;
 
     /**
+     * @var \Magento\Catalog\Model\ProductFactory
+     */
+    private $productFactory;
+
+    /**
      * @param Context $context
      * @param \Magento\Framework\Url\EncoderInterface $urlEncoder
      * @param \Magento\Framework\Json\EncoderInterface $jsonEncoder
@@ -74,6 +81,7 @@ class View extends AbstractProduct implements \Magento\Framework\DataObject\Iden
      * @param ProductRepositoryInterface|\Magento\Framework\Pricing\PriceCurrencyInterface $productRepository
      * @param \Magento\Framework\Pricing\PriceCurrencyInterface $priceCurrency
      * @param array $data
+     * @param \Magento\Catalog\Model\ProductFactory $productFactory
      * @codingStandardsIgnoreStart
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
@@ -88,7 +96,8 @@ class View extends AbstractProduct implements \Magento\Framework\DataObject\Iden
         \Magento\Customer\Model\Session $customerSession,
         ProductRepositoryInterface $productRepository,
         \Magento\Framework\Pricing\PriceCurrencyInterface $priceCurrency,
-        array $data = []
+        array $data = [],
+        \Magento\Catalog\Model\ProductFactory $productFactory = null
     ) {
         $this->_productHelper = $productHelper;
         $this->urlEncoder = $urlEncoder;
@@ -99,6 +108,8 @@ class View extends AbstractProduct implements \Magento\Framework\DataObject\Iden
         $this->customerSession = $customerSession;
         $this->productRepository = $productRepository;
         $this->priceCurrency = $priceCurrency;
+        $this->productFactory = $productFactory
+            ?: ObjectManager::getInstance()->get(ProductFactory::class);
         parent::__construct(
             $context,
             $data
@@ -127,7 +138,7 @@ class View extends AbstractProduct implements \Magento\Framework\DataObject\Iden
     {
         $this->getLayout()->createBlock(\Magento\Catalog\Block\Breadcrumbs::class);
         $product = $this->getProduct();
-        if (!$product) {
+        if (!$product->getId()) {
             return parent::_prepareLayout();
         }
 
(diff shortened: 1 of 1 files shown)


orlangur:
`if (!empty($product))` is equivalent to `if ($product)` ;)

In which situation do we need `null` instead of product model at all? Is it not easier to register dummy product model instead?

BlackIkeEagle:
@orlangur creating a dummy would work I think

BlackIkeEagle:
@orlangur then only thing I don't want to do is to fill in the 35 constructor parameters manually. so that would mean I have to get the objectManager in that view which is nasty.

So maybe better to keep it simple and check if its empty?

orlangur:
> fill in the 35 constructor parameters manually

You don't have to do it :) Just use `ProductFactory`.

BlackIkeEagle:
@orlangur thanks, I'm currently not working on a magento project, but updated the code to use the ProductFactory instead of checking for empty product in several places

BlackIkeEagle:
I'm going to fix the tests later, have to do some other work first

orlangur:
@BlackIkeEagle, yeah, before investing efforts into tests fixing let's wait for an official review also.

BlackIkeEagle:
@maghamed Do you want me to fix the unittests or do you first have to review and come back to this?

app/code/Magento/Catalog/Block/Product/View.php
```diff
@@ -63,6 +63,11 @@ class View extends AbstractProduct implements \Magento\Framework\DataObject\Iden
     protected $productRepository;
 
     /**
+     * @var \Magento\Catalog\Model\ProductFactory
+     */
+    protected $_productFactory;
```
maghamed: please don't introduce protected access modifiers, use private one instead.
Also don't use _ prefix for attribute names, as this violates PSR naming

app/code/Magento/Catalog/Block/Product/View.php
```diff
@@ -88,6 +94,7 @@ public function __construct(
         \Magento\Customer\Model\Session $customerSession,
         ProductRepositoryInterface $productRepository,
         \Magento\Framework\Pricing\PriceCurrencyInterface $priceCurrency,
+        \Magento\Catalog\Model\ProductFactory $productFactory,
```
maghamed: new dependency should be added last one to the class , even after $data array

and should be optional, default value is null

this is all done to prevent backward incompatibility

maghamed (changes requested):

app/code/Magento/Catalog/Block/Product/View.php
BlackIkeEagle: @maghamed I will make the changes tomorrow. But in general you are ok with the changes as they are presented now?

BlackIkeEagle:
@maghamed could you confirm the style is as desired now?

app/code/Magento/Catalog/Block/Product/View.php
```diff
@@ -99,6 +106,7 @@ public function __construct(
         $this->customerSession = $customerSession;
         $this->productRepository = $productRepository;
         $this->priceCurrency = $priceCurrency;
+        $this->productFactory = $productFactory;
```
orlangur: You need to add something like
```
$this->productFactory = $productFactory ?: ObjectManager::getInstance()->get(ProductFactory::class);
```
here for BC.

app/code/Magento/Catalog/Block/Product/View.php
```diff
@@ -88,7 +94,8 @@ public function __construct(
         \Magento\Customer\Model\Session $customerSession,
         ProductRepositoryInterface $productRepository,
         \Magento\Framework\Pricing\PriceCurrencyInterface $priceCurrency,
-        array $data = []
+        array $data = [],
+        \Magento\Catalog\Model\ProductFactory $productFactory
```
orlangur: Please add default `null` value for BC.

app/code/Magento/Catalog/Block/Product/View.php
```diff
@@ -173,7 +181,12 @@ public function getProduct()
             $product = $this->productRepository->getById($this->getProductId());
             $this->_coreRegistry->register('product', $product);
         }
-        return $this->_coreRegistry->registry('product');
+        $product = $this->_coreRegistry->registry('product');
+        if (! $product) {
```
orlangur: Please remove space, it is not strictly forbidden yet but inconsistent with other Magento code.

Please avoid unnecessary access to registry.

Please cover this new logic with unit test if it's not too hard.

orlangur (changes requested):

app/code/Magento/Catalog/Block/Product/View.php
BlackIkeEagle: @orlangur sorry I forgot that

app/code/Magento/Catalog/Block/Product/View.php
BlackIkeEagle: @orlangur thanks I'll fix that

app/code/Magento/Catalog/Block/Product/View.php
BlackIkeEagle: @orlangur I'll check if its unnescessary and remove the space its just a personal preference for readability

BlackIkeEagle:
for the test I should at least add:

~~~ php
    public function testGetProductWithProductInRegistry()
    {
    }

    public function testGetProductWithoutProductInRegistryWithProductId()
    {
    }

    public function testGetProductWithoutProductInRegistryWithoutProductId()
    {
    }
~~~

I'll add these later, first have to do some other work

ishakhsuvarov:
@BlackIkeEagle Closing this PR due to inactivity.
Please reopen when you wish to continue with it.
Thanks!

PR #8883, opened 2017-03-14
Fix a typo

evgk:
### Description
Fix a typo in backend create-order functionality


vrann:
@evgk This might be backward incompatible as far as somebody might have already created own style for this element. I would say the importance of this change is lower than the possibility to break some design.

okorshenko:
@evgk  thank you for your contribution
@avoelkl thank you for processing this PR!

PR #8906, opened 2017-03-16
Correctly consider .html files in "template" path to for translation JSON generation.

Rane2k:
Certain static html files were not found. For example src/app/design/frontend/MyVendor/mytheme/MyVendor_MyModule/web/template
while files in
src/app/design/frontend/MyVendor/mytheme/MyVendor_MyModule/web/templates
were found.
To my knowledge, "template" is the correct naming for this path, not "templates".

This caused a problem when the translation JSON was generated, translatable strings from files in such directories were not considered.

### Description

This pull request adds the proper proper path /template, as well as leaving the /templates path in there to prevent BC breaks.

(Could not find the issue on github)

### DIFF
--- a/lib/internal/Magento/Framework/App/Utility/Files.php
+++ b/lib/internal/Magento/Framework/App/Utility/Files.php
@@ -810,14 +810,21 @@ class Files
                 }
             }
         }
+        
+         // Original paths, to preserve backwards compatibility:
+        $legacyThemePaths = $this->getThemePaths($area, $namespace . '_' . $module, '/web/templates');
+        // Correct paths:
         $themePaths = $this->getThemePaths($area, $namespace . '_' . $module, '/web/template');
+
         $files = self::getFiles(
             array_merge(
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8906) <br/>All committers have signed the CLA.

Rane2k:
This might actually be incorrect.

The confusion stems from the fact that the Magento_UI module has both directories:

src/vendor/magento/module-ui/view/frontend/web/template
src/vendor/magento/module-ui/view/frontend/web/templates

Both have content.

If one wants to override one of the templates under src/vendor/magento/module-ui/view/frontend/web/templates we are forced to use the wrong directory name in our theme as well.

I am unsure if this should be merged. It might be more prudent to fix the inconsistency in the Magento_Ui module.

vrann:
@Rane2k agree, there is no expectations on using "template" directory.

PR #8913, opened 2017-03-16
magento/magento2#3356: XML syntax to change block's template

nsergo:
Fixed setting block template with arguments node.

### Description
Fixed setting the template with arguments node and is prioritised over setTemplate action.

### Fixed Issues
1. magento/magento2#3356: XML syntax to change block's template

### Manual testing scenarios
1. Change a block's template with arguments node in layout XML

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/View/Layout/Generator/Block.php
+++ b/lib/internal/Magento/Framework/View/Layout/Generator/Block.php
@@ -211,12 +211,15 @@ class Block implements Layout\GeneratorInterface
 
         // create block
         $className = $attributes['class'];
-        $block = $this->createBlock($className, $elementName, [
-            'data' => $this->evaluateArguments($data['arguments'])
-        ]);
-        if (!empty($attributes['template'])) {
+        $arguments = $this->evaluateArguments($data['arguments']);
+        $block = $this->createBlock($className, $elementName, ['data' => $arguments]);
+
+        if (!empty($arguments['template'])) {
+            $block->setTemplate($arguments['template']);
+        } elseif (!empty($attributes['template'])) {
             $block->setTemplate($attributes['template']);
         }
+
         if (!empty($attributes['ttl'])) {
             $ttl = (int)$attributes['ttl'];
             $block->setTtl($ttl);


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8913) <br/>All committers have signed the CLA.

lib/internal/Magento/Framework/View/Layout/Generator/Block.php
```diff
@@ -211,12 +211,15 @@ protected function generateBlock(
...
         $className = $attributes['class'];
-        $block = $this->createBlock($className, $elementName, [
-            'data' => $this->evaluateArguments($data['arguments'])
-        ]);
-        if (!empty($attributes['template'])) {
+        $arguments = $this->evaluateArguments($data['arguments']);
+        $block = $this->createBlock($className, $elementName, ['data' => $arguments]);
+
+        if (!empty($arguments['template'])) {
+            $block->setTemplate($arguments['template']);
```
okorshenko, line 218: the template file will be set from arguments by this logic: https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/View/Element/Template.php#L167 called from line 215
`$block = $this->createBlock($className, $elementName, ['data' => $arguments]);`

The issue is that it will be overridden by https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/View/Layout/Generator/Block.php#L218 if the template file was configured in the attributes.

Provided change fixes the issue. But by this fix we will set template file 2 times. 

On the other hand, by this PR we are changing priority of configuration: attributes vs arguments which is backward incompatible change. 

The correct fix for this issue should be the following:
1. add optional attribute to blockReferenceType https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/View/Layout/etc/elements.xsd#L317 . By this we will allow template overriding from other modules:
`<referenceBlock name="block.name" template="my/custom/template">`
Having template attribute in  referenceBlock node will eliminate ambiguity of block configuration. 
2. Update Layout configuration processor to support new attribute
3. Configuration of block template via arguments will be treated as a misuse of framework
4. Action setTempalte is not recommended for this case

So the correct fit of this issue is a little bit more complicated

Thank you!

okorshenko (changes requested):

okorshenko:
@nsergo thank you for your contribution. We are closing this PR for now since it requires significant refactoring/changes

PR #8923, opened 2017-03-17
Check of null result value in swatch-renderer.js

aholovan:
Check of null result value in swatch-renderer.js

### Description
There is an error when product has more than one kind of swatches:
Uncaught TypeError: Cannot read property 'oldPrice' of undefined - swatch-renderer.js:843

### Manual testing scenarios
1) Set Special Price for some Simple of the Configurable product which has more than one kind of swatches, for example SKU: WJ03
2) Go to PDP (../augusta-pullover-jacket.html)
3) Select some Color and look to the console - there You will be able to see the error. This is because 'result' is null before all of the swatches are selected.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
+++ b/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
@@ -840,13 +840,13 @@ define([
                 }
             );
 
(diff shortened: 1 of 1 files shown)


okorshenko:
@aholovan thank you for your contribution!

PR #8930, opened 2017-03-17
Improved check when CategoryProcessor attempts to create a new category

ccasciotti:
Improved check perfomed by CategoryProcessor when it checks if it has to create a new category or not. 

### Description

The original check didn't consider that a category must be supplied with uppercase or lowecase characters. In this case, if you're trying to import a category named Foo, and a category named FOO, the system will create both categories but fails when attempts to create URL keys because the first url already exists causing an exception. Checking path existence by lowercase-ing it will avoid duplicate url keys creation attempts. I've also added a method to clear failed categories between bunches import, and updated related unit test file. Lastly, i've fixed a typo in CategoryProcessor's PHPDoc.

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product.php
@@ -1529,6 +1529,9 @@ class Product extends \Magento\ImportExport\Model\Import\Entity\AbstractEntity
             $existingImages = $this->getExistingImages($bunch);
 
             foreach ($bunch as $rowNum => $rowData) {
+                // reset category processor's failed categories array
+                $this->categoryProcessor->clearFailedCategories();
+
                 if (!$this->validateRow($rowData, $rowNum)) {
                     continue;
                 }
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
@@ -79,7 +79,10 @@ class CategoryProcessor
                     for ($i = 1; $i < $pathSize; $i++) {
                         $path[] = $collection->getItemById((int)$structure[$i])->getName();
                     }
-                    $index = implode(self::DELIMITER_CATEGORY, $path);
+                    /** @var string $index */
+                    $index = $this->standardizeString(
+                        implode(self::DELIMITER_CATEGORY, $path)
+                    );
                     $this->categories[$index] = $category->getId();
                 }
             }
@@ -123,13 +126,16 @@ class CategoryProcessor
      */
     protected function upsertCategory($categoryPath)
     {
-        if (!isset($this->categories[$categoryPath])) {
+        /** @var string $index */
+        $index = $this->standardizeString($categoryPath);
+
+        if (!isset($this->categories[$index])) {
             $pathParts = explode(self::DELIMITER_CATEGORY, $categoryPath);
             $parentId = \Magento\Catalog\Model\Category::TREE_ROOT_ID;
             $path = '';
 
             foreach ($pathParts as $pathPart) {
-                $path .= $pathPart;
+                $path .= $this->standardizeString($pathPart);
                 if (!isset($this->categories[$path])) {
                     $this->categories[$path] = $this->createCategory($pathPart, $parentId);
                 }
@@ -138,7 +144,7 @@ class CategoryProcessor
             }
         }
 
-        return $this->categories[$categoryPath];
+        return $this->categories[$index];
     }
 
     /**
@@ -171,7 +177,7 @@ class CategoryProcessor
      * @param string $category
      * @param \Magento\Framework\Exception\AlreadyExistsException $exception
      *
-     * @return array
+     * @return $this
      */
     private function addFailedCategory($category, $exception)
     {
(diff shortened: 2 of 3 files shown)


app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
```diff
@@ -204,4 +221,21 @@ public function getCategoryById($categoryId)
...
         return isset($this->categoriesCache[$categoryId]) ? $this->categoriesCache[$categoryId] : null;
     }
+
+    /**
+     * Standardize a string by lowercase-ing it
+     *
+     * @param string $string
+     * @return string
+     */
+    protected function standardizeString(string $string)
```
tinogo: Can you remove the "string" scalar type hint, please? :)
This is only possible with PHP 7+ and Magento's minimum supported PHP version is 5.6.*.

ccasciotti:
@tinogo i've removed type hint :)

app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
```diff
@@ -204,4 +221,21 @@ public function getCategoryById($categoryId)
...
+    {
+        if (function_exists('mb_strtolower')) {
```
ishakhsuvarov: I believe `composer.json` for Magento already requires `ext-mbstring` so you may remove this check.

app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
ishakhsuvarov: Removing the `function_exists` check would eliminate the need in the additional method.

ishakhsuvarov (changes requested):

app/code/Magento/CatalogImportExport/Model/Import/Product.php
```diff
@@ -1529,6 +1529,9 @@ protected function _saveProducts()
             $existingImages = $this->getExistingImages($bunch);
 
             foreach ($bunch as $rowNum => $rowData) {
+                // reset category processor's failed categories array
+                $this->categoryProcessor->clearFailedCategories();
```
ishakhsuvarov, line 1533: Could you please explain this change?

app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
```diff
@@ -204,4 +221,19 @@ public function getCategoryById($categoryId)
...
+
+    /**
+     * Standardize a string by lowercase-ing it
+     *
+     * @param string $string
+     * @return string
+     */
+    protected function standardizeString($string)
+    {
+        if (function_exists('mb_strtolower')) {
```
ishakhsuvarov: I believe `composer.json` for Magento already requires `ext-mbstring` so you may remove this check.

ccasciotti: you're right, i've removed the check

app/code/Magento/CatalogImportExport/Model/Import/Product.php
ccasciotti: this change clears failed categories' array for every bunch, avoiding to log the same failed categories every time a bunch is imported. See `app/code/Magento/CatalogImportExport/Model/Import/Product.php::processRowCategories` that is called for every bunch's import called in the same file at line 1603

app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
ccasciotti: i've removed the check

app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
```diff
@@ -204,4 +221,15 @@ public function getCategoryById($categoryId)
...
         return isset($this->categoriesCache[$categoryId]) ? $this->categoriesCache[$categoryId] : null;
     }
+
+    /**
+     * Standardize a string by lowercase-ing it
+     *
+     * @param string $string
+     * @return string
+     */
+    protected function standardizeString($string)
```
ishakhsuvarov: Do you still need a method which just calls `strtolower`? I guess it is redundant now

ishakhsuvarov (changes requested):

app/code/Magento/CatalogImportExport/Model/Import/Product/CategoryProcessor.php
ccasciotti: Yes but if in the future we have to change it with more complex checks, we just have to change the function's code instead of changing all the pieces of code where appears `mb_strtolower`. It's more efficient i think.

ishakhsuvarov: Please, at least make it private and provide a reason behind this method in docblock. Thanks.

ccasciotti: ok, done.

ishakhsuvarov (approved):

okorshenko:
@ccasciotti thank you for your contribution!

PR #8950, opened 2017-03-20
Remove deadcode

jipjop:
### Description

Removed unused dependencies from several classes. Got the warnings from the PHP CS Fixer.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Braintree/Gateway/Response/CardDetailsHandler.php
+++ b/app/code/Magento/Braintree/Gateway/Response/CardDetailsHandler.php
@@ -5,7 +5,6 @@
  */
 namespace Magento\Braintree\Gateway\Response;
 
-use \Braintree\Transaction;
 use Magento\Braintree\Gateway\Config\Config;
 use Magento\Payment\Gateway\Helper\ContextHelper;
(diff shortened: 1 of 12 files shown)


orlangur:
Such kind of PR is useless as we can remove such occurrences all over the code automatically. https://github.com/magento/magento2/pull/8950/commits/3fc476be306d8d2b1ff0ec379dcd33b708946180 looks like you do it manually.

The crux is that we don't enforce that no such new code appear currently. I believe we shall implement static test enforcing `php-cs-fixer` rules first and then do automatic fix of the whole codebase only once at a suitable moment.

ishakhsuvarov:
@orlangur I totally agree, that we should remove automatically and enforce with the test, however:
While such occurrences exist there is no harm removing those in any possible way.

ishakhsuvarov (approved):

vrann:
@jipjop Thank you for the contribution!

orlangur:
@ishakhsuvarov I was mostly scared by manual nature of these changes :) It's ok to run `unused_use` fixer fully or partially but not safe to do same changes manually. Hope we have enough automation for changed files.

PR #8953, opened 2017-03-21
Log level for caught exception

flancer64:
"Broken reference" exception is logged every time when customer goes to /checkout/ page (in my case). There are some other [reports](http://magento.stackexchange.com/questions/133618/got-error-critical-broken-reference-no-element-found-with-id-checkout-header) about this behavior. Please decrease log level for caught exceptions or don't catch them if these are really critical cases.


flancer64:
Hello, guys! I am impressed!! This PR is 3 days old only and is already processed. Thanks and good luck, @magento-team & @vrann (y)

ishakhsuvarov:
@flancer64 Thanks for contributing 👍

PR #8965, opened 2017-03-21
Reduce calls to SplFileInfo::realpath() in the Magento\Setup\Module\Di\Code\Reader\ClassesScanner class

kschroeder:
Reduces calls to SplFileInfo::realpath() in the Magento\Setup\Module\Di\Code\Reader\ClassesScanner class.

### Description
I have had issues with fairly slow setup:di:compile calls.  In the course of my investigation, I ran the XDebug profiler and found that there were several million calls to SplFileInfo::realpath().  This change reduced my compile time from 9-10 minutes to 3 minutes.

[update]

I have been able to find several 10's of seconds (100+ seconds on my test box) of efficiency to be gained by creating a single extraction class that looks only for the first namespace/class match.

[update]

The single extraction class has been changed so that it now scans the entire PHP source file to find unconventionally defined classes as well

### Manual testing scenarios
This test simply reduces the number of system calls on the filesystem.  There is a functional change, not a logical change.

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8965) <br/>All committers have signed the CLA.

piotrkwiecinski:
@kschroeder nice one :) I've tested in on dev setup and for me it went down from 7 to 4 min and on CI server from 2:30 to 46 sec. Much appreciated you've saved quite a few CPU cycles :).

kschroeder:
@piotrkwiecinski Sweet, I was hoping it wasn't just me!  :-)

schmengler:
Nice, verifiable PR with huge benefits! Unfortunately your change in `ClassesScanner.php` triggered the cyclomatic complexity check in the static test suite. This is not really a problem with your change, but rather with the original method. Let's take this as a chance to refactor the method a bit more. A simple **extract method** in the loop should do it.

kschroeder:
@schmengler I updated the PR with a new commit that separates the extract() method.  If this one doesn't pass I'll separate the $classNames loop.

schmengler:
```
--------------------------------------------------------------------------------
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
--------------------------------------------------------------------------------
 59 | ERROR | Missing function doc comment
--------------------------------------------------------------------------------
```

PieterCappelle:
The best community ever. Great PR.

kschroeder:
Note that I found another 2 minutes or so I could squeeze out of it, which is where those new commits are coming from.  However, I found an edge case that is breaking it a little bit so it is not ready to be merged yet.

AydinHassan:
Nice - these new improvements will probably fix #8782 also.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
```diff
@@ -42,12 +43,8 @@ protected function extract()
...
         foreach ($tokens as $key => $token) {
-            if (!is_array($token)) {
-                if ($token == '{') {
-                    $paramNestingLevel++;
-                } else if ($token == '}') {
-                    $paramNestingLevel--;
-                }
+            if ($token == '{') {
+                return $classes;
             }
```
schmengler: If I understand that code correctly, it will break at least the following special cases:

- `namespace X { }` syntax
- multiple classes per file

While both do not happen in the core, we have no control over third party code. Keep in mind that the Magento DI framework can also used to integrate third party libraries that are not Magento modules.

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/FileClassScannerTest.php
```diff
@@ -117,42 +117,15 @@ public function test()
...
-    {
-        return self::class;
-    }
-
-}
-PHP
-        );
-        /* @var $scanner FileClassScanner */
-
+        $filename = __DIR__ . '/../../../../../../../../../../app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/AbstractEav.php';
```
schmengler: what is the weird exception case with that file?

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -16,11 +16,25 @@ class ClassesScanner implements ClassesScannerInterface
...
+     */
+
+    protected $directoryList;
+
+    /**
+     * @param DirectoryList $directoryList
      * @param array $excludePatterns
      */
-    public function __construct(array $excludePatterns = [])
+    public function __construct(DirectoryList $directoryList, array $excludePatterns = [])
```
schmengler: To increase the chance of getting this released as soon as possible, please make new constructor arguments optional so that there is no backwards incompatibility. Then if it is null, use `ObjectManager::getInstance()->get()` to retrieve the object (yes, that's the recommended  practice)

schmengler: Nevermind, you already removed it :)

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -1,12 +1,12 @@
 <?php
 /**
- * Copyright © 2013-2017 Magento, Inc. All rights reserved.
+ * Copyright © 2016 Magento. All rights reserved.
```
schmengler: the **** copyright header is always the current year...

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -43,7 +49,14 @@ public function addExcludePatterns(array $excludePatterns)
      */
     public function getList($path)
     {
+
         $realPath = realpath($path);
+        $isGeneration = strpos($realPath, DIRECTORY_SEPARATOR . 'generation' . DIRECTORY_SEPARATOR) !== false;
```
schmengler: Is it possible to restrict that check to the relative path below the Magento root. Just in case, somebody installs Magento below a directory called "generation"? It's unlikely but you never know...

schmengler (changes requested):

schmengler:
Thank you for the update! This is neat, but I see some problems with potential edge cases. Please also check the review comments marked as "outdated", I just commented them in previous commits but they are still valid.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: heh, I need to be better about my copy and paste, will change

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: I'm working on refactoring this part of the code, there is some interaction with some of the tests somehow that did not like this.  I had intended to use DirectoryList class to get the full path but introducing that dependency broke lots of other tests so I opted to just do a string match.  I will make the check more intentional.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: I wasn't aware of that practice, I have been trying like the plague to avoid using the ObjectManager like that.

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/FileClassScannerTest.php
kschroeder: It is similar to #8782.  The class keyword has the same token number as the class definition keyword.  That means that the code used to determine the class name has to understand the different contexts.  I had tried to write the test so that I could mock the code but I couldn't get it to replicate.  So I just referenced the one file that was breaking.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
kschroeder: Yeah, this code, which is slightly a hack to deal with the ::class issue, expects that the code people integrate follows the most basic of Magento coding standards.  I was also hoping that I wouldn't have to iterate over the entire file.  But if this is an important enough edge case (and I will presume it is unless told otherwise) then I will scan the whole file.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -43,7 +80,16 @@ public function addExcludePatterns(array $excludePatterns)
      */
     public function getList($path)
     {
+
         $realPath = realpath($path);
+        $isGeneration = $this->isGeneration($realPath);
+
+        // Generation folders should not have their results cached since they may actually change during compile
+        if (!$isGeneration) {
+            if (isset($this->fileResults[$realPath])) {
```
piotrkwiecinski: Instead of 2 nested ifs I would be better to  have them as !$isGenerated && isset

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/FileClassScannerTest.php
```diff
@@ -0,0 +1,244 @@
...
+        /* @var $scanner FileClassScanner */
+
+        $result = $scanner->getClassNames();
+
+        self::assertCount(2, $result);
+        self::assertContains('This\Is\My\Ns\ThisIsMyTest', $result);
+        self::assertContains('This\Is\My\Ns\ThisIsForBreaking', $result);
+    }
+
+    public function testBracketedNamespacesAndClasses()
```
piotrkwiecinski, line 163: @kschroeder what do you think about extracting classes to data provider(s) here? It would be more flexible to add more cases in future. I think you could make these tests cleaner this way.

kschroeder: Perhaps, but I was following the output of the original code so-as to minimize any downstream pain.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: Yes.  Pushing soon.

kschroeder:
@schmengler @piotrkwiecinski I keep triggering the cyclomatic complexity warning for PHPMD.  Because of the recursive and iterative nature of parsing tokens (because of the inconsistent representation in the `token_get_all()` array), I'm hesitant to make it all perfect because those method calls do add up.  I haven't measured it recently but at one point there was a method called 55 million times.  With 55 million calls micro-optimizations do add up.  So I would like a little guidance on what you would prefer: reduce the individual method complexity or notate MD to ignore CC?

schmengler:
> So I would like a little guidance on what you would prefer: reduce the individual method complexity or notate MD to ignore CC?

If reducing the method complexity has performance impact, I'd say go for optimized performance in this case and mark the class as ignored.

davidalger:
> If reducing the method complexity has performance impact, I'd say go for optimized performance in this case and mark the class as ignored.

I would agree, but include a comment indicating why it's ignored, so it doesn't eventually get refactored by some random dude trying to get rid of all the cyclomatic complexity warnings. ;)

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
schmengler: Don't get me wrong, continue avoiding it. And please still add any dependencies to the constructor (i.e. DirectoryList), just make them optional as `DirectoryList $directoryList = null` and fall back to the object manager. This way it's fully backwards compatible - and the fallback can be removed later.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -43,7 +88,14 @@ public function addExcludePatterns(array $excludePatterns)
...
     {
+
         $realPath = realpath($path);
+        $isGeneration = $this->isGeneration($realPath);
+
+        // Generation folders should not have their results cached since they may actually change during compile
+        if (!$isGeneration && isset($this->fileResults[$realPath])) {
+            return $this->fileResults[$realPath];
+        }
         if (!(bool)$realPath) {
```
orlangur, line 71: This condition is really confusing.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -52,46 +104,66 @@ public function getList($path)
...
+            $classes = array_merge($classes, $classNames);
         }
         return $classes;
     }
 
+    protected function includeClasses(array $classNames, $fileItemPath)
+    {
+        foreach ($classNames as $className) {
+            if (!class_exists($className)) {
+                require_once $fileItemPath;
```
orlangur: Why do we need `require_once` here and not `require` since we follow PSR-2?

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * Warnings are suppressed for this method due to a micro-optimization that only really shows up when this logic
+     * is called several millions of times, which can happen quite easily with even moderately sized codebases.
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD.NPathComplexity)
+     * @return array
+     */
+
+    protected function extract()
```
orlangur: Better use PHP Parser.

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/ClassesScannerTest.php
```diff
@@ -12,9 +12,18 @@ class ClassesScannerTest extends \PHPUnit_Framework_TestCase
...
+     *
+     * @var string
+     */
+
+    private $generation;
+
     protected function setUp()
     {
-        $this->model = new \Magento\Setup\Module\Di\Code\Reader\ClassesScanner();
+        $this->generation = realpath(__DIR__ . '/../../_files/var/generation');
```
orlangur, line 27: Why need `realpath` here?

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/FileClassScannerTest.php
```diff
@@ -0,0 +1,243 @@
...
+<?php
+
+echo 'hello world';
+
+if (class_exists('some_class')) {
+    \$object = new some_class();
+}
+PHP
+        );
+        /* @var $scanner FileClassScanner */
```
orlangur: PHPDoc shall start with `/**`, please fix all occurrences.

orlangur (changes requested):

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -16,13 +17,32 @@ class ClassesScanner implements ClassesScannerInterface
     protected $excludePatterns = [];
 
     /**
+     * @var array
+     */
+
+    protected $fileResults = [];
```
orlangur: All new properties shall be `private`.

orlangur (changes requested):

orlangur:
> a single extraction class that looks only for the first namespace/class match

Does it mean the FQCN matching class itself? Why we cannot obtain it from file path itself?

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: It simply excludes the generation directory from the local result cache if it exists.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: Presumably because of the possibility of an incorrect class definition in one file affecting another operation.  This is is how it was done in the original code.  https://github.com/magento/magento2/blob/364635691c8baa5cc30d58c1c1879fc309f5b048/setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
kschroeder: Why?

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/ClassesScannerTest.php
kschroeder: To get a clear indication that it is the path that is broken if there is an error during debugging.

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/FileClassScannerTest.php
kschroeder: Will do

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: Will do

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
orlangur: I noticed `require_once` derived from old code, of course.

> possibility of an incorrect class definition

But why not fail fast in such case? I'm not saying such change must be a part of your PR, just asking if you see some logic in `_once` as you are already in context of this code.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
orlangur: Because there is no need to reinvent the wheel. Proposed implementation is quite low-level and hard to maintain while this problem is easily solvable with PHP Parser (I did some prototyping with it and FQCN resolution in particular is shown in documentation).

But please clarify https://github.com/magento/magento2/pull/8965#issuecomment-289475460 first - I thought this `extract` is intended to find all class references at first glance.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
orlangur: I mean it should have been `if (!$realPath)` ;)

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -16,13 +17,32 @@ class ClassesScanner implements ClassesScannerInterface
...
+     */
+
+    private $generationDirectory;
+
+    /**
      * @param array $excludePatterns
+     * @param string $generationDirectory
      */
-    public function __construct(array $excludePatterns = [])
+
```
orlangur: Please remove all such empty lines. I'm not sure it is enforced by static tests currently but is definitely inconsistent with Magento 2 codebase.

orlangur (changes requested):

schmengler (approved):

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -16,11 +17,29 @@ class ClassesScanner implements ClassesScannerInterface
...
+    public function __construct(array $excludePatterns = [], DirectoryList $directoryList = null)
     {
         $this->excludePatterns = $excludePatterns;
+        if (!$directoryList instanceof DirectoryList) {
```
maghamed: it's better to compare with null than to make instanceof opeartion.
It's better from teh performance point of view

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -52,46 +78,66 @@ public function getList($path)
...
-                }
-                $classes[] = $className;
-            }
+            $this->includeClasses($classNames, $fileItemPath);
+            $classes = array_merge($classes, $classNames);
         }
         return $classes;
     }
 
+    protected function includeClasses(array $classNames, $fileItemPath)
```
maghamed: newly created method should be private , not protected

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
```diff
@@ -0,0 +1,179 @@
...
+namespace Magento\Setup\Module\Di\Code\Reader;
+
+class FileClassScanner
+{
+    /**
+     * The filename of the file to introspect
+     *
+     * @var string
+     */
+
```
maghamed: please remove this blank lines

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Extracts the fully qualified class name from a file.  It only searches for the first match and stops looking
+     * as soon as it enters the class definition itself.
+     *
+     * Warnings are suppressed for this method due to a micro-optimization that only really shows up when this logic
+     * is called several millions of times, which can happen quite easily with even moderately sized codebases.
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
```
maghamed, line 67: we forbid SuppressWarning for Cyclomatic complexity in newly created code

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Looks forward from the current index to determine if the namespace is nested in {} or terminated with ;
+     *
+     * @param integer $index
+     * @return bool
+     */
+
+    protected function isBracedNamespace($index)
```
maghamed: don't use protected for newly created methods

maghamed (changes requested):

orlangur:
@kschroeder @schmengler 
> With 55 million calls micro-optimizations do add up. So I would like a little guidance on what you would prefer: reduce the individual method complexity or notate MD to ignore CC?

> If reducing the method complexity has performance impact ...

Performance impact needs to be demonstrated first.

I believe implementation based on PHP Parser would be both better maintainable and has better performance. But before doing this please clarify https://github.com/magento/magento2/pull/8965#issuecomment-289475460

schmengler:
@orlangur Regarding #8965, I explicitly requested to not do this. Quoting my review:

> If I understand that code correctly, it will break at least the following special cases:
> 
> - `namespace X { }` syntax
> - multiple classes per file
> 
> While both do not happen in the core, we have no control over third party code. Keep in mind that the Magento DI framework can also used to integrate third party libraries that are not Magento modules.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
schmengler: @orlangur Since micro optimization is a thing here, I would not introduce a full fledged parser until it's proven that the impact is negligible. Both should have linear complexity, but the current solution is just one loop that does not exactly what's needed and not more.

orlangur:
@schmengler I understand this, but do we need to support PSR-0,1,4 incompatible third-party code?

Some 99.99% of libraries would not need source code parsing, for other proper wrappers could be created so that we do not scan bad smelling code ever.

schmengler:
@orlangur I prefer a "no surprises" approach here. I've been in the 0.01 percent group before in a similar case. Debugging issues in the compiler and finding workarounds is a pain. Could you write wrappers for these libraries? Sure. If you *know* that this is the problem. But it will be hidden very well and we have no reliable way to warn the user about it.

orlangur:
@schmengler makes sense to me, thanks, I'm not even sure all libraries currently used directly in core are PSR-0,1,4 compliant.

Added TODO for myself to check later for possibility of this kind of improvement keeping developer experience seamless.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
orlangur: > micro optimization is a thing here

This needs to be shown in numbers... For the cyclomatic complexity, not for alternative solution with third-party parser.

>  I would not introduce a full fledged parser until it's proven that the impact is negligible

Ok, no objections :) I was just asking to evaluate possibility of PHP Parser usage here and now, it can be improved later iteratively.

orlangur (commented):
"general feedback without explicit approval" :)

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
kschroeder: Oh, well, you'd have to ask the original author of the class.  I didn't write that line.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
kschroeder: See https://github.com/magento/magento2/pull/8965#issuecomment-289130082

schmengler:
@maghamed the PR still needs your approval, please leave feedback. This is an important change and it would be a shame if we don't get it into the next release.

setup/src/Magento/Setup/Module/Di/Code/Reader/ClassesScanner.php
```diff
@@ -16,11 +17,26 @@ class ClassesScanner implements ClassesScannerInterface
...
+     * @param string $generationDirectory
      */
-    public function __construct(array $excludePatterns = [])
+    public function __construct(array $excludePatterns = [], DirectoryList $directoryList = null)
     {
         $this->excludePatterns = $excludePatterns;
+        if ($directoryList === null) {
+            $directoryList = ObjectManager::getInstance()->get(DirectoryList::class);
+        }
+        $this->generationDirectory = $directoryList->getPath(DirectoryList::GENERATION);
```
maghamed, line 39: it's not allowed to make function calls in the constructor, just assignments to class variables.

here we receive  $directoryList and after that make a call to get Path

setup/src/Magento/Setup/Module/Di/Code/Reader/InvalidFileException.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Magento\Setup\Module\Di\Code\Reader;
+
+class InvalidFileException extends \InvalidArgumentException
```
maghamed, line 5: Custom Exceptions are not recommended to be introduced in each module.

Here is more discussion on this subject -  https://github.com/magento/magento2/pull/9315/files#r112514206

setup/src/Magento/Setup/Test/Unit/Module/Di/Code/Reader/ClassesScannerTest.php
```diff
@@ -5,16 +5,31 @@
...
      * @var \Magento\Setup\Module\Di\Code\Reader\ClassesScanner
      */
     private $model;
 
+    /**
+     * the /var/generation directory realpath
+     *
+     * @var string
+     */
+
```
maghamed, line 22: empty line

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
maghamed: > > So I would like a little guidance on what you would prefer: reduce the individual method complexity or notate MD to ignore CC?

> If reducing the method complexity has performance impact, I'd say go for optimized performance in this case and mark the class as ignored.

@schmengler  Performance and Complexity are two different metrics. And in this specific case these metrics do not correlate. Because high Cyclomatic complexity doesn't lead to high performance and vice versa.

In Magento we have a rule, that any new code should not have SuppressWarnings for CouplingBetween objects and CyclomaticComplexity. Because it's from the beginning has a Technical Debt. 

High CyclomaticComplexity often says that some functionality has not been fully covered with automated testing. Because it's too hard to cover all the possible conditional logic and exit points for this method. 

Moreover such code is fragile and really hard to maintain in future.

Some time ago we even actively used [CRAP](https://testing.googleblog.com/2011/02/this-code-is-crap.html) metrics for this.

So, this method should be decomposed on some smaller once

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
```diff
@@ -0,0 +1,171 @@
...
+        $filename = realpath($filename);
+        if (!file_exists($filename) || !\is_file($filename)) {
+            throw new InvalidFileException(
+                sprintf(
+                    'The file "%s" does not exist or is not a file',
+                    $filename
+                )
+            );
+        }
+        $this->filename = $filename;
```
maghamed, line 47: magento constructors should contain just assigments, and no business logic allowed inside

maghamed (changes requested):

kschroeder:
Forget this.  I'm done.  I have about a thousand other things I need to be working on.

mzeis:
@kschroeder Thank you for putting so much work into this pull request! I hope we still can get your changes into core because - as @schmengler said - it would be a shame to not have it in future Magento versions.

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
mzeis: @maghamed I'd like to argue that this is a valid use case according to 2.3.1 of the [Technical Guidelines](http://devdocs.magento.com/guides/v2.1/coding-standards/technical-guidelines/technical-guidelines.html):

> 2.3.1. Constructor SHOULD throw an exception when validation of an argument has failed.

In the example for 2.3.1, an array of options is validated and an `InvalidArgumentException` is thrown if one of the objects in the array doesn't implement a particular interface.

In the given code the provided filename has to be an existing file so for me, this is a valid validation. Which alternative approach would you suggest?

setup/src/Magento/Setup/Module/Di/Code/Reader/FileClassScanner.php
mzeis: @maghamed See https://github.com/magento/magento2/pull/8965#issuecomment-289122432 for the explanation by @kschroeder why there might be a performance impact when decomposing this into more methods. How can we proceed here? Should we try to get hard numbers on the difference in performance?

setup/src/Magento/Setup/Module/Di/Code/Reader/InvalidFileException.php
mzeis: For `Magento\Setup\Module\Di\Code\Reader\FileClassScanner::__construct` I'd suggest to go with `\Magento\Framework\Exception\FileSystemException`. It's used the same way in `\Magento\Framework\Filesystem\Directory\Write::assertIsFile`.

For `Magento\Setup\Module\Di\Code\Reader\FileClassScanner::isBracedNamespace`, I'm a bit undecided which of the existing exeptions is the best fit. Maybe `\Magento\Framework\Exception\NotFoundException`, at least for one of the two cases?

joshuaadickerson:
@kschroeder I was about to jump to your defense on rewriting that method and I still think it should be accepted and then it can be reworked at any point in the future. Then I checked out the class and I looked at your argument. Taking a method and rewriting it to use internal (to the class) methods, even at 55 million calls, should not make any noticeable difference. I think all of those variables you're declaring in the extract() method are properties of the class. When you call extract(), they represent the state of that object. The class's only purpose is for that method (which is good).

I don't care to submit a PR or follow the coding guidelines, but I created a test of what it would look like and how that would affect performance: https://gist.github.com/joshuaadickerson/240f3633d99106fb7aed44a98a0ef0fd

It would be interesting to see what the difference between your class and my class are for the build that you said requires 55m calls.

mzeis:
@maghamed Please can you have a look at the comments I gave on April 23rd? Thanks!

antonkril:
The PR improves the state of the product. Let's merge.

antonkril (approved):

Issue #3784, opened 2016-03-15
js/theme & js/responsive files in wrong requirejs-config file?


gavinlimely:
Should the js/theme & js/responsive dependencies be located in the Magento_Theme requirejs-config file? I've created a new theme that doesn't extend the blank theme & Magento is still trying to include js/theme & js/responsive even though I don't require them in the theme I'm creating. 

Would it make more sense for them to be included in the blank theme?

gavinlimely:
any updates?

piotrekkaminski:
@elenleonova  can you look into it? is it a bug?

minlare:
This is a bug, workaround for now is to create blank files js/responsive.js and js/theme.js in the custom theme.

anton-huz:
Yes, it is an issue. We are going to fix it in scope of MAGETWO-56478.
Please stay in touch.

markdavies:
Any movement on this? Running into the same problem today.

ghost:
Does anyone know if this issue has been/will be fixed? The suggested work around isn't really viable.

PR #8982, opened 2017-03-23
Move blank theme dependencies out of Magento_Theme requirejs-config

mikeoloughlin:
### Description

Creating a new theme that does not fallback to the blank theme causes two files to 404 because Magento_Theme module depends on blank theme assets `js/theme.js` and `js/responsive.js`

1. #3784: js/theme & js/responsive files in wrong requirejs-config file?

### Reproduce bug:

1. Follow the steps to create a new theme from the [devdocs](http://devdocs.magento.com/guides/v2.1/frontend-dev-guide/themes/theme-create.html)
2. Don't define a parent theme:
```
<theme xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Config/etc/theme.xsd">
     <title>New theme</title>
 </theme>
```
3. Clear cache and go to the frontend to see the 404s

### Manual testing steps:
1. New theme should no longer attempt to load `theme.js` or `responsive.js`
2. Switch to blank theme and clear cache. Storefront should load `theme.js` and `responsive.js`


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8982) <br/>All committers have signed the CLA.

app/design/frontend/Magento/blank/requirejs-config.js
```diff
@@ -0,0 +1,11 @@
+/**
```
Igloczek: I'd say that adding `requirejs-config.js` at theme root should be considered as anti-pattern and Magento core shouldn't promote it, b/c it's against "everything is a module" aproach.

By default this config file is stored in `Theme` module of `Magento` vendor, so following files structure, it should be added as `app/design/frontend/Magento/blank/Magento_Theme/requirejs-config.js` and JS files should be moved from `app/design/frontend/Magento/blank/web/js` to `app/design/frontend/Magento/blank/Magento_Theme/web/js`.

mikeoloughlin: That's a good point and I agree with you. I've updated the code and tested it.

Thanks :)

Igloczek (approved):

okorshenko:
@mikeoloughlin thank you for your contribution!

PR #8989, opened 2017-03-23
Remove redundant check in if-condition

FabianLauer:
Removed the part of an if condition that tests whether a quote is virtual or not. The part that was removed is redundant because of line 43.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8989) <br/>All committers have signed the CLA.

ishakhsuvarov:
@FabianLauer Thank you for your contribution!

Issue #3400, opened 2016-02-14
\Magento\Framework\Module\PackageInfo assumes locations


fooman:
This code 

```
private function load()
{
    if ($this->packageModuleMap === null) {
        $jsonData = $this->reader->getComposerJsonFiles()->toArray();
        foreach ($this->componentRegistrar->getPaths(ComponentRegistrar::MODULE) as $moduleName => $moduleDir) {
            $key = $moduleDir . '/composer.json';
```

assumes that the registered path also holds the composer.json file which is not necessarily the case (for example https://github.com/fooman/sameorderinvoicenumber-m2/tree/v2.0.2/src). I am hoping there is a way to query composer itself for all packages of type magento2-module.

davidverholen:
I'm not sure if you can query a composer instance for these paths, but it would be pretty easy to find them.

Since the composer.json is always in the root directory of the package, it just has to go to the parent dir until it finds the composer.json. 

This could be limited to the vendor dir, so that it does not go any higher in the directory structure. The vendor-dir can be read from the root composer.json (default: vendor)

fooman:
I think the underlying problem is that Magento has an internal name for a module which is defined via `registration.php`. This definition is not formally tied to the composer name (it usually follows the convention `VendorName_ModuleName`), however for their own modules they employ some trickery here https://github.com/fooman/magento2/blob/1535766cb5d9146485793462c9cc9270c970cf0e/lib/internal/Magento/Framework/Module/PackageInfo.php#L177.

Just trialling something here https://github.com/magento/magento2/compare/develop...fooman:component-manager-reader-improvements?expand=1, where at least the parent dir is checked. It works but it's not quite right as this currently determines the name from the autoload section, and I think it would be better if it would get back the module name from the reader.

davidverholen:
but, to read the module name from the composer.json, it would have first to know, where it is.
Currently, there is no direct connection between the registered component and a module.

Searching it in the parent dirs would work but also is not really nice.

It would be possible to create a map in the magento-composer-installer: composer-name to magento module name.

The it would be easy to find the composer.json

fooman:
> but, to read the module name from the composer.json, it would have first to know, where it is.

That is what Magento currently tries to do, even though the `composer.json` name does not include the module name as it is known to Magento under.

> Currently, there is no direct connection between the registered component and a module.

This is the overall process of how I currently have it pieced together:

1.) Magento application gets started
2.) includes Composer autoload information
[...]

PR #8990, opened 2017-03-23
Fall back to parent directory if composer.json not found in module directory

schmengler:
### Description
This change allows modules to follow the [pds/skeleton](https://github.com/php-pds/skeleton) standard and have their source code in a `src` subdirectory of the repository.

### Manual testing scenarios

1. create a module with `composer.json` and `registration.php` in the root directory and source code in `src`

    registration.php
    ```
        'Vendor_Module',
        __DIR__ . '/src'
    );
    ```

    composer.json
    ```
      "autoload": {
        "files": [
          "registration.php"
        ],
        "psr-4": {
          "Vendor\\Module\\": "src"
        }
      },
      "suggest": {
        "vendor/module-sample-data": "Sample Data version:1.0.0"
      },
    ```
2. call `bin/magento sampledata:deploy` and see if `vendor/module-sample-data` is going to be installed.

### Contribution checklist


schmengler:
Another case where Magento reads `composer.json` files from modules is `\Magento\Framework\Module\PackageInfo`. This is a bit more complicated to change and test, so it would be a different PR

fooman:
hey @schmengler love the direction of this PR. It is currently [already possible to have the code under src/](https://github.com/fooman/printorderpdf-m2) with the registration file sitting under `src/registration.php` and then reference it from a root `composer.json` file as `src/registration.php`. The only place where this falls over as you mentioned is in `\Magento\Framework\Module\PackageInfo` (I currently solve it by providing an alternative `src/composer.json` which is not ideal).

I had looked into this here https://github.com/magento/magento2/issues/3400. My current thought to solve this across the board would be to add a new optional param to `\Magento\Framework\Component\ComponentRegistrar::register()` which holds the `composer.json` file location which can then be used in `PackageInfo` to find the correct file. If empty it's assumed to be in the same folder as `registration.php` (aka current behaviour).

schmengler:
@fooman Good idea with the optional parameter, that would give even more flexibility! Putting registration.php in `src` was my first approach too but I remember this caused trouble because of assumptions in Magento. Duplicating composer.json is a workaround but that cannot be the solution

okorshenko:
Hi @schmengler thank you for your contribution. Could you please fix broken static tests. Also we are getting Fatal error with this changes. 
`PHP Fatal error:  Uncaught Error: Cannot instantiate interface Magento\Framework\Filesystem\DriverInterface`
Could you please verify your changes. Thank you

Ctucker9233:
@magento-team can this be ported to 2.2?

PR #8997, opened 2017-03-24
Use page result instead of rendering layout directly and add page tit…

mrkhoa99:
### Description

- Use page result instead of rendering layout directly in Advanced Search controller.
- Add Title page on Advanced search page.
### Fixed Issues:

-  Missing Title page on Advanced search page.

### DIFF
--- a/app/code/Magento/CatalogSearch/Controller/Advanced/Result.php
+++ b/app/code/Magento/CatalogSearch/Controller/Advanced/Result.php
@@ -9,9 +9,17 @@ namespace Magento\CatalogSearch\Controller\Advanced;
 use Magento\CatalogSearch\Model\Advanced as ModelAdvanced;
 use Magento\Framework\App\Action\Context;
 use Magento\Framework\Controller\ResultFactory;
+use Magento\Framework\UrlFactory;
 
 class Result extends \Magento\Framework\App\Action\Action
 {
+
+    /**
+     * Url factory
+     *
+     * @var UrlFactory
+     */
+    protected $_urlFactory;
 
     /**
      * Catalog search advanced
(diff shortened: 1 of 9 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=8997) <br/>All committers have signed the CLA.

app/code/Magento/CatalogSearch/Controller/Advanced/Result.php
```diff
@@ -32,33 +25,35 @@ class Result extends \Magento\Framework\App\Action\Action
      *
      * @param Context $context
      * @param ModelAdvanced $catalogSearchAdvanced
-     * @param UrlFactory $urlFactory
```
ishakhsuvarov: Due to Backwards Compatibility Policy we can not remove any arguments from the constructor at the moment

ishakhsuvarov (changes requested):

app/code/Magento/CatalogSearch/Controller/Advanced/Result.php
mrkhoa99: @ishakhsuvarov I will fix them. Seem that Travis build failed? Can you give your idea?

ishakhsuvarov:
@mrkhoa99 You can check the [details link](https://travis-ci.org/magento/magento2/jobs/214623508) for the Travis check in this PR. At the moment it failed on unit tests job with the following Fatal Error:
> `PHP Fatal error:  Uncaught Error: Call to a member function getConfig() on null in /home/travis/build/magento/magento2/app/code/Magento/CatalogSearch/Controller/Advanced/Result.php:66`

mrkhoa99:
@ishakhsuvarov seem that we cannot use `Magento\Framework\Controller\ResultFactory` to set the title page in controller? I think it should be ok.

PR #9013, opened 2017-03-25
Added JS Jasmine tests to Travis CI

Igloczek:
JS tests are already written, but not used on Travis CI pipeline, so I added the necessary config to enable them.


dev/travis/js/config.php
```diff
@@ -0,0 +1,360 @@
+<?php
```
ishakhsuvarov: Adding a hard-coded configuration may be too fragile, since it may be a subject for modification. Please consider installing to get the generated file

dev/travis/before_script.sh
```diff
@@ -76,4 +76,20 @@ case $TEST_SUITE in
...
+    js)
+        curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
+        export NVM_DIR="$HOME/.nvm"
+        [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
+        nvm install --lts
+        nvm use --lts
+
+        cp package.json.sample package.json
+        cp Gruntfile.js.sample Gruntfile.js
+        npm install -g yarn
```
ishakhsuvarov: Is it really viable to install `yarn` just to install `grunt-cli`?

ishakhsuvarov:
Hi @Igloczek 
Thank you for this PR. It is really a great idea to add JS Jasmine tests to the Travis-ci.

Could you please look into some questions I've left in the comments? Thanks!

Igloczek:
@ishakhsuvarov I'm pretty sure you forgot to hit "Submit review" 😉

ishakhsuvarov:
@Igloczek You are so right

dev/travis/before_script.sh
```diff
@@ -76,4 +76,20 @@ case $TEST_SUITE in
 
         cd ../../..
         ;;
+    js)
+        curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
+        export NVM_DIR="$HOME/.nvm"
+        [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
+        nvm install --lts
```
orlangur: I believe it would be more robust to specify some concrete version, like in https://github.com/BanzaiMan/travis_production_test/blob/9c02aef/.travis.yml#L15, so that we see in environment variable which version we test against

.travis.yml
```diff
@@ -21,16 +21,22 @@ env:
...
+    - php: 5.6.29
+      env: TEST_SUITE=js
 before_install: ./dev/travis/before_install.sh
 install: composer install --no-interaction --prefer-dist
 before_script: ./dev/travis/before_script.sh
 script:
   - test $TEST_SUITE = "static" && TEST_FILTER='--filter "Magento\\Test\\Php\\LiveCodeTest"' || true
-  - phpunit -c dev/tests/$TEST_SUITE $TEST_FILTER
+  - if [ $TEST_SUITE != "js" ]; then phpunit -c dev/tests/$TEST_SUITE $TEST_FILTER; fi
+  - if [ $TEST_SUITE == "js" ]; then grunt spec; fi
```
orlangur, line 45: Looking at https://travis-ci.org/magento/magento2/jobs/215020317 - is there some less verbose mode? Display only failed tests maybe...

dev/travis/before_script.sh
Igloczek: Not only `grunt-cli`, all dependencies defined in `package.json` are downloaded using `yarn` in line 90 (`yarn` is replacement for `npm install`).

dev/travis/js/config.php
Igloczek: I was following [docs](https://github.com/magento/devdocs/blob/develop/guides/v2.2/config-guide/cli/config-cli-subcommands-static-view.md#deploy-static-view-files-without-installing-magento-deploy_without_db) in this matter, b/c installing the whole application, just to get statics, it's serious overhead for me.

Maybe simpler option will be to work on source files, rather than processing them through M core, before running tests? It should be also easier for development purposes.

.travis.yml
```diff
@@ -21,16 +21,22 @@ env:
     - TEST_SUITE=integration INTEGRATION_INDEX=2
     - TEST_SUITE=integration INTEGRATION_INDEX=3
     - TEST_SUITE=static
+    - TEST_SUITE=js
 cache:
   apt: true
-  directories: $HOME/.composer/cache
+  directories:
+    - $HOME/.composer/cache
+    - $HOME/.nvm
```
Igloczek: I forgot about cashing `node_modules` and maybe also `yarn.lock` (it's like `composer.lock` for Yarn, but works better 😜 ).

orlangur (changes requested):

dev/travis/js/config.php
ishakhsuvarov: The documentation in question is about generating static on a non-production machine, so this is a slightly different case. Providing a hard-coded configuration file is not the best practice and needs to be maintained all the time.

I would still suggest installing application for now, at least for consistency, since internally this is what we do to run Jasmine tests.

dev/travis/before_script.sh
ishakhsuvarov: Thanks for clarification.

dev/travis/before_script.sh
Igloczek: I'm not sure if we really need to care about precise versioning, b/c in Oct 2017 we should be ready to use v7 without any issues and necessary updates in code base.

Node.js LTS plan is precisely described here - https://github.com/nodejs/LTS

dev/travis/js/config.php
orlangur: @ishakhsuvarov could you move some internal code to public maybe so that we don't have two different implementations?

From https://github.com/magento/magento2/blob/develop/dev/travis/before_script.sh#L61 it looks it is not so easy to steal application installation from integration tests and it is placed somewhere inside framework.

As to me hardcoded config seems fine at least as a temporary solution if we add comment on top of it how to regenerate file.

.travis.yml
Igloczek: No idea, for now I just focuses on make this test running, then we can think about some adjustments / improvements.

PS. I really hate Grunt, so if I don't have to modify this, I will be more than happy to leave it as it is 😉

dev/travis/js/config.php
ishakhsuvarov: @orlangur It is executed under different infrastructure then Travis CI. Would not solve the issue in question.
"Stealing" application from the integration tests is not a good solution either. Clean install should do the trick for now, while no better solution is available.

dev/travis/before_script.sh
orlangur: If I understood the doc right v7 is not LTS and will not be used in such code.

The point is that I don't see any value in dynamic `Node.js` version, it can only bring us some headache when the LTS version will switch. Explicit is better than implicit.

As soon as `v8` stable is released we can change value in our config if the tests still pass. Current version is like specifying `php: stable` version instead of concrete value.

dev/travis/js/config.php
orlangur: @ishakhsuvarov what do you mean by clean install? :) I thought installation in scope of integration tests is clean. Is it possible to install application without database at all, for example?

.travis.yml
orlangur: Ok, I didn't use it as well, and didn't google for a solution just in case you know the answer.

It can wait until it become a problem "output is too big" some day (hopefully never).

dev/travis/before_script.sh
Igloczek: You are right, v8, not v7.
LTS means more than stable, b/c stable v8 will be released in April 2017, but will be marked as LTS in Oct 2017. It's a different approach than PHP takes.

orlangur: Is it guaranteed that using v6 for tests before, we will not have any issues under v8 when it become LTS? What's BC police?

Igloczek: OFC there are BC between major releases, but major becomes LTS after half year and most of the packages get updates during this period, so migration is in most of the cases are seamless.

dev/travis/js/config.php
Igloczek: @ishakhsuvarov Could you share part of your tests configuration related to installation? Will be much easier to copy-paste(-adjust) working solution 😉

dev/travis/before_script.sh
orlangur: It does not mean all our tests will work without changes.

Just answer the simple question - what is the value of NOT specifying a concrete version but using LTS instead?

ishakhsuvarov: By the way, `package.json.sample` bundled with Magento also has version requirement.

dev/travis/js/config.php
ishakhsuvarov: @Igloczek So, I've done some brief experiments with Travis CI and here is what seems to be working:

1. Instead of bundling `config.php` add the minimum installation as follows (it does not add too much time to the execution):
> `echo "Installing Magento"`
> `mysql -uroot -e 'CREATE DATABASE magento2;'`
> `php bin/magento setup:install -q --admin-user="admin" --admin-password="123123q" --admin-email="admin@example.com" --admin-firstname="John" --admin-lastname="Doe"`
2. You may do some experiments with the static content deployment settings to add a bit more performance. Here is what worked for me (may not be the best solution, just experiment):
> `echo "Deploying Static Content"`
> `php bin/magento setup:static-content:deploy -f -q -j=2 --no-css --no-less --no-images --no-fonts --no-misc --no-html-minify`

dev/travis/before_script.sh
Igloczek: You are always up to date, don't have to follow releases schedule etc. When new LTS version will be up, your tests will start using it.

But to just quit discussion, I'll change it to `6` and this will be good enough till next LTS.

Docs about `engines` option from `package.json`
> this field is advisory only will produce warnings when your package is installed as a dependency.

So in our case, we can simply remove this param.

ishakhsuvarov (approved):

ishakhsuvarov:
@Igloczek Looks good to me now! Thanks for the fixes.
I am going to restart Travis build, to get rid of the `Some checks were not successful` error, and proceed with the merge.

Igloczek:
You are faster than my thoughts, I was going to do same thing 😄

dev/travis/before_script.sh
orlangur: Thanks!

> When new LTS version will be up, your tests will start using it.

Yeah, that's a scary part as to me, as put our builds a bit out of control.

orlangur (approved):

Igloczek:
Please stop the merging process, I found a bug related to NVM and we are now using some built-in Node.js version.

ishakhsuvarov:
@Igloczek Thanks for reporting

Igloczek:
Fixed.

dev/travis/before_script.sh
```diff
@@ -80,8 +80,8 @@ case $TEST_SUITE in
         curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
         export NVM_DIR="$HOME/.nvm"
         [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
-        bash -c "nvm install $NODE_JS_VERSION" || true
```
orlangur: Why there was `bash` used at all in the first place?

Igloczek: Just b/c I'm a dumb copy-paster 🙄

orlangur: That's absolutely fine as I had just an approximate understanding of what is going on when reviewed your code as well 👍 

Hope we have some decent DevOps in Magento Community or in Magento Team to review Travis scripts :)

okorshenko:
@Igloczek thank you for your contribution!

PR #5518, opened 2016-07-07
Remove superfluous realpath call

tkn98:
Calling realpath is not necessary for __DIR__

Refs:
- http://php.net/manual/language.constants.predefined.php

PR #9025, opened 2017-03-27
Remove superfluous realpath call

tkn98:
Calling realpath is not necessary for __DIR__

Refs:

- http://php.net/manual/language.constants.predefined.php

(Previously reported against wrong base branch in #5518)

### DIFF
--- a/pub/index.php
+++ b/pub/index.php
@@ -10,7 +10,7 @@ use Magento\Framework\App\Bootstrap;
 use Magento\Framework\App\Filesystem\DirectoryList;
 
 try {
-    require realpath(__DIR__) . '/../app/bootstrap.php';
+    require __DIR__ . '/../app/bootstrap.php';
 } catch (\Exception $e) {
     echo <<<HTML
 <div style="font:12px/1.35em arial, helvetica, sans-serif;">


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9025) <br/>All committers have signed the CLA.

pub/index.php
```diff
@@ -10,7 +10,7 @@
 use Magento\Framework\App\Filesystem\DirectoryList;
 
 try {
-    require realpath(__DIR__) . '/../app/bootstrap.php';
```
okorshenko, line 13: It make sense to change the code in the following way:
`realpath(__DIR__ . '/../app/bootstrap.php')`
in this case we will get `/root/folder/app/bootstrap.php` insead of `/root/folder/pub/../app/bootstrap.php`

okorshenko (changes requested):

okorshenko:
@tkn98 could you please sign CLA? Thank you

pub/index.php
tkn98: On closer look I'd say that such a suggestion is not necessary. Require deals fine with such paths, it does not need any further assistance. Or do you have a scenario at hand where that is not the case? I personally don't know of any nor can I imagine one.

tkn98:
@okobchenko: reg. CLA: Wrong author-email was in the commit, that's why the CLA was shown not-signed. I have signed that long ago, should come green then, updated the commits author.

tkn98:
@okobchenko: reg. CLA: (update: now fixed) Wrong author-email was in the commit, that's why the CLA was shown not-signed. I have signed that long ago, should come green then, updated the commits author.

pub/index.php
orlangur: Passing whole string to `realpath` is a bad idea from debugging perspective, see http://stackoverflow.com/a/21035675.

I saw a lot `realpath` calls all over the code, but for `__FILE__` or `__DIR__` it seems useless.

okorshenko (approved):

okorshenko:
@tkn98 thank you for your contribution. Your changes successfully merged https://github.com/magento/magento2/commit/485adc5b9c245cd67a99255b2d2da324263deb50#diff-4aa58a2cf6c6b39adfc4e63689aea3c0

Issue #7000, opened 2016-10-14
Braintree doesn't work when PayPal is enabled and no shipping is required


tehXor:
### Steps to reproduce

1. Install Magento 2.1.2
2. Configure Braintree and enable Credit Card as well as PayPal with Braintree
3. In the frontend add only downloadable products to your cart (where no hardware shipping is required)
4. Try to place the order
### Expected result
1. On the selection screen for payment methods you should be able to select credit card as well as PayPal
2. When you select one you should be prompted to enter your billing address
### Actual result
1. On the selection screen for payment methods the loading spinner doesn't finish
2. In the background you can see that none of the braintree methods are loaded
3. There is one error in the javascript log which occurs in /Magento_Braintree/js/view/payment/method-renderer/paypal.js:
   `paypal.js:276 Uncaught TypeError: Cannot read property '0' of undefined`
### Additional Notes
1. When PayPal is disabled but credit card is enabled, credit card works like expected
2. We are in sandbox mode
### Workaround
1. Patch `getShippingAddress` in /Magento_Braintree/js/view/payment/method-renderer/paypal.js in any way that doesn't rely on the existenz of `address.street`
2. Even if the address is completely empty billing data is pulled from PayPal and correctly submitted to Magento, so this workaround seems safe

iivashchenko:
@tehXor, thank you for reporting, but unfortunately, I can't reproduce that issue. I've tried to reproduce it on Google Chrome and, as a result, on checkout I have three payment methods available (PayPal (Braintree),  Credit Card (Braintree) and Check / Money order).

Let me suggest you clear browser cache and try to reproduce that issue again.

tehXor:
Clearing caches doesn't help and when I look at the paypal.js implementation the problem seems logic to me:
When you look at line 298 in the paypal.js ( https://github.com/tehXor/magento2/blob/develop/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/paypal.js#L298 ) you can clearly see, that the shipping address is used as address. And it is also evident, that there are cases, where the system cannot know the shipping address yet: A new/guest user purchases a product where shipping is not required (virtual or downloadable product) and therefore hasn't entered his address anywhere.
In that case line 307 has to throw an exception ( https://github.com/tehXor/magento2/blob/develop/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/paypal.js#L307 ).

[...]

PR #9029, opened 2017-03-27
Generalize fix from GitHub issue #7000

scottsb:
Issue #7000 fixed a particular case where the Braintree PayPal payment method renderer would cause a JavaScript crash when trying to use unavailable address information.

However, there are other edge cases where this still occurs (such as when a registered user doesn't have a default shipping address set). This generalizes the fix to use a try/catch, which avoids having to test for every single possibility in the conditional. **Any time** that we can't retrieve the necessary address information, the best option is to simply return an empty object to PayPal, as this now does.

### DIFF
--- a/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/paypal.js
+++ b/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/paypal.js
@@ -296,21 +296,20 @@ define([
         getShippingAddress: function () {
             var address = quote.shippingAddress();
 
-            if (_.isNull(address.postcode) || _.isUndefined(address.postcode)) {
-
+            try {
+                return {
+                    recipientName: address.firstname + ' ' + address.lastname,
+                    streetAddress: address.street[0],
+                    locality: address.city,
+                    countryCodeAlpha2: address.countryId,
+                    postalCode: address.postcode,
+                    region: address.regionCode,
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9029) <br/>All committers have signed the CLA.

YevSent:
Hi, @scottsb, thanks for contribution. Could you cover your fix with [js test](http://devdocs.magento.com/guides/v2.1/test/js/jasmine.html)?
I think it would be useful to cover the next few cases:

- the address does not contain some properties - test covers the negative scenario
- address contains all properties - test covers the positive scenario

scottsb:
I will add this test, but it will probably be a couple weeks before I can get to it.

YevSent (changes requested):
Changes should be covered by js test.

ishakhsuvarov:
Hi @scottsb 
Are you still interested in this PR? Any help needed?

scottsb:
@ishakhsuvarov, I am still interested in doing this, but we are heads down on a new launch with a tight deadline, so I won't be able to get back to this until after we launch.

ishakhsuvarov:
@scottsb Thank you for the update. Let us know if you need any help.

okorshenko:
Hi @scottsb thank you for your contribution. Please, create new PR once ready. Closing due to inactivity

PR #9035, opened 2017-03-28
Remove unused navigation-menu.js in blank theme

mikeoloughlin:
### Description
Removing unused js file `navigation-menu.js` from blank theme. Unless I'm mistaken, this was replaced pre-alpha.


okorshenko:
@mikeoloughlin unfortunately we can not remove it right now. According to our backward compatibility policy, we must deprecate it first.
In the scope of next minor release, we will be able to delete it.
If you would like you can change this PR and deprecate it

okorshenko:
@mikeoloughlin please, create new PR if you would like to deprecate this file. Thank you

PR #8307, opened 2017-01-27
Allow digits in communication class type definition

cmuench:
Our company vendor prefix "N98" is not allowed. The XSD prohibits the usage of digits.

PR #5420, opened 2016-07-02
Fix issue #4470: import.xsd doesn't allow numbers in vendor name and class name

nhp:
It was possible to have model definitions that were not parsable
in the xml because it validated the xsd specificatons.
There was no Problem in adding those Models and use them but
not with that .xsd file.

PR #9044, opened 2017-03-29
Remove superfluous character in class

tkn98:
The backslash character ("`\`")  was put more than once in the same
regex character class across some XSD files. Per each character class
there is no need to specify it twice.

I ran over this while looking into an issue with class-name validation
which do not allow digits within class names.

Refs:

- https://www.w3.org/TR/xmlschema11-2/#cces

- #4470

- #5420

- #8307


vrann:
Is this PR relies on the #5420? As far as I see ImportExport is not fixed here.

tkn98:
@vrann No it's not fixed here, I didn't want to hijack it and appreciate some feedback first. My idea is to collect places (removing superfluous the characters in the pattern can help with finding these) where this is in place in one issue and make the pattern the same for the different places so it is more maintainable in the future. Please let me know what you think.

okorshenko:
@tkn98 thank you for your contribution!

PR #9052, opened 2017-03-29
Define custom paths for logs without writing extra handlers

noglitchyo:
Enhances the base log system and provide a way to define custom log directories/files through the di.xml file without having to create extra handlers just to change the path of log files.

This enhancement can lead to broken behaviors for custom handlers defined with $filename using the full path. I do not know if it is considered as a BC break or not.

### Description

_Problem 1_
Actually, there is no way to define custom paths for log handler without creating an extra handler redefining the paths.

This article presents a good way to define custom paths for logs: https://www.maxpronko.com/blog/how-to-create-custom-logger-in-magento-2
But this implementation lacks of flexibility and forces developers to create an extra class. It can become fastidious if you have multiple modules and have to create the same handler just with the filename property changing. Plus, for each new handler defined you have to keep tracking of eventuals BC breaks.
In my opinion, developers should be free to custom this kind of things with ease without writing new classes.

_Problem 2_
With $filePath, we could think at first that the developer can choose a full custom location for its log paths but instead, $filePath is concatened with the $fileName property, which is kind of a misleading from my point of view.

_Solution_
The provided changes allow developers to define custom log files using the di.xml file and do not force them to implement extra class. The developer can choose to:
- Store the log just in a custom directory in var/log
- Store the log somewhere else in the server with the path of his choice (directory and filename)

### Manual testing scenarios
Here is a way to define custom paths for existing handlers with the provided changes:
**1. Edit the _di.xml_ file of your module and add the below code**
```
<virtualType name="customLogger" type="Magento\Framework\Logger\Monolog">
   <arguments>
        <argument name="handlers"  xsi:type="array">
            <item name="debug" xsi:type="object">customDebugHandler</item>
        </argument>
    </arguments>
</virtualType>

<virtualType name="customDebugHandler" type="Magento\Framework\Logger\Handler\System">
    <arguments>
        <argument name="fileDirectory" xsi:type="string">/myvendor/mymodule</argument>
    </arguments>
</virtualType>
```

**2. Use the customLogger defined as a depency of an existing or newly created class**

**Edit your _di.xml_**

```
<type name="MyVendor\MyModule\MyClass">
        <arguments>
            <argument name="logger">customLogger</argument>
        </arguments>
</type>
```
**Create a class _MyVendor/MyModule/MyClass.php_**
```
<?php
namespace MyVendor\MyModule;

class MyClass {

    /**
     * @var \Psr\Log\LoggerInterface
     */
    protected $logger;
    
    public function __construct(\Psr\Log\LoggerInterface $logger)
    {
        $this->logger = $logger;
        $this->logger->debug('Logged to my custom debug path');
    }
}
```
As I said above, this enhancement might lead to broken paths for module with handlers re-defining the $fileName property by using a full path. (i.g: /var/log/mycustomfile.log) and would force developers to update theirs modules and change the $fileName property of their handlers.

Tell me what you think of this enhancement, if the break described should be taken in consideration, if you find this enhancement useful or not, etc...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Developer/Model/Logger/Handler/Debug.php
+++ b/app/code/Magento/Developer/Model/Logger/Handler/Debug.php
@@ -37,15 +37,17 @@ class Debug extends \Magento\Framework\Logger\Handler\Debug
      * @param ScopeConfigInterface $scopeConfig
      * @param DeploymentConfig $deploymentConfig
      * @param string $filePath
+     * @param string $fileDirectory
      */
     public function __construct(
         DriverInterface $filesystem,
         State $state,
         ScopeConfigInterface $scopeConfig,
         DeploymentConfig $deploymentConfig,
-        $filePath = null
+        $filePath = null,
+        $fileDirectory = null
     ) {
-        parent::__construct($filesystem, $filePath);
+        parent::__construct($filesystem, $filePath, $fileDirectory);
 
         $this->state = $state;
         $this->scopeConfig = $scopeConfig;
--- a/lib/internal/Magento/Framework/Logger/Handler/Base.php
+++ b/lib/internal/Magento/Framework/Logger/Handler/Base.php
@@ -13,6 +13,13 @@ use Monolog\Logger;
 
 class Base extends StreamHandler
 {
+    const DEFAULT_LOG_DIRECTORY = '/var/log';
+
+    /**
+     * @var string
+     */
+    protected $fileDirectory;
+
     /**
      * @var string
      */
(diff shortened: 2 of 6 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9052) <br/>All committers have signed the CLA.

vrann:
@noglitchyo My main concern with this approach is that changing the di.xml to change the path to the log file is no better than creating a new handler: you still write code which should be part of some module and checked into the repository.

Ideally, this should be resolved on the level of the deployment configuration, it should be possible to configure paths somewhere in env.php. In this way, it will allow having different configuration for the development environment and on the production.

Does this makes sense?

okorshenko:
@noglitchyo thank you for your contribution. Please consider the point of @vrann .
Closing the pull request due to inactivity.

korostii:
Hi @vrann, 

I think there's a misunderstanding of the "developer can choose a full custom location" phrase.
An `env.php` setting would allow the Magento instance to store logs into a directory other than `var/log` but the log files would be still shared between all the core and custom and 3rd-party modules. Right?

I think that the most common use-case (and, likely, the task which author of this PR was trying to simplify) for manipulating the `Handler`'s filepath is in order to have a *module-specific* logs' location.

Having module's logs to be stored separately from the Magento's own makes a lot of sense because you'll see a ton of INFO and DEBUG entries appear in both `system.log` and `debug.log` out of the box, at every page load. 
Good luck searching through _that_.

For a real-life example, consider porting "as is" a module which does a lot of logging via  from Magento1 to Magento2.
Last time I checked, in Magento2 this basic task requires creating at least **two** extra classes (Logger and Handler) plus some mandatory `di.xml` entries _in each module_ and it's just too messy. Now compare that to `Mage::log($message, Zend_Log::DEBUG, 'some_module.log');` in Magento1.

IMHO a Logger class that would either automatically create a single Handler with module-specific filename (e.g. `var/log/vendor_module.log`) or which could be configured full in whichever of the modules' XML config files - `di.xml` or otherwise) provided by the framework out of the box would be a big improvement for module developers' routine.

PR #9060, opened 2017-03-30
Fix wrong usage of namespaces for the Cli class

adragus-inviqa:
Not cool, M2. Not. Cool.

Relates to https://github.com/magento/magento2/issues/9058.

### DIFF
--- a/bin/magento
+++ b/bin/magento
@@ -19,7 +19,7 @@ try {
 try {
     $handler = new \Magento\Framework\App\ErrorHandler();
     set_error_handler([$handler, 'handler']);
-    $application = new Magento\Framework\Console\Cli('Magento CLI');
+    $application = new \Magento\Framework\Console\Cli('Magento CLI');
     $application->run();
 } catch (\Exception $e) {
     while ($e) {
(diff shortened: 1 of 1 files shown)


AydinHassan:
I'm not sure this is necessary, `bin/magento` is not namespaced, so it would work: https://3v4l.org/WCqpl.

I think #9058 is related to `2.1` where `Cli` is referenced relatively but is not imported: https://github.com/magento/magento2/blob/2.1/bin/magento. This seems to be fixed in `develop`

bin/magento
```diff
@@ -19,7 +19,7 @@ try {
 try {
     $handler = new \Magento\Framework\App\ErrorHandler();
     set_error_handler([$handler, 'handler']);
-    $application = new Magento\Framework\Console\Cli('Magento CLI');
+    $application = new \Magento\Framework\Console\Cli('Magento CLI');
```
orlangur, line 22: Try `Alt+Enter` + `Import Class` in PhpStorm ;)

Hope to see `T_NAMESPACE_SEPARATOR` banned anywhere besides `namespace` and `use` statements someday...

orlangur (changes requested):

adragus-inviqa:
@AydinHassan - you're right. I just chaged the `master` version, though. Needless now, I guess. But I still don't like not having a leading slash. Less expressive.

PR #9062, opened 2017-03-30
Upgrade JS dependencies

Igloczek:
Upgrade of all of JS dependencies required to run Grunt.js task. Some packages have BC due to major releases, so I had to adjust configuration.

I also changed accepted version range from `^` to `~`, b/c it's safer to use only hotfix upgrades, than minor (extremely important when we are not maintaining any lockfile for these dependencies).

Some tasks don't work, b/c was broken before changes i.e. `grunt styledocco:documentation`.
Some of them work just weird like `grunt replace:documentation`, b/c on every run they add license comments, even it already are added, but also it's not related to packages upgrade.

Nice things - styles compilation works faster 😉


omiroshnichenko (approved):

magento-team:
@Igloczek thank you for your contribution to Magento 2 project

PR #7593, opened 2016-11-28
Translation for validate messages

redelschaap:
Added phrase functions to validate messages.

PR #9087, opened 2017-04-01
Make email validation only validate @ sign presence, fixes #4547

### LINKED ISSUE #4547: Support for new `.tech` TLD and others required
## Problem

New TLD's are not supported by Magento
## Solution

Add the domain names from the [IANA list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt).
## Steps to reproduce
1. Install Magento 2.0.4 preferablly using the Digital Ocean image.
2. Change the admin email to something like `[redacted]@payload.tech`.
3. Read the errors
## Expected result
1. Changes successfully saved
## Actual result
1. 'payload.tech' is not a valid hostname for email address '[redacted]@payload.tech'
2. 'payload.tech' appears to be a DNS hostname but cannot match TLD against known list
3. 'payload.tech' appears to be a local network name but local network names are not allowed
4. 'payload.tech' is not a valid hostname for email address '[redacted]@payload.tech' 'payload.tech' appears to be a DNS hostname but cannot match TLD against known list 'payload.tech' appears to be a local network name but local network names are not allowed

sambolek:
### Description
New TLDs aren't valid because of Zend Framework Email validator, which is too strict. As some pointed out, this might not be necessary. See discussion: https://github.com/magento/magento2/pull/8420/files#r99476657

1. magento/magento2#4547: Support for new `.tech` TLD and others required

### Manual testing scenarios
1. During customer registration, try to register with email test@test.swiss
2. Observe an error being thrown http://prntscr.com/er5qdb
`"Email" is not a valid hostname. 'test.swiss' looks like a DNS hostname but we cannot match the TLD against known list. 'test.swiss' looks like a local network name, which is not an acceptable format.
`

### Contribution checklist


PieterCappelle:
+1

sambolek:
Will fix build errors in case this pull request is considered a good way forward. Awaiting comments.

orlangur:
Cool! 👍 Especially that you didn't forget about a JavaScript part.

I'm adding some comments for current changeset and let's proceed after we have an official confirmation such approach is generally acceptable.

app/code/Magento/Customer/Model/Metadata/Form/AbstractData.php
```diff
@@ -336,42 +336,42 @@ protected function _validateInputRule($value)
                     __("'%value%' appears to be a DNS hostname but cannot extract TLD part")
                     __("'%value%' appears to be a DNS hostname but cannot match TLD against known list")
                     */
-                    $validator = new \Zend_Validate_EmailAddress();
+                    $validator = new \Magento\Framework\Validator\EmailAddress();
```
orlangur: We won't have anything besides `"%1" is not a valid email address.` validator error anymore thus all irrelevant cases shall be removed.

app/code/Magento/Eav/Model/Attribute/Data/AbstractData.php
```diff
@@ -364,42 +364,42 @@ protected function _validateInputRule($value)
                     __("'%value%' appears to be a DNS hostname but cannot extract TLD part")
                     __("'%value%' appears to be a DNS hostname but cannot match TLD against known list")
                     */
-                    $validator = new \Zend_Validate_EmailAddress();
+                    $validator = new \Magento\Framework\Validator\EmailAddress();
```
orlangur: We won't have anything besides `"%1" is not a valid email address.` validator error anymore thus all irrelevant cases shall be removed.

app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
```diff
@@ -419,13 +419,13 @@ define([
...
             function (value) {
-                return utils.isEmptyNoTrim(value) || /^([a-z0-9,!\#\$%&'\*\+\/=\?\^_`\{\|\}~-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z0-9,!\#\$%&'\*\+\/=\?\^_`\{\|\}~-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*@([a-z0-9-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z0-9-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*\.(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]){2,})$/i.test(value);//eslint-disable-line max-len
+                return utils.isEmptyNoTrim(value) || value.indexOf('@') !== -1;
             },
             $.mage.__('Please enter a valid email address (Ex: johndoe@domain.com).')
         ],
         'validate-emailSender': [
             function (value) {
-                return utils.isEmptyNoTrim(value) || /^[\S ]+$/.test(value);
+                return utils.isEmptyNoTrim(value) || value.indexOf('@') !== -1;
```
orlangur: After such change there is no difference between `validate-email` and `validate-emailSender`.

I believe it is wrongly copy-pasted from the bottom file:
```
['validate-emailSender', 'Please use only visible characters and spaces.', function (v) {
                  return Validation.get('IsEmpty').test(v) ||  /^[\S ]+$/.test(v)		                  return Validation.get('IsEmpty').test(v) ||  /^[\S ]+$/.test(v)
```

So, please just fix message and do not change validation function here.

lib/internal/Magento/Framework/Validator/EmailAddress.php
```diff
@@ -9,4 +9,13 @@
 
 class EmailAddress extends \Zend_Validate_EmailAddress implements \Magento\Framework\Validator\ValidatorInterface
```
orlangur, line 14: There is no need to keep inheritance from Zend Framework 1 class. Please mark this class as `@deprecated` without changes and create a new one with proper logic.

As we don't have `EmailAnything` except `EmailAddress`, I believe simply `Email` would be a good name for such class.

lib/internal/Magento/Framework/Validator/EmailAddress.php
```diff
@@ -9,4 +9,13 @@
...
 {
+    public function isValid($value)
+    {
+        if (false === \strpos($value, '@')) {
```
orlangur: 1) Please avoid Yoda conditions - JavaScript logic you wrote is much more readable ;) What would be really useful here is unit test
2) Remove leading backslash for function, there is no static test which forbids or enforces this but there is no other occurrence of `\strpos` currently. Let's be consistent.
3) `strrpos` will work faster than `strpos` at average as domain is usually shorter than account name ;D

lib/web/mage/validation.js
```diff
@@ -586,13 +586,13 @@
...
             function (v) {
-                return $.mage.isEmptyNoTrim(v) || /^([a-z0-9,!\#\$%&'\*\+\/=\?\^_`\{\|\}~-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z0-9,!\#\$%&'\*\+\/=\?\^_`\{\|\}~-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*@([a-z0-9-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z0-9-]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*\.(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]){2,})$/i.test(v); //eslint-disable-line max-len
+                return $.mage.isEmptyNoTrim(v) || v.indexOf('@') !== -1;
             },
             $.mage.__('Please enter a valid email address (Ex: johndoe@domain.com).')
         ],
         'validate-emailSender': [
             function (v) {
-                return $.mage.isEmptyNoTrim(v) || /^[\S ]+$/.test(v);
+                return $.mage.isEmptyNoTrim(v) || v.indexOf('@') !== -1;
```
orlangur: Same as https://github.com/magento/magento2/pull/9087/files#r109294015

orlangur (changes requested):

sambolek:
Hi @orlangur 
created a new validator for Emails:
`https://github.com/sambolek/magento2/blob/dc8867d40ca763a0a11012b1700659251b5ed378/lib/internal/Magento/Framework/Validator/Email.php `

that follows Currency validator practice: 
`https://github.com/sambolek/magento2/blob/dc8867d40ca763a0a11012b1700659251b5ed378/lib/internal/Magento/Framework/Validator/Currency.php`

Is this okay, or should I look into how AllowedProcotols validator is implemented, and have Email validator extend from AbstractValidator and implement ValidatorInteface as well, like here: `https://github.com/sambolek/magento2/blob/dc8867d40ca763a0a11012b1700659251b5ed378/lib/internal/Magento/Framework/Validator/AllowedProtocols.php` ?

sambolek:
Attempting to implement this validator by extending AbstractValidator and implementing ValidatorInteface.

As AbstractValidator is missing some key components to compete with Zend_Validate_Abstract, I have added some methods to this Email validator, so we can add messages by key, like we can for all Zend validators.

If we're looking to implement our own validators for some parts of the system, moving forward, we should have an implementation for setMessage, _error, and _createMessage in AbstractValidator. That way, adding new validators will be easier.

orlangur:
@sambolek thanks, I'll look deeper into implementation today and share my thoughts.

Anticipatorily I believe we should not have anything ZF-specific like `setMessage` outside the Magento Framework thus such email-related code in core modules shall be adopted to our own  narrowed validator interface. But I can change my mind after looking into existing implementation.

Also, we need to think about ZF1 validators replacement. I'll check how ZF2 validators differ from ZF1, for instance.

sambolek:
@orlangur any news on this one?

orlangur:
@sambolek sorry, forgot to review that evening and then had no time. Will review asap today and this time will not forget.

app/code/Magento/Customer/Model/Metadata/Form/AbstractData.php
```diff
@@ -324,90 +324,12 @@ protected function _validateInputRule($value)
...
-                    $validator->setMessage(
-                        __("'%value%' looks like a local network name, which is not an acceptable format."),
-                        \Zend_Validate_Hostname::LOCAL_NAME_NOT_ALLOWED
-                    );
-                    $validator->setMessage(
-                        __(
-                            "'%value%' appears to be a DNS hostname, but the given punycode notation cannot be decoded."
-                        ),
-                        \Zend_Validate_Hostname::CANNOT_DECODE_PUNYCODE
+                        \Magento\Framework\Validator\Email::INVALID
```
orlangur: Please encapsulate it in validator.

app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
orlangur: Please fix misprint with validate-emailSender if you don't mind (or tell me and I'll report it separately if you believe it is out of scope).

lib/internal/Magento/Framework/Validator/Email.php
```diff
@@ -0,0 +1,106 @@
+<?php
+/**
+ * Email address validator
+ *
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\Validator;
+
+class Email extends AbstractValidator implements \Magento\Framework\Validator\ValidatorInterface
```
orlangur: I believe the plan was to avoid inheritance from `AbstractValidator` whenever possible, by just implementing a couple of `\Zend_Validate_Interface` methods.

orlangur:
Hi @sambolek! I left some comments in review and let me clarify approach I find the best one here.

> that follows Currency validator practice:
https://github.com/sambolek/magento2/blob/dc8867d40ca763a0a11012b1700659251b5ed378/lib/internal/Magento/Framework/Validator/Currency.php

This one is a bad example I believe, it does not implement any interface in fact :D

I believe when the `\Magento\Framework\Validator\ValidatorInterface` was introduced additional methods like `setMessage` or `error` were intentionally omitted. For this simplified to teh max validator we don't need such infrastructure and it seems a good idea to me to encapsulate the ultimate possible validation error message within validator.

The closest good example I see is `\Magento\Framework\Validator\DataObject` - you need just to implement `isValid` and `getMessages` methods. Maybe we need to extend `\Magento\Framework\Validator\AbstractValidator` instead of just implementing `\Zend_Validate_Interface` as `DataObject` does, but as I didn't spot any consistent pattern in class/interface hierarchy of validators I leave such decision up to you.

orlangur (changes requested):

schmengler:
Thank you for your contribution! Can you adjust the failing tests to the new implementation? And please revert the changes in jQuery, modifications of third party libraries should not be allowed (see also: https://github.com/magento/magento2/pull/7593#issuecomment-286854054)

So we either have to live with that regex validation or replace its usages.

orlangur:
> modifications of third party libraries should not be allowed

Is it officially prohibited somewhere? As to me changing third-party libraries behavior is perfectly fine until we do not break upgradeability.

Probably, we can move this change to the bottom of file so that it is consistent with other basic functionality extensions there. But moving such change to a separate file is not a good idea as third-party developers can accidentally use original jQuery version instead of patched one.

chreds:
Any movement on this? Lots of new TLD's that aren't supported in Magento2 right now.

orlangur:
@chreds yes. As there are two possible implementation approaches an approval from Magento Product Team was requested to make an informed decision. After that we will be able to proceed with this PR acceptance.

okorshenko:
Hi @orlangur and @sambolek Thank you for working on this issue. ZF2 resolves the problem described in magento/magento2#4547  (see [hostname validator](https://github.com/zendframework/zend-validator/blob/master/src/Hostname.php#L1297) and [support of IANA](https://github.com/zendframework/zend-validator/blob/master/src/Hostname.php#L75))

We would like to update validator library from ZF1 to ZF2 in the entire application and eliminate direct knowledge about ZF1 validator from the code base. We should introduce Magento adapters and interfaces (where needed) and use only Magento code though the code base. 

Update of ZF1 validator to ZF2 is a big task and we would like to have it done for sure. 
Please, let me know if you would like to contribute to Magento upgrade of this component and we will start working with you on this upgrade. We already have several similar upgrade tasks that were implemented by Magento Community. See the project [Up For Grabs](https://github.com/magento/magento2/projects/10)

Closing this PR. We would like to avoid the validation only by `@` sign. 
Thanks everyone for collaboration!

orlangur:
@okorshenko that's sad to hear as validating only `@` is the best approach as to me and will never break down again :) Thanks for the reply though!

Don't forget to report https://github.com/magento/magento2/pull/8420/files#r99476657 as technical debt then ;)

korostii:
> We would like to avoid the validation only by `@` sign.

Could you please share the motivation for this decision?

IMHO it is *much* more acceptable from UX perspective to have some invalid emails pass through rather than deny a customer from using a perfectly valid address. And thus having a less restrictive validator is much better than having a _broken_ one. 
Please also see the article referenced by @schmengler: https://hackernoon.com/the-100-correct-way-to-validate-email-addresses-7c4818f24643

It puzzles me: why not use this one at least temporarily, until a Magento-style (c), adapter-wrapped, ZF2 validator is actually implemented (which might take months or years for all we know)

Issue #2802, opened 2015-12-24
Sitemap generation in wrong folder when vhost is connected to pub folder


jasperzeinstra:
In my case my domain (vhost) is connected to the pub folder instead of the root folder, for security reasons. Now when I try to generate a sitemap.xml in te root of the url. www.mydomain.com/sitemap.xml I need to fill in `/` as path. But the sitemap is created in the root folder now, instead of the pub folder. And when I set `/pub/` as my path I get the URL www.mydomain.com/pub/sitemap.xml returned in the backend. This is an incorrect URL.

antboiko:
Hi @jasperzeinstra Thanks for reporting this issue. We've created an internal ticket MAGETWO-47607 to process it.

navarr:
A good work around for this would be to create a symbolic link pointing to the root sitemap.xml from pub, but if you are creating multiple sitemaps that will get unwieldy fast.

hostep:
Duplicate - https://github.com/magento/magento2/issues/1146

okorshenko:
Fixed in #9094

magento-team:
Internal ticket to track issue progress: MAGETWO-67054

PR #9094, opened 2017-04-02
Issue #2802, #1146: Fixing sitemap generation folder

JosephMaxwell:
### Description
Magento assumes for the above tickets that `robots.txt` and the sitemap paths are to be found in the root directory. To get the sitemap to be generated in the `pub` folder (per Magento's recommendation), you need to set the base directory to be at least `pub/`. This resolves the issue by adding new configuration to `app/etc/env.php`:

```
directories/
    document_root_is_pub (bool)
```

The default keeps backward compatibility: if the value is not specified, original functionality remains (use of the root directory). If you specify `document_root_is_pub` and set it to `true`, the sitemap is located in the correct directory.

### Fixed Issues
This is a fix for:
1. https://github.com/magento/magento2/issues/2802
2. https://github.com/magento/magento2/issues/1146

### Manual testing scenarios
*Before:*
1. Ensure that apache / nginx has the document root set to use `/pub`. 
2. Create a sitemap.
3. Click the generate link.
4. Try to access it: you will get a 404 error.

*After:*
1. Sitemap should be generated, and you will not get a 404 error.

### DIFF
--- a/app/code/Magento/Config/Model/Config/Backend/Admin/Robots.php
+++ b/app/code/Magento/Config/Model/Config/Backend/Admin/Robots.php
@@ -9,8 +9,6 @@
  */
 namespace Magento\Config\Model\Config\Backend\Admin;
 
-use Magento\Framework\App\Filesystem\DirectoryList;
-
 class Robots extends \Magento\Framework\App\Config\Value
 {
     /**
@@ -39,12 +37,13 @@ class Robots extends \Magento\Framework\App\Config\Value
         \Magento\Framework\App\Config\ScopeConfigInterface $config,
         \Magento\Framework\App\Cache\TypeListInterface $cacheTypeList,
         \Magento\Framework\Filesystem $filesystem,
+        \Magento\Config\Model\Config\Reader\Source\Deployed\DocumentRoot $documentRoot,
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
         array $data = []
     ) {
         parent::__construct($context, $registry, $config, $cacheTypeList, $resource, $resourceCollection, $data);
-        $this->_directory = $filesystem->getDirectoryWrite(DirectoryList::ROOT);
+        $this->_directory = $filesystem->getDirectoryWrite($documentRoot->getPath());
         $this->_file = 'robots.txt';
     }
 
--- a/app/code/Magento/Config/Model/Config/Backend/Admin/Robots.php
+++ b/app/code/Magento/Config/Model/Config/Backend/Admin/Robots.php
@@ -9,6 +9,9 @@
  */
 namespace Magento\Config\Model\Config\Backend\Admin;
 
+use Magento\Config\Model\Config\Reader\Source\Deployed\DocumentRoot;
+use Magento\Framework\App\ObjectManager;
+
 class Robots extends \Magento\Framework\App\Config\Value
 {
     /**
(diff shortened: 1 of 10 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9094) <br/>All committers have signed the CLA.

app/code/Magento/Config/Model/Config/Backend/Admin/Robots.php
```diff
@@ -41,10 +42,13 @@ public function __construct(
         \Magento\Framework\Filesystem $filesystem,
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
+        \Magento\Config\Model\Config\Reader\Source\Deployed\DocumentRoot $documentRoot = null,
```
okorshenko: please, update doc block

app/code/Magento/Config/Model/Config/Reader/Source/Deployed/DocumentRoot.php
```diff
@@ -0,0 +1,51 @@
...
+use Magento\Framework\Config\ConfigOptionsListConstants;
+use Magento\Framework\App\Filesystem\DirectoryList;
+
+/**
+ * Class DocumentRoot
+ * @package Magento\Config\Model\Config\Reader\Source\Deployed
+ */
+class DocumentRoot
+{
+    private $config;
```
okorshenko: add doc block here

app/code/Magento/Sitemap/Block/Adminhtml/Grid/Renderer/Link.php
```diff
@@ -34,10 +41,13 @@ public function __construct(
         \Magento\Backend\Block\Context $context,
         \Magento\Sitemap\Model\SitemapFactory $sitemapFactory,
         \Magento\Framework\Filesystem $filesystem,
+        \Magento\Config\Model\Config\Reader\Source\Deployed\DocumentRoot $documentRoot = null,
```
okorshenko: please, update doc block

okorshenko (changes requested):

app/code/Magento/Sitemap/Block/Adminhtml/Grid/Renderer/Link.php
rodrigowebjump: done

app/code/Magento/Config/Model/Config/Reader/Source/Deployed/DocumentRoot.php
rodrigowebjump: done

app/code/Magento/Config/Model/Config/Backend/Admin/Robots.php
rodrigowebjump: done

JosephMaxwell:
@okorshenko: we updated the PR per your comments.

okorshenko (approved):

magento-team:
@JosephMaxwell thank you for your contribution. Your Pull Request has been successfully merged

skovalenk:
@magento/community-gatekeepers  @JosephMaxwell how other developers should learn about this configuration in env.php
ENV.PHP is file that should be generated, so by this logic there should be command which allows to switch root folder.
Or PR to documentation with this feature should be added.

okorshenko:
Hi @sereban 
We are working on documentation for this functionality.
Thank you

avoelkl:
Hi @JosephMaxwell,
I just came across this PR because I was wondering if there was already a fix for the sitemap.xml issue - and there is! Thank you :)

I added the following to my env.php:
```
    'directories' =>
        array(
            'document_root_is_pub' => true
        ),
```

After clicking "Generate" the URL still contains /pub/ but the sitemap works with and without /pub:
- /pub/sitemap/cz/cs/sitemap.xml 
- /sitemap/cz/cs/sitemap.xml

Is this the expected result?

avstudnitz:
@okorshenko What about said documentation? I can't find it...

webspeaks:
@okorshenko Any documentation available? This issue still exists in 2.2.5.

jalogut:
It would be great to find documentation about that in the dev docs. I only found that:

- https://devdocs.magento.com/guides/v2.2/cloud/trouble/robots-sitemap.html

Is the `'document_root_is_pub' => true` needed for something else too?

OvalMedia:
2.2.9 and the issue persists. Nginx/Apache Docroot is set to /pub folder.
Sitemap entry in the robots.txt still looks like this:
"Sitemap: https://www.example.com/pub/sitemap.xml"

document_root_is_pub in env.php does not make any difference.

What am I missing here?

PR #9102, opened 2017-04-02
Improve performances by improving composer autoload

jacquesbh:
Improve the autoload by removing a lot of `file_exists()` calls.

### Description

I just added the classmap element in autoload in `composer.json`.

If the directory doesn't exist composer will complaint about it. So I also added a `.gitkeep` file.
Which for me is a good way to make it work.
It's worth the improvement.

### Fixed Issues

Didn't find a relevant one.

### Manual testing scenarios

Use `composer dumpautoload -o` before the change, load the page (all caches enabled, but works also without).
Then calculate the number of `file_exists` calls.

Make the change, re-run the dump of the autoload, then call your page again.

Don't forget of course to flush the cache between operations.

Here is the comparison with cache enabled: https://blackfire.io/profiles/compare/7a733717-695d-495f-8bfe-0ffc65154894/graph

The `autoload_classmap.php` has +3k lines after this. So, in best scenarios it can avoid 3k `file_exists`.

### Contribution checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -243,6 +243,9 @@
         "files": [
             "app/etc/NonComposerComponentRegistration.php"
         ],
+        "classmap": [
+            "var/generation/Magento/"
+        ],
         "exclude-from-classmap": [
             "**/dev/**",
             "**/update/**",


vrann:
@jacquesbh I noticed that you added ``"var/generation/Magento/"`` to the classpath. The latest develop branch uses ``generated/Magento`` directory instead. Have you tried that on the latest codebase?

jacquesbh:
Hum, not on develop. I'll fix that.

Thanks.

Le ven. 14 avr. 2017 à 22:36, Eugene Tulika <notifications@github.com> a
écrit :

adragus-inviqa:
Looks like the path can be dinamycally changed when calling the compilation CLI command (note the `--generation` option): [link](http://devdocs.magento.com/guides/v2.1/config-guide/cli/config-cli-subcommands-compiler.html#config-cli-subcommands-run). Will/must this PR take that into consideration, too?

jacquesbh:
Ok, if the path can be changed, this PR doesn't apply anymore. But we can
put this tip somewhere in the documentation I think.

It adds more than 3k files in the autoload files of composer. Which result
in less is_file() calls for each page (not 3k obviously).



Jacques Bodin-Hullin

-- 
Monsieur Biz <https://monsieurbiz.com>
@jacquesbh <https://twitter.com/jacquesbh>
+33 (0)6 75 54 11 97
GPG : https://keybase.io/jacquesbh

🦄

Le sam. 15 avr. 2017 à 11:20, adragus-inviqa <notifications@github.com> a
écrit :

ishakhsuvarov:
@jacquesbh Closing per your comment.
Thank you for collaboration.

mslabko:
@jacquesbh , thank you, this is a good improvement. Could you please create PR to https://github.com/magento/devdocs?

mslabko:
Hi @jacquesbh 
I've checked [di:compile](https://github.com/magento/magento2/blob/2.2-develop/setup/src/Magento/Setup/Console/Command/DiCompileCommand.php#L107) command and there are no extra options. Our devdocs is outdated and all described options related to [DiCompileMultiTenantCommand.php](https://github.com/magento/magento2/blob/2.0.9/setup/src/Magento/Setup/Console/Command/DiCompileMultiTenantCommand.php), which available only for 2.0.x version.

It would be really great if you will create a new PR to 2.3-develop (and after merging backport it for  2.2-develop) with proposed changes. 
And I think it makes sense to add a mapping for "generated/code/", so, not only Magento modules will be processed.

Also, it will be very cool, if you update our [devdocs](https://github.com/magento/devdocs/blob/develop/guides/v2.2/config-guide/cli/config-cli-subcommands-compiler.md)  for 2.2.x and 2.3.x  versions.

mslabko:
@jacquesbh what do you think about this? Will you have time to prepare this PR for 2.3-develop branch.
If you don't have time now, would you don't mind if we implement such changes by ourself?

jacquesbh:
Hi @mslabko :)

I put that in my todo list. I will do it. This week probably.

PR #9121, opened 2017-04-04
Remove wrong '_setup' replace when getting DB connection

jalogut:
If a custom database resource contains '_setup', default is returned as resourceName does not match after replacement.

### DIFF
--- a/lib/internal/Magento/Framework/App/ResourceConnection/Config.php
+++ b/lib/internal/Magento/Framework/App/ResourceConnection/Config.php
@@ -63,9 +63,7 @@ class Config extends \Magento\Framework\Config\Data\Scoped implements ConfigInte
     {
         $this->initConnections();
         $connectionName = \Magento\Framework\App\ResourceConnection::DEFAULT_CONNECTION;
-
-        $resourceName = preg_replace("/_setup$/", '', $resourceName);
-
+        
         if (!isset($this->_connectionNames[$resourceName])) {
             $resourcesConfig = $this->get();
             $pointerResourceName = $resourceName;


okorshenko:
@jalogut thank you for your contribution.
Could you please provide more details about your use case?
Currently in Magento 2 we have separate connection for Setup application. Frontend or Backend of the application should not use setup connection to do their logic. Why do you need to user smth_setup connection on frontend?

okorshenko:
Closing this PR due to inactivity

jalogut:
Hi @okorshenko ,

Sorry for the delay in my reply, I did not see you answer.
Our case is simply to add a second custom database into `etc/env.php` like that:

```php
array (
    'table_prefix' => '',
    'connection' => 
    array (
      'custom' => 
      array (
        'host' => 'localhost',
        'dbname' => '<db_custom>',
        'username' => '<user>',
        'password' => '<pass>',
        'engine' => 'innodb',
        'initStatements' => 'SET NAMES utf8;',
        'active' => '1',
      ),
      'default' => 
      array (
        'host' => 'localhost',
        'dbname' => '<db_default>',
        'username' => '<user>',
        'password' => '<pass>',
        'model' => 'mysql4',
        'engine' => 'innodb',
        'initStatements' => 'SET NAMES utf8;',
        'active' => '1',
      ),
    ),
  ),
  'resource' => 
  array (
    'custom_setup' => 
    array (
      'connection' => 'custom',
    ),
    'default_setup' => 
    array (
      'connection' => 'default',
    ),
  ),
```

After that I would expect to use the custom database by resource name `custom_setup`. However, as `_setup` is always replaced, the system does not find any resource with this name. Then, I need to use a different name without `_setup`. From my point of view, there is no need to replace that. In fact, the only reason why `default_setup` database works is because `default` connection is the fall back. When calling resource `default_setup` is also not found but it falls back to `default`.

So my question here. What is exactly the purpose of replacing `_setup`? I checked the code and I do not see any need to have that.

jalogut:
Could you please open the PR again and have another look after my comment?

Thanks

jalogut:
I created a new PR with my last comment. I am not sure if someone has checked out my explanation as the PR was already closed.

- https://github.com/magento/magento2/pull/9726

PR #9124, opened 2017-04-04
Solve issues with API

paales:
### Description
Our team is working on a new API service layer (GraphQL) that intensively uses the API retrospection classes and noticed that there are some inconsistencies in the API

- Some comment sort orders aren't correct, fixed for `ProductAttributeMediaGalleryEntryInterface`, `ConfigurableProductManagementInterface`.
- Replaced di injections and argument types of `SearchCriteria` with `SearchCriteriaInterface`

### DIFF
--- a/app/code/Magento/Catalog/Api/Data/ProductAttributeMediaGalleryEntryInterface.php
+++ b/app/code/Magento/Catalog/Api/Data/ProductAttributeMediaGalleryEntryInterface.php
@@ -138,7 +138,7 @@ interface ProductAttributeMediaGalleryEntryInterface extends ExtensibleDataInter
     /**
      * Set media gallery content
      *
-     * @param $content \Magento\Framework\Api\Data\ImageContentInterface
+     * @param \Magento\Framework\Api\Data\ImageContentInterface $content
      * @return $this
      */
     public function setContent($content);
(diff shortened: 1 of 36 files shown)


generated/.htaccess
```diff
@@ -1,2 +0,0 @@
-Order deny,allow
-Deny from all
```
ishakhsuvarov: What is the reason to remove `Deny` from the `generated` direcotry?

ishakhsuvarov (changes requested):

generated/.htaccess
paales: That, uhm, shouldn't be removed..  deleted the generated folder out of habit I think.

paales: @ishakhsuvarov Changed the PR.

ishakhsuvarov (approved):

ishakhsuvarov:
@paales Did you use `Replace in Path` or created some automated tool to achieve the purpose?

paales:
Hi @ishakhsuvarov 

I did a find and replace at some point. The functionality was validated as it generates a complete schema of all the complete service layer in GraphQL and the endpoints now work as expected :)

ishakhsuvarov:
Hi @paales 
We are still in the process of review, as potentially this PR may cause backwards compatibility issue for some of the 3rd party extensions. We will keep you updated.
Thanks.

paales:
Understandable :)

vrann:
@paales I've been researching GraphQL for Magento myself and found it very promising for the next generation of the APIs. Would be curious to see your perspective on this subject! If there is anything you want to share, we can connect via slack.

paales:
Hi @vrann, sure! I've send you a message on Slack (on the super secret Magento chat)

magento-team:
@paales thank you for your contribution

PR #9125, opened 2017-04-04
Add getShippingAddress() to order interface (resolves #6919)

### LINKED ISSUE #6919: API improvement: OrderInterface should have a getShippingAddress method
As a developer who writes an order export for Magento2 you would expect that besides the getBillingAddress there would be also a getShippingAddress method. While all the [g/s]etShipping... methods exists in the Magento\Sales\Model\Order model (which is the actual implementation of the OrderInterface) those elementary shipping-address methods are not part of the API interface.

Example: $order is declared as type Magento\Sales\Api\Data\OrderInterface. Practically the code "$order->getShippingAddress()" works but from a declarative view the method does not exist in the API and the implementation therefore is technically wrong.

While it is understood that not all orders have a shipping address (virtual orders/goods) the majority of orders processed in ecommerce actually have a shipping address. So the shipping-address methods should become part of the OrderInterface like it is already the case for the billing-address methods.

schmengler:
### Description

Adding `getShippingMethod()` to the order interface, analogous to the existing `getBillingMethod()`.

1. magento/magento2#6919: API improvement: OrderInterface should have a getShippingAddress method #6919

### Manual testing scenarios

1. GET an order via API
2. See if there is a "shipping_address"

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Api/Data/OrderInterface.php
+++ b/app/code/Magento/Sales/Api/Data/OrderInterface.php
@@ -1529,6 +1529,13 @@ interface OrderInterface extends \Magento\Framework\Api\ExtensibleDataInterface
     public function getBillingAddress();
 
     /**
+     * Gets the shipping address, if any, for the order.
+     *
+     * @return \Magento\Sales\Api\Data\OrderAddressInterface|null Billing address. Otherwise, null.
+     */
+    public function getShippingAddress();
+
+    /**
      * Sets the billing address, if any, for the order.
      *
      * @param \Magento\Sales\Api\Data\OrderAddressInterface $billingAddress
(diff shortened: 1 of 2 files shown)


Vinai:
Since it is a data object, each getter should have a matching setter.

adragus-inviqa:
@Vinai - Dunno. I wouldn't want setters in my interfaces. They invite mutability. Let the implementation figure it out, not the interface. You shouldn't really know or care that the implementation is a data object anyway.

schmengler:
For consistency, I added the setter as well

ishakhsuvarov:
@schmengler Unfortunately, due to backward compatibility policy we can not accept this Pull Request now, as it modifies an interface.
Please use `ShippingAssignment` extension attribute to process shippingAddress.
Thanks.

PR #9129, opened 2017-04-04
Product Wizard: Use result type Layout instead of page layout

klein0r:
To avoid unnecessary removal of default layout handle and to fit type hint

### Description

It doesn't make sense at all to use a page layout and remove the default handle. You can get the same result by using the Layout type directly.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9129) <br/>All committers have signed the CLA.

magento-team:
@klein0r thank you for your contribution to Magento 2 project

PR #9131, opened 2017-04-04
Fix to allow Zend_Db_Expr as column default

scottsb:
### Description
Currently if you attempt to use a `Zend_Db_Expr` as the default value for a column, it fails because it is accidentally converted to a string too early, which results in a malformed SQL query.

This is due to a check that attempts to avoid double quoting but isn't careful to ensure the passed value is a string, triggering an implicit type conversion.

### Manual testing scenarios
Create a column in a setup script similar to this:

        /** @var \Magento\Framework\Setup\SchemaSetupInterface $setup */
        $setup->getConnection()->addColumn(
            'test_table',
            'test_column',
            [
                'type' => Table::TYPE_DATETIME,
                'default' => new \Zend_Db_Expr('CURRENT_TIMESTAMP'),
                'comment' => 'Test column comment'
            ]
        );

Before this PR that fails with the exception:

> Syntax error or access violation: 1067 Invalid default value for 'test_column'

After this PR the column is created correctly.


magento-team:
@scottsb thank you for your contribution

PR #9140, opened 2017-04-05
remove duplicate calls to initObjectManager in bootstrap class

sivajik34:
There are duplicate calls for initobjectManager method,i think we can initialize objectmanager in constructor,so that object manager class will be available at the time of bootstrap object creation.

### DIFF
--- a/lib/internal/Magento/Framework/App/Bootstrap.php
+++ b/lib/internal/Magento/Framework/App/Bootstrap.php
@@ -373,20 +373,7 @@ class Bootstrap
     {
         $handler = new ErrorHandler();
         set_error_handler([$handler, 'handler']);
-    }
-
-    /**
-     * Initializes object manager
-     *
-     * @return void
-     */
-    private function initObjectManager()
-    {
-        if (!$this->objectManager) {
-            $this->objectManager = $this->factory->create($this->server);
-            $this->maintenance = $this->objectManager->get(\Magento\Framework\App\MaintenanceMode::class);
-        }
-    }
+    }   
 
     /**
      * Getter for error code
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9140) <br/>All committers have signed the CLA.

lib/internal/Magento/Framework/App/Bootstrap.php
```diff
@@ -279,7 +279,6 @@ protected function assertMaintenance()
         if (null === $isExpected) {
             return;
         }
-        $this->initObjectManager();
         /** @var \Magento\Framework\App\MaintenanceMode $maintenance */
         $this->maintenance = $this->objectManager->get(\Magento\Framework\App\MaintenanceMode::class);
```
Bartlomiejsz, line 282: I think this could also be removed :) Maintenance is now also initialized in constructor.

sivajik34: @Bartlomiejsz  i think we can remove  Maintenance initialization  from constructor .
I don't know why Maintenance initialization happens in two times (assertMaintenance method,initObjectManager method)
i think Its enough in assertMaintenance.
anyway I'm looking for core team feedback?

ishakhsuvarov: @sivajik34 @Bartlomiejsz To me it looks like it is enough only once. Seems to be some artifacts from legacy code. Going to look into it more deeply.

okorshenko:
@sivajik34 thank you for your contribution!

PR #7417, opened 2016-11-13
CMS Page and Block - Load by identifier and store id - fix issue 6570

convenient:
The following should fix issue https://github.com/magento/magento2/issues/6570

This should allow developers to access a cms block or a cms page by identifier and store ID.

This change should be backwards compatible, if no `storeId` is passed in then the functionality will work as it previously did.

If a `storeId` is passed in then it should get picked up by the logic in `Magento\Cms\Model\ResourceModel\Page::_getLoadSelect` and `Magento\Cms\Model\ResourceModel\Block::_getLoadSelect`.

A bit unrelated, but PHPStorm notified me of a bug which I fixed in commit 594871f648c50dadd92adbf92627e36da6cd41e9

Issue #6570, opened 2016-09-11
There is no ability to pass store ID during getting CMS page/block by identity.


ytorbyk:
### Preconditions
1. There could be more then one CMS block/page with the same identity (for each store view).
### Expected result
1. There must be ability to get CMS block/page by identity for specific store view.
### Actual result

`\Magento\Cms\Api\PageRepositoryInterface::getById($pageId)`
`\Magento\Cms\Api\BlockRepositoryInterface::getById($blockId)`
1. Both methods return the first found item in DB with passed identity.

davidalger:
The GitHub issue tracker is intended for technical issues only. Please refer to the [Community Forums](https://community.magento.com) or [Magento Stack Exchange](http://magento.stackexchange.com/) site for technical questions. In your case, the [programming questions forum](https://community.magento.com/t5/Programming-Questions/bd-p/programming-magento-2) is likely the most appropriate. Feel free to reopen this issue if you think you have encountered a bug in Magento 2.

ytorbyk:
@davidalger Magento 2 doesn't provide ability to get CMS block/page. Isn't it a bug?

ytorbyk:
Sorry, my bad. Poor explanation  

### Preconditions

Magento 2.1.1

### Steps to reproduce
1. Create at least two store views
2. Create CMS block for each of the store views with the same identity
3. Go to template/block (no matter which one)
4. Call \Magento\Cms\Api\BlockRepositoryInterface::getById($blockId) for retrieving CMS block

### Expected result
1. You have ability to get CMS block by identity for specific store ID

### Actual result
1. The interface method requires only identity and no ability to pass store ID
2. The method returns the first CMS block found in DB

SerhiyShkolyarenko:
reopened issue due to updated steps.

veloraven:
I'm closing this issue as it was fixed in #7218.
If you can still reproduce this issue, please create a new GitHub issue report.

convenient:
@veloraven this issue should not have been close IMO.

- this issue says you cannot load a specific cms block by identifier, when that identifier is shared between multiple store views. 
- #7218 affected **saving** of cms blocks allowing you greater control over the `store_id`
- Pending PR #7417 should fix this issue, and allow you to specify both a `store_id` and `identifier` combination to allow for granular **loading** of cms blocks.

Just for clarity 👍

magento-team:
Internal ticket to track issue progress: MAGETWO-66748

[...]

PR #9163, opened 2017-04-07
[Cms] Added a getByIdentifier() method to the BlockRepository

enrico69:


adragus-inviqa:
So you never tried loading a block via its identifier? Try it. [It already works](https://github.com/magento/magento2/blob/5640494293ed26876700bb7c5f9578132bc5b0e6/app/code/Magento/Cms/Model/ResourceModel/Block.php#L108-L112).

app/code/Magento/Cms/Api/BlockRepositoryInterface.php
```diff
@@ -32,6 +32,16 @@ public function save(Data\BlockInterface $block);
...
 
     /**
+     * Load Block data by given Block identifier
+     *
+     * @param string $blockIdentifier
+     * @param int $storeId
+     * @return \Magento\Cms\Api\Data\BlockInterface
+     * @throws \Magento\Framework\Exception\NoSuchEntityException
+     */
+    public function getByIdentifier($blockIdentifier, $storeId);
```
ishakhsuvarov: Unfortunately, due to backwards compatibility policy, currently we can not add any new methods to the interfaces, specially the ones marked with the `@api` annotation. You may extract this method to a separate interface/class, if it is really needed.

ishakhsuvarov (changes requested):

app/code/Magento/Cms/Api/BlockRepositoryInterface.php
enrico69: What do you suggest? Should I simply remove this part from the interface and let the implementation in the model ( `app/code/Magento/Cms/Model/ResourceModel/Block.php` ) ?

adragus-inviqa: Actually, your PR is redundant, as `Magento\Cms\Api\BlockRepositoryInterface::getById('my-block-identifier')` already works. As is `Magento\Cms\Api\BlockRepositoryInterface::getById(12)`, too. 

The resource model I showed you is just the place where the logic decides which route to go: identifier or numeric id? Again, try it yourself.

enrico69: Yes but there is also this point, given by the user _convenient_

> Might be useful to be aware of #6570 and #7417 in which identifier is not enough to update cms blocks by identifier when that identifier is shared across multiple stores.

enrico69 (commented):
Done

enrico69:
@ishakhsuvarov changes are completed.

okorshenko:
Hi @enrico69 
Thank you for your contribution. Unfortunately, we can not accept this PR. I agree with the point of @adragus-inviqa . Looks like this will be duplication of the functionality.
Thank you for your efforts.

Just FYI: we are working on https://github.com/magento/magento2/pull/7417  It should resolve your issue

PR #9165, opened 2017-04-07
Improved text of exception message in case of error in module's composer.json

vovayatsyuk:
### Description

When some third-party module has syntax error inside its composer.json, it's very hard to determine which
module caused an exception. 

This PR changes default message with the following:

```
Some_Module's composer.json error: Decoding failed: Syntax error
```

where `Decoding failed: Syntax error` - is a message of the original exception

### Manual testing scenarios

> Sample data must be installed to test this behavior.
1. Make an error in `app/code/Magento/Cms/composer.json`:
2. Run `php bin/magento sampeladata:reset`. You'll receive a message:

    ```
    [Zend_Json_Exception]          
    Decoding failed: Syntax error
    ```

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Module/PackageInfo.php
+++ b/lib/internal/Magento/Framework/Module/PackageInfo.php
@@ -82,7 +82,15 @@ class PackageInfo
             foreach ($this->componentRegistrar->getPaths(ComponentRegistrar::MODULE) as $moduleName => $moduleDir) {
                 $key = $moduleDir . '/composer.json';
                 if (isset($jsonData[$key]) && $jsonData[$key]) {
-                    $packageData = \Zend_Json::decode($jsonData[$key]);
+                    try {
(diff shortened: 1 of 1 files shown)


adragus-inviqa:
I wouldn't concat any other string with the module name; it'd keep it completely separate, to make it more obvious.
Something like`Error decoding composer.json for module Some_Module`.

vovayatsyuk:
I'd like to keep original exception message without change, because there are multiple possible variants for it. So we need a template that will suite for all kind of messages.

adragus-inviqa:
I'm not sure what you mean by "a template that will suite for all kind of messages"; it just has to be a good format. 

But your recent change looks good. I just didn't like the interpolation: `$moduleName's`.

magento-team:
@vovayatsyuk thank you for your contribution. Your Pull Request has been successfully merged

PR #9210, opened 2017-04-12
Do not di:compile tests/ folder

kassner:
### Description

`setup:di:compile` failes when you have `magento/data-migration-tool` as a regular dependency and you haven't installed the dev dependencies with composer (--no-dev).

### Manual testing scenarios

1. Install dependencies using `composer install --no-dev`
2. `composer require magento/data-migration-tool:2.1.4` and enable the module
3. `php bin/magento setup:di:compile`

### Current result

    PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in /var/www/html/vendor/magento/data-migration-tool/tests/static/testsuite/Migration/Php/LiveCodeTest.php on line 19

### Expected result

`setup:di:compile` runs without failures.


magento-team:
@kassner thank you for your contribution

PR #9215, opened 2017-04-12
Remove unused and invalid method

adragus-inviqa:
a) there is no `$this->_getDomConfigModel()` method
b) method is unused

### DIFF
--- a/lib/internal/Magento/Framework/Config/View.php
+++ b/lib/internal/Magento/Framework/Config/View.php
@@ -134,16 +134,6 @@ class View extends \Magento\Framework\Config\Reader\Filesystem
     }
 
     /**
-     * Return copy of DOM
-     *
-     * @return \Magento\Framework\Config\Dom
-     */
-    public function getDomConfigCopy()
-    {
-        return clone $this->_getDomConfigModel();
-    }
-
-    /**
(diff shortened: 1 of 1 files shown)


sivajik34:
@adragus-inviqa  may be we can add deprecated, without deleting the method.

adragus-inviqa:
Yeah, I suppose we could, but I looked for its usage, and couldn't find one. If you deprecate it, it's still callable. But calling it would give you a fatal error, as there's no `$this->_getDomConfigModel()` method.

I'd rather we get rid of it.

vrann:
@adragus-inviqa per our backward compatibility policy method should deprecate method first and then it will be removed in the next release. Please deprecate it.

adragus-inviqa:
Ok, I will. But what about the fact that people can still call it, and trigger a fatal error? Shouldn't I remove the method body at least?

PS: Do you actually remove deprecated stuff in the subsequent release? 🤔

vrann:
@adragus-inviqa ah, good point. If there is fatal exception from using it, then better to remove.

magento-team:
@adragus-inviqa thank you for your contribution. Your Pull Request has been successfully merged

PR #9189, opened 2017-04-10
Avoid duplicate ltrim function on not complied mode.

sivajik34:
Case 1: not complied
1.Magento\Framework\ObjectManager\ObjectManager -> create method will ltrim $type.
2.Magento\Framework\ObjectManager\Config\Config -> getPreference method again ltrim $type.
so in order to avoid duplicate ltrim calls i removed from one file,added to necessary file.

PR #9217, opened 2017-04-12
Optimizing getting parents by given type

sivajik34:
We can optimize type's  parents info  by storing into an array,so that next time onwards we can return this array with out calculating whole logic.


adragus-inviqa:
You shouldn't use underscores for private/protected members.

Do you have some numbers? How much time does this save?

sivajik34:
@adragus-inviqa 
Magento core team using underscores.
i don't know exactly how much time it will save or not.maybe useful to store into array without calculating again whole thing.

adragus-inviqa:
Yeah, they shouldn't be using underscores. Must be some legacy code. But I'll wait for them to confirm.

Whenever you optimise stuff, you need to come up with numbers, to actually _prove_ that your change is needed. Is it saving a fraction of a second or 2 minutes? 

Every refactoring can introduce bugs, so your PR needs to be justified. With numbers.

sivajik34:
@adragus-inviqa  i agree. can you give some idea to check the improvement.
I mean some tools,script to get actual numbers?
i think i can use profiler.

lib/internal/Magento/Framework/Code/Reader/ClassReader.php
```diff
@@ -8,6 +8,11 @@
 class ClassReader implements ClassReaderInterface
 {
     /**
+     * @var array
+     */
+    protected $_classNameParents;
```
ishakhsuvarov: Please avoid using underscores in property names, it is a legacy standard.
There is no need for this property to be protected as we do not encourage inheritance-based APIs

lib/internal/Magento/Framework/Code/Reader/ClassReader.php
```diff
@@ -56,27 +61,32 @@ public function getConstructor($className)
...
-                if ($parentInterfaces) {
-                    $result = array_values(array_diff($interfaces, $parentInterfaces));
-                } else {
-                    $result = array_values($interfaces);
-                }
-            }
-            array_unshift($result, $parentClass);
+        if (isset($this->_classNameParents[$className])) {
+            return $this->_classNameParents[$className];
         } else {
```
ishakhsuvarov: Looks like there is no more need for the `else`

lib/internal/Magento/Framework/ObjectManager/Config/Compiled.php
```diff
@@ -120,6 +120,7 @@ public function getInstanceType($instanceName)
      */
     public function getPreference($type)
     {
+        $type = ltrim($type, '\\');
```
ishakhsuvarov, line 123: This does not look like a part of in-object cache, which is the purpose of this PR

lib/internal/Magento/Framework/ObjectManager/ObjectManager.php
```diff
@@ -53,7 +53,6 @@ public function __construct(FactoryInterface $factory, ConfigInterface $config,
      */
     public function create($type, array $arguments = [])
     {
-        $type = ltrim($type, '\\');
         return $this->_factory->create($this->_config->getPreference($type), $arguments);
```
ishakhsuvarov, line 56: This does not look like a part of in-object cache, which is the purpose of this PR

lib/internal/Magento/Framework/Code/Reader/ClassReader.php
```diff
@@ -56,27 +61,32 @@ public function getConstructor($className)
...
-                $parentInterfaces = class_implements($parentClass);
-                if ($parentInterfaces) {
-                    $result = array_values(array_diff($interfaces, $parentInterfaces));
-                } else {
-                    $result = array_values($interfaces);
-                }
-            }
-            array_unshift($result, $parentClass);
+        if (isset($this->_classNameParents[$className])) {
+            return $this->_classNameParents[$className];
```
ishakhsuvarov: Please consider updating tests to verify this case.

ishakhsuvarov (changes requested):

ishakhsuvarov:
@sivajik34 Thank you for the PR. Idea looks good overall, however please check review comments.
Some numbers would be good as well.

lib/internal/Magento/Framework/ObjectManager/Config/Compiled.php
adragus-inviqa: I think OP has mixed up branches a bit: https://github.com/magento/magento2/pull/9189#issuecomment-292954874

lib/internal/Magento/Framework/Code/Reader/ClassReader.php
sivajik34: @ishakhsuvarov so can i use private $classNameParents?

lib/internal/Magento/Framework/ObjectManager/ObjectManager.php
sivajik34: @ishakhsuvarov can i remove these changes and sent only changes related to this PR?

lib/internal/Magento/Framework/Code/Reader/ClassReader.php
ishakhsuvarov: Yes, that is what I meant :)

lib/internal/Magento/Framework/ObjectManager/ObjectManager.php
ishakhsuvarov: Yes, that would be perfect.

ishakhsuvarov:
Hi @sivajik34 
Are you going to proceed with this PR?
Thanks!

sivajik34:
@ishakhsuvarov if its not that much helpful we can close it.no issues.

ishakhsuvarov:
@sivajik34 
Ok. Thank you for collaboration

Issue #8202, opened 2017-01-19
Cannot save attribute


maximehuran:
### Preconditions

1. Have an attribute with a source model Magento\Eav\Model\Entity\Attribute\Source\Boolean
2. Have it's frontend_input with "select".

### Steps to reproduce

1. Go in backend and try to save the attribute

### Expected result

1. The attribute must be saved with success

### Actual result

1. Error 500 : 
```
1 exception(s):
Exception #0 (Exception): Notice: Undefined index: value in /var/www/preprod/vendor/magento/module-catalog/Controller/Adminhtml/Product/Attribute/Validate.php on line 155

Exception #0 (Exception): Notice: Undefined index: value in /var/www/preprod/vendor/magento/module-catalog/Controller/Adminhtml/Product/Attribute/Validate.php on line 155
#0 /var/www/preprod/vendor/magento/module-catalog/Controller/Adminhtml/Product/Attribute/Validate.php(155): Magento\Framework\App\ErrorHandler->handler(8, 'Undefined index...', '/var/www/prepro...', 155, Array)
#1 /var/www/preprod/vendor/magento/module-catalog/Controller/Adminhtml/Product/Attribute/Validate.php(108): Magento\Catalog\Controller\Adminhtml\Product\Attribute\Validate->checkUniqueOption(Object(Magento\Framework\DataObject), Array)
#2 /var/www/preprod/var/generation/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Validate/Interceptor.php(24): Magento\Catalog\Controller\Adminhtml\Product\Attribute\Validate->execute()
#3 /var/www/preprod/vendor/magento/framework/App/Action/Action.php(102): Magento\Catalog\Controller\Adminhtml\Product\Attribute\Validate\Interceptor->execute()
#4 /var/www/preprod/vendor/magento/module-backend/App/AbstractAction.php(226): Magento\Framework\App\Action\Action->dispatch(Object(Magento\Framework\App\Request\Http))
#5 /var/www/preprod/vendor/magento/module-catalog/Controller/Adminhtml/Product/Attribute.php(79): Magento\Backend\App\AbstractAction->dispatch(Object(Magento\Framework\App\Request\Http))
#6 /var/www/preprod/vendor/magento/framework/Interception/Interceptor.php(74): Magento\Catalog\Controller\Adminhtml\Product\Attribute->dispatch(Object(Magento\Framework\App\Request\Http))
[...]

PR #9221, opened 2017-04-12
Avoid: Undefined index: value in app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Validate.php on line 157 in Ajax return

mhauri:
When saving an Attribute and the array $options doesn't contain a key named 'value' or 'delete' it will return a Notice Warning instead of the expected ajax result in `admin/catalog/product_attribute/validate/attribute_id/XXX/` 

### Description
This PR will check if the keys 'value' and 'delete' are set.

1. #8202: Cannot save attribute
2. #8271: Visibility save attribute error

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Validate.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Validate.php
@@ -154,11 +154,11 @@ class Validate extends \Magento\Catalog\Controller\Adminhtml\Product\Attribute
      */
     private function checkUniqueOption(DataObject $response, array $options = null)
     {
-        if (
-            is_array($options)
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9221) <br/>All committers have signed the CLA.

adragus-inviqa:
Hi. Appreciate the time, but there's a small suggestion I'd like to make. The style should be:
```php
if (foo &&
    bar &&
    baz) {
//  ...
}
```
For the life of me I can't find the style guide (I was looking in PSR-2), but I'm sure I can find it somewhere. 😬

magento-team:
@mhauri Thank you for the contribution

PR #9250, opened 2017-04-14
Fixed coding standard violations in the Framework\Archive namespace

dverkade:
Fixed coding standard violations in the Framework\Archive namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes: 

- Removed @codingStandardsIgnoreFile from the file still containing this at the head of the file
- Fixed indentation
- Removed unused "use" namespace at the top of the file.


magento-team:

PR #9254, opened 2017-04-14
Fixed coding standard violations in the Framework\Encryption namespace

dverkade:
Fixed coding standard violations in the Framework\Encryption namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Removed @codingStandardsIgnoreFile
- Fixed indentation


magento-team:

PR #9258, opened 2017-04-15
Fixed coding standard violations in the Framework\Filesystem namespace

dverkade:
Fixed coding standard violations in the Framework\Filesystem namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Removed @codingStandardsIgnoreFile
- Removed the is_null function in order to do a normal compare.


magento-team:

PR #9260, opened 2017-04-15
Fixed coding standard violations in Framework\HTTP namespace:

dverkade:
Fixed coding standard violations in the Framework\HTTP namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Removed @codingstandardsIgnoreFile from header of the file.
- Corrected line length


magento-team:
@dverkade thank you for your contribution. Your Pull Request has been successfully merged

PR #9262, opened 2017-04-15
Remove zend json from theme

dmanners:
### Description
Since Zend Framework1 is EOL we should replace the usages. In this PR I replace the usage of Zend_Json in Magento Theme with the new serializer class. Part of the #9236 task.

1.   magento/magento2#9236 
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Theme/Block/Adminhtml/Wysiwyg/Files/Tree.php
+++ b/app/code/Magento/Theme/Block/Adminhtml/Wysiwyg/Files/Tree.php
@@ -23,19 +23,29 @@ class Tree extends \Magento\Backend\Block\Template
     protected $urlEncoder;
 
     /**
+     * @var \Magento\Framework\Serialize\Serializer\Json
+     */
+    private $serializer;
+
+    /**
      * @param \Magento\Backend\Block\Template\Context $context
      * @param \Magento\Theme\Helper\Storage $storageHelper
      * @param \Magento\Framework\Url\EncoderInterface $urlEncoder
      * @param array $data
+     * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @throws \RuntimeException
      */
     public function __construct(
         \Magento\Backend\Block\Template\Context $context,
         \Magento\Theme\Helper\Storage $storageHelper,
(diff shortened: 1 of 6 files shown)


adragus-inviqa:
All this looks ok, but I was just wondering: seeing how the M2 JSON serializer [is only a dumb wrapper over json decode/encode](https://github.com/magento/magento2/blob/e3fc5b793af60c354341ca877832bfc06736308c/lib/internal/Magento/Framework/Serialize/Serializer/Json.php), how would we handle encoding/decoding errors? We'd have to use `json_last_error()` now? At least Zend_Json had exceptions...

I'm saying this, because this PR removes a try/catch block that would've been otherwise useful. And said try/catch block was around a _cookie_ read no less. Aka user input. You should never trust that. Imagine `json_decode($_GET['param'])` and _not_ checking for errors. Not cool.

dmanners:
Yup @adragus-inviqa I was thinking about something similar to Zend https://github.com/zendframework/zend-json/blob/3426c66e689120a384a370df2cf35745f1b15b2c/src/Json.php#L230 or symfony https://github.com/symfony/serializer/blob/80e172d62cdac03df4a0fd8bcef8513bf9d2aa83/Encoder/JsonDecode.php#L80 and then process the exceptions.

adragus-inviqa:
Well, yes, that's what I mean. But that also means that the behaviour of the serializer must change, breaking compatibility, because it doesn't throw at the moment. 

And since it's a BC break, they won't allow it. Your best option is a _new_ `SerializeWithThrow` decorator or something, wrapping the original one. That should also be a new PR, btw.

----

Or maybe we can add a new method to the existing Serializer called `getErrors()` or something? This way, we won't be breaking BC. I'm not sure about this way of doing things, as people might forget to check the errors, but it's an alternative.

dmanners:
Yeah, I think having another class specifically to deal with json throwing an exception is a bit of overkill. I will put together a PR for the serializer with the exception handling and see what the core Magento team say.

okorshenko:
@dmanners @adragus-inviqa thank you for raising the issue with Exceptions. This is known issue for us and Magento Core Team is looking for proper solution right now.

magento-team:
@dmanners Thank you for the contribution

Issue #9200, opened 2017-04-11
Create new CLI command: Enable DB logging


vrann:
In order to help developers to work with the Magento, create new commands which will simplify the debugging of the application. Currently, enabling of these setting require doing that either with the UI or changing the configuration of the di.xml.

Create CLI command which will make possible to enable/disable DB query logging

ClodClod:
I think could be very useful

vrann:
@federivo volunteered for this contribution.
I would assume the command will be implemented in ``\Mageno\Developer`` module. As an example of unit tests for such command, you can check tests in the following directory: app/code/Magento/Developer/Test/Unit/Console/Command

federivo:
Great @vrann , thank you. 
I'll briefly describe my thoughts on how to implement in a bit.

federivo:
Hey @vrann

I would love your feedback on these.


- Enable SQL queries logging command:

I see this one as a setting in the core_config_data table as much as the "dev/debug/template_hints_storefront" setting for template hints.
So the command should set that setting as true or false depending on current value and we'll need to check for that value in the `getStats` method in `Magento\Framework\DB\Logger\LoggerAbstract` in order to return the log data or an empty string.

One thing to consider is that the user will still need to change the preference for the dependency. In my point of view, if the command somehow set which implementation of the DB logger to use it will override the user preference who may be using `Magento\Framework\DB\Logger\File` or his own implementation. So the user should set the preference for the logger and then can enable and disable it by using the command.
Maybe we can check if the preference set in the di.xml is `Magento\Framework\DB\Logger\Quiet` and send a warning to the user?

Thoughts?


- Enable / disable template hints:

This should change the setting "dev/debug/template_hints_storefront" in the core_config_data table. Set to true if it is false and viceversa.


- Enable / disable magento profiler:

I'm in doubt with this one. Should this enable / disable the outcome of this: http://devdocs.magento.com/guides/v2.1/config-guide/db-profiler/db-profiler.html ?

Thanks!

vrann:
@federivo Hi, can you join the slack team https://magentocommeng.slack.com/, I created the room to discuss this work.

Let's attack the first issue: query logging.
There are two problems which I see
- there is no way dynamically change DI configuration by the CLI command. 
- the configuration for the query logging cannot really be stored in the database because you need it before the classes used to perform database connection are created

[...]

PR #9264, opened 2017-04-15
Enable/Disable DB query logging commands

federivo:
### Description

Added console commands for enabling/disabling the db query logging.

1. magento/magento2#9200: Create new CLI command: Enable dev settings, i.e DB logging, Profiler, Template Hints

### Manual testing scenarios

1. Run bin/magento dev:query-log:enable command.
2. Clear configuration cache.
3. Make any request to the Magento site and verify that the DB query log is being written.
4. Run bin/magento dev:query-log:disable command.
5. Clear configuration cache
6. Make any request to the site and verify that the DB query log have no new log data.

### DIFF
--- a/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
+++ b/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
@@ -21,6 +21,11 @@ class QueryLogDisableCommand extends Command
     const COMMAND_NAME = 'dev:query-log:disable';
 
     /**
+     * Success message
+     */
+    const SUCCESS_MESSAGE = "DB query logging disabled.";
+
+    /**
      * @var Writer
      */
     private $deployConfigWriter;
@@ -58,6 +63,6 @@ class QueryLogDisableCommand extends Command
         $data = [LoggerProxy::PARAM_ALIAS => LoggerProxy::LOGGER_ALIAS_DISABLED];
         $this->deployConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
 
-        $output->writeln("<info>DB query logging disabled.</info>");
+        $output->writeln("<info>". self::SUCCESS_MESSAGE . "</info>");
     }
 }
--- a/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
+++ b/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Developer\Console\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Magento\Framework\App\DeploymentConfig\Writer;
+use Magento\Framework\Config\File\ConfigFilePool;
+
+class QueryLogDisableCommand extends Command
+{
+    /**
+     * command name
+     */
+    const COMMAND_NAME = 'dev:query-log:disable';
+
+    /**
+     * File logger alias
+     */
+    const QUIET_LOGGER_ALIAS = 'quiet';
+
+    /**
+     * @var Writer
+     */
+    private $deploymentConfigWriter;
+
+    public function __construct(
+        Writer $deploymentConfigWriter,
+        $name = null
+    )
+    {
+        parent::__construct($name);
+        $this->deploymentConfigWriter = $deploymentConfigWriter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this->setName(self::COMMAND_NAME)
+            ->setDescription('Disable DB query logging');
+
+        parent::configure();
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws \InvalidArgumentException
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $data = ['db_logger_alias' => 'quiet'];
+        $this->deploymentConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
+
+        $output->writeln("<info>DB query logging disabled.</info>");
+    }
+}
--- a/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
+++ b/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
@@ -61,7 +61,7 @@ class QueryLogDisableCommand extends Command
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $data = [LoggerProxy::PARAM_ALIAS => LoggerProxy::LOGGER_ALIAS_DISABLED];
-        $this->deployConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
+        $this->deployConfigWriter->saveConfig([ConfigFilePool::APP_ENV => [LoggerProxy::CONF_GROUP_NAME => $data]]);
 
         $output->writeln("<info>". self::SUCCESS_MESSAGE . "</info>");
     }
--- a/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
+++ b/app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
@@ -7,11 +7,11 @@
 namespace Magento\Developer\Console\Command;
 
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Magento\Framework\App\DeploymentConfig\Writer;
 use Magento\Framework\Config\File\ConfigFilePool;
+use Magento\Framework\DB\Logger\LoggerProxy;
 
 class QueryLogDisableCommand extends Command
 {
@@ -21,22 +21,21 @@ class QueryLogDisableCommand extends Command
     const COMMAND_NAME = 'dev:query-log:disable';
 
     /**
-     * File logger alias
-     */
-    const QUIET_LOGGER_ALIAS = 'quiet';
-
-    /**
      * @var Writer
      */
-    private $deploymentConfigWriter;
+    private $deployConfigWriter;
 
+    /**
+     * QueryLogDisableCommand constructor.
+     * @param Writer $deployConfigWriter
+     * @param null $name
+     */
     public function __construct(
-        Writer $deploymentConfigWriter,
+        Writer $deployConfigWriter,
         $name = null
-    )
-    {
+    ) {
         parent::__construct($name);
-        $this->deploymentConfigWriter = $deploymentConfigWriter;
+        $this->deployConfigWriter = $deployConfigWriter;
     }
 
     /**
@@ -56,8 +55,8 @@ class QueryLogDisableCommand extends Command
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $data = ['db_logger_alias' => 'quiet'];
-        $this->deploymentConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
+        $data = [LoggerProxy::PARAM_ALIAS => LoggerProxy::LOGGER_ALIAS_DISABLED];
+        $this->deployConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
 
         $output->writeln("<info>DB query logging disabled.</info>");
     }
--- a/app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
+++ b/app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
@@ -107,7 +107,9 @@ class QueryLogEnableCommand extends Command
         $data[LoggerProxy::PARAM_QUERY_TIME] = number_format($logQueryTime, 3);
         $data[LoggerProxy::PARAM_CALL_STACK] = (int)($logCallStack != 'false');
 
-        $this->deployConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
+        $configGroup[LoggerProxy::CONF_GROUP_NAME] = $data;
+
+        $this->deployConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $configGroup]);
 
         $output->writeln("<info>". self::SUCCESS_MESSAGE . "</info>");
     }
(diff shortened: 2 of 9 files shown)


app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
```diff
@@ -0,0 +1,64 @@
...
+        parent::configure();
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws \InvalidArgumentException
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $data = ['db_logger_alias' => 'quiet'];
```
vrann: - please use constants for db_logger_alias and quiet
- db_logger_alias is not the best name for the config parameter, because we used it to set the mode of the logger: "writing to file" vs "quiet". I would suggest something like "db logger mode"
- for the same reason, "quiet" might be renamed to "disabled"

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * @var Writer
+     */
+    private $deploymentConfigWriter;
+
+    public function __construct(
+        Writer $deploymentConfigWriter,
+        $name = null
+    )
+    {
```
vrann: braces should be on the same line

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,111 @@
...
+        parent::configure();
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws \InvalidArgumentException
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $data = ['db_logger_alias' => 'file'];
```
vrann: you already have constant for the "file" which can be used here. Though, I'd suggest to move it out from the command to the proxy or interface

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->deploymentConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
+
+//        $data = [
```
vrann: dead code should be eliminated

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * command name
+     */
+    const COMMAND_NAME = 'dev:query-log:enable';
+
+    /**
+     * File logger alias
+     */
+    const FILE_LOGGER_ALIAS = 'file';
```
vrann: you can use constant from the proxy instead of this one

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
vrann: just noticed that you have constant in proxy, please use it here

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+namespace Magento\Framework\DB\Logger;
+
+class LoggerProxy extends LoggerAbstract
+{
+    /**
+     * Logger alias param name
+     */
+    const PARAM_ALIAS = 'db_logger_alias';
```
vrann: per comment earlier please rename the value of the param from "db_logger_alias" to "db_logger_mode" or something similar.

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $data = ['db_logger_alias' => 'file'];
+        $this->deploymentConfigWriter->saveConfig([ConfigFilePool::APP_ENV => $data]);
+
+//        $data = [
+//            'db_logger' => [
+//                'logger_alias' => 'file',
+//                'log_all_queries' => false,
```
vrann: For simplicity, we might just default these paratemeters to:
- log all queries => true
- log query time => 0.001
- log call stack => true
This will be the minimum requirement.

Two other options requires more implementation:
- parameters passed to the proxy in the same way as the alias for the logger through the DI and env.php
- create virtual types for the file writer with the different configuration, i,e FileLogAllWithStack and FileLogNoStack

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
```diff
@@ -0,0 +1,87 @@
...
+     * @param bool $logCallStack
+     */
+    public function __construct(
+        FileFactory $loggerFileFactory,
+        QuietFactory $loggerQuietFactory,
+        $loggerAlias,
+        $logAllQueries = false,
+        $logQueryTime = 0.05,
+        $logCallStack = false
+    )
```
vrann: braces should be on the same line

vrann (changes requested):

app/code/Magento/Developer/Console/Command/QueryLogDisableCommand.php
federivo: Fixed

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Fixed

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Fixed

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Fixed

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Fixed

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Chose to handle the params the same as the alias. Need your feedback on that approach.

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
federivo: Fixed

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
federivo: Fixed

vrann (approved):

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
```diff
@@ -0,0 +1,113 @@
...
+     * LoggerProxy constructor.
+     * @param File $file
+     * @param Quiet $quiet
+     * @param bool $loggerAlias
+     * @param bool $logAllQueries
+     * @param float $logQueryTime
+     * @param bool $logCallStack
+     */
+    public function __construct(
+        File $file,
```
vrann: please avoid passing dependency which will never be used. It is better to pass Factory instead, which will create the right logger based on alias the first time logger is requested.

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
```diff
@@ -0,0 +1,113 @@
...
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\DB\Logger;
+
+class LoggerProxy extends LoggerAbstract
+{
+    /**
+     * Logger alias param name
+     */
+    const PARAM_ALIAS = 'db_logger_mode';
```
vrann: please group Deployment configuration options related to logging together and rename them:

'db_logger' => [
     'output' => 'file',
     'log_everything' => 1,
     'query_time_threshold' => '0.001',
     'include_stacktrace' => 1
]

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this->setName(self::COMMAND_NAME)
+            ->setDescription('Enable DB query logging');
+
+        $this->addArgument(
```
vrann: please change arguments to options, so that it will allow named parameters vs positional.

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,114 @@
...
+{
+    /**
+     * input parameter log-all-queries
+     */
+    const INPUT_ARG_LOG_ALL_QUERIES = 'log-all-queries';
+
+    /**
+     * input parameter log-query-time
+     */
+    const INPUT_ARG_LOG_QUERY_TIME = 'log-query-time';
```
vrann: please remove "log-" prefix on all these options names
rename option to query-time-threshold

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,114 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Magento\Framework\App\DeploymentConfig\Writer;
+use Magento\Framework\Config\File\ConfigFilePool;
+
+class QueryLogEnableCommand extends Command
+{
+    /**
+     * input parameter log-all-queries
+     */
+    const INPUT_ARG_LOG_ALL_QUERIES = 'log-all-queries';
```
vrann: please rename to include-all-queries

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * input parameter log-query-time
+     */
+    const INPUT_ARG_LOG_QUERY_TIME = 'log-query-time';
+
+    /**
+     * input parameter log-call-stack
+     */
+    const INPUT_ARG_LOG_CALL_STACK = 'log-call-stack';
```
vrann: please rename to include-call-stack

vrann (changes requested):

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
```diff
@@ -0,0 +1,113 @@
...
+        switch ($loggerAlias) {
+            case self::LOGGER_ALIAS_FILE:
+                $this->logger = $file;
+                break;
+            default:
+                $this->logger = $quiet;
+                break;
+        }
+
+        parent::__construct($logAllQueries, $logQueryTime, $logCallStack);
```
vrann: I noticed that the ``$logCallStack`` option does not work. I think the reason behind it is that File and Quiet should be constructed with this parameter passed to the constructor. 

This call to the parent method is not really needed because that's File or Quiet should be constructed with these parameters

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
```diff
@@ -0,0 +1,113 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\DB\Logger;
+
+class LoggerProxy extends LoggerAbstract
```
vrann: LoggerProxy should not extend the abstract, instead it should implement the LoggerInterface and delegate all calls to the class which extends the abstract, such as File or Quiet

vrann (changes requested):

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
federivo: Done

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
federivo: Done. Passed logger factories instead.

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Done

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Done.

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Done

app/code/Magento/Developer/Console/Command/QueryLogEnableCommand.php
federivo: Done

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
federivo: Done.

lib/internal/Magento/Framework/DB/Logger/LoggerProxy.php
federivo: You're right. Now it's working.

vrann (approved):

magento-team:
@federivo thank you for your contribution. Your Pull Request has been successfully merged

mslabko:
Hi @federivo, 
Thank you for your contribution! Here is some code we use internally for debugging sql\app which is nice to have in CLI command tool to make it multifunctional like a Swiss Army  knife :) 
- Adds executed sql queries to the bottom of page, so you can determinate during development which queries are executed on the page https://gist.github.com/mslabko/06b8dd2e80670f8e1abbf19f70ffd99f  [how it looks](http://prnt.sc/ez40ex)


- Allows to add watches for specific classes, methods, etc and monitor elapsed time, cpu, sql and more https://gist.github.com/kandy/7ae16d74e2bdc35ffd7b524f089259c2
But this one require installed tideways extension (https://tideways.io/profiler/docs/setup/installation) and there is no open documentation how it works, so need to digg into [source code ](https://github.com/tideways/php-profiler-extension/blob/master/tideways.c) for find some usefull features. [how it looks](http://prnt.sc/ez40mo)

Dedicated "Up For Grabs" GitHub issue - https://github.com/magento/magento2/issues/9341

PR #1201, opened 2015-04-19
Add customer_validate event

dfry22:
Intended to try to minimize need for rewrites when custom validation is required for the customer.

**Motivation**
Analyzed 5996 Magento 1.x Extensions and determined the most common rewrites and hope to add events to Magento2 to make these rewrites unnecessary (https://gist.github.com/dfry22/474f727495228baccf28).
Nearly 1 in 100 extensions was rewriting Mage_Customer_Model_Customer, and after analyzing a large sample, found that many were rewriting the validation.

**Solution**
Dispatch customer_validate event and pass the errors array so it can be modified by custom validators and passed back into the function. This allows for localized validation and any other custom customer validation

PR #9266, opened 2017-04-15
A fix for issue #9265

evgk:
Fixes issue #9265

### Description

This line: $quantity = $quantity ?: 1;
in ./vendor/magento/module-catalog/Pricing/Price/TierPrice.php, line 75,
uses the ternary shorthand operator to set the quantity to 1 when
the quantity that is passed can be converted to false.
At this point, the quantity comes from the SaleableInterface item - the simple product model itself.
Note that the type signature for the getQty method in SaleableInterface indicates that it should always
return a float:

/**
 * Returns quantity of saleable item
 *
 * @return float
 */
public function getQty();

However, on the grouped-product page this item comes from this collection:
./vendor/magento/module-catalog/Model/ResourceModel/Product/Link/Product/Collection.php
(a collection of linked products). So the quantity that is set in the product model comes from the
"Default quantity" field which is stored as an attribute of the link(catalog_product_link_attribute_decimal) -
however, since neither Magento nor the underlying Zend libraries have any conversion from the MySQL to PHP types when loading the collections,
and MySQL client/server communication is done via a text-based protocol, this "qty" field ends up being a string in PHP.
Thus, the getQty function will return a string representation of a decimal value("0.0000000") in a clear violation of the SaleableInterface, and it leads to this issue with the tier price above(since the ternary shorthand operator will treat "0.000000" string as true, not setting the quantity to 1 as a result).
Suggested fix does a forceful conversion of the link-attributes with type=decimal to a PHP float-value on the link-product-collection afterLoad event.
It is possible to do the same for the integer type as well, but doing it just for the type=decimal seems least intrusive, and it would solve this interface-mismatch and tier-price issue.

1. magento/magento2#9265: Tier price / grouped product issue

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/Product/Link/Product/Collection.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/Product/Link/Product/Collection.php
@@ -365,6 +365,7 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Product\Collection
     /**
      * After the collection has been loaded
      *
+     * @SuppressWarnings(PHPMD.CamelCaseMethodName)
      * @return $this
      */
     protected function _afterLoad()
--- a/app/code/Magento/Catalog/Model/ResourceModel/Product/Link/Product/Collection.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/Product/Link/Product/Collection.php
@@ -361,4 +361,28 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Product\Collection
                 );
         }
     }
+    
+    /**
+     * After the collection has been loaded
+     *
+     * @return $this
+     */
+    protected function _afterLoad()
+    {
+        if($this->getLinkModel()) {
+            $attributes = $this->getLinkAttributes();
+            foreach($this as $item) {
+                foreach ($attributes as $attribute) {
+                    $code = $attribute['code'];
+                    $value = $item->getData($code);
+                    if($value !== null) {
+                        if($attribute['type'] == 'decimal') {
+                            $item->setData($code,(float)$value);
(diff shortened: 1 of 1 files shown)


evgk:
Hi, 
  I can see the PR wasn't processed in April. Do you need any additional information?

ishakhsuvarov:
Hi @evgk 
We have tried and could not reproduce the issue in question (#9265) on the latest code from the `develop` branch, may be it it specific to `2.1.x` versions only?

evgk:
Hi,
  I have now tried to reproduce it on the latest codebase(before, I tested on 2.1.6) - and I can reproduce the bug. I'm attaching the screenshots.

root@manage /var/www/magento2/last_dev/magento2 # git log -n 1
commit bd22a2c3f61bb596d5b8535ee01bc4b062bd96f5
Merge: 397f87b 015424a
Author: Oleksii Korshenko <okorshenko@magento.com>
Date:   Thu Jun 15 19:33:40 2017 -0500

    Merge pull request #1201 from magento-engcom/develop-prs

    Public Pull Requests

    magento/magento2#9943
    magento/magento2#9941


This is a screenshot of an incorrect product-page without the suggested fix:
price_without_patch

This is a screenshot of a correct product page with the suggested fix applied:
price_with_patch

Those are the settings for the grouped product:
grouped_product_settings

Those are the settings for the test2 product:
test2_product_settings

Tier price for the test2 product:
test2_product_tier_price

Do you have the same settings when testing and do you see the same in frontend?

evgk:
The "Default Quantity" field in the "Grouped Products" section on the grouped product should be 0

Please note that this is essential to reproduce the bug.

PR #9281, opened 2017-04-17
Fixed coding standard violations in Framework\Image namespace

dverkade:
Fixed coding standard violation in the Framework\Image namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Fixed line length


magento-team:

PR #9299, opened 2017-04-18
Improved check for attribute codes in "additional_attributes" field

ccasciotti:
This PR adds the capability to handle uppercased attribute codes in "additional_attributes" field.
Assume that you have this content in additional_attributes field: 

attributecode1=value1,ATTRIBUTECODE2=value2

and in your database both attributes have code in lowercase.
With current behavior the second attribute is skipped, with this PR the system also handles the second attribute without skipping it.

I've added also two new test cases to test both methods `parseAttributesWithoutWrappedValues` and `parseAttributesWithWrappedValues`


magento-team:
@ccasciotti thank you for your contribution. Your Pull Request has been successfully merged

PR #9302, opened 2017-04-18
Fixed coding standard violations in the Framework\Module namespace 

dverkade:
Fixed coding standard violation in the Framework\Module namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:

- Fixed line length
- Changed is_null to === null comparison


magento-team:

PR #9303, opened 2017-04-18
Fixed coding standard violations in the Framework namespace 

dverkade:
Fixed coding standard violation in the Framework namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:

- Fixed line length (expect command in the Url class where the comment is to long, so ignoring that specific line)
- Fixed identation
- Changed all is_null functions to a === null compare.


magento-team:

PR #9304, opened 2017-04-18
Fixed coding standard violations in all Factory classes located in app/code

dverkade:
Fixed coding standard violation for all factory classes in the app/code folder, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:

- Fixed line length
- Changed is_null function to a === null compare.
- Grouped the same constants together


magento-team:

PR #9306, opened 2017-04-19
Fix PaymentTokenFactory interface to have the "Interface" at the end of the name.

dverkade:
Fixed coding standard violation for an interface which was not ending the name with "Interface". In order to let CI check this file everytime changed the following:
- Removed the @codingstandardsIgnoreFile from the head of the file.
- Renamed the file and the name of the interface to PaymentTokenFactoryInterface so that it ends with the name "Interface"
- Refactored references to this class in other Magento modules so that this name change is reflected there as well.


dverkade:
@magento-team For this code violation fix I changed the name of a interface. Please look into this, since I think this might introduce a breaking change and tell me how you want to deal with this. 

A possible solution I can think of is to keep the old class name and let it extend to the refactored new class name. We can add a "deprecated" message in the old class name, so that it can eventually be removed. 

Please let me know your thoughts so I can change this in order to get this accepted.

ishakhsuvarov:
Hi @dverkade 
> A possible solution I can think of is to keep the old class name and let it extend to the refactored new class name. We can add a "deprecated" message in the old class name, so that it can eventually be removed.

This is pretty much the way we tend to deal with these issues at the moment. Would be nice if you could update your PR in the proposed way and we will look into it.
Thanks!

dverkade:
Hi @ishakhsuvarov,

I checked this again and made some changes (refactored to much code in the first place) and added the deprecated file which I suggested. So this PR is complete from my side. Hope you can have a look at it.

ishakhsuvarov:
@dverkade Thanks for the update. We are in the process of discussion of the proposed changes, since there are some concerns regarding the interface in question. I will provide an update for you when we come up with some ideas.

maghamed:
Hey @dverkade ,

Making naming consistency along the code base is very important for us, but we can't introduce changes for API very often. That’s why I want you to ask to accompany your changes with refactoring of the PaymentTokenFactoryInterface, because this contract is far from ideal, and does not represent ideal state dictated by our  Technical Vision. Which I'll describe below.
Because we don’t want to introduce changes to API interfaces often (like change the same interface in each next release). The less changes introduced for API – the less 3rd party devs are affected and need to update their code. That’s why if we change API interface it should be changed to the ideal state dictated by Technical Vision. 

The current problems I see are next:
- Usage of Inheritance Based API for AbstractPaymentTokenFactory and its successors
- Violation of Single Responsibility principle by **PaymentTokenFactoryInterface** which contains two methods create() and getType(). And getType has service purposes only (aka protected contract), it's not used by business logic from outside, but instead used for setting correct type for Model\PaymentToken during the instantiation
https://github.com/magento/magento2/blob/99e85cbc45223baa3551e4c534b650c0d2c6358b/app/code/Magento/Vault/Model/AbstractPaymentTokenFactory.php#L40
- This code 
```php
        $paymentToken = $this->objectManager->create(PaymentTokenInterface::class);
        $paymentToken->setType($this->getType());
```
also has a "smell" , because in our case instance of Model\PaymentToken doesn't have sense without setting type. But we could potentially create Model\PaymentToken object not specifying one. If we will inject dependency on Model\PaymentToken in Constructor DI of some other object.

So, what I propose to do:
1. Let’s accept $type as a constructor parameter and set it within constructor method of Model\PaymentToken, to make sure that our object is well formed and valid just after instantiation.
2. Let's introduce new interface  
```php
<?php
/**
 * Copyright © 2013-2017 Magento, Inc. All rights reserved.
 */
namespace Magento\Vault\Api\Data;

/**
 * Interface PaymentTokenFactoryInterface
 */
interface PaymentTokenFactoryInterface
{ 
    /**
     * Create payment token entity
     * @param $type account|card
     * @return PaymentTokenInterface
     */
    public function create($type);
}
```
which will have correct factory contract. Without getType() method, and create method which accepts type as parameter.

3. Old implementation will proxy call to new one like this:
```php
/**
 * Class CreditCardTokenFactory
 * @api
 * @deprecated 
 * @see PaymentTokenFactoryInterface
 */
class CreditCardTokenFactory extends AbstractPaymentTokenFactory
{
    /**
     * @var string
     */
    const TOKEN_TYPE_CREDIT_CARD = 'card';


    public function __construct(
        ObjectManagerInterface $objectManager
        PaymentTokenFactoryInterface $newTokenFactory = null
    ) {
        $this->objectManager = $objectManager;
        $this->tokenFactory = $newTokenFactory ?: \Magento\Framework\App\ObjectManager::getInstance()
            ->get(PaymentTokenFactoryInterface::class);
    }

    /**
     * Create payment token entity
     * @return PaymentTokenInterface
     */
    public function create()
    {
        return $this->tokenFactory->create($this->getType());
    }
    /**
     * @inheritdoc
     */
    public function getType()
    {
        return self::TOKEN_TYPE_CREDIT_CARD;
    }
}
```
 
Doing all of the above, we wouldn't introduce Backward Incompatible changes, and will introduce useful refactoring which moves our Vault Token Factory interface to the desired vision. So, all the 3rd party devs will need to make an upgrade just once.

Hope that makes sense for you.

dverkade:
Hi @maghamed,

Thanks for your information and have checked the solution you are suggesting. However I do see a problem with this. In your example you have added a "create" function to the CreditCardTokenFactory class. This class extends the AbstractPaymentTokenFactory class which in its turn implements the PaymentTokenFactoryInterface. The interface has changed and it's now manditory to submit the $type parameter to it, so the proxy create function you are suggestion in the CreditCardTokenFactory will not be compatible with the interface it needs to implement. This will not work and throws an error since the create function is incompatible.

How about we change the behaviour of the AbstractPaymentTokenFactory class to something like this, where we have a fallback on  the type argument. It can be null, when it is, it will call the $this->getType() function, otherwise it will pass the argument given to it. 

```
abstract class AbstractPaymentTokenFactory implements PaymentTokenFactoryInterface
{
    /**
     * Create payment token entity
     * @return PaymentTokenInterface
     */
    public function create($type = null)
    {
        /** @var PaymentTokenInterface $paymentToken */
        $paymentToken = $this->objectManager->create(PaymentTokenInterface::class, ($type) ? $type : $this->getType());
        return $paymentToken;
    }
}
```

maghamed:
@dverkade 

>  In your example you have added a "create" function to the CreditCardTokenFactory class.

No, I didn't add "create" function to CreditCardTokenFactory. Function create already there, it was inherited from AbstractPaymentTokenFactory with next implementation.
```php
    /**
     * Create payment token entity
     * @return PaymentTokenInterface
     */
    public function create()
    {
        /** @var PaymentTokenInterface $paymentToken */
        $paymentToken = $this->objectManager->create(PaymentTokenInterface::class);
        $paymentToken->setType($this->getType());
        return $paymentToken;
    }
```
So, CreditCardTokenFactory already has a contract of `public function create()`
All I did is overwritten default implementation, providing new one which will proxy the call to new interface `PaymentTokenFactoryInterface::create($type)`

So, we not changing existing interface 
```php
interface PaymentTokenInterfaceFactory
{
    /**
     * Create payment token entity
     * @return PaymentTokenInterface
     */
    public function create();

    /**
     * Return type of payment token
     * @return string
     */
    public function getType();
}
```
which we consider as deprecated.
But make its implementation work through the newly created interface.
```php
/**
 * Interface PaymentTokenFactoryInterface
 */
interface PaymentTokenFactoryInterface
{ 
    /**
     * Create payment token entity
     * @param $type account|card
     * @return PaymentTokenInterface
     */
    public function create($type);
}
```

Regarding changes proposed by you
```php
abstract class AbstractPaymentTokenFactory implements PaymentTokenFactoryInterface
{
    /**
     * Create payment token entity
     * @return PaymentTokenInterface
     */
    public function create($type = null)
```
we can't change contract of the AbstractPaymentTokenFactory , because that will introduce backward incompatible changes. Which is forbidden.

Does it make sense for you?

maghamed:
Hi @dverkade  , will you help us to improve existing APIs of Vault module based on the output above? That would be very valuable for M2 and improve code quality of Magento Service Contracts.

dverkade:
Hi @maghamed 

Yes, I would like to update this PR and get is accepted. But i'm a little confused. You are mentioning that the CreditCardTokenFactory already has a create function. But this is not in de develop branch i'm look in. Is this something that is in another branch already?

Hope you can point me in the right direction so we can solve this issue.

maghamed:
In the Develop branch, CreditCardTokenFactory is extended from AbstractPaymentTokenFactory
https://github.com/magento/magento2/blob/99e85cbc45223baa3551e4c534b650c0d2c6358b/app/code/Magento/Vault/Model/CreditCardTokenFactory.php#L12

AbstractPaymentTokenFactory has `create()` method https://github.com/magento/magento2/blob/99e85cbc45223baa3551e4c534b650c0d2c6358b/app/code/Magento/Vault/Model/AbstractPaymentTokenFactory.php#L36
 
This method also part of the PaymentTokenInterfaceFactory interface
https://github.com/magento/magento2/blob/99e85cbc45223baa3551e4c534b650c0d2c6358b/app/code/Magento/Vault/Api/Data/PaymentTokenInterfaceFactory.php

Thus, CreditCardTokenFactory class already has a method create() which it inherits from the AbstractPaymentTokenFactory.

Does it make sense?

dverkade:
Hi @maghamed,

Thank you for your anwer. I will look into this over the weekend.

dverkade:
Hi @maghamed,

I see you are at  the Meet Magento NL contribution day as well. I'll be there, so let's discuss this PR and #9407 at the event, so I can contribute to this.

app/code/Magento/Vault/Model/AbstractPaymentTokenFactory.php
```diff
@@ -11,7 +11,8 @@
 
 /**
  * Class AbstractPaymentTokenFactory
- * @api
+ * @deprecated
+ * @see PaymentTokenFactory
```
maghamed: @ see PaymentTokenFactoryInterface

app/code/Magento/Vault/Model/AbstractPaymentTokenFactory.php
```diff
@@ -21,12 +22,18 @@
...
+     * @var PaymentTokenFactoryInterface
+     */
+    private $paymentTokenFactory;
+
+    /**
      * AccountPaymentTokenFactory constructor.
      * @param ObjectManagerInterface $objectManager
      */
-    public function __construct(ObjectManagerInterface $objectManager)
+    public function __construct(ObjectManagerInterface $objectManager, PaymentTokenFactoryInterface $paymentTokenFactory)
```
maghamed: PaymentTokenFactoryInterface $paymentTokenFactory should be optional parameter
PaymentTokenFactoryInterface $paymentTokenFactory = null

to preserve backward compatibility

app/code/Magento/Vault/Model/AccountPaymentTokenFactory.php
```diff
@@ -7,7 +7,8 @@
 
 /**
  * Class AccountPaymentTokenFactory
- * @api
+ * @deprecated
+ * @see PaymentTokenFactory
```
maghamed: @ see PaymentTokenFactoryInterface

app/code/Magento/Vault/Model/CreditCardTokenFactory.php
```diff
@@ -7,7 +7,8 @@
 
 /**
  * Class CreditCardTokenFactory
- * @api
+ * @deprecated
+ * @see PaymentTokenFactory
```
maghamed: @ see PaymentTokenFactoryInterface

app/code/Magento/Vault/Model/PaymentTokenFactory.php
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Vault\Model;
+
+use Magento\Vault\Api\Data\PaymentTokenFactoryInterface;
+
+/**
+ * Interface PaymentTokenFactory
```
maghamed: please provide a precise description.
as it's not an interface

maghamed (changes requested):

ishakhsuvarov:
Hi @dverkade 
Do you wish to continue working on this PR or should we close it?
Thanks

dverkade:
Hi @ishakhsuvarov 

I did not get a reply on this issue where I'm mentioning the failing tests: 
https://github.com/magento/magento2/issues/9407

When I get an answer I can work on this PR again

dverkade:
Hi @ishakhsuvarov @maghamed,

I have finished this PR and tests are now passing with the latest changes made.

ishakhsuvarov:
@dverkade 
Thanks. We will proceed with the review shortly

app/code/Magento/Vault/Model/AbstractPaymentTokenFactory.php
```diff
@@ -21,12 +23,25 @@
...
      * @param ObjectManagerInterface $objectManager
+     * @param PaymentTokenFactoryInterface $paymentTokenFactory
      */
-    public function __construct(ObjectManagerInterface $objectManager)
-    {
+    public function __construct(
+        ObjectManagerInterface $objectManager,
+        PaymentTokenFactoryInterface $paymentTokenFactory = null
+    ) {
+        if (!$paymentTokenFactory) {
```
maghamed: please use === null instead

app/code/Magento/Vault/Model/AbstractPaymentTokenFactory.php
```diff
@@ -35,9 +50,11 @@ public function __construct(ObjectManagerInterface $objectManager)
...
-        $paymentToken = $this->objectManager->create(PaymentTokenInterface::class);
-        $paymentToken->setType($this->getType());
-        return $paymentToken;
+        return $this->paymentTokenFactory->create($this->getType());
     }
+
+    /**
+     * @return string
+     */
+    abstract public function getType();
```
maghamed: no need to specify abstract public getType method here
we get it from PaymentTokenInterfaceFactory 
```php
interface PaymentTokenInterfaceFactory
{
    /**
     * Create payment token entity
     * @return PaymentTokenInterface
     */
    public function create();

    /**
     * Return type of payment token
     * @return string
     */
    public function getType();
}

```

app/code/Magento/Vault/Api/Data/PaymentTokenFactoryInterface.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Payment Token types
+     * @var string
+     */
+    const TOKEN_TYPE_ACCOUNT = 'account';
+    const TOKEN_TYPE_CREDIT_CARD = 'card';
+
+    /**
+     * Create payment token entity
+     * @param $type string
```
maghamed: please specify Null as a possible input value

app/code/Magento/Vault/Model/PaymentTokenFactory.php
```diff
@@ -0,0 +1,62 @@
...
+     * @var array
+     */
+    private $tokenTypes = [];
+
+    /**
+     * PaymentTokenFactory constructor.
+     * @param ObjectManagerInterface $objectManager
+     * @param array $tokenTypes
+     */
+    public function __construct(ObjectManagerInterface $objectManager, $tokenTypes = [])
```
maghamed: please specify 
array for $tokenTypes

dverkade: I don't know what you mean here. array has been definied in the doc comments and is specified in the param as default value was well. What and how do you want it to be described?

maghamed: I mean explicitly specify type hint for $tokenTypes
like
`array $tokenTypes = []`

maghamed (approved):
Good job, well done! 
Thanks for contribution

maghamed:
Btw @dverkade  you made a PR which we accepted and will deliver to mainline soon.
This PR introduced a valuable refactoring for our Core Payment APIs.
And now Payment APIs are closer to the desired state than they were before your contribution.

Maybe you have some time to write a blog post regarding this Pull Request and share with Magento community the experience you gained. How it's complicated to write APIs in a backward compatible way, and that it's possible to do so, not breaking anybody. Not a lot of contributors introduce changes for Magento APIs and you are one of those.

I believe your experience would be helpful for many Mangeto contributors.

PR #9315, opened 2017-04-19
[WIP] introduce newsletter service contract module #9233

davidverholen:
### Description
First proposal for Newsletter Service Contract

### Referencing Issue
1. magento/magento2#9233


app/code/Magento/NewsletterApi/Api/Data/NewsletterSubscriptionInterface.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    const SUBSCRIPTION_ID = 'subscription_id';
+    const EMAIL = 'email';
+    /**#@-*/
+
+    /**
+     * Set ID
+     *
+     * @param int $id
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
```
maghamed: setter functions should return void

That's a good practice for CQS and CQRS design principles

app/code/Magento/NewsletterApi/Api/Data/NewsletterSubscriptionInterface.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function setId($id);
+
+    /**
+     * Set E-Mail Address
+     *
+     * @param string $email
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
```
maghamed: same as above

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionManagementInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace Magento\NewsletterApi\Api\Data;
+
+/**
+ * Newsletter Subscription Management Interface
+ *
+ * @api
+ */
+interface NewsletterSubscriptionManagementInterface
```
maghamed: what for duplicate Newsletter prefix for all interfaces? 
We already have one in the namespace

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionManagementInterface.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @param \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface $newsletterSubscription
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
```
maghamed: ["command"](https://martinfowler.com/bliki/CQRS.html) in terms of CQRS
suppose to return void or ID of newly created entity

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+ * @api
+ */
+interface NewsletterSubscriptionRepositoryInterface
+{
+    /**
+     * Save newsletter subscription.
+     *
+     * @param \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface $newsletterSubscription
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
+     * @throws \Magento\Framework\Exception\LocalizedException
```
maghamed: Localized is too generic.
next exception could be thrown:
     * @throws \Magento\Framework\Exception\InputException
     * @throws \Magento\Framework\Exception\StateException
     * @throws \Magento\Framework\Exception\CouldNotSaveException

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @api
+ */
+interface NewsletterSubscriptionRepositoryInterface
+{
+    /**
+     * Save newsletter subscription.
+     *
+     * @param \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface $newsletterSubscription
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
```
maghamed: save should not return object being saved , that violates CQRS

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+    public function save($newsletterSubscription);
+
+    /**
+     * Retrieve newsletter subscription.
+     *
+     * @param int $newsletterSubscriptionId
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function getById($newsletterSubscriptionId);
```
maghamed: don't think we need to use prefix "newsletter" for variables as well

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function save($newsletterSubscription);
+
+    /**
+     * Retrieve newsletter subscription.
+     *
+     * @param int $newsletterSubscriptionId
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface
+     * @throws \Magento\Framework\Exception\LocalizedException
```
maghamed: * @throws \Magento\Framework\Exception\NoSuchEntityException
should be used instead

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public function getById($newsletterSubscriptionId);
+
+    /**
+     * Retrieve newsletter subscriptions matching the specified criteria.
+     *
+     * @param \Magento\Framework\Api\SearchCriteriaInterface $searchCriteria
+     * @return \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionSearchResultsInterface
+     * @throws \Magento\Framework\Exception\LocalizedException
```
maghamed: in what cases exception being thrown?

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public function getList($searchCriteria);
+
+    /**
+     * Delete newsletter subscription.
+     *
+     * @param \Magento\NewsletterApi\Api\Data\NewsletterSubscriptionInterface $newsletterSubscription
+     * @return bool true on success
+     * @throws \Magento\Framework\Exception\LocalizedException
```
maghamed: next exceptions could be thrown:
     * @throws \Magento\Framework\Exception\InputException
     * @throws \Magento\Framework\Exception\StateException
     * @throws \Magento\Framework\Exception\NoSuchEntityException

Localized is too generic

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionRepositoryInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public function delete($newsletterSubscription);
+
+    /**
+     * Delete newsletter subscription by ID.
+     *
+     * @param int $newsletterSubscriptionId
+     * @return bool true on success
+     * @throws \Magento\Framework\Exception\NoSuchEntityException
+     * @throws \Magento\Framework\Exception\LocalizedException
```
maghamed: same as above

app/code/Magento/NewsletterApi/Exception/AlreadySubscribedException.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
```
maghamed: we don't encourage to introduce a custom exception which supposed to be handled the same way as already existing localized exceptions. 

Consider throwing existing exceptions residing in the Framework\LocalizedException namespaces instead of creating custom exception types.

Do create and throw custom exceptions if you have an error condition that can be programmatically handled in a different way than any other existing exceptions. Otherwise, throw one of the existing exceptions.

Do not create and throw new exceptions just to have your team's exception.

maghamed (changes requested):
Please accompany your interfaces with Use-Cases which are covered by current APIs.
Because now it's not clear whether all desirable cases are covered.

For example, our current Newsletter module implementation provides:
- ability to manage the status of subscription 
https://github.com/magento/magento2/blob/develop/app/code/Magento/Newsletter/Model/Subscriber.php#L257
There are 4 such statuses: SUBSCRIBED|UNSUBSCRIBED|NCONFIRMED|NOT_ACTIVE
All that done with confirmation/unsubsctiption links generation
- subscribe/unsubscribe Customer by ID
- Email Template management
https://github.com/magento/magento2/blob/develop/app/code/Magento/Newsletter/Model/Template.php
- Allowance of Subscription for Guests 
https://github.com/magento/magento2/blob/develop/app/code/Magento/Newsletter/Model/Subscriber.php#L52

maghamed:
README.md which describes module's duties and responsibilities is not provided

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionManagementInterface.php
davidverholen: A class name (also without the namespace) should already be very clear on what it's purpose is to improve IDE auto completion and readability of the code using it

davidverholen:
@maghamed thank you for your review, it's very helpful especially regarding magento's best practices. 

I would prefer the longer interface names for code readability and auto completion, on the other hand, the name Subscription is still pretty unique (also without the newsletter prefix)

app/code/Magento/NewsletterApi/Exception/AlreadySubscribedException.php
adragus-inviqa: Well, you're suggesting us to use M2's exceptions, when you guys have created your own? Not fair.

How would you differentiate between some `LocalizedException` down the stack and a `LocalizedException` that's only triggered if there already is a subscriber? Existing exceptions are too generic than this unique usage.

> Do create and throw custom exceptions if you have an error condition that can be programmatically handled in a different way than any other existing exceptions.

**Every** exception can be handled that way, but that's not the point of specialised exceptions. It's about granular handling. There are cases when you can handle an `AlreadySubscribedException` exception, and choose not to handle a `LocalizedException`, so you throw it further up the stack.

Specific exceptions are a good thing. Don't make us catch and throw generic exceptions. Maybe @davidverholen can extend `\Magento\Framework\Exception\AlreadyExistsException` instead, but I still consider this exception useful. Heck, I'll go as far as saying that every _module_ needs its own exception namespace.

ishakhsuvarov:
@davidverholen 
* Please add the type hints to the setter method signatures which accept specific interface.
* Comments in PHPDoc should be present for every method and be as specific as possible.
* In the `README` please specify the module's duties and the reasons why service contract is located in a separate module.

app/code/Magento/NewsletterApi/Exception/AlreadySubscribedException.php
davidverholen: I'm agreeing a little bit with both of you.

On the one hand I would always prefer a specialized exception over a generic one.

On the other hand, in the service contract it might be better to be a little more generic while in the implementation itself the integrator can then introduce more specialized exceptions, that are extended from the more generic exception

adragus-inviqa: > On the other hand, in the service contract it might be better to be a little more generic while in the implementation itself the integrator can then introduce more specialized exceptions, that are extended from the more generic exception

`AlreadySubscribedException` _is_ the most generic exception you can throw from the Newsletter service contract. 

Furthermore, we, as clients of the API, cannot use the _implementation's_ exception, like you said, as we would break the contract. We would always have to catch (and use) some `LocalizedException`, thus catching more exception than we should.

davidverholen: working further with the interfaces brought some new thoughts up for me.

I think the Newsletter Repository should throw the given internal Exceptions, as there are already plenty of good / speaking exceptions for the persistence layer itself.

the newsletter management on the other hand is not part of the persistence layer anymore and should have it's own exceptions.

If I delete a subscription entity over the repository, it is pretty clear what can happen and I can throw the default exceptions.

if I unsubscribe over the newsletter management, it is up to the integrator what happens then (e.g. delete entity or maybe just change subscription status to unsubscribed)

So my conclusion is, for the management, there should be specialized exceptions.

davidverholen:
@maghamed I just added a new commit. There are some features that should not be in this module I think since they are more implementation details (they still could be part of the Magento_Newsletter Modules Api later)

For example the template management. Most Newsletter integrations won't use the magento e-mail templating

Settings like allow guest subscriptions could be a part of this module, but I think it's also better to leave these in the implementation module (I will of course add all given features as api later in the Magento_Newsletter Module itself, just want to make sure not to restrict future implementations using this service contract module)

app/code/Magento/NewsletterApi/Api/NewsletterSubscriptionManagementInterface.php
maghamed: But here we just duplicate Newsletter prefix for all entities and variables. Which doesn't bring a lot of additional sense.

Moreover we have a static check which says that variable names should not be longer than 20 symbols. Because long variable name usually says about "smell" in code.

For example, we have "Catalog" module, but Product entity within that module called
Catalog\Api\Data\ProductInterface
not
Catalog\Api\Data\CatalogProductInterface

Using Catalog or Newsletter prefixes we will come up to names format we had in Magento 1, before Namespaces came to Magento 2.

Namespace prefix with Class name represent fully qualified name of the class.

app/code/Magento/NewsletterApi/etc/webapi.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
```
ishakhsuvarov, line 1: Here you could also add:
* Endpoints for the customer to subscribe/unsubscribe himself, so no admin privileges needed.
* Guest resources
* More of the API, e. g. subscribe by id, unsubscribe, etc.

app/code/Magento/NewsletterApi/Api/Data/SubscriptionInterface.php
```diff
@@ -0,0 +1,104 @@
...
+     * @return void
+     */
+    public function setEmail(string $email);
+
+    /**
+     * Set Status
+     *
+     * set the status of the subscription
+     *
+     * @param int $status
```
ishakhsuvarov: How should the `int` status be mapped to the actual status?

app/code/Magento/NewsletterApi/Api/Data/SubscriptionInterface.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Set E-Mail Address
+     *
+     * set the e-mail address of the subscriber
+     *
+     * @param string $email
+     * @return void
+     */
+    public function setEmail(string $email);
```
ishakhsuvarov: Unfortunately we currently support only the type hints which were available pre PHP 7 (Class/Interface name, array, callable).

ishakhsuvarov:
@davidverholen 
> Settings like allow guest subscriptions could be a part of this module, but I think it's also better to leave these in the implementation module

But there is not really much difference between subscribing a customer and a guest? Why?

> There are some features that should not be in this module I think since they are more implementation details (they still could be part of the Magento_Newsletter Modules Api later)

Even if we could consider that some features should be specific API for the newsletter module, current implementation is missing some really general features, confirmation links/statuses, actual messages queue processing.
Additionally, `NewsletterApi` may be confusing with `Newsletter/Api` namespace available.

maghamed:
@davidverholen 

> There are some features that should not be in this module I think since they are more implementation details (they still could be part of the Magento_Newsletter Modules Api later)

That's why I propose to start from Use-Cases and cover all existing use-cases, covered by current Newsletter module.
And show how these use-cases matched with proposed APIs. If we will see that we need to decompose module on two or three modules at the final implementation, following modularity best practices - it's okay. 
But what's the most important for us - is to prevent of API gap, when new API would be brought to the codebase, but these APIs are not enough to cover needed business cases, so 3rd party devs will still rely on implementation, but not APIs.
The drawback of such approach is that we will introduce Backward Incompatible changes several times with each API updates. So, instead of updating dependencies once during the migration to new Newsletter API module, 3rd party devs would have to do it each next release when we will change/update something in Newsletter API. 
That's why if we introduce new service contracts - we are trying to make them compatible with our Technical Vision (Desirable State) of some specific module.

So, we need to see the final state of Service Contracts before accepting such PR

app/code/Magento/NewsletterApi/Api/Data/SubscriptionInterface.php
davidverholen: you cannot predefine the status in the api since different implementations might have different status. Int was just a decision because the current Magento_Newsletter implementation uses integer constants

davidverholen:
@maghamed 

> That's why I propose to start from Use-Cases and cover all existing use-cases, covered by current Newsletter module.

But my point is. If I cover all use cases that the Magento_Newsletter Module is currently providing, I also force 3rd party integrators to implement these use cases. That's why for example the templating feature should not be covered in this module. Also the status mapping can't be defined here for that reason

app/code/Magento/NewsletterApi/Exception/AlreadySubscribedException.php
maghamed: @adragus-inviqa 
> Well, you're suggesting us to use M2's exceptions, when you guys have created your own? Not fair.

Actually not, we are trying not to introduce custom exception types in Magento modules as well. And rely on \Magento\Framework\Exception\*
(successors of LocalizedException).

One of the reason - is that it doesn't make sense to have an Exception Interface, but it make sense to have just Exception implementation. But logic behind our Service Contracts is to expose API interfaces only. Without any implementation.

Also, practice of creation a lot of custom exceptions will lead to building a business logic on exception throwing/catching for business invariants implementation instead of usage of conditional logic with business statuses etc.
Business logic (both application and domain) MUST NOT be managed with exceptions. Conditional statements should be used instead.

@davidverholen 
> I think the Newsletter Repository should throw the given internal Exceptions, as there are already plenty of good / speaking exceptions for the persistence layer itself.

That's not correct, on each layer should be defined own exceptions hierarchy.
Exception hierarchy for Service Contract level is successors of LocalizedException. It is allowed to throw exceptions only which are defined on the same layer.
If an exception is caught on the layer which is different from the one where it has been thrown, and it should be re-thrown, then a new exception instance appropriate to the current layer should be created (in this case, the original exception must be passed to a new instance with "previous" argument).

For this specific case I don't recommend to use LocalizedException , because it's too generic and  COULD only be thrown in presentation layer (Controllers, Blocks).

For this case, I propose to use `Magento\Framework\Exception\StateException`.  Potentially we could consider to add `\Magento\Framework\Exception\AlreadyExistsException`, but as for me, it's a subtype of StateException.

app/code/Magento/NewsletterApi/Api/Data/SubscriptionInterface.php
maghamed: if so, on the level of API we have to provide the ability to extend the status list. 
And provide a meaningful way of moving the state between all of these statuses. Like vending machine. [State](https://sourcemaking.com/design_patterns/state) design pattern could be used here if needed.
But that should be declared explicitly enough on the level of APIs.
Because if we leave that as implementation details. Some could want to extend list of statuses and add custom one. For these purposes he/she will have to extend/customize private implementation, but not the API contract, because such extension point is not provided

maghamed:
@davidverholen 

we need to make sure that all functionality would be preserved. Some 3rd party integrators rely just on basic functionality. Other extensions provide own implementation/customize existing one on top of Magento API. We need to satisfy all possible use cases, as we deliver a framework solution.

If templating should not be covered by this module - it should be clearly specified where it goes. And what APIs for templating functionality should be introduced and where.

Regarding status mapping I already mentioned above

app/code/Magento/NewsletterApi/Exception/AlreadySubscribedException.php
korostii: > Business logic (both application and domain) MUST NOT be managed with exceptions. Conditional statements should be used instead.

It's funny you mention that. Please take a look at the [following piece of code](https://github.com/magento/magento2/blob/99e85cbc45223baa3551e4c534b650c0d2c6358b/app/code/Magento/Quote/Model/Webapi/ParamOverriderCartId.php#L58-L60)
Could you guess _which kind_ of entity wasn't found? Hint: it's not necessarily the Customer, it might be Product as well.

Would you say this code is badly written and should avoid potentially provoking an exception as means to implement conditional logic? There's [a lot of similar code](https://github.com/magento/magento2/search?q=catch+NoSuchEntityException+&type=Code&utf8=%E2%9C%93) all around. Is there a sane alternative to prefer in such cases?

app/code/Magento/NewsletterApi/Api/SubscriptionManagementInterface.php
```diff
@@ -0,0 +1,107 @@
...
+     * Subscribe a customer by id
+     *
+     * Adds a new Newsletter Subscription for the given customer
+     * by provided customer id
+     *
+     * @param int $customerId the id of the customer to be subscribed
+     *
+     * @return bool true on success
+     *
+     * @throws \Magento\NewsletterApi\Exception\AlreadySubscribedException when subscription already exists
```
antonkril: Since there is no easy way to check if customer/email is already subscribed, throwing the exception if user is already subscribed will promote using exceptions for conditional logic. Do not throw the exception.

davidverholen:
Hey @maghamed @ishakhsuvarov,

are you by any chance attending the DE contribution days in leipzig this weekend?
I got a little Stuck here, didn't have much time for this.

If you are in Leipzig this Weekend, maybe we can figure out the API as the most critical part of the issue and when I find some time again I can continue with the implementation

ishakhsuvarov:
Hi @davidverholen 
We will be attending Contribution Days and Meet Magento in Leipzig, so will figure out something with this API for sure.
See you there!

davidverholen:
@ishakhsuvarov awesome! looking forward to working with you guys

ishakhsuvarov:
Hi @davidverholen 
Are you still working on this issue or should I close this PR?

ishakhsuvarov:
@davidverholen I am closing this PR now due to inactivity.
Please reopen if you want to continue any time.
Thank you.

PR #9318, opened 2017-04-19
Fixed coding standard violations in the Framework\Backup namespace

dverkade:
Fixed coding standard violations in the Framework\Backup namespace, so that they will be included in the CI tools. Fixed the following:
- Removed @codingStandardsIgnoreFile from the head of the files.
- Fixed line length
- Fixed identation
- Changed class comment to the top of the class where it is in every other file instead of the top of the file.


magento-team:

PR #9325, opened 2017-04-20
RenderPool  performance improvment using object cache

sivajik34:
We can store classNames,block templates into multi dimensional array based on code,type.
next time onwards  we can retrieve by code,type.


adragus-inviqa:
Again, @sivajik34, do you know how much time are you saving with these? 

In my testing, they save 0.009s. How much time was saved on your side? Do we really need them?

sivajik34:
@adragus-inviqa 
let me frank i don't know how to calculate like u.
can you give me  information how you are getting 0.009s.
so i can calculate my side once.

orlangur:
@sivajik34 how do you call it a "performance improvement" not getting the exact numbers? I believe @adragus-inviqa can probably share some snippet of testing code so that you can try it on your side.

Generally, we cannot add performance optimizations blindly, ideally we should test performance impact on some user scenario or at least make a synthetic isolated test. Especially in such case when we increase memory consumption.

sivajik34:
@orlangur understood.
@adragus-inviqa could you please share testing code,so next time onwards it will be helpful for me.

sivajik34:
@orlangur so overall these type of pull requests won't required?
So its going to be rejected?
So i can ignore these type of code refactor?

adragus-inviqa:
Sorry, peeps, I lied. I did no testing whatsoever. It was more to convince @sivajik34 to give a counter to my imaginary number. 😝 

I think my point was made anyway.

I saw you've added some optimisations in your latest PRs, @sivajik34. That's great. It's something I wanted to do, too: be on the lookout for bottlenecks in M2's codebase and try to come up with better algorithms/data structures/logic. 

But you always needs to _test first_. For example: your storing data in class members might look ok, but remember that a) you're increasing memory consumption (whenever you store stuff) and b) there is block caching for that. So it's not always a good idea to do static caching in class members.

As for testing, I'm not sure. Since this looks like it's changing the checkout, try testing it with and without your change and average them out, with no cache enabled. The usual stuff.

----

> so overall these type of pull requests won't required?
> So its going to be rejected?
> So i can ignore these type of code refactor?

You should continue optimising M2's codebase if you want and have the time, but you must back them up with test results. I don't think any of the core members will have any problems merging your code if it proves to be efficient.

orlangur:
> So overall these type of pull requests won't required?
> So its going to be rejected?
> So i can ignore these type of code refactor?

No-no-no, things are much brighter :)

Sometimes performance improvement is pretty straightforward to evaluate, even though it could be negligible in a full scenario. Like, `is_null` is x3 times slowed than `=== null` and there is no additional memory consumption.

However, when you have some idea for optimization and there is a trade-off - for example, code become faster but it uses more memory - you should evaluate such change by calculating performance boost in action.

Generally it's better to start from code profiling to spot bottlenecks. Probably it only takes some 0.001% of CPU time to execute code you touched, so, it's not worth changing. For example, I have a dream to get rid of `strpos` call in `DataObject::getData` but if I remember correctly it cannot really help as only some 0.5-1% of time spent inside this method for regular request. There is also some `underscoreCache` within `DataObject` and I'm not sure it's valuable nowadays.

adragus-inviqa:
http://codepad.org/OOHq7oEo

Copied from [this post](http://stackoverflow.com/a/8229059/394589). Just sayin'.

ishakhsuvarov:
@sivajik34 I've ran some performance tests against this PR and did not find any noticeable benefit on any of the scenarios. Do you have any numbers regarding cpu time and memory usage with these changes applied?

sivajik34:
For example if we are in product listing page with more than 30 products per page with same type of  products can get little bit benefit.
I mean if we do these type of changes may be overall we get some noticeable benefit.

orlangur:
> may be overall we get some noticeable benefit.

... or maybe not and we will just waste some memory. Just do some benchmarking and show your numbers.

@adragus-inviqa your link is obviously wrong as 10000 iterations are nothing. `is_null` is a function unlike `===` operator thus it is slower by design. Here are some fairly looking numbers to me: http://ua2.php.net/manual/en/function.is-null.php#84161

Not like it can give some noticeable performance boost by itself, just an example of a safe optimization which can make our code more consistent. Same for the `==` vs `===`: the latter is slightly faster and also makes your code more robust.

adragus-inviqa:
@orlangur - I meant that it's not worth it. Just like that php.net comment said. So I don't see how that's an argument. 

I was trying to suggest that this (calls vs operators) is not the kind of optimisation I'm hoping to see in PRs. I think M2 has bigger problems than too many _is_null_ calls...

Who cares about 10 _million_ iterations anyway? It's `add to cart`, not `create DNA sequence of E.T.`, sheesh.

ishakhsuvarov:
@sivajik34 
> For example if we are in product listing page with more than 30 products per page with same type of products can get little bit benefit.

I've done some brief profiling with the described scenario and FPC disabled and could not find any measurable improvement. Do you have any numbers which would show some improvement?
Thanks

sivajik34:
@ishakhsuvarov  We can close it if its not that much helpful.no issues.

ishakhsuvarov:
@sivajik34 
Ok, closing this PR for now. Thank you for collaboration.

PR #9329, opened 2017-04-20
Fixed coding standard violations in the Framework\Stdlib namespace

dverkade:
Fixed coding standard violations in the Framework\Stdlib namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Removed @codingStandardsIgnoreFile at the head of the files
- Fixed indentation
- Fixed class property which type was not defined.
- Wrapped header of PhpCookieManagerTest with a codingstandardignorestart and end, because another namespace is given in the same class file, which is not allowed by PSR-2

### DIFF
--- a/lib/internal/Magento/Framework/Stdlib/DateTime/TimezoneInterface.php
+++ b/lib/internal/Magento/Framework/Stdlib/DateTime/TimezoneInterface.php
@@ -4,8 +4,6 @@
  * See COPYING.txt for license details.
  */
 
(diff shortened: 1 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9329) <br/>All committers have signed the CLA.

magento-team:
@dverkade thank you for your contribution. Your Pull Request has been successfully merged

PR #9330, opened 2017-04-20
Fixed coding standard violations in the Framework\Data namespace

dverkade:
Fixed coding standard violations in the Framework\Data namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Removed @codingStandardsIgnoreFile at the head of the files
- Fixed indentation
- Changed is_null function call to a === null compare


magento-team:

PR #9343, opened 2017-04-21
Add logging to contact us form

JamesonNetworks:
### Description

If contact us fails to process, the exception is swallowed

I did not open an issue for this problem

### Manual testing scenarios

1. Create conditions for the contact us form to fail. I did this by changing the code in the controller to explicitly throw an exception, however, I first experienced this in the Magento Cloud where no useful information is being retained.

### Contribution checklist
 - [Y] Pull request has a meaningful description of its purpose
 - [Y] All commits are accompanied by meaningful commit messages
 - [N/A] All new or changed code is covered with unit/integration tests (if applicable)
 - [Y] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Contact/Controller/Index/Post.php
+++ b/app/code/Magento/Contact/Controller/Index/Post.php
@@ -49,13 +49,13 @@ class Post extends \Magento\Contact\Controller\Index
         Context $context,
         ConfigInterface $contactsConfig,
         MailInterface $mail,
-        DataPersistorInterface $dataPersistor,
+        DataPersistorInterface $dataPersistor = null,
         LoggerInterface $logger = null
     ) {
         parent::__construct($context, $contactsConfig);
         $this->context = $context;
         $this->mail = $mail;
-        $this->dataPersistor = $dataPersistor;
+        $this->dataPersistor = $dataPersistor ?: \Magento\Framework\App\ObjectManager::getInstance()->get(DataPersistorInterface::class);
         $this->logger = $logger ?: \Magento\Framework\App\ObjectManager::getInstance()->get(LoggerInterface::class);
     }
 
--- a/app/code/Magento/Contact/Controller/Index/Post.php
+++ b/app/code/Magento/Contact/Controller/Index/Post.php
@@ -50,13 +50,13 @@ class Post extends \Magento\Contact\Controller\Index
         ConfigInterface $contactsConfig,
         MailInterface $mail,
         DataPersistorInterface $dataPersistor,
-        LoggerInterface $logger
+        LoggerInterface $logger = null
     ) {
         parent::__construct($context, $contactsConfig);
         $this->context = $context;
         $this->mail = $mail;
         $this->dataPersistor = $dataPersistor;
-        $this->logger = $logger;
+        $this->logger = $logger ?: \Magento\Framework\App\ObjectManager::getInstance()->get(LoggerInterface::class);
     }
 
     /**
--- a/app/code/Magento/Contact/Controller/Index/Post.php
+++ b/app/code/Magento/Contact/Controller/Index/Post.php
@@ -14,6 +14,7 @@ use Magento\Framework\App\Request\DataPersistorInterface;
 use Magento\Framework\Controller\Result\Redirect;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\HTTP\PhpEnvironment\Request;
+use Psr\Log\LoggerInterface;
 
 class Post extends \Magento\Contact\Controller\Index
 {
@@ -33,21 +34,29 @@ class Post extends \Magento\Contact\Controller\Index
     private $mail;
 
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Context $context
      * @param ConfigInterface $contactsConfig
      * @param MailInterface $mail
      * @param DataPersistorInterface $dataPersistor
+     * @param LoggerInterface $logger
      */
     public function __construct(
         Context $context,
         ConfigInterface $contactsConfig,
         MailInterface $mail,
-        DataPersistorInterface $dataPersistor
+        DataPersistorInterface $dataPersistor,
+        LoggerInterface $logger
     ) {
         parent::__construct($context, $contactsConfig);
         $this->context = $context;
         $this->mail = $mail;
         $this->dataPersistor = $dataPersistor;
+        $this->logger = $logger;
     }
 
     /**
@@ -70,6 +79,8 @@ class Post extends \Magento\Contact\Controller\Index
             $this->messageManager->addErrorMessage($e->getMessage());
             $this->getDataPersistor()->set('contact_us', $this->getRequest()->getParams());
         } catch (\Exception $e) {
+            $this->logger->addError('An error occurred processing a posted message to contact us in Controller\\Index\\Post: ' . $e->getMessage());
+            $this->logger->critical($e);
             $this->messageManager->addErrorMessage(
                 __('We can\'t process your request right now. Sorry, that\'s all we know.')
             );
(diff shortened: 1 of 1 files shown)


app/code/Magento/Contact/Controller/Index/Post.php
```diff
@@ -33,21 +34,29 @@ class Post extends \Magento\Contact\Controller\Index
...
      * @param DataPersistorInterface $dataPersistor
+     * @param LoggerInterface $logger
      */
     public function __construct(
         Context $context,
         ConfigInterface $contactsConfig,
         MailInterface $mail,
-        DataPersistorInterface $dataPersistor
+        DataPersistorInterface $dataPersistor,
+        LoggerInterface $logger
```
ishakhsuvarov: Due to backwards compatibility policy we can not add a new required argument to the constructor.
You may use the injection with the optional parameter, like this:
```php
public function __construct(\NewDependencyInterface $newDependency = null) {
    $this->newDependency = $newDependency ?: \Magento\Framework\App\ObjectManager::getInstance()->get(\NewDependencyInterface::class);
}
     
```

JamesonNetworks: Instantiate the object directly with object manager? My understanding is that is a big no no

JamesonNetworks: Regardless, updated. Would like clarification on when it is ok to use object manager. Maybe only in core code?

app/code/Magento/Contact/Controller/Index/Post.php
```diff
@@ -33,21 +34,29 @@ class Post extends \Magento\Contact\Controller\Index
...
      * @param MailInterface $mail
      * @param DataPersistorInterface $dataPersistor
+     * @param LoggerInterface $logger
      */
     public function __construct(
         Context $context,
         ConfigInterface $contactsConfig,
         MailInterface $mail,
-        DataPersistorInterface $dataPersistor
+        DataPersistorInterface $dataPersistor,
```
maghamed, line 52: current parameter should be optional as well.

$dataPersistor = null

maghamed (changes requested):

app/code/Magento/Contact/Controller/Index/Post.php
JamesonNetworks: What is the reasoning behind this?

app/code/Magento/Contact/Controller/Index/Post.php
ishakhsuvarov: You are right, using ObjectManager directly is indeed a "big no no". However, some existing extension or customization may possibly extend this class, thus, by changing the constructor signature you would break it.

ishakhsuvarov: Such direct usages of the ObjectManager should be refactored eventually, but that may only happen we module version is bumped.

app/code/Magento/Contact/Controller/Index/Post.php
JamesonNetworks: Addressed. I'm concerned about the future look of M2 if these kind of changes continue though. Is the plan just to have optionally parameterized constructors on every class with direct calls to object manager?

app/code/Magento/Contact/Controller/Index/Post.php
JamesonNetworks: Makes sense, thanks for the clarification!

maghamed (approved):

app/code/Magento/Contact/Controller/Index/Post.php
maghamed: we have next requirements dictated by Backward Compatibility policy:
Add the new optional parameter to the constructor **at the end** of arguments list.
In the constructor body, if new dependency is not provided (value of introduced argument is null) fetch the dependency using `Magento\Framework\App\ObjectionManager::getInstance(). `

```php
class ExistingClass
{
    /** @var \New\Dependency\Interface */
    private $newDependency;
  
    public function __construct(
        \Old\Dependency\Intreface $oldDependency,
        $oldRequiredConstructorParameter,
        $oldOptinalConstructorParameter = null,
        \New\Dependency\Interface $newDependency = null
    ) {
        ...
        $this->newDependency = $newDependency ?: \Magento\Framework\App\ObjectManager::getInstance()->get(\New\Dependency\Interface::class);
    }
     
    public function existingFunction()
    {
        // Existing functionality
        ...
        ...
        
        // Use $this->newDependency wherever the new dependency is needed
        ...
        ...
    }
}
   
// Sample unit test code snippet follows
class ExistingClassTest extends \PHPUnit_Framework_TestCase
{
    private $existingClassObject;
   
    protected function setUp()
    {
        ...
        // Create dependency mcoks with $this->getMock() or $this->getMockBuilder()
        $newDependencyMock = $this->getMock(\New\Dependency\Interface::class, [], [], '', false);      
          
        $objectManager = new \Magento\Framework\TestFramework\Unit\Helper\ObjectManager($this);
        $this->existingClassObject = $objectManager->getObject(
            ExistingClass::class,
            [
                'oldDependency' => $oldDependencyMock,
                'oldRequiredConstructorParameter' => 'foo',
                'oldOptinalConstructorParameter' => 'bar',
                'newDependency' => $newDependencyMock, 
            ]
        );
    }
   
    public function testExistingFunction()
    {
        ...
        ...
    }
}
```

maghamed: @JamesonNetworks  oh sorry, my fault 
I didn't notice that this parameter has already been there before.
` DataPersistorInterface $dataPersistor`

Sorry, for misleading you.
I was wrong. No need to make it optional.

JamesonNetworks: rgr, no worries, corrected and pushed

JamesonNetworks: I'm not confused about what you are doing, I'm confused as to why. It seems to me you would just mark the PRs that change the constructors and tag them to be released in releases that have the potential to break compatibility. Instead, we're building in tech debt here and expecting a maintainer to come in behind and remove these problems at the time of releases

maghamed: adding an optional parameter to the constructor doesn't bring Backward Incompatible changes. Because code will work according to its contract nevertheless dependency was passed or not (in this case we will instantiate one with ObjectManager::getInstance static call)

app/code/Magento/Contact/Controller/Index/Post.php
```diff
@@ -70,8 +79,10 @@ public function execute()
             $this->messageManager->addErrorMessage($e->getMessage());
             $this->getDataPersistor()->set('contact_us', $this->getRequest()->getParams());
         } catch (\Exception $e) {
+            $this->logger->addError('An error occurred in Controller\\Index\\Post: ' . $e->getMessage());
```
adragus-inviqa: a) why do we need this one? wouldn't the stack trace by enough? 
b) if you're going to do it, why aren't you using `__CLASS__`?

JamesonNetworks: a) I'm not sure what M2 logging best practices are so I was throwing it in the two places I would expect to see it.
b) Should have used `__CLASS__` here. 

I removed this line since the critical log could be enough? If you have any pointers to docs on best practices for core logging, I would appreciate it. 

Thanks!

JamesonNetworks:
@adragus-inviqa @ishakhsuvarov @maghamed Thanks everyone for the review and input! I appreciate the feedback

magento-team:
@JamesonNetworks thank you for your contribution. Your Pull Request has been successfully merged

PR #9350, opened 2017-04-21
Fixed coding standard violations in the Framework\App namespace

dverkade:
Fixed coding standard violations in the Framework\App namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:

- Move ActionFake to it's own file, because PSR-2 dicates only one class per file.
- Fixed indentation


magento-team:

PR #9352, opened 2017-04-21
Fixed coding standard violations in the Framework\File namespace

dverkade:
Fixed coding standard violations in the Framework\File namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Fixed indentation
- Fixed missing parameter variable in docblock comment
- Fixed camel case strings in function


magento-team:

PR #9353, opened 2017-04-21
Fixed coding standard violations in the Framework\DB namespace

dverkade:
Fixed coding standard violations in the Framework\DB namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Fixed indentation
- Changed is_null to strict === null compare 
- Added codeignorestart and end around \Zend_Db_Adapter_Pdo, because it's not using namespace which is not allowed by PSR-2
- Removed commented out code from MysqlTest class


magento-team:

PR #9354, opened 2017-04-21
Fixed coding standard violations in the Framework\Translate namespace

dverkade:
Fixed coding standard violations in the Framework\Translate namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Fixed indentation
- Wrapped __() functions with a codingStandardsIgnoreStart and end comment, because these will fail on the short function name.


magento-team:

PR #9358, opened 2017-04-22
Fixed coding standard violations in the Framework\Code namespace

dverkade:
Fixed coding standard violations in the Framework\Code namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:

- Removed @codingStandardsIgnoreFile from the head of the file.
- Changed interfaces in tests to actually end with the name "Interface".


orlangur:
Please do not create PRs without green builds ;)

PR #9362, opened 2017-04-22
Fixed coding standard violations in the Framework\ObjectManager namespace

dverkade:
Fixed coding standard violations in the Framework\ObjectManager namespace, so that it will be checked bij PHP CS and no longer be ignored while doing CI checks. Made the following changes:
- Removed @codingStandardsIgnoreFile from the head of the file.

### DIFF
--- a/lib/internal/Magento/Framework/ObjectManager/Code/Generator/Persistor.php
+++ b/lib/internal/Magento/Framework/ObjectManager/Code/Generator/Persistor.php
@@ -4,8 +4,6 @@
  * See COPYING.txt for license details.
  */
 
-// @codingStandardsIgnoreFile
-
 namespace Magento\Framework\ObjectManager\Code\Generator;
 
 /**
--- a/lib/internal/Magento/Framework/ObjectManager/Code/Generator/Repository.php
+++ b/lib/internal/Magento/Framework/ObjectManager/Code/Generator/Repository.php
@@ -4,8 +4,6 @@
  * See COPYING.txt for license details.
  */
 
-// @codingStandardsIgnoreFile
-
 namespace Magento\Framework\ObjectManager\Code\Generator;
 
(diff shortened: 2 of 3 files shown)


orlangur:
This PR can be closed as I've done such changes automatically.

dverkade:
@orlangur This changes are still in the develop branch. I cann't see any open pull requests for these files. So this can be merged. If you are finished with your work you can reference a pull request in this commit and it's up to @magento-team which PR they are accepting.

dverkade:
@magento-team I still have the ConfigDomMock class which has the @codingStandardsIgnoreFile in the head of the file. I was not able to remove this one, because it throws an error because the class is in the root namespace and does not contain a vendor name. What is the technical vision of Magento for these type of classes? 

I'm suggesting the following solution: Put this file in the "Magento\Framework\ObjectManager\Test\Unit\Config\Reader" namespace and update the test including/referincing this file to the correct new class name.

orlangur:
@dverkade,
>  I cann't see any open pull requests for these files

I do not post unprepared pull requests ;) Just trying to minimize conflicts with useless manual changes. Thanks for your other today PRs as they really cover what I postponed for now :+1: I've already finished stabilization of draft branch, creating clean branch out of `develop` is a matter of hours (would be some 30 minutes with SSD env :smile: ).

> What is the technical vision of Magento for these type of classes?

`class ConfigDomMock extends \PHPUnit_Framework_TestCase` this is some evil mix of stub and test case. Technically you can put it somewhere near `\Magento\Framework\ObjectManager\Test\Unit\Config\Reader\DomTest` but the better way would be to refactor a unit test in such a way that you test code behavior and not flow. We shall just test output of `$this->model->read();` call.

orlangur:
This change is covered by #9367 by the way.

PR #9371, opened 2017-04-24
Update tax.js so that "Display Zero Tax Subtotal = No" works 

raivisdejus:
Making comparison less strict as tax value from this.getPureValue() when empty is 0.000 so that "Display Zero Tax Subtotal = No" works 

### Description

Making comparison less strict as tax value from this.getPureValue() when empty is 0.000 

1. Makes Display Zero Tax Subtotal = NO work

### Manual testing scenarios

Tested manually:
1. Set Display Zero Tax Subtotal = NO
2. Open cart

### Contribution checklist

### DIFF
--- a/app/code/Magento/Tax/view/frontend/web/js/view/checkout/cart/totals/tax.js
+++ b/app/code/Magento/Tax/view/frontend/web/js/view/checkout/cart/totals/tax.js
@@ -21,7 +21,7 @@ define([
          * @override
          */
         ifShowValue: function () {
-            if (this.getPureValue() === 0) {
+            if (this.getPureValue() == 0) {
                 return isZeroTaxDisplayed;
             }
 


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9371) <br/>All committers have signed the CLA.

app/code/Magento/Tax/view/frontend/web/js/view/checkout/cart/totals/tax.js
```diff
@@ -21,7 +21,7 @@ define([
          * @override
          */
         ifShowValue: function () {
-            if (this.getPureValue() === 0) {
+            if (this.getPureValue() == 0) {
```
ishakhsuvarov, line 24: This change makes the logic less robust, we should avoid using non-strict comparison even though JS allows this.

ishakhsuvarov (changes requested):
Please check the cause, why strict comparison does not work and try fixing the root cause of the issue.
Thank you

raivisdejus:
Correct fix for the issue proposed in new pull request -> https://github.com/magento/magento2/pull/9538

PR #9380, opened 2017-04-24
Removed unnecessary code and namespaces from import validators

ccasciotti:
Removed `init` method declaration from subclasses of `AbstractImportValidator` that simply calls `parent::init($context)` and removed unnecessary use of `Magento\CatalogImportExport\Model\Import\Product\Validator\AbstractImportValidator` from some import validators


magento-team:
@ccasciotti Thank you for the contribution

PR #9387, opened 2017-04-25
Update AttributeRepository.php

redelschaap:
Fixes a bug in case `$attribute->getOption()` returns NULL when it does not have any options. When NULL is given to `AttributeRepository::getOptionArray`, PHP complains about a TypeError.

### DIFF
--- a/app/code/Magento/Customer/Ui/Component/Listing/AttributeRepository.php
+++ b/app/code/Magento/Customer/Ui/Component/Listing/AttributeRepository.php
@@ -95,12 +95,13 @@ class AttributeRepository
             if ($entityTypeCode == AddressMetadataInterface::ENTITY_TYPE_ADDRESS) {
                 $attributeCode = self::BILLING_ADDRESS_PREFIX . $attribute->getAttributeCode();
             }
+            $attributeOptions = $attribute->getOption();
             $attributes[$attributeCode] = [
                 AttributeMetadataInterface::ATTRIBUTE_CODE => $attributeCode,
                 AttributeMetadataInterface::FRONTEND_INPUT => $attribute->getFrontendInput(),
                 AttributeMetadataInterface::FRONTEND_LABEL => $attribute->getFrontendLabel(),
                 AttributeMetadataInterface::BACKEND_TYPE => $attribute->getBackendType(),
-                AttributeMetadataInterface::OPTIONS => $this->getOptionArray($attribute->getOptions()),
+                AttributeMetadataInterface::OPTIONS => is_array($attributeOptions) ? $this->getOptionArray($attributeOptions) : [],
                 AttributeMetadataInterface::IS_USED_IN_GRID => $attribute->getIsUsedInGrid(),
                 AttributeMetadataInterface::IS_VISIBLE_IN_GRID => $attribute->getIsVisibleInGrid(),
                 AttributeMetadataInterface::IS_FILTERABLE_IN_GRID => $management->canBeFilterableInGrid($attribute),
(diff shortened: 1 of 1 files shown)


app/code/Magento/Customer/Ui/Component/Listing/AttributeRepository.php
```diff
@@ -95,12 +95,13 @@ protected function getListForEntity(array $metadata, $entityTypeCode, MetadataMa
             if ($entityTypeCode == AddressMetadataInterface::ENTITY_TYPE_ADDRESS) {
                 $attributeCode = self::BILLING_ADDRESS_PREFIX . $attribute->getAttributeCode();
             }
+            $attributeOptions = $attribute->getOption();
```
ishakhsuvarov: Original code used `$attribute->getOptions()` did you mean it?

app/code/Magento/Customer/Ui/Component/Listing/AttributeRepository.php
```diff
@@ -95,12 +95,13 @@ protected function getListForEntity(array $metadata, $entityTypeCode, MetadataMa
...
             $attributes[$attributeCode] = [
                 AttributeMetadataInterface::ATTRIBUTE_CODE => $attributeCode,
                 AttributeMetadataInterface::FRONTEND_INPUT => $attribute->getFrontendInput(),
                 AttributeMetadataInterface::FRONTEND_LABEL => $attribute->getFrontendLabel(),
                 AttributeMetadataInterface::BACKEND_TYPE => $attribute->getBackendType(),
-                AttributeMetadataInterface::OPTIONS => $this->getOptionArray($attribute->getOptions()),
+                AttributeMetadataInterface::OPTIONS => is_array($attributeOptions) ? $this->getOptionArray($attributeOptions) : [],
```
ishakhsuvarov: This line is too long. Magento code style standard requires lines to be no longer then 120 characters.

ishakhsuvarov (changes requested):
Service contract for `\Magento\Customer\Api\Data\AttributeMetadataInterface::getOptions` (which is the `$attribute` in this case) states that the only possible return type is `OptionInterface[]`. It is better to check why is it `null` in the specific case and fix it.

app/code/Magento/Customer/Ui/Component/Listing/AttributeRepository.php
redelschaap: No, don't know how it got there

app/code/Magento/Customer/Ui/Component/Listing/AttributeRepository.php
redelschaap: Updated.

redelschaap:
@ishakhsuvarov Updated the code with your change requests.

ishakhsuvarov:
@redelschaap Please check the general comment as well:
> Service contract for \Magento\Customer\Api\Data\AttributeMetadataInterface::getOptions (which is the $attribute in this case) states that the only possible return type is OptionInterface[]. It is better to check why is it null in the specific case and fix it.

I am not sure this is the most optimal fix at the moment

redelschaap:
I think this fix is optimal, because we have a situation where that method does not return an array of `OptionInterface`s, but `NULL` instead. I don't remember what exactly caused it, but it the `option` property of that object was never filled and thus `NULL`.

ishakhsuvarov:
@redelschaap Service contract claims that no `NULL` value should be returned. Could you please check the case you have encountered? We may need to fix that as it may affect much more client code.

redelschaap:
@ishakhsuvarov That's the theory, practice shows otherwise ;). I can't reproduce the bug right now as we're working towards the launch of the website.

Where do you suggest this should be fixed?

ishakhsuvarov:
@redelschaap I would suggest finding exact case, where it reproduces, and fixing the reason why `null` is returned.

redelschaap:
Well in my case an extension was responsible, that I do know. You can't say "it's the extension's fault, they should fix it", when Magento's core makes it possible to leave the `options` property `NULL`.

ishakhsuvarov:
@redelschaap Closing this PR for now. Please reopen if this issue occurs again with the reproducible steps. We will definitely look for ways to solve the problem.
Thank you for collaboration.

PR #9429, opened 2017-04-27
Fix not detecting current store using store code in url using $storeResolver->getCurrentStoreId()

mimarcel:
### Description

Current store id is not correctly identified when ```$storeResolver->getCurrentStoreId()``` is used.
```$storeManager->getStore()->getId()``` should be used instead.

### Fixed Issues

1. magento/magento2#8732: The country drop-down list display incorrect after upgrade to 2.1.4 in Admin

#### PRE-REQUISTIES:
1. Set *STORES -> Configuration -> GENERAL -> Web -> Url Options -> Add Store Codes to Urls* to *Yes*.
2. The system has at least two websites.
* The first website is set as default website
* The first website has a store view with code *default*
* The second website has a store view with code *storeview2*.
stores___settings___stores___magento_admin_
3. Set *STORES -> Configuration -> GENERAL -> General -> Country Options -> Allow Countries* to *Estonia* for first website.
4. Set *STORES -> Configuration -> GENERAL -> General -> Country Options -> Allow Countries* to *Romania* for second website.

#### STESP TO REPLICATE
1. Access the website **directly** using storeview2 store code, e.g. http://local-magento2/storeview2/
2. Add any product to cart
3. Go to checkout

#### EXPECTED
Available country for checkout: Romania

#### ACTUAL
Available country for checkout: Estonia
Checkout Page

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/Block/Checkout/DirectoryDataProcessor.php
+++ b/app/code/Magento/Checkout/Block/Checkout/DirectoryDataProcessor.php
@@ -6,6 +6,7 @@
 namespace Magento\Checkout\Block\Checkout;
 
 use Magento\Directory\Helper\Data as DirectoryHelper;
+use Magento\Store\Api\StoreResolverInterface;
 use Magento\Store\Model\StoreManagerInterface;
 
 /**
@@ -49,19 +50,21 @@ class DirectoryDataProcessor implements \Magento\Checkout\Block\Checkout\LayoutP
     /**
      * @param \Magento\Directory\Model\ResourceModel\Country\CollectionFactory $countryCollection
      * @param \Magento\Directory\Model\ResourceModel\Region\CollectionFactory $regionCollection
-     * @param StoreManagerInterface $storeManager
+     * @param StoreResolverInterface $storeResolver
      * @param DirectoryHelper $directoryHelper
+     * @param StoreManagerInterface $storeManager
      */
     public function __construct(
         \Magento\Directory\Model\ResourceModel\Country\CollectionFactory $countryCollection,
         \Magento\Directory\Model\ResourceModel\Region\CollectionFactory $regionCollection,
-        StoreManagerInterface $storeManager,
-        DirectoryHelper $directoryHelper
+        StoreResolverInterface $storeResolver,
+        DirectoryHelper $directoryHelper,
+        StoreManagerInterface $storeManager = null
     ) {
         $this->countryCollectionFactory = $countryCollection;
         $this->regionCollectionFactory = $regionCollection;
-        $this->storeManager = $storeManager;
         $this->directoryHelper = $directoryHelper;
+        $this->storeManager = $storeManager ?: \Magento\Framework\App\ObjectManager::getInstance()->get(StoreManagerInterface::class);
     }
 
     /**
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by magento-cicd2** [comment]
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9429) <br/>All committers have signed the CLA.

**Comment by hostep** [comment]
We just were hit by this bug by upgrading a store from Magento 2.1.5 to 2.1.6, but at yet another place.

Namely in the layered navigation. In version 2.1.6 Magento added some extra caching for the layered navigation select attributes, and they also use the super buggy `StoreResolver::getCurrentStoreId()` method, which doesn't do its job at all, because it always returns the default store id instead of the one on which you are browsing.
See here: https://github.com/magento/magento2/blob/2.1.6/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php#L255
This was added in this commit: https://github.com/magento/magento2/commit/67b5ff743a11449fe97d0fcf711be30df6ec878b

Replacing that with `$this->storeManager->getStore()->getId()` like is done in this PR, resolves the issue.

I think the real solution is to make sure the `StoreResolver::getCurrentStoreId()` does it job like it is supposed to do, instead of replacing it all over the codebase, but maybe that will cause new issues...

Tnx for the PR @mimarcel, saved me a lot of time and debugging!

**Comment by mimarcel** [comment]
@hostep It's indeed confusing that `getCurrentStoreId` does not return the final store id. However, the class name is called `StoreResolver`, therefore it's responsible for **resolving** the current store id. The `StoreManager` class is responsible for **managing** the current store id. The `StoreManager` uses the`StoreResolver`, but the current store id can be overwritten by other logic as well by calling `setCurrentStore`. This is quite tricky since it seems that the issue does replicate in more than one place; it seems it's not really easy to catch even for developers from Magento.

Glad this solution was useful to someone!

**Comment by hostep** [comment]
Ok I think this makes sense, thanks for the clarification!
I'll try to create a PR for the Magento 2.1.6 issue we encountered today (maybe tonight, maybe in one of the coming days).

File: app/code/Magento/Checkout/Block/Checkout/DirectoryDataProcessor.php [hunk 1]
```diff
@@ -49,18 +49,18 @@ class DirectoryDataProcessor implements \Magento\Checkout\Block\Checkout\LayoutP
...
      * @param \Magento\Directory\Model\ResourceModel\Region\CollectionFactory $regionCollection
-     * @param StoreResolverInterface $storeResolver
+     * @param StoreManagerInterface $storeManager
      * @param DirectoryHelper $directoryHelper
      */
     public function __construct(
         \Magento\Directory\Model\ResourceModel\Country\CollectionFactory $countryCollection,
         \Magento\Directory\Model\ResourceModel\Region\CollectionFactory $regionCollection,
-        StoreResolverInterface $storeResolver,
+        StoreManagerInterface $storeManager,
```
ishakhsuvarov: Unfortunately, due to backwards compatibility policy we may not change constructor arguments. Please add another optional dependency and deprecate the old one if it is not used anymore.

ishakhsuvarov (changes requested):

PR #9450, opened 2017-04-28
Fix to include an else to display a "static block" only

nathanchick:
### Description
Whilst working, we noticed that when we set the category to show static block only nothing showed on the fronted. When taking a look at the view file we noticed this was not included.

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/templates/category/products.phtml
+++ b/app/code/Magento/Catalog/view/frontend/templates/category/products.phtml
@@ -16,4 +16,6 @@
 ?>
 <?php if (!$block->isContentMode() || $block->isMixedMode()): ?>
     <?php echo $block->getProductListHtml() ?>
+<?php else: ?>
+    <?php echo $block->getCmsBlockHtml() ?>
 <?php endif; ?>


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9450) <br/>All committers have signed the CLA.

ishakhsuvarov:
Hi @nathanchick 
I have tried to reproduce the issue you have described on the latest `develop` branch and did not succeed. Could you please provide detailed steps for manual testing?

Additionally, `Catalog/view/frontend/templates/category/products.phtml` already contains pretty same logic. Please check if it is what you have been looking for. 
Thank you.

nathanchick:
The contents of the file does not contain the else I have added. 

If it is not in content mode or is in mixed it will show products otherwise nothing. 

So if the admin panel is set to 'Static Block' only it will not show the static block that has been selected on the product list page.

> Additionally, Catalog/view/frontend/templates/category/products.phtml already contains pretty same logic. Please check if it is what you have been looking for.

Please check the diff you will see what has been added and what it is currently missing.

ishakhsuvarov:
@nathanchick Could you please provide exact steps you follow to reproduce the problem? 
We can not achieve the same result as you describing.
Thank you!

nathanchick:
@ishakhsuvarov Yeah sure, it is only a small oversight but does need adding to the core.

1. Login to the admin.
2. Create a static block.
3. Navigate to: 'Products' > 'Categories'. 
4. Expand 'Display Settings'.
5. Select 'Static block only'.
6. Select your new static block in 'Add CMS Block'.
7. Save

**Expected result**
You see your static block on the frontend

**Actual result**
No static block is shown

**The reason why**
As you can see the code does not yet include a needed else block, which I have added.

https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/view/frontend/templates/category/products.phtml

I hope this is enough, I really can't explain this any further. As I said the issue is really small but I thought it is still needed.

korostii:
Hi @nathanchick,
There seems to be a [separate cms.phtml](https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/view/frontend/templates/category/cms.phtml) block [included](https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/view/frontend/layout/catalog_category_view.xml#L14) nearby which is supposed to do render that part. Are you sure you haven't removed any of its ancestors in your theme or something like that?

ishakhsuvarov:
@nathanchick We still can not reproduce this problem with the `develop` branch. All of the possible Display Settings work as expected. Could you please provide more details and check the comment above?

PR #9480, opened 2017-05-03
Refactored Source For ImportExport module, related to #7469

IvanChepurnyi:
### Description
The following pull request provides a refactored Import/Export source file parsing.
These are the major new features of this pull requests:
* Developer can create custom parsers in any format by implementing `\Magento\ImportExport\Model\Import\Source\FileParser\ParserInterface`
* Developer can create custom parser factories, so any stream or external resource can be feed as import. For adding new factory, developer needs to implement `\Magento\ImportExport\Model\Import\Source\FileParser\ParserFactoryInterface` and add it to `\Magento\ImportExport\Model\Import\Source\FileParser\CompositeParserFactory` as an argument in `di.xml`
* Zip files are now properly implemented
* Whole file parser logic is covered with tests and driven by TDD

### Fixed Issues
magento/magento2#7468: CatalogImportExport doesn't support empty row values
    *  Added possiblity to specify "null" option for CsvParser, but this option must be passed into file creation factory
    * Added extension points `\Magento\ImportExport\Controller\Adminhtml\Import\Validate::getSourceAdapterOptions()` and to `\Magento\ImportExport\Model\Import::getSourceAdapterOptions()` for override by thirdparty, as it is not yet possible to completely replace current creation logic without breaking code that is not covered with tests.
### Manual testing scenarios

1. Run integration tests with Import/Export related functionality
2. Run unit tests with Import/Export related functionality

### Contribution checklist


koenner01:
Hey IvanChepurnyi, thanks for incorporating the fix for my reported issue.
I have just found a troublesome situation when using the empty value logic (https://github.com/magento/magento2/issues/7468#issuecomment-310016729).

Just wanted to let you know in case you might run into a similiar problem

IvanChepurnyi:
@koenner01 did you checked it with "null" option passed into CSV Parser Factory and then used this as an empty value for a special price?

app/code/Magento/ImportExport/Controller/Adminhtml/Import/Validate.php
```diff
@@ -121,6 +121,19 @@ private function getImport()
     }
 
     /**
+     * Returns options for source adapter
+     *
+     * @param $data
+     * @return array
+     */
+    protected function getSourceAdapterOptions($data)
```
maghamed, line 129: we don't recommend to use Inheritance Based API, so all the methods should be whether public or private

app/code/Magento/ImportExport/Model/Import.php
```diff
@@ -785,4 +785,16 @@ public function getDeletedItemsCount()
     {
         return $this->_getEntityAdapter()->getDeletedItemsCount();
     }
+
+    /**
+     * Returns options for source adapter
+     *
+     * @return array
+     */
+    protected function getSourceAdapterOptions()
```
maghamed, line 794: we don't recommend to use Inheritance Based API, so all the methods should be whether public or private

app/code/Magento/ImportExport/Model/Import/Adapter.php
```diff
@@ -14,6 +18,15 @@
  */
 class Adapter
 {
+    /** @var FileFactory */
+    private $fileSourceFactory;
+
+    public function __construct(FileFactory $fileSourceFactory)
```
maghamed, line 24: doc block absent for this constructor

app/code/Magento/ImportExport/Model/Import/Adapter.php
```diff
@@ -25,29 +38,12 @@ class Adapter
      * @return AbstractSource
      *
      * @throws \Magento\Framework\Exception\LocalizedException
+     * @deprecated
+     * @see Adapter::createSourceByPath()
```
maghamed, line 42: It's better to avoid mixing up Deprecated logic and Newly created one in one class.
Thus, we recommend to introduce new Factory entity and mark it with @ api.
Current class to mark as @ deprecated.

Inject newly created factory as DI injection to Adapter class.
And make current class to proxy calls to newly created API

app/code/Magento/ImportExport/Model/Import/Adapter.php
maghamed, line 24: Violation of BC policies, all the parameters should be optional (nullable)

app/code/Magento/ImportExport/Model/Import/Adapter.php
```diff
@@ -58,9 +54,72 @@ public static function factory($type, $directory, $source, $options = null)
...
+
+        if (strpos($path, '.') === false) {
+            return $fileName;
+        }
+
+        $fileExtension = substr($fileName, strrpos($fileName, '.') + 1);
+        return $fileExtension;
+    }
+
+    private function throwUnsupportedFileException($path)
```
maghamed, line 102: there is no PHP Doc Block

app/code/Magento/ImportExport/Model/Import/Adapter.php
```diff
@@ -58,9 +54,72 @@ public static function factory($type, $directory, $source, $options = null)
...
+
+    /**
+     * Finds source for import by file path
+     *
+     * @param $path
+     * @param $options
+     *
+     * @return AbstractSource
+     */
+    public function createSourceByPath($path, $options = [])
```
maghamed, line 75: this should be a responsibility of new API entity

app/code/Magento/ImportExport/Model/Import/Adapter.php
```diff
@@ -58,9 +54,72 @@ public static function factory($type, $directory, $source, $options = null)
...
+            $this->throwUnsupportedFileException($path);
+        } catch (CorruptedFileException $e) {
+            $this->throwUnsupportedFileException($path);
+        }
+
+        return $parser;
+    }
+
+
+    private function extractFileExtension($path)
```
maghamed, line 90: It's better to use PHP core function for these purposes than to implement one by yourself

```php
$path_parts = pathinfo('/www/htdocs/inc/lib.inc.php');
echo $path_parts['extension'];
```

http://php.net/manual/en/function.pathinfo.php

app/code/Magento/ImportExport/Model/Import/Source/Csv.php
```diff
@@ -7,6 +7,8 @@
 
 /**
  * CSV import adapter
+ *
+ * @deprecated
```
maghamed, line 11: Reason of depreciation should be provided and @ see tag specifying the new, correct version of current API

app/code/Magento/ImportExport/Model/Import/Source/File.php
```diff
@@ -0,0 +1,31 @@
...
+use Magento\ImportExport\Model\Import\AbstractSource;
+
+class File extends AbstractSource
+{
+    private $parser;
+
+    public function __construct(FileParser\ParserInterface $parser)
+    {
+        $this->parser = $parser;
+        parent::__construct($this->parser->getColumnNames());
```
maghamed, line 18: it's not correct to make computation in constructors, Magento encourages developers to use constructors just for assignments

app/code/Magento/ImportExport/Model/Import/Source/File.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Magento\ImportExport\Model\Import\Source;
+
+use Magento\ImportExport\Model\Import\AbstractSource;
+
+class File extends AbstractSource
+{
+    private $parser;
+
+    public function __construct(FileParser\ParserInterface $parser)
```
maghamed, line 15: PHP Doc Block absent

app/code/Magento/ImportExport/Model/Import/Source/File.php
```diff
@@ -0,0 +1,31 @@
...
+        $this->parser = $parser;
+        parent::__construct($this->parser->getColumnNames());
+    }
+
+    protected function _getNextRow()
+    {
+        return $this->parser->fetchRow();
+    }
+
+    public function rewind()
```
maghamed, line 26: Doc Blocks are absent

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+namespace Magento\ImportExport\Model\Import\Source\FileParser;
+
+use Magento\Framework\App\Filesystem\DirectoryList;
+use Magento\Framework\Filesystem;
+
+class ZipParserFactory implements ParserFactoryInterface
+{
+    private $filesystem;
```
maghamed, line 14: doc block are absent

app/code/Magento/ImportExport/Model/Import/Source/FileFactory.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param array $options
+     * @return AbstractSource
+     */
+    public function createFromFilePath($path, array $options = [])
+    {
+        return $this->createFromFileParser(
+            $this->parserFactory->create($path, $options)
+        );
+    }
+
```
maghamed, line 42: why do we have two different methods of creation in the same class? 
even dependencies and input parameters are different. 
It's better to segregate this code onto two different classes.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
```diff
@@ -0,0 +1,62 @@
...
+
+        foreach ($parserFactories as $parserFactory) {
+            $this->addParserFactory($parserFactory);
+        }
+    }
+
+    public function create($path, array $options = [])
+    {
+        foreach ($this->parserFactories as $parserFactory) {
+            $parser = $this->createParserIfSupported($parserFactory, $path, $options);
```
maghamed, line 30: this does not look like a proper Object Oriented way of polymorphic object creation. 

because we instantiate in advance all the factories and just looping through all of them trying to instantiate needed parser

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        throw new UnsupportedPathException($path);
+    }
+
+    private function createParserIfSupported(ParserFactoryInterface $parserFactory, $path, array $options)
+    {
+        try {
+            $parser = $parserFactory->create($path, $options);
+        } catch (UnsupportedPathException $e) {
+            return false;
```
maghamed, line 57: this does not look like a proper way of work with Exceptions

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
```diff
@@ -0,0 +1,62 @@
...
+    /** @var ParserFactoryInterface[] */
+    private $parserFactories;
+
+    public function __construct(array $parserFactories = [])
+    {
+        $this->parserFactories = [];
+
+        foreach ($parserFactories as $parserFactory) {
+            $this->addParserFactory($parserFactory);
+        }
```
maghamed, line 24: there should not be logic in the constructors , just assignments

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
```diff
@@ -0,0 +1,62 @@
...
+                continue;
+            }
+
+            return $parser;
+        }
+
+        $this->thereWasNoParserFound($path);
+    }
+
+    public function addParserFactory(ParserFactoryInterface $parserFactory)
```
maghamed, line 42: this method should not be public, actually I doubt we need it at all , because all it does is just add item to the array

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CorruptedFileException.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Magento\ImportExport\Model\Import\Source\FileParser;
+
+class CorruptedFileException extends \RuntimeException
+{
+    private $fileName;
+
+    public function __construct($fileName = '', $message = '')
+    {
+        $this->fileName = $fileName;
+        parent::__construct($message ?: sprintf('File "%s" is corrupted', $fileName));
```
maghamed, line 16: something wrong with this code, because it has two default parameters 
$fileName = '', $message = ''
and creating an object with **default** parameters we got a message 
`'File "" is corrupted'`

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CorruptedFileException.php
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+
+    public function getFileName()
+    {
+        return $this->fileName;
+    }
```
maghamed, line 23: why do we need public getter? 
This violates encapsulation in OOP

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParser.php
```diff
@@ -0,0 +1,113 @@
...
+     * List of CSV parsing options
+     *
+     * @var array
+     */
+    private $options = [
+        'delimiter' => ',',
+        'enclosure' => '"',
+        'escape' => '\\',
+        'null' => null
+    ];
```
maghamed, line 24: why did we hardcoded options? 
you suppose to use DI configuration instead

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParser.php
```diff
@@ -0,0 +1,113 @@
...
+     * @var string[]
+     */
+    private $columns;
+
+    public function __construct(
+        Filesystem\File\ReadInterface $file,
+        $options = []
+    ) {
+        $this->file = $file;
+        $this->options = $options + $this->options;
```
maghamed, line 45: The + operator returns the right-hand array appended to the left-hand array; **for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored.**

```php
$a = array("a" => "apple", "b" => "banana");
$b = array("a" => "pear", "b" => "strawberry", "c" => "cherry");

$c = $a + $b; // Union of $a and $b
echo "Union of \$a and \$b: \n";
var_dump($c);

Union of $a and $b:
array(3) {
  ["a"]=>
  string(5) "apple"
  ["b"]=>
  string(6) "banana"
  ["c"]=>
  string(6) "cherry"
}
```
This is a **bug**, because you probably want to provide an ability to overwrite existing options

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParser.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->columns = $this->fetchCsvLine();
+
+        if ($this->columns === false) {
+            throw new \InvalidArgumentException('CSV file should contain at least 1 row');
+        }
+    }
+
+    public function getColumnNames()
+    {
+        return $this->columns;
```
maghamed, line 55: public getter violates encapsulation

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParserFactory.php
```diff
@@ -0,0 +1,59 @@
...
+
+        if (substr($filePath, -4) !== '.csv') {
+            throw new UnsupportedPathException($filePath);
+        }
+
+        $directory = $this->filesystem->getDirectoryRead($directoryCode);
+        $filePath = $directory->getRelativePath($filePath);
+
+        if (!$directory->isFile($filePath)) {
+            throw new \InvalidArgumentException(sprintf('File "%s" does not exists', $filePath));
```
maghamed, line 48: exist

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParserFactory.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->filesystem = $filesystem;
+        $this->objectManager = $objectManager;
+    }
+
+    public function create($filePath, array $options = [])
+    {
+        $directoryCode = $options['directory_code'] ?? DirectoryList::ROOT;
+
+        if (substr($filePath, -4) !== '.csv') {
```
maghamed, line 40: it's better to use getExtension function, moreover you already have one

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+    private $parserFactory;
+
+    public function __construct(
+        Filesystem $filesystem,
+        ParserFactoryInterface $parserFactory,
+        $isZipAvailable = null
+    ) {
+        $this->filesystem = $filesystem;
+        $this->parserFactory = $parserFactory;
+        $this->isZipAvailable = $isZipAvailable ?? extension_loaded('zip');
```
maghamed, line 25: why do we use extension_loaded if we can use composer required ?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+class ZipParserFactory implements ParserFactoryInterface
+{
+    private $filesystem;
+    private $isZipAvailable;
+    private $parserFactory;
+
+    public function __construct(
+        Filesystem $filesystem,
+        ParserFactoryInterface $parserFactory,
+        $isZipAvailable = null
```
maghamed, line 21: how is available variable could be NULL ?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+        foreach ($files as $file) {
+            $destinationFilename = $this->extractFileIntoDirectory($zip, $writeDirectory, $file);
+
+            try {
+                $parser = $this->parserFactory->create(
+                    $destinationFilename,
+                    ['directory_code' => DirectoryList::TMP] + $options
+                );
+            } catch (UnsupportedPathException $e) {
+                continue;
```
maghamed, line 55: It's so.. soo weird I can't even imagine which would work first continue or finally

Wrong handling of exceptions

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $zip = new \ZipArchive();
+
+        if ($zip->open($sourceDirectory->getAbsolutePath($path)) !== true) {
+            throw new CorruptedFileException($path);
+        }
+
+        $files = $this->fetchFileNamesFromZipFile($zip);
+
+        foreach ($files as $file) {
```
maghamed, line 46: is this a work around for 
```
.
..
filename.csv 
```
?

because from the method name looks like it supports several files archived in one archive, but it actually does not

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParser.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private function mapRowData($row)
+    {
+        $result = [];
+        foreach ($this->columns as $index => $column) {
+            $result[] = $this->mapNullValue(
+                $this->extractRowValue($row, $index)
+            );
+        }
```
maghamed, line 92: could be implemented more elegant in Functional way with array map

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ParserFactoryInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\ImportExport\Model\Import\Source\FileParser;
+
+/**
+ * File parser factory instance
+ *
+ */
```
maghamed, line 12: should be marked as @ api

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ParserInterface.php
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\ImportExport\Model\Import\Source\FileParser;
+
+/**
+ * File parser for import of source format
+ *
```
maghamed, line 11: should be marked as @ api

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ParserInterface.php
```diff
@@ -0,0 +1,37 @@
...
+     * @return string[]
+     */
+    public function getColumnNames();
+
+    /**
+     * Reads a single row from a file
+     *
+     * Must return an array of values in the same order as in getColumnNames call
+     *
+     * @return array|false
```
maghamed, line 27: why do we return **false** in negative scenarios here but not throw an exception? 
Even in your description you say

>  Must return an array of values 

We make the code of business logic always add conditional logic for this **false** value

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
```diff
@@ -0,0 +1,24 @@
...
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\ImportExport\Model\Import\Source\FileParser;
+
+class UnsupportedPathException extends \RuntimeException
+{
+    private $path;
+
+    public function __construct($path = '', $message = '')
```
maghamed, line 13: looks like both of parameters should be mandatory, but not optional 

because creating object with default params , which should be okay, you create a message 
'Path "" is not supported

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
```diff
@@ -0,0 +1,24 @@
...
+    {
+        parent::__construct($message ?: sprintf('Path "%s" is not supported', $path));
+
+        $this->path = $path;
+    }
+
+    public function getPath()
```
maghamed, line 20: looks like we have this public method just to use one in **Unit tests.**
We have a lot of public methods in this Pull Request which were added just to be used in Unit Tests.
They are not used in business logic.
Such approach violates the Object Encapsulation and contradicts with Object Oriented design

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+
+            return $parser;
+        }
+
+        throw new UnsupportedPathException($path);
+    }
+
+    private function assertZipFileIsParsable($path)
+    {
+        if (substr($path, -4) !== '.zip') {
```
maghamed, line 68: we need to use getExtension function or checkIsArchive 
This code should be easy to read

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
```diff
@@ -0,0 +1,118 @@
...
+
+            $files[] = $fileName;
+        }
+
+        return $files;
+    }
+
+    private function isDirectoryZipEntry($fileName)
+    {
+        return substr($fileName, -1) === '/';
```
maghamed, line 102: We need to use Core PHP functions, they are faster and more reliable 
`is_dir ` is a good candidate here 
http://php.net/manual/en/function.is-dir.php

app/code/Magento/ImportExport/Model/Import/Source/Zip.php
```diff
@@ -7,6 +7,8 @@
 
 /**
  * Zip import adapter.
+ *
+ * @deprecated
```
maghamed, line 11: the clear reason of depreciation should be specified , and if there is an alternative appeared , it should be specified in @ see tag

app/code/Magento/ImportExport/Test/Unit/Model/Import/AdapterTest.php
```diff
@@ -5,33 +5,157 @@
...
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Filesystem;
+use Magento\ImportExport\Model\Import\Adapter;
+use Magento\ImportExport\Model\Import\Source\FileFactory;
+use Magento\ImportExport\Model\Import\Source\FileParser\CsvParserFactory;
+use Magento\ImportExport\Model\Import\Source\FileParser\ParserFactoryInterface;
+use Magento\ImportExport\Model\Import\Source\FileParser\ZipParserFactory;
+use Magento\ImportExport\Test\Unit\Model\Import\Source\FileParser\FakeObjectManager;
+use Magento\ImportExport\Test\Unit\Model\Import\Source\FileParser\FakeParser;
+use Magento\ImportExport\Test\Unit\Model\Import\Source\FileParser\FakeParserFactory;
```
maghamed, line 18: This is not Unit test in terms of Magento 2,
This is an Integration test. Think we need to keep consistency in our approaches

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileFactoryTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $fileFactory = new FileFactory(
+            new FakeParserFactory(),
+            new FakeObjectManager()
+        );
+
+        $fileFactory->createFromFilePath('test.csv');
+    }
+
+    public function testGivenParserFactoryConfiguredWhenSourceIsCreatedByPathThenRightInstanceIsCreated()
```
maghamed, line 33: it's not so easy to read this method name.
It's like 10 years ago when SMS cost a lot, you try to save space (140 symbols) for text not using spaces at all.

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileParser/FakeParserFactory.php
```diff
@@ -0,0 +1,54 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\ImportExport\Test\Unit\Model\Import\Source\FileParser;
+
+use Magento\ImportExport\Model\Import\Source\FileParser\ParserFactoryInterface;
+use Magento\ImportExport\Model\Import\Source\FileParser\UnsupportedPathException;
+
+class FakeParserFactory implements ParserFactoryInterface
```
maghamed, line 12: We have so many Fake objects here, and we need to update their code updating code of the real objects as well 

Currently, we don't use this approach for Unit testing. And definitely it  needs separate approval from AC 

Because for now we introduced additional entities to maintain. But why do we need that if we already have Integration tests?

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileParser/_files/test_options.csv
```diff
@@ -0,0 +1 @@
+column1|;column2;|column3
```
maghamed, line 1: why do we need to use CSV files for Unit testing? 

Integration tests are made for these purposes

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
```diff
@@ -0,0 +1,62 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\ImportExport\Model\Import\Source\FileParser;
+
+/**
+ * File Parser Composite
```
maghamed, line 10: should be marked as @  api , because used as an Extension Point

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileParser/FakeObjectManager.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $constructorSignature = new \ReflectionMethod($type, '__construct');
+        $resolvedArguments = [];
+        foreach ($constructorSignature->getParameters() as $parameter) {
+            $arguments = $this->assertParameterValue($arguments, $parameter);
+
+            $resolvedArguments[] = $arguments[$parameter->getName()] ?? $parameter->getDefaultValue();
+        }
+
+        return $resolvedArguments;
```
maghamed, line 45: when we provide an implementation of Fake objects, who and how will test this implementation? 

For example, current function is **10 lines of code**. How can we make sure that this code is safe to rely upon?

maghamed (approved):
My main concerns for this Pull Request:

- Incorrect following of Backward Compatibility policies (new Interfaces should be added independently to legacy code). No need to mix up legacy and new implementation in one class.
- Not proper work with Exceptions. Sometimes Exceptions substituted with FALSE value and vice versa.
- Not proper Object design for factories and parser instantiation
- Re-implemented some core functions responsible for work with filesystem
- Lot of public getters are introduced just for testing purposes (violating object encapsulation)
- All existing Unit tests are actually Integration ones, in terms of Magento 2

maghamed (changes requested):
My main concerns for this Pull Request:

- Violation of Backward compatibility policies. No need to mix up the legacy (deprecated) code with newly created APIs. It's better to segregate classes.
- There are many public getters added just for testing purposes and not used in code of business logic (violation of object encapsulation) 
- Not proper work with exceptions. Some of the exceptions are substituted with FALSE, and vice versa
- Not proper Object Oriented design for ParserFactory instantiation
- Some core functions for working with Filesystem are re-implemented on PHP level. Which is incorrect
- All Unit tests looks like Integration tests in terms of Magento 2 
- Lot of Fake objects are introduced. These fake objects contain business logic, which is not covered with any kind of test. These fake objects should be also maintained in future

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileFactoryTest.php
IvanChepurnyi: Method name looks like this because "\_" are prohibited by Magento code style guideline. Previous versions had "\_" as space.

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileParser/_files/test_options.csv
IvanChepurnyi: There are no component level integration tests in Magento 2 repository and the one you are referring to requires to work with real database and other components enabled. However, this test case is quite isolated and does not leave any side effects visible to other test cases. Unit scope for this one is parsing a ZIP file, but integration test is more about testing integration of components versus just I/O operation.

IvanChepurnyi: Would love to get an advice how to overcome a problem of running fast "integration" test witout bootstrapping all database and other components.

app/code/Magento/ImportExport/Test/Unit/Model/Import/Source/FileParser/FakeObjectManager.php
IvanChepurnyi: I will add a test case for it, and also probably move it into framework, as FakeObjectManager must be available for other factory instantiation in the test harness as well.

app/code/Magento/ImportExport/Model/Import/Adapter.php
IvanChepurnyi: Is it required to have DocBlock on private methods that describe its purpose in the name?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
IvanChepurnyi: `addParserFactory` validates type of the factory passed into the constructor. What would be your advice on doing it, unless variadic types are not supported by DI?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
IvanChepurnyi: Can you elaborate on this?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CorruptedFileException.php
IvanChepurnyi: The file name can be of a help for client code that catches this specific exception. I can remove it if filename is redundant error information for parsing files.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
IvanChepurnyi: How do you then pass valuable information related to an exception? Shall we then expect message parsing in UI for specific error that needs to be generated from domain exception?

app/code/Magento/ImportExport/Model/Import/Adapter.php
IvanChepurnyi: What do you see as a comment for a constructor?

maghamed: DocBlock is not just comment, but specification of all input types.

app/code/Magento/ImportExport/Model/Import/Adapter.php
maghamed: it's required for all methods and properties, no matter whether it private or public one

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: `is_dir` works only with real file. but here directory is detected from ZIP file format list, not extracted files.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: it does allow multiple files and takes first supported file format in Zip archive file. Adding new format to composite factory automatically adds support in zip archive.

IvanChepurnyi: Zip list is different from file system directory.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: I will move continue outside of catch.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
maghamed: why do we consider this as a correct behavior to take archive and process any first file which we are able to process in this archive?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
maghamed: you can still use [pathinfo()](http://php.net/manual/en/function.pathinfo.php)
'dirname' and 'basename'
Example:
```php
$path = 'zipfile.zip'  

$zip = new ZipArchive;

if ($zip->open($path) === true) {
    for($i = 0; $i < $zip->numFiles; $i++) {
        $filename = $zip->getNameIndex($i);
        $fileinfo = pathinfo($filename);
        copy("zip://".$path."#".$filename, "/your/new/destination/".$fileinfo['basename']);
    }                   
    $zip->close();                   
}
```
https://stackoverflow.com/questions/25959507/check-if-user-has-compressed-folder-or-files

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
maghamed: you don't process this exception anywhere currently. It's "processed" just in Unit tests.

Also, don't think that we want to show some Path information in the successor of LocalizedException, which is shown on UI. Because it's insecure to reveal such internal details.
All we will probably do is just log this exception when happened.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
maghamed: The root cause of the problem described above

> this does not look like a proper Object Oriented way of polymorphic object creation.
> 
> because we instantiate in advance all the factories and just looping through all of them trying to instantiate needed parser

Because we don't use proper Object Oriented design to choose correct factory to be used and instantiate needed parser. We just look over all possible factories and try to instantiate parser with the help of each next factory in the list. So, our Exception is not actually exception (because Exception represents unexpected behavior, but here we have expected false results) that's why we have a juggling of Exception -> FALSE
Which is bad practice working with Exceptions

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
maghamed: Variadic types are not supported by DI and I doubt it should be supported.
What is correct that DI should make a type validation for arrays, and it probably would do that in future.

For now, there are two possibilities to make a type check : 

1. Use xsi:type="object" not xsi:type="string". But doing that way you can't implement lazy loading, because ObjectManager will create an entity of needed type for you

2. Use fast type check in the constructor. It does not look well, but quite fast and adding comment that it's just for type check that would be easy to remove that in future when DI would be responsible for this operation
```php
// Type check for array to make sure 
// that all parser factories implement ParserFactoryInterface
$this->parserFactories = (function(ParserFactoryInterface ...$parserFactories) {
	return $parserFactories;
})(...$parserFactories);
```

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: But `pathinfo` does not tell if it is directory or not. So functionality will be broken.

IvanChepurnyi: Here is a proof it won't work, as internal PHP api trims the trailing slashes:
https://3v4l.org/BtgEi

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
IvanChepurnyi: It is a valid case for actual CSV file parser factory or for ZIP file parser factory as if they are used directly by a client, returning false is a terrible thing. Adding if ($status) is worse practice as it increases cyclomatic complexity of client code and hidden bugs for unchecked conditions, exceptions look better in this case in my opinion.

Also, a CompositeFactory does throw the same exception if no other parser factories gave a valid parser instance, so unexpected path gets notified up the chain. 

We cannot use here common for Magento 2 "file extension based" strategy pattern, as file path might be a stream or even a temporary file without the extension, so offloading logic to parser factory looks more clean and extensible. As well factory always throws UnsupportedPathException if the file cannot be read by any of the parsers. 

If you have a better idea how to solve this issue, I would be happy to change the code to that structure.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
maghamed: pathinfo tells enough info for your specific task, please refer to manual.
Moreover, any hardcoded values, such as '/' are forbidden to use. You should use constants instead - like DIRECTORY_SEPARATOR

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParser.php
IvanChepurnyi: Actually, it is the other way around, so it is expected behaviour, as `$options` are passed in constructor via a factory and `$this->options` are default arguments for a CSV file. I can move them to Factory to remove confusion if you'd like.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParser.php
IvanChepurnyi: It has a client `FileSource` that requires column names for proper Source implementation.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: Did you see result of `pathinfo` with proper ZIP file list information?
It does not work, here is a proof inlined:
```
<?php

var_dump(pathinfo('path/to/a/zip/dir/'));

array(3) {
  ["dirname"]=>
  string(13) "path/to/a/zip"
  ["basename"]=>
  string(3) "dir"
  ["filename"]=>
  string(3) "dir"
}
```
As you see path info treats dir in the same way as a file. This makes impossible to use it for this functionality. If I use it breaks the tests as it is unaware of ZIP file format specifics.

`DIRECTORY_SEPARATOR` is a bad idea as well, as it is OS specific, but ZIP file format is not. Functionality will be broken if run on Windows.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: Because it is standard behavior for merchants to do. User experience must be taken into account, so no special file naming and crafting of zip archive should be used, it should be simple and intuitive. If you want to import products, you should be able just to compress any file or folder with a file and import it into Magento. That is why  code must be smart enough to ignore non valid files like OS specific ".Thumbs", or not supported files.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
IvanChepurnyi: It is done for testing only, never used in DI. I can remove this code checks altogether if zip is already a required PHP extension.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
maghamed: Idea is to have a decision point. The decision point in your case is your composite factory. 
The decision point will detect which Factory should be used in each particular case and then just instantiate **the only** needed factory to process the document.
In this case, we will instantiate just one factory, instead all of the factories in advance (even if we need the only one). And will not try to process document with factories which does not fit for this document.

In this case, Exception would be a real exceptional state, not like it's now when **Exception is Expected** behavior.

>  returning false is a terrible thing.

Yes, that's why I am advising you to make your code more Object Oriented friendly to get rid of such code
```php
$parser = $this->createParserIfSupported($parserFactory, $path, $options);
if ($parser === false) {
    continue;
 }
```

_Problems you have with Exceptions and False handling is just a **side-effect of wrong OO design** with choosing right Factory to handle the precise business case_

app/code/Magento/ImportExport/Model/Import/Source/FileParser/ZipParserFactory.php
maghamed: I mean, what if the archive contains two or more valid files (products_1.csv, products_2.csv and so on).
Like before compression merchant split the original file into several files and after that compressed them all together.
Why do we choose first random file which fits for the parser, like  products_2.csv
but not all?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
IvanChepurnyi: I think you should consider different presentations layers, not only web. 
Mapping to LocalizedException doesn't mean just wrapping exception `getPath()` method. Mapper must prepare exception for specific presentation layer depending on its context. If path revealing is a bad idea for web UI, but there are valid use cases for shell application.
For instance, if I start importing multiple files from a shell script, I want to know complete path to a failed file to import. 

It is just a nice feature to have. If you recommend not having it, I am fine with removing it.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CompositeParserFactory.php
IvanChepurnyi: If you'd like I can move it to RegExp based detection on a path via factory selection strategy. But I wouldn't change actual CSV and ZIP factories for path validation, as they must throw an exception if the file is not in the expected format. 
Also, I will extract RegExp detection into FactoryDetector strategy, so any other factory detection can be replaced (e.g. in future based on actual file first bytes).

app/code/Magento/ImportExport/Model/Import/Source/FileParser/CsvParserFactory.php
IvanChepurnyi: WIll extract into separate exception class, forgot to do it here.

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
IvanChepurnyi: What if I leave path as the only required argument?

app/code/Magento/ImportExport/Model/Import/Source/FileParser/UnsupportedPathException.php
IvanChepurnyi: BTW, just checked the code and there should be no concern over security. The path is not an absolute one, as the import is going to be sandboxed into DirectoryList directory code, and the relative path extraction is used in the factory code. So no actual system path traversal can be achieved via this exception.

ishakhsuvarov:
@IvanChepurnyi @maghamed 
Is there any activity on this PR at the moment or shall I close it for now?

ishakhsuvarov:
@IvanChepurnyi @maghamed I am closing this PR for now due to inactivity. Please reopen if you wish to continue work on it.
Thanks

PieterCappelle:
Hi @IvanChepurnyi, we are talking about this PR in the Magento2 Slack Channel. Is it possible to join this channel and join the channel #import_export. Thanks!

PR #6257, opened 2016-08-20
Fix "none" is not meant to be translated in catalog_product_view layout file

azanelli:
Fix this issue: #5886

Issue #6218, opened 2016-08-18
Label 'None' should not be translated in short description of product info


yariksheptykin:
Product view page [contains a block](https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml#L92) that displays short product description. This block renders `product/view/attribute.phtml` template and [passes](https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml#L97) `at_label` to it set to `"none"`. Magento is however instructed to translate the label before passing it to the template (`translate="true"`). In the template file the label is displayed [only if its value is not](https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/view/frontend/templates/product/view/attribute.phtml#L37) `"none"`. But since it is being translated, this condition only works for English locales.

yariksheptykin:
Getting rid of `translate="true"` solves the problem.

andimov:
@yariksheptykin 
Thank you for reporting!
I think this issue related to #5886.

yariksheptykin:
Yeah, this one is a duplicate. Feel free to close.

andimov:
Ok, thanks! I'll closed this ticket as a duplicate.

PR #9502, opened 2017-05-04
update attribute.phtml to respect translation

steros:
`$_attributeLabel` contains the translated label which then fails marvelously if your language isn't English.
This is more of a quick fix cause if you do not use the translation of 'none' for an empty attribute it will still fail :(

2. ...

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/templates/product/view/attribute.phtml
+++ b/app/code/Magento/Catalog/view/frontend/templates/product/view/attribute.phtml
@@ -34,7 +34,7 @@ if ($_attributeType && $_attributeType == 'text') {
 
 <?php if ($_attributeValue): ?>
 <div class="product attribute <?php /* @escapeNotVerified */ echo $_className?>">
-    <?php if ($_attributeLabel != 'none'): ?><strong class="type"><?php /* @escapeNotVerified */ echo $_attributeLabel?></strong><?php endif; ?>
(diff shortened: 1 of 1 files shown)


davidmonichi:
For me the fix above was not enough. The label of **short_description** was translated so the fixed helped but then I had some troubles with the attributeLabel of field **description** which is not translated. Therefore I fixed it with:

`<?php if ($_attributeLabel != __('none') && $_attributeLabel != 'none'): ?><strong class="type"><?php /* @escapeNotVerified */ echo $_attributeLabel?></strong><?php endif; ?>`

Don't know if its okay to do it this way. Only to let you know.

steros:
Yeah you are right I just stumbled over this too!

davidmonichi:
Hi, I did again some research and found out that the main problem is that the short_description should actually not be translated, see issue #6218. Not sure which fix should be merged.

steros:
Ah at work I don't have that much time to research and even so I wouldn't for magento to be honest.
If it's true that the label shouldn't be translated then #6257 would be the better fix I guess.

mzeis:
Hi @steros, thanks for creating the ticket!

As the PR #6257 was accepted and merged into `develop`, would you be okay if I close this PR?

steros:
That's all right. Thanks everyone for the involvement.

PR #9503, opened 2017-05-04
Make imageAsset protected to allow extensions to access the original …

rolftimmermans:
…file

I am a developer on a Magento 2 extension that optimizes images.

Since 2.1.6 there have been some changes to how product images are stored. I am trying to make the extension work with the latest changes, but the current implementation does not allow us to access the underlying `imageAsset` in order to retrieve/change the path of the images. This is required in order to make the extension work.

Please make `imageAsset` protected instead of private so we can update our extension and make it work with the latest Magento 2 versions again.

2. ...

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Image.php
+++ b/app/code/Magento/Catalog/Model/Product/Image.php
@@ -183,7 +183,7 @@ class Image extends \Magento\Framework\Model\AbstractModel
     /**
      * @var \Magento\Framework\View\Asset\LocalInterface
      */
-    private $imageAsset;
+    protected $imageAsset;
 
     /**
      * @param \Magento\Framework\Model\Context $context


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9503) <br/>All committers have signed the CLA.

ihor-sviziev:
@rolftimmermans why do you need to change that variable? Why don't you use [Magento 2 Plugins (Interceptors)](http://devdocs.magento.com/guides/v2.1/extension-dev-guide/plugins.html) instead?

korostii:
Hi @ihor-sviziev, I am not the OP but I after taking a look at the code in question I would guess it's because of the following reasons:
There's no other way to inject a different class into that variable because the class is provided by [a private method](https://github.com/magento/magento2/blob/647367e9e31980d15a3a0951bc13427e84988b17/app/code/Magento/Catalog/Model/Product/Image.php#L900), which is inaccessible through plugins. Additionally, this method calls the `ObjectManager` directly, which effectively blocks the attempts to safely use inject anything into it as a constructor argument using DI.

Thus an alternative solution would be to move the initialization of `$viewAssetImageFactory` up into the `__construct()`. After doing so, extension developers will be free to inject an `\Magento\Catalog\Model\View\Asset\ImageFactory` descendant of their choosing. 

The way it's currently implemented seems to be intentionally unfriendly to external changes for reasons unknown. Seems like a step in a wrong direction to me.

okorshenko:
Hi @rolftimmermans 

We would like to avoid inheritance based API in Magento 2. So making it protected will promote inheritance from this class.  

You can refactor the code in the following way and create new PR or update this PR:
1. remove private method getViewAssetImageFactory. it is private. easy to remove
2. inject factory via constructor. Please, follow this rule: http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter
3. in your extension you can redefine/re-configure factory for constructor and use custom implementation

okorshenko:
Hi @rolftimmermans We are closing this PR. Please feel free to create new PR if you would like to provide a solution for your case. 

Thank you

rolftimmermans:
Thanks for your response. I've created PR #9670 based on your suggestion.

PR #9507, opened 2017-05-04
Reduced CSS File Size Output for Luma & Blank Theme by Not Compiling Mixins Library into CSS

laurnts:
### Updated Description
Magento 2 is using _extends.less to add theme specific mixins which is being called by
@import (reference) 'sources/_extends.less'

However for some reason .abs- prefixed classes went through the compilation mode which end up in the CSS file, bloating the size between 100kb up to 200kb which is particularly alot for just a CSS file. After long investigation this happens because some of the .abs- prefixed mixin is being called within a .media query method which enforces the parsing of a mixin although they are not being used.

### Manual testing scenarios
1. grunt exec:luma && grunt:less luma will compile css file for luma theme
2. All of the compiled css files are located in pub/static/frontend/...
3. Here styles-m.css and styles-l.css contains ie: .abs-revert-to-action-secondary
4. While searching through the entire Magento2 directory, this class .abs-revert-to-action-secondary is never being called directly within the phtml. It's only being called via &:extend(.abs-revert-to-action-secondary).
5. Replace the whole [.media.....] with [& when (@media-common = true)] in both Luma & Blank
5. By not compiling the whole '.abs' classes, saved about 100 - 200kb in css file size.

### Contribution checklist
 - [yes] Pull request has a meaningful description of its purpose
 - [yes] All commits are accompanied by meaningful commit messages

### DIFF
--- a/app/design/frontend/Magento/blank/web/css/source/_extends.less
+++ b/app/design/frontend/Magento/blank/web/css/source/_extends.less
@@ -12,7 +12,7 @@
 //  ---------------------------------------------
 
 & when (@media-common = true) {
-    .abs-reset-list() {
+    .abs-reset-list {
         .lib-list-reset-styles();
         > li {
             margin: 0;
@@ -25,7 +25,7 @@
 //  ---------------------------------------------
 
 & when (@media-common = true) {
-    .abs-action-link-button() {
+    .abs-action-link-button {
         .lib-button();
         .lib-link-as-button();
         .lib-css(border-radius, @button__border-radius);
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9507) <br/>All committers have signed the CLA.

laurnts:
After few hours digging into it, the pull request solution above didn't actually fix the issue. It simply making all of the .abs- prefix classes / mixins not being compiled. The reason why the CSS gets larger is because some of the .abs- prefixed mixins are rendered inside a media query besides the common media query. I'm not quite sure how to solve this yet (best way), but Luma / Blank theme works just fine without using _extends.less. Within my project I created a child theme of Luma and the layout didn't change abit even with overriding the _extends.less.

okorshenko:
@laurnts thank you for your contribution. Unfortunately, these fixed break UI on store front.
screen shot 2017-05-18 at 11 14 05

PR #9530, opened 2017-05-05
Alter Sitemap's items with results from new event

navarr:
### Description

Add an event "sitemap_prepare_items" that sends the items and the store id out, allowing customizations to add, modify, or remove items for the sitemap generation.

### Manual testing scenarios
1. Create an observer for sitemap_prepare_items
2. Add to the items an DataObject in the same format as the sitemap expects
3. Ensure the new item is in the generated sitemap.xml file

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sitemap/Model/Sitemap.php
+++ b/app/code/Magento/Sitemap/Model/Sitemap.php
@@ -277,6 +277,16 @@ class Sitemap extends \Magento\Framework\Model\AbstractModel
                 self::CLOSE_TAG_KEY => '</urlset>',
             ],
         ];
+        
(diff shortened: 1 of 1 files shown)


ishakhsuvarov:
Hi @navarr 
Usually we avoid adding dispatching any new events in the core code, just maintain the existing ones. Do you think it is possible to refactor the code in such way to allow usage of plugins to achieve the same result?
Thanks

navarr:
Absolutely.

Let me review the newish core code style stuff and I'll make a new PR in the future.

Is there a good way to discuss suggested implementation, or should I just approach with one and use the feedback from that to refine it?

ishakhsuvarov:
@navarr I think It would be good if you submit one and we discuss.
Thank you!

PR #9539, opened 2017-05-07
fix for "Class Magento\Framework\Console\CLI not found" in case sensitive scenarios

obuchowski:
`CLI ` is replaced with `Cli`


magento-team:
@EObukhovsky Thank you for the contribution

Issue #9518, opened 2017-05-04
Chrome version 58 causes problems with selections in the tinymce editor


hostep:
### Preconditions
1. Magento CE 2.1.6 or develop branch, doesn't matter
2. Google Chrome browser, version 58.0.3029.81 or 58.0.3029.96 or 58.0.3029.110

### Steps to reproduce
1. Start editing a CMS page in the backend of Magento
2. Make sure you have a CMS page which contains an image and/or a widget
3. Try to select the image or the widget, by clicking it.

### Expected result
1. Image or widget is selected

### Actual result
1. Nothing is selected, and we get an error in the console:
```
tiny_mce_src.js:1173 Uncaught DOMException: Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 1.
    at Editor.<anonymous> (https://magento2-test.whatever/static/version1493836737/adminhtml/Magento/backend/en_US/tiny_mce/tiny_mce_src.js:1173:27)
    at Dispatcher.dispatch (https://magento2-test.whatever/static/version1493836737/adminhtml/Magento/backend/en_US/tiny_mce/tiny_mce_src.js:538:13)
    at DOMUtils.eventHandler (https://magento2-test.whatever/static/version1493836737/adminhtml/Magento/backend/en_US/tiny_mce/tiny_mce_src.js:12740:34)
    at HTMLDocument.cb (https://magento2-test.whatever/static/version1493836737/adminhtml/Magento/backend/en_US/tiny_mce/tiny_mce_src.js:7066:14)
```

### Discussion
This doesn't happen in Safari 10.1 or Firefox 53.0. I'm pretty sure this used to work in Chrome 57 but haven't verified this.
The [commit log](https://chromium.googlesource.com/chromium/src/+log/57.0.2987.133..58.0.3029.81?pretty=fuller&n=10000) for Chrome between version 57.0.2987.133 and 58.0.3029.81 contains a bunch of updates/fixes to how they implemented the [Selection API](https://developer.mozilla.org/en-US/docs/Web/API/Selection)
More specific: https://www.chromestatus.com/feature/5696359768260608

From the source code of TinyMCE, it looks like they used the `setBaseAndExtent` method as a workaround for [an old bug](http://bugs.webkit.org/show_bug.cgi?id=12250) in the WebKit rendering engine.

If I remove this workaround and use a normal `select` call, everything seems to be working again in Chrome 58, and also keeps working in Safari 10.1
So this might be the solution then:
```diff
[...]

PR #9540, opened 2017-05-07
Removed workaround for old Webkit bug in the TinyMCE editor for selec…

hostep:
…tions of images and hr elements, because this workaround now causes problems in Google Chrome version 58.

### Description
This PR attempts to solve a new bug which popped up by upgrading Google Chrome to version 58.
In the TinyMCE editor, we can no longer select images (or widgets, which are represented by an image), so we can no longer edit images (or widgets) using the TinyMCE editor.
See https://github.com/magento/magento2/issues/9518 for more detailed info with links to a bunch of external sources.

1. https://github.com/magento/magento2/issues/9518: Chrome version 58 causes problems with selections in the TinyMCE editor

### Manual testing scenarios
1. Start editing a CMS page in the backend of Magento using Google Chrome version 58
2. Make sure you have a CMS page which contains an image and/or a widget
3. Try to select the image or the widget, by clicking it
4. You won't succeed, because a JS error is being thrown: `Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 1.`

### Contribution checklist

### Discussion
I only briefly tested these changes, only on Chrome 58 and Safari 10.1, so it would be highly appreciated if someone else can test all browsers which Magento supports and see if this doesn't break any other browser.

I applied the changes to all tiny_mce scripts I could find, although I think only the `tiny_mce_src.js` one is used in the backend of Magento. But since you bundle all the other scripts in Magento, It could be possible that someone is using another script, so I assumed this should be fixed in all the tiny_mce scripts.
Also: the minified tiny_mce scripts were manually edited, I don't know the command which was being used to generate the minified scripts, so please double check that my manual edits are ok.

Since this bug will probably popup with a lot of shopowners in the near future, this fix should get a bit of a higher priority and should also get backported to Magento versions 2.1 and 2.0 as soon as possible.
If this PR is accepted, I can create a backport for version 2.1, but not sure how backporting to 2.0 is supposed to work?


hostep:
I just noticed this problem is also manifested in Magento 1 shops in Chrome 58. So if Magento releases a new version for version 1, this should best be included as well I believe (maybe even in a SUPEE patch, so we can apply it to older Magento versions as well?).

@okorshenko: any reasons why this PR is being ignored, or is it still under investigation by Magento?
I believe it's a pretty high priority problem, since the Google Chrome browser is now the dominant player and most people will already have upgraded to version 58 and will no longer be able to use the TinyMCE editor in the backend of Magento...

Thanks!

omiroshnichenko (approved):

magento-team:
@hostep thank you for your contribution. Your Pull Request has been successfully merged

PR #9545, opened 2017-05-08
Enable filter type extension in search request

markoweb:
Extending the filter type is not possible due to strict fields and methods restriction. Also would be better to use builders for creating the new filter instances with possibility to define filter types through dependency injection.

### DIFF
--- a/lib/internal/Magento/Framework/Search/Request/Cleaner.php
+++ b/lib/internal/Magento/Framework/Search/Request/Cleaner.php
@@ -14,7 +14,7 @@ class Cleaner
     /**
      * @var array
      */
-    private $requestData;
+    protected $requestData;
 
     /**
      * @var array
@@ -24,7 +24,7 @@ class Cleaner
     /**
      * @var array
      */
-    private $mappedFilters;
+    protected $mappedFilters;
 
     /**
      * @var AggregationStatus
@@ -220,7 +220,7 @@ class Cleaner
      * @param array $filterReference
      * @return array
      */
-    private function processFilterReference($filterReference)
+    protected function processFilterReference($filterReference)
     {
         foreach ($filterReference as $key => $value) {
             $this->cleanFilter($value['ref']);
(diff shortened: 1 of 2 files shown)


maghamed:
In accordance with [Magento 2 technical vision](http://devdocs.magento.com/guides/v2.1/coding-standards/technical-guidelines/technical-guidelines.html) 

> 2.6. Inheritance SHOULD NOT be used. Composition SHOULD be used for code reuse.

We don't encourage to use Inheritance Based API, and strongly encourage to use composition instead

maghamed:
The potential issue I see is that we don't have an interface for 
\Magento\Framework\Search\Request\Builder
to provide an ability substitute out of the box implementation with a custom one.
But could you please provide your use-case why do you want to make customization at this level?

markoweb:
I have implemented a custom filter type for the Elasticsearch (EE) module (nested datatype). It implements the `\Magento\Framework\Search\Request\FilterInterface` interface.
The problem is that in the following places it will crash because of the unrecognized data type:
* https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/Search/Request/Cleaner.php#L196
* https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/Search/Request/Mapper.php#L233
E.g. there is a filter instance creation hardcoded for a predefined types in the switch. That could be solved by creating builders and defining them through the dependency injection.

maghamed:
the problem with Filters why we don't have FilterBuilder as a part of the contract in Framework is that each specific filter is very different from others.
For example, we have such global Builder for MySQL implementations.
Which operates with existing filter Builders
https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/Search/Adapter/Mysql/Filter/Builder.php

maghamed:
@markoweb I agree that we have an architectural gap here, not providing the ability to introduce new Filters and throwing Exceptions if they added.
I will think about how that could be fixed. 
@markoweb are you interested to continue work on this issue if I'll provide more sophisticated but more robust solution from the M2 Architecture perspective?

okorshenko:
Hi @markoweb 
Are you going to proceed with this PR or we should close it?
Thank you

markoweb:
Hi @okorshenko 
We can close it

Issue #9445, opened 2017-04-28
Cart 860 does not contain item 1204


niranjanleanswift:
### Preconditions

1. Magento version - 2.1.6

### Steps to reproduce

1. When a product is add to cart it produces this error(Cart 860 does not contain item 1204)
2. 860- quote Id, 1204-quote item id.

### Expected result

1. The cart should be added successfully.

### Actual result

1. [Screenshot, logs]

korostii:
The issues which cannot be consistently reproduced on a clean installation of Magento 2 CE aren't likely to get resolved in here.
Please try add as much information as possible, including the environment you use, 3rd party themes and modules installed (if any), preconditions, exact steps to follow, screenshots and stack trace from browser console and even the custom code you're using. These raise the chances of the bug getting reproduced and acknowledged. 
If you don't provide enough information, it is likely that the issue will receive a "needs update" label and will get closed in a couple weeks.

cygnet-aamunshi:
Hello,

We can reproduce this issue. For every fresh quote for the first time when a fresh new session is initialise and if we try to add multiple products programmatically then it will give us the given error after adding the first product to cart.

After the first error is reported it works smoothly for every time we try to add the bunch of products. ut, if we clear browser cache or place an order and then try to add the same bunch of products then it gives the error after adding the first product. 

It seems that after adding the first product, when we go to add the second one and the code that checks for the existing items in cart is creating issue.

magento-team:
Internal ticket to track issue progress: MAGETWO-69005

magento-team:
Internal ticket to track issue progress: MAGETWO-70267

PR #9553, opened 2017-05-08
Change when item existence is checked in order to update / add new item, see #9445

frosit:
### Description
This commit corrects a validation within a quote save method. Because this function checks whether `getItemById($itemId)` returns an item too late (while already being in the update part) the exception message `Cart %1 does not contain item %2` is shown. By doing this validation one line earlier, it can continue in the else part if this item is not known to the quote. In the else part, a new product is added to the cart as it should.

It's not easy to reproduce. In order to reproduce and verify this, i've setup a small module in a repository that can be used. The README file contains more in-depth explanation.

https://github.com/frosit/module-devpoc.git

1. magento/magento2#9445: Cart % does not contain item %

### Manual testing scenarios
1. Follow the write-up and use the POC module at https://github.com/frosit/module-devpoc.git

### Contribution checklist

### DIFF
--- a/app/code/Magento/Quote/Model/Quote/Item/CartItemPersister.php
+++ b/app/code/Magento/Quote/Model/Quote/Item/CartItemPersister.php
@@ -58,8 +58,7 @@ class CartItemPersister
         $itemId = $item->getItemId();
         try {
             /** Update existing item */
-            if (isset($itemId)) {
-                $currentItem = $quote->getItemById($itemId);
+            if (isset($itemId) && $currentItem = $quote->getItemById($itemId)) {
                 if (!$currentItem) {
                     throw new NoSuchEntityException(
                         __('Cart %1 does not contain item %2', $cartId, $itemId)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9553) <br/>All committers have signed the CLA.

okorshenko:
Hi @frosit 
Thank you for your contribution. Unfortunately we can not accept this pull request. By design we should not create new cart item based on existing item (when item id is defined). If we will accept your fix, it will add incorrect behavior to the system. We executed extended set of the automated tests against your change and a lot of tests failed due to incorrect system behavior.

cygnet-aamunshi:
Hello,

I have replicated the issue. If we try to add multiple products programmatically then we get the said error for the first instance. After that it works smoothly and all the products are added to cart without any issue.

But, if we clear the browser cache and again try to add the same bunch of products then it will give same error for the first time and after that it again works fine.

As reported by @frosit getItemById($itemId) returns null even if the item is there in the cart. I have verified it in database as well.

PR #9574, opened 2017-05-10
no need to create customer once u got the quote object

sivajik34:
we can avoid the loading of customer once u have quote object.
we need to create customer object if quote object not able to get by customer id.


magento-team:
@sivajik34 thank you for your contribution

PR #9588, opened 2017-05-10
Support controller src_type for head links

kassner:
`\Magento\Framework\View\Page\Config\Generator\Head::SRC_TYPE_CONTROLLER` was foreseen but never implemented. This small PR allows to define `<link>` tags with dynamic URLs linking back to the current store. 

## Example

```
    <head>
        <css src="customcss/render/css" src_type="controller" />
    </head>
</page>
```

This will render:

    <link  rel="stylesheet" type="text/css"  media="all" href="http://magento.dev/customcss/render/" />

Given that `http://magento.dev/` is the Base URL.

### DIFF
--- a/lib/internal/Magento/Framework/View/Page/Config/Generator/Head.php
+++ b/lib/internal/Magento/Framework/View/Page/Config/Generator/Head.php
@@ -58,13 +58,22 @@ class Head implements Layout\GeneratorInterface
     protected $pageConfig;
 
     /**
+     * @var \Magento\Framework\UrlInterface
+     */
+    protected $url;
+
+    /**
      * Constructor
      *
      * @param \Magento\Framework\View\Page\Config $pageConfig
+     * @param \Magento\Framework\UrlInterface $url
      */
-    public function __construct(\Magento\Framework\View\Page\Config $pageConfig)
-    {
+    public function __construct(
+        \Magento\Framework\View\Page\Config $pageConfig,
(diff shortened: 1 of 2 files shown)


ihor-sviziev:
Why it's needed? Any custom css should be compiled during setup:static-content:deploy.

kassner:
I understand that compiling is the best practice, but in the real world
clients want to add _some_ CSS by themselves. We have a lot of customers
with this feature in M1 and now that we are proceeding with upgrades they
keep asking for it.

The project still needs to provide a controller for that, so just merging
this PR doesn't give them the ability to add custom CSS via controller.

Also, this patch influences the <link> tag, and more 'non-crazy real world'
scenarios can be supported, like
https://gist.github.com/kevinSuttle/1997924#html-link-tags. I'm just trying
to give the ability to the developer to render a in-store URL with that.

ihor-sviziev:
In this case it's better to create some file(s) inside static directory with your code & give link to it.

kassner:
How can we do that without writing down the baseurl inside the template? As
far as I researched, it isn't possible, and it won't cover the other cases.

OZZlE:
which versions of magento was it merged to? Is it fixed in 2.1.x?

ihor-sviziev:
Hi @OZZlE,
This issue was fixed in 2.2.x branch, looks like it wasn't backported to 2.1-develop branch

PR #9598, opened 2017-05-11
Use CSS style instead of width and height attributes for logo image

shanavas786:
### Description
width and height attributes does not override aspect ratio of logo image.
Using css styles does that.

### Manual testing scenarios
1. Upload a logo image with a different aspect ratio
2. user defined width and size are forcefully applied now

### Contribution checklist

### DIFF
--- a/app/code/Magento/Theme/view/frontend/templates/html/header/logo.phtml
+++ b/app/code/Magento/Theme/view/frontend/templates/html/header/logo.phtml
@@ -19,8 +19,8 @@
 <?php endif ?>
         <img src="<?php /* @escapeNotVerified */ echo $block->getLogoSrc() ?>"
              alt="<?php /* @escapeNotVerified */ echo $block->getLogoAlt() ?>"
-             <?php echo $block->getLogoWidth() ? 'width="' . $block->getLogoWidth() . '"' : '' ?>
-             <?php echo $block->getLogoHeight() ? 'height="' . $block->getLogoHeight() . '"' : '' ?>
+             style="<?php echo $block->getLogoWidth() ? 'width:' . $block->getLogoWidth() . 'px;' : '';
+                    echo $block->getLogoHeight() ? 'height:' . $block->getLogoHeight() . 'px;' : '' ?>"
         />
 <?php if ($block->isHomePage()):?>
     </strong>


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9598) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=9598) before we can accept your contribution.<br/>

brendanfalkowski:
I think it's preferable to set neither `width/height` attributes or `style` properties set on the logo.

Under normal circumstances, anyone making a logo graphic and uploading it will expect that the dimensions used in their graphics app will be exactly the dimensions it renders at. The browser does this automatically, so there's no need to tell it the dimensions using either technique. 

Caveat: setting `width/height` property **can** slightly improve rendering speed because the browser's layout engine can reserve space and move on before the image loads, but this is (in a practical sense) a completely negligible operation — especially considering the asset will be cache after the first load.

Under abnormal circumstances (i.e. when I want to transform the logo or swap it at different breakpoints) having defined properties or styles doesn't help either. That would always be handled directly in CSS.

shanavas786:
> I think it's preferable to set neither width/height attributes or style properties set on the logo.

Magento does accept with and height properties while uploading a logo and sets the `width` and `height` attributes accordingly.

> Under abnormal circumstances (i.e. when I want to transform the logo or swap it at different breakpoints) having defined properties or styles doesn't help either. That would always be handled directly in CSS.

I think this is not true as inline styles have higest priority than other styles (unless forced by `!important `)

brendanfalkowski:
@shanavas786 — True, my point was neither is actually necessary. And if you do assign inline styles then it may break CSS that is being used to transform the logo dimensions at different breakpoints.

It feels bad to have a CSS definition be dependent on config that's set properly (or not) but only stored in the database. As a developer I would just ignore those methods to prevent somebody in the admin breaking the site.

eug123:
As already mentioned by @brendanfalkowski

These changes will break default responsive image behavior: maximum width is set to 100%, so the image will not go out of the container in case of page resize, setting the height in the inline styles will override "auto" value set in external CSS and page resize may cause image distortion.

The browser will render image properly without attributes, but it's still recommended that image dimensions have to be specified to improve rendering speed.

You can override logo.phtml for your theme if you want to control image ratio in the admin panel, but still, it is better to do in the image editor.

shanavas786:
@eug123 @brendanfalkowski makes sense. Then it would be better to remove width and height attribute fields from logo upload form

okorshenko:
@shanavas786 could you please sign **Contributor License Agreement**? Thank you

okorshenko:
@eug123 thank you for feedback. Closing this PR.

PR #9261, opened 2017-04-15
Remove zend json from weee

dmanners:
### Description
Since Zend Framework 1 is at end of life we should remove the usage of Zend_Json. This review is part of #9236 

### Fixed Issues (if relevant)

1. magento/magento2#9236: Upgrade ZF components. Zend_Json (Part fix)

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #9600, opened 2017-05-11
Do not hardcode product link types

kassner:
When creating a new relation type, it is not possible to save it from the product edit form on admin, since `Magento\Catalog\Controller\Adminhtml\Product\Initialization\Helper` only iterates over the original link types (`related`, `crosssell`, `upsell`) because they are hardcoded.


fooman:
Hey @kassner - thanks for the pull request. Makes sense.

Can you please take a look at the Travis failures along with my comment on the constructor.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
```diff
@@ -90,14 +95,16 @@ public function __construct(
         StockDataFilter $stockFilter,
         \Magento\Catalog\Model\Product\Initialization\Helper\ProductLinks $productLinks,
         \Magento\Backend\Helper\Js $jsHelper,
-        \Magento\Framework\Stdlib\DateTime\Filter\Date $dateFilter
+        \Magento\Framework\Stdlib\DateTime\Filter\Date $dateFilter,
+        \Magento\Catalog\Model\Product\LinkTypeProvider $linkTypeProvider
```
fooman: To introduce a new constructor argument as backwards compatible as possible can you please follow this pattern used here: https://github.com/magento/magento2/pull/9261/files#diff-e4882753f27f1777c744e441e2b4f9e2R103

kassner: It seems to be a bit counter intuitive to use ObjectManager here when "the good practices" tell you to avoid touching it, specially for a class that is not part of the `@api`.

fooman: Indeed - but I guess it's currently better to be able to move forward without breaking things until a deprecation strategy is adopted.

fooman:
If you could apply the changes you added to #9601 to here that be great, as develop gets processed before the backport.

magento-team:
@kassner thank you for your contribution to Magento 2 project

PR #9601, opened 2017-05-11
Do not hardcode product link types [2.1 backport]

kassner:
Backport for #9600

--
When creating a new relation type, it is not possible to save it from the product edit form on admin, since `Magento\Catalog\Controller\Adminhtml\Product\Initialization\Helper` only iterates over the original link types (`related`, `crosssell`, `upsell`) because they are hardcoded.


magento-team:
@kassner thank you for your contribution

PR #9603, opened 2017-05-11
Di config refactor

sivajik34:
i think we can create separate files like preference.xml,type.xml,virtualtype.xml,plugin.xml instead of single di.xml file.
its not complete PR.
Just basic idea to move further.
Magento itself creating events.xml,routes.xml,menu.xml files instead of single config.xml file.
any feedback welcome.


adragus-inviqa:
You keep coming with PRs, but you never give a reason as to WHY, or any advantages for your change. M2 splits the XMLs by their _concern_ (events, di, menu etc.) Why do you want the di split even further? What advantages would that give? Don't you think we have enough XMLs already?

I mean, it's ok to propose stuff, but you're just wasting time if most of your PRs are going to be dismissed out right. Give more thought to your changes/refactoring.

I saw many of your PRs were made just for the sake of contributing. You won't get many PRs accepted that way.

sivajik34:
@adragus-inviqa understood.
If you observe out of 11-PRs  4-PRs are invalid.remaining are ok.
Present PR Advantages:  it will increase readability.
anyway oneday definitely i will contribute PR with valid reasons.

adragus-inviqa:
Sorry for sounding too salty. It's just that I hate seeing devs putting up time in their work, and their effort going in vain.

I have no problem with the current di.xml structure. But that's just me. Maybe the core devs have a different opinion.

PS: use `git pull --ff`

orlangur:
Looks like a BC break to me, at least I don't see how `app/etc/di.xml` will continue to work after such change. But let's wait for the core devs feedback before investing more efforts into this.

okorshenko:
@sivajik34 thank you for this proposal. Unfortunately we can not accept this design due to the following reason:

1. This is will be backward incompatible change if we will implement in desired state in the first iteration. In order to make it in a backward compatible way, we need at least 2.3 (first iteration. Adding new functionality + deprecation) and 2.4 release (remove deprecated code)

2. >Magento itself creating events.xml,routes.xml,menu.xml files instead of single config.xml file.

This is not the same case. Events, routes and menu configuration can be used intedenyantly, so there are no requirements to keep it in the single configuration file/configuration type. 

3. DI configuration type is solid: preferences, types, virtualtypes and plugins do not make sense without each other. We can not split DI configuration type on separate types and use them independently. We will always use them together. So there is no sense to separate them. On the other hand, we may have DI configuration type which will be populated by data from 4 different file types, but in this case there are no benefits for application. An only **potential** improvement can be - developer experience. But in this case backward compatibility is much more important.

Closing this PR.

Thank you

Issue #8018, opened 2017-01-02
Running indexer:reindex catalog_category_product fails due to limit 500


magnetic5355:
### Preconditions
1. Latest Magento2 pull from github as of 1/2/17
2. PHP 7, NGINX, mariadb

### Steps to reproduce

1. run indexer:reindex catalog_category_product

### Expected result
1.  index is rebuilt past 500 products per category resulting in products missing from site.

### Actual result
1. products are not included in the catalog_category_product_index resulting in products not appearing on site.

**Query in error:**
INSERT INTO `catalog_category_product_index` (`category_id`, `product_id`, `position`, `is_parent`, `store_id`, `visibility`) SELECT `cc`.`entity_id` AS `category_id`, `ccp`.`product_id`, `ccp`.`position`, 1 AS `is_parent`, 1 AS `store_id`, IFNULL(cpvs.value, cpvd.value) AS `visibility` FROM `catalog_category_entity` AS `cc`
 INNER JOIN `catalog_category_product` AS `ccp` ON ccp.category_id = cc.entity_id
 INNER JOIN `catalog_product_website` AS `cpw` ON cpw.product_id = ccp.product_id
 INNER JOIN `catalog_product_entity` AS `cpe` ON ccp.product_id = cpe.entity_id
 INNER JOIN `catalog_product_entity_int` AS `cpsd` ON cpsd.entity_id = cpe.entity_id AND cpsd.store_id = 0 AND cpsd.attribute_id = 94
 LEFT JOIN `catalog_product_entity_int` AS `cpss` ON cpss.entity_id = cpe.entity_id AND cpss.attribute_id = cpsd.attribute_id AND cpss.store_id = 1
 INNER JOIN `catalog_product_entity_int` AS `cpvd` ON cpvd.entity_id = cpe.entity_id AND cpvd.store_id = 0 AND cpvd.attribute_id = 96
 LEFT JOIN `catalog_product_entity_int` AS `cpvs` ON cpvs.entity_id = cpe.entity_id AND cpvs.attribute_id = cpvd.attribute_id AND cpvs.store_id = 1 WHERE (cc.path LIKE '1/2/%') AND (cpw.website_id = '1') AND (IFNULL(cpss.value, cpsd.value) = 1) AND (IFNULL(cpvs.value, cpvd.value) IN (2, 3, 4)) AND (`cc`.`entity_id` > '1051') ORDER BY `cc`.`entity_id` ASC
[...]

Issue #7968, opened 2016-12-27
Magento 2.1.3 all products not displaying after upgrade from 2.1.2


adityag186:
### Preconditions

1. MAgento 2.1.3
2. Since website has been upgraded, products are not showing correctly.

### Steps to reproduce

1. Go to www.atlanticbooks.com.
2. Select Subjects.
3. Select Business Management.

### Expected result

1. 4706 products should be displayed

### Actual result

1. Only 500 products are being displayed

grischunya:
I have the same error after reindex console.
If I add 1 new product it is added and displayed but after manual re-indexing - 500 products again

perry3181:
Same problem in 2.0.11 #7948

ndsina:
Same problem 2.1.3EE.
After investigation found, that the 'selectsByRange' method(\Magento\Framework\DB\Adapter\Pdo\Mysql.php) works incorrectly.

adityag186:
Do you know the fix for the issue?

ndsina:
I've just rewritten the 'isRangingNeeded' method(vendor/magento/module-catalog/Model/Indexer/Category/Product/AbstractAction.php) to return false, but it is temporary solution, I am looking for correct one.

adityag186:
Please update me if a permanent solution is released. Thanks.

ndsina:
Ok

perry3181:
Hello @adityag186, I just go to your store, seems you already fixed the issue.
Could you let me know how to fix it? Thanks!

lennartbee:
@ndsina Thanks, this works perfect for now

perry3181:
@lennartbee, I also tried rewrite 'isRangingNeeded' method to return false. But it's not work for me.

lennartbee:
@perry3181 is your magento crun running? Try reindexing manually also
 `php bin/magento indexer:reindex`

perry3181:
@lennartbee Yes, I tried rewrite and then reindex manually, but it still not work for me.
But I found that if any item changed to "Out of Stock", it will becomes visible in frontend category.

refaelgold:
@ndsina  you save my life! thanks!!!

siment:
Duplicate: #8018

shuaiZend:
@ndsina fixed,thank you for share.

xmav:
I’m closing this issue because it is a duplicate. Please reopen this issue as needed.
https://github.com/magento/magento2/issues/8018

PR #9621, opened 2017-05-13
Correct limit processing in query batch iterator

flancer64:
https://github.com/magento/magento2/issues/8018
https://github.com/magento/magento2/issues/7968

We should increase LIMIT to collect all records with ID=$max that are over the limit in selection.

### DIFF
--- a/lib/internal/Magento/Framework/DB/Query/BatchIterator.php
+++ b/lib/internal/Magento/Framework/DB/Query/BatchIterator.php
@@ -164,8 +164,32 @@ class BatchIterator implements BatchIteratorInterface
             ]
         );
         $row = $this->connection->fetchRow($wrapperSelect);
-        $this->minValue = $row['max'];
-        return intval($row['cnt']);
+
+        /* we need collect all records with ID=$upTo over LIMIT ($this->>batchSize)*/
+        $upTo = $row['max'];
+        $selectTail = $this->initSelectObject();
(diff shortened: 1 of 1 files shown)


okorshenko:
Hi @flancer64 
Could you please provide more information about this PR following the PR [description standard](https://github.com/magento/magento2/blob/develop/PULL_REQUEST_TEMPLATE.md)? 

Also, could you please fix broken tests and cover your changes with integration tests.

Thank you

flancer64:
Hello, guys.

No, I could not. I had solved my own problem - missed products on the front in the Magento 2 store. I have shared my solution as this PR and as [standalone module](https://github.com/flancer32/mage2_fix_pr9621). I can answer your questions if you will ask me. But I will not fix your unit tests and create integration test. This is your project and this is your responsibility. You should analyze this solution (it might be wrong fix and the unit tests errors would say about it), you should fix your unit tests if solution is acceptable (or you can offer a better solution based on this one) and you should create your integration test (I know nothing about your integration tests) that integrationally tests this case in the suite of the other similar cases (which cases? I don't know). This is your job, not mine. I can help you to do it, but I will not do it for you.

Sincerely yours,

Alex Gusev.

okorshenko:
Hi @flancer64 Thank you for the quick response. Unfortunately, we can not accept this PR as is without tests. We will consider your solution for this issue. Thank you for your contribution.
Closing this PR.

PR #9655, opened 2017-05-16
Backport of MAGETWO-69152: Removed workaround for old Webkit bug in t…

hostep:
…he TinyMCE editor for selections of images and hr elements, because this workaround now causes problems in Google Chrome version 58.

(cherry picked from commit 4d10f05c4d015e59e6628bd58782a811b61b13a6)

### Description
This is a backport of issue MAGETWO-69152 for Magento 2.1

1. https://github.com/magento/magento2/issues/9518: Chrome version 58 causes problems with selections in the tinymce editor

Reference to the commit on the develop branch:
4d10f05c4d015e59e6628bd58782a811b61b13a6

### Discussion

Be aware, this is a slightly different commit then the original one, because on the develop branch we have this commit which was applied: https://github.com/magento/magento2/commit/ead58ef5e87fbccfe62b49f5e432a4d59e7eadeb, but it isn't applied on the 2.1-develop branch, so there were conflicts in the minified files, please review this again, because I didn't actually test it.

Also: I believe this bug is pretty important to be fixed in Magento 2.0.x as well, are you guys taking care of that?
Also: The same is true for Magento 1.x, can you guys try to communicate this to the M1 support people and try to get it included in the next release of Magento 1.x, and maybe even create a SUPEE-patch for this. Since not every Magento 1 shop will be running on the latest version. For example, we currently maintain around 60 Magento 1 shops which aren't on the very last version (ranging between 1.7.0.2 and 1.9.1.0)

Thanks!


omiroshnichenko (approved):

magento-team:

hostep:
Thanks!

@magento-team or @omiroshnichenko: can you please give some more info about backporting this to Magento 2.0.x and Magento 1.x
So we at least know if this is going to happen or not. And if it is going to happen, could you also provide some estimated timeframe for fixing this in these older Magento versions, that would be very much appreciated! So we can decide if we should put time into this ourselves, or if we can wait for you guys to provide new versions/patches for this issue?

Thanks! :)

PR #3401, opened 2016-02-14
decouple product view gallery images config from Gallery Block

davidverholen:
Last week we had the Problem, that we wanted to add some additional image types (picture tag for different viewports) to the Product View Gallery Images.

Since the creation of the Data array for the json configuration is hardcoded in the block, we had to implement a Plugin.
https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/Block/Product/View/Gallery.php#L58

Although it also works with a Plugin, I thought it would be nicer if this configuration is decoupled from the Gallery Block, since this configuration also depends on the template having the images defined in the view.xml

PR #9670, opened 2017-05-17
Allow injection of Magento\Catalog\Model\View\Asset\ImageFactory

rolftimmermans:
It is not really possible to override the behaviour of Magento\Catalog\Model\View\Asset\ImageFactory.

Based on the suggestions given in https://github.com/magento/magento2/pull/9503#issuecomment-301195050 this implementation aims to fix that by allowing injection of a factory via the constructor.


adragus-inviqa:
No need for an underscore for member names, yeah?

rolftimmermans:
> No need for an underscore for member names, yeah?

I followed the conventions in the rest of the class...

app/code/Magento/Catalog/Model/Product/Image.php
```diff
@@ -173,7 +173,7 @@ class Image extends \Magento\Framework\Model\AbstractModel
     /**
      * @var \Magento\Catalog\Model\View\Asset\ImageFactory
      */
-    private $viewAssetImageFactory;
+    protected $_viewAssetImageFactory;
```
ihor-sviziev: Could you revert this change? Variable names should not be prefixed with underscore according to PSR-2. Also protected variables should be replaced with private for all cases where it's possible

app/code/Magento/Catalog/Model/Product/Image.php
```diff
@@ -226,6 +228,13 @@ public function __construct(
         $this->_assetRepo = $assetRepo;
         $this->_viewFileSystem = $viewFileSystem;
         $this->_scopeConfig = $scopeConfig;
+        if ($viewAssetImageFactory == null) {
```
ihor-sviziev: Would be great to replace this if block with ternary operators

adragus-inviqa:
I know, but some of it is legacy/wrong. M2 uses [PSR-2](http://devdocs.magento.com/guides/v2.0/coding-standards/code-standard-php.html) for its code style, and that _suggests_ not to use underscores.

E.g. https://github.com/magento/magento2/pull/3401#discussion_r111334856

app/code/Magento/Catalog/Model/Product/Image.php
okorshenko: please, revert this change. The property must be private

okorshenko (changes requested):

app/code/Magento/Catalog/Model/Product/Image.php
```diff
@@ -226,6 +228,13 @@ public function __construct(
...
+            $this->_viewAssetImageFactory = ObjectManager::getInstance()->get(
+                \Magento\Catalog\Model\View\Asset\ImageFactory::class
+            );
+        } else {
+            $this->_viewAssetImageFactory = $viewAssetImageFactory;
```
orlangur: Please check this out: https://github.com/magento/magento2/blob/develop/app/code/Magento/Backend/Block/System/Account/Edit/Form.php#L60

- new dependency shall be added as last one
- there is no need in `if` statement - just make a one-liner
- always use imported class name like `OptionInterface::class`

orlangur (changes requested):
Please read https://github.com/magento/magento2/pull/9503#issuecomment-300261918 more carefully.

So, even if you preferred inheritance for your customization and extended this class, you don't need to change property visibility to protected.

You may simply specify another class for this argument in `di.xml` for your new class.

rolftimmermans:
I have made changes according to the requests by @okorshenko and @orlangur. Please let me know if this is sufficient to be considered for merging. Thanks!

ihor-sviziev (approved):

ihor-sviziev (approved):

okorshenko (approved):

orlangur (approved):

magento-team:
@rolftimmermans thank you for your contribution to Magento 2 project

PR #9672, opened 2017-05-17
Revert minimum stability to stable, tasks #4359

ktomk:
Stability revert from alpha to stable. In 79c9052 set to alpha due to earlier required composer/composer version which was not yet stable.

Composer upgrade was blocked due to changes w/ composer autoloading and a PHP 5.3 regression allowing slash-prefixed class names in the past and the unlucky coincidence that di.xml files made use of such class-names, fixed in 0b243b8.

Now composer was updated in 0364bf9 and alpha degradation can be lifted.

Refs:

- #4359

- 79c90524290601c51c501fb7cf4c5dd2454a2a34

- 0b243b899f2c94a9eab18bf8faeb2b87b120f272

- 0364bf93b363fe7923b2fa95ab61f3b715828565


orlangur (approved):
Following packages are updated:
- `justinrainbow/json-schema` 5.2.0 -> 5.2.1, we don't use it directly, changelog: https://github.com/justinrainbow/json-schema/releases
- `phpseclib/phpseclib` 2.0.4 -> 2.0.5, changelog: https://github.com/phpseclib/phpseclib/blob/master/CHANGELOG.md#205---2017-05-07

which seem fine to me.

PR #9681, opened 2017-05-18
Issue 9680: Use parent name for variations

PascalBrouwers:
### Description
Use parent name for variations when creating configurables

1. magento/magento2#9680: Adding product configurations uses sku based name

### Manual testing scenarios
1. Create a configurable product with some configurations (simple products)
2. Use name 'NAME'
3. Use sku 'SKU'
4. Create configurations and Generate products
5. Confirm that name of the configurations (simple products) are based on the name of the parent product

### Contribution checklist

### DIFF
--- a/app/code/Magento/ConfigurableProduct/view/adminhtml/web/js/variations/steps/summary.js
+++ b/app/code/Magento/ConfigurableProduct/view/adminhtml/web/js/variations/steps/summary.js
@@ -91,6 +91,7 @@ define([
             var productSku = this.variationsComponent().getProductValue('sku'),
                 productPrice = this.variationsComponent().getProductPrice(),
                 productWeight = this.variationsComponent().getProductValue('weight'),
+                productName = this.variationsComponent().getProductValue('name'),
                 variationsKeys = [],
                 gridExisting = [],
                 gridNew = [],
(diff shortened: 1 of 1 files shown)


miguelbalparda:
Hi @PascalBrouwers! Thanks for your contribution! 
I have checked #9680 and it seems it might not be a bug entirely, but I still believe this might be interesting to merge. Would you mind adding some more info on why this should be included instead of the SKU variation of the product name? Thanks again!

PascalBrouwers:
I see what you mean. Using the sku could also be intended behaviour. 
But to me it's a bug because I would expect the simples to use the parent name. For merchants it is very confusing to see the simples in the product overview based on sku and would rather see it based on name. If this PR could be merged it would help Magento to be a better product.

PieterCappelle:
I agree with @pascalbrouwers.

iancassidyweb:
Agree with the @PascalBrouwers. To me it makes more sense for Merchants to work with names and not SKU's.

Having worked merchant side for an organisation with multiple admin users across several business units it's common place for employees to know the product ranges and names as they deal with these on a daily basis however it's not as common for employees to naturally know the SKU's for products unless they were involved in a more warehousing / logistics / purchasing role.

However that's just my two cents :-)

miguelbalparda:
This is breaking some tests, @vrann what are the next steps?

miguelbalparda:
This is being worked on as we speak, sorry for the late reply!

Issue #9679, opened 2017-05-18
Translation for layered navigation attribute option not working


popatkaran:
I am using Magento CE 2.1.6 and having multilingual structure of website. When I switch the store to Arabic language all text getting translated  but only attribute options in layered navigation is not translating.

### Preconditions
1. Magento CE 2.1.6 
2. php 7.0.x 

### Steps to reproduce
1. Open website with default store.
2. Switch website with store switcher or paste direct url of another store
3. Check layered navigation

### Expected result
1. All the text should be translated to applied language to relevant store. 

### Actual result
1. All other text in website getting translated properly but attribute option name is not getting translated.
- Working well with English store.
![m216en](https://cloud.githubusercontent.com/assets/5829021/26188792/e0360704-3bbe-11e7-9098-29afc8a18388.png)
- Not working with Arabic one.
![m216ar](https://cloud.githubusercontent.com/assets/5829021/26188793/e036a2b8-3bbe-11e7-890b-f42c2b6c9101.png)

tsifra:
Hello, did you translate it in backend - store - attributes - [attr name]?
If not - please read docs. And close the issue.

popatkaran:
@tsifra  yes, I have already put translated text for the same.

I have found that if I flush cache using command after switching store translation works but, If i move back to English store attribute options remains in Arabic instead of English. This happens only for Layered navigation.

tsifra:
My bad. Actually i didn't understand your problem right way. And translation of attribute was "hardcoded" by me to [atttribute] - Manage labels. Because i use only one language. Have no idea how to make different attribute labels for different languages. Maybe you have to create different stores with different labels and default languages? Please check [atttribute] - Properties tab. There are Admin and store columns for each attr value.

popatkaran:
Yes, I have multiple store which having different language as default. Store change and other front-end area are translating well but only layered navigation is causing an issue.

hostep:
We ran into the same problem today after upgrading from Magento 2.1.5 to 2.1.6.
This is a new bug introduced in 2.1.6 by this commit: 67b5ff7

It's because the wrong `storeId` is being used in the `cacheKey`, because the `storeResolver->getCurrentStoreId()` isn't always returning the current store id.

I'm currently working on a PR which will fix this.

okorshenko:
The fix was delivered to 2.1-develop and develop branch. Thank you @hostep for your contribution!

magento-team:
[...]

PR #9704, opened 2017-05-19
Fixes regression bug introduced in Magento 2.1.6 where the layered navigation options are sometimes being cached using the wrong store id.

hostep:
### Description
Fixes problem with layered navigation options being cached using the wrong store id.

This problem only exists on Magento 2.1.6, not on the develop branch.
The problem was introduced in https://github.com/magento/magento2/commit/67b5ff743a11449fe97d0fcf711be30df6ec878b

1. https://github.com/magento/magento2/issues/9679: Translation for layered navigation attribute option not working

### Manual testing scenarios
1. Set up a new Magento 2.1.6
2. Make sure you have 2 storeviews, I used the codes: 'en' and 'nl' (English and Dutch)
3. Go to Stores => Configuration => Web => Url Options and set 'Add Store Code to Urls' to 'Yes'
4. Create a new category, I used 'Test'
5. Open the 'Color' attribute and add 2 options:

| Admin | English | Dutch |
| ------ | ------- | ------ |
| Red | Red | Rood |
| Black | Black | Zwart |

6. Add the 'Color' attribute to the 'Default' attributeset
7. Create 2 products, assign them to the 'Test' category, make sure the first one has color 'Red' and the second one has color 'Black'
8. Reindex
9. Flush cache
10. **VERY IMPORTANT**: Do **NOT** use the storeswitcher on the frontend in the next steps
11. In the frontend, go the the test category in English: http(s)://mydomain/en/test.html
12. Check the color options, it will say: 'Red' and 'Black', good!
13. Now go to the test category in Dutch, directly using the url, not the storeswitcher: http(s)://mydomain/nl/test.html
14. Check the color options, it will say: 'Red' and 'Black', **not** good!
15. Flush the cache
16. Refresh the Dutch page
17. Check the color options, it will say: 'Rood' and 'Zwart', good!
18. Again, go to the English page using the url
19. Check the color options, it will say: 'Rood' and 'Zwart', **not** good!

### Discussion
This problem exists because the `$cacheKey` is using the default store id every time if you don't use the storeswitcher. So the wrong info is cached and being used from the cache in the above cases.

The bug happens, because the original code calls `$this->storeResolver->getCurrentStoreId()`, this function checks if either the url has a '___store=xxx' param or if a cookie named 'store' exists. If neither of these things exist, it will return the default store view id.
When you use the storeswitcher, it will set a cookie and the functionality will work. But what about visitors going to the url directly just after the cache was flushed? They will put the wrong info into the cache.

This PR targets the `2.1-develop` branch, as the `develop` branch hasn't gotten this MAGETWO-65484 feature yet, which is strange...

Inspiration of this PR came from this other PR, which fixes some similar issues: https://github.com/magento/magento2/pull/9429

Would be really great if this could get included in the next minor release, as this is a pretty important bug which should get fixed ASAP.

It might be a good idea if this functionality is covered by some kind of test.
I have almost 0 experience with writing tests, so I won't do this for now, if somebody else is willing to do this, be my guest :)

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php
+++ b/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php
@@ -13,6 +13,7 @@ namespace Magento\Eav\Model\Entity\Attribute\Frontend;
 
 use Magento\Framework\App\CacheInterface;
 use Magento\Store\Api\StoreResolverInterface;
+use Magento\Store\Model\StoreManagerInterface;
 use Magento\Framework\App\ObjectManager;
 use Magento\Eav\Model\Cache\Type as CacheType;
 use Magento\Eav\Model\Entity\Attribute;
@@ -20,14 +21,21 @@ use Magento\Eav\Model\Entity\Attribute;
 abstract class AbstractFrontend implements \Magento\Eav\Model\Entity\Attribute\Frontend\FrontendInterface
 {
     /**
+     * Default cache tags values
+     * will be used if no values in the constructor provided
+     * @var array
+     */
+    private static $defaultCacheTags = [CacheType::CACHE_TAG, Attribute::CACHE_TAG];
+
+    /**
      * @var CacheInterface
      */
     private $cache;
 
     /**
-     * @var StoreResolverInterface
+     * @var StoreManagerInterface
      */
-    private $storeResolver;
+    private $storeManager;
 
     /**
      * @var array
@@ -51,21 +59,21 @@ abstract class AbstractFrontend implements \Magento\Eav\Model\Entity\Attribute\F
      * @param CacheInterface $cache
      * @param StoreResolverInterface $storeResolver
      * @param array $cacheTags
+     * @param StoreManagerInterface $storeManager
      * @codeCoverageIgnore
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
     public function __construct(
         \Magento\Eav\Model\Entity\Attribute\Source\BooleanFactory $attrBooleanFactory,
         CacheInterface $cache = null,
         StoreResolverInterface $storeResolver = null,
-        array $cacheTags = [
-        CacheType::CACHE_TAG,
-        Attribute::CACHE_TAG,
-        ]
+        array $cacheTags = null,
+        StoreManagerInterface $storeManager = null
     ) {
         $this->_attrBooleanFactory = $attrBooleanFactory;
         $this->cache = $cache ?: ObjectManager::getInstance()->get(CacheInterface::class);
-        $this->storeResolver = $storeResolver ?: ObjectManager::getInstance()->get(StoreResolverInterface::class);
-        $this->cacheTags = $cacheTags;
+        $this->cacheTags = $cacheTags ?: self::$defaultCacheTags;
+        $this->storeManager = $storeManager ?: ObjectManager::getInstance()->get(StoreManagerInterface::class);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php
```diff
@@ -49,22 +49,22 @@
...
-     * @param StoreResolverInterface $storeResolver
+     * @param StoreManagerInterface $storeManager
      * @param array $cacheTags
      * @codeCoverageIgnore
      */
     public function __construct(
         \Magento\Eav\Model\Entity\Attribute\Source\BooleanFactory $attrBooleanFactory,
         CacheInterface $cache = null,
-        StoreResolverInterface $storeResolver = null,
+        StoreManagerInterface $storeManager = null,
```
maghamed: that's a backward incompatible change,
and we can't just remove existing dependency and add another one instead.

So, existing dependency should be @deprecated 
while new dependency added at the end of the list

hostep: Really? It only existed in Magento 2.1.6, in no other version. It would very much surprise me if any 3rd party already extends this class.

But if you really want me to add it as a new param, then I can do that.
Please let me know.

hostep:
@maghamed: I applied your requested changes, please let me know if everything is ok now. Thanks! :)

_UPDATE_: I forgot to mark the `$storeResolver` param in the constructor as `@deprecated`, but I don't know exactly how to do that? I can't seem to find any examples for marking a param as `@deprecated` at first sight in Magento's current code?

hostep:
I just noticed the 3 failures in the static tests.

I don't understand the first two, because there is a doc comment about the `$storeManager` variable?
I understand the third one, but how can I solve it? It is being generated because we want to remain backwards compatible by not removing the `$storeResolver` from the constructor params. But that param isn't being used anymore, so it makes sense that a warning is being thrown.

Please let me know how to resolve these static test failures. Thanks! :)

hostep:
I just see that this feature (with the same bug) has landed on the develop branch: https://github.com/magento/magento2/commit/938313ef90fae31a9f3250d7266d0885054ff9d3

hostep:
After figuring out how to run these tests locally, I was able to suppress the third warning.

The first two warnings feel like a bug in PHPCS itself, if I change the default assignment of the `$cacheTags` param to an empty array in the constructor, the static tests go through fine, so PHPCS isn't properly detecting the different params, I think.
Just a super wild guess here, but it might be fixed by upgrading PHPCS to at least version 2.0.0RC1, where they switched to using a comment tokenizer instead of a parser: https://github.com/squizlabs/PHP_CodeSniffer/commit/9113cd6b642e047358904bce1f176e55c264fdf0 (but this is only guessing, I haven't actually tested this)

Anyway, I just pushed an update to suppress the third warning.

hostep:
I just pushed another update, which should fix all static tests.

I took the approach like was done in https://github.com/magento/magento2/commit/938313ef90fae31a9f3250d7266d0885054ff9d3 with the `$defaultCacheTags` member.
Not entirely sure if this doesn't break your backwards compatibility rules, but I don't know another way of fixing the static tests (which I still believe occur because of a bug in PHP CodeSniffer).

I'm also still waiting on feedback about how to mark the `$storeResolver` param as `@deprecated`.

Also: since this bug now also exists in the develop branch, who will take care of this? Should I create a new PR against the develop branch, or are you guys looking into this?

Thanks!

hostep:
@maghamed or @okorshenko: it would be great if I could get some feedback on this PR, because it feels like this PR doesn't get any attention anymore from Magento. And since it fixes a regression bug it should get included in the next minor release preferably.

Thanks!

okorshenko:
Hi @hostep , @ishakhsuvarov is working on this PR. Initial testing successfully passed. We are working right now on more detailed testing of this pull request.

Thank you

magento-team:

hostep:
@magento-team or @okorshenko: can you let me know who will take care of fixing this on the develop branch? Are you guys looking into this, or should I create a new PR for this?

Thanks!

ishakhsuvarov:
@hostep I don't think it is in progress internally at the moment, so feel free to submit a PR for that.
Thank you

hostep:
@ishakhsuvarov: new PR created for the develop branch: https://github.com/magento/magento2/pull/9873

Issue #6455, opened 2016-09-01
Cookie Restriction Mode Overlay is cached by Varnish


bka:
Ok, here is another one related to full page caching with varnish. It's related to these other issues: #6441, #5596. Using cookie restriction mode in combination with varnish does not work, because this block is not declared as non-cachable. Having cookie restriction mode enabled, customer is asked by a dialog to accept cookies. This cookie notice block contains the question: _"We use cookies to make your experience better. To comply with the new e-Privacy directive, we need to ask for your consent to set the cookies. Learn more."_ and gives you a button to _"Allow Cookies"_. Using varnish it might happen that this dialog won't appear anymore. Here is how:
## Steps to reproduce
1. Tested on Magento 2.1 CE
2. Use Varnish as Full Page Cache
3. Enable Cookie Restriction Mode under General->Web->Default Cookie Settings
4. Flush Cache
5. Visit Store Front
6. Accept Cookies by clicking on Allow Cookies button
7. Flush Cache
8. Visit Store Front (Your acceptance will be cached)
9. Delete your Cookies
10. Visit Store Front again (I'm an unkown visitor now with no cookies)
## Expected result
1. Cookie notice block should appear again 
## Actual result
1. Cookie notice block **does not** appear (and never will for anybody until you clear cache again)

ishakhsuvarov:
Reopening, as fix for `2.1` is not implemented yet.

magento-team:
Internal ticket to track issue progress: MAGETWO-69372

sydekumf:
We still experience that in 2.2.6 with normal Magento cache (no varnish). Why?

PR #9711, opened 2017-05-20
Cookie Restriction Mode Overlay should not be cached by Varnish #6455

bka:
### Description
Changed user-dependend `isUserNotAllowSaveCookie` condition in notice.ptml to fix varnish caching issue #6455

The proposed solution for this problem is that alertinfo container is always present in dom, no matter what cookie data the user has set, it should only depend on the setting if cookie restriction mode is enabled.

1. magento/magento2#6455

### Manual testing scenarios
Follow steps in issue #6455


vrann:
@bka merged, thanks!

magento-team:
@bka thank you for your contribution. Your Pull Request has been successfully merged

PR #9714, opened 2017-05-20
Can't delete last item in cart if Minimum Order is Enable #6151

storbahn:
### Description
During editing the cart (adding or removing products from cart) the minimum order amount (mam) will be checked. This caused that the last product could not be removed from cart. In this fix the validation of the mam is removed during editing the cart. The mam will checked later during checkout.

1. magento/magento2#6151

### Manual testing scenarios
See steps to reproduce #6151:
1. Put item in cart
2. Go to admin -> SALES -> sales
3. Enable -> Minimum Order Amount
4. Minimum Amount = 10 or something
5. Remove item from cart
6. Expected result: cart is empty

### DIFF
--- a/app/code/Magento/Quote/Model/QuoteManagement.php
+++ b/app/code/Magento/Quote/Model/QuoteManagement.php
@@ -140,7 +140,7 @@ class QuoteManagement implements \Magento\Quote\Api\CartManagementInterface
     /**
      * @var \Magento\Quote\Model\Quote\Validator\MinimumOrderAmount\ValidationMessage
      */
-    private $minimumAmountErrorMessage;
+    private $minimumAmountMessage;
 
     /**
      * @param EventManager $eventManager
@@ -163,7 +163,7 @@ class QuoteManagement implements \Magento\Quote\Api\CartManagementInterface
      * @param \Magento\Customer\Model\Session $customerSession
      * @param \Magento\Customer\Api\AccountManagementInterface $accountManagement
      * @param QuoteFactory $quoteFactory
-     * @param Quote\Validator\MinimumOrderAmount\ValidationMessage $minimumAmountErrorMessage
+     * @param Quote\Validator\MinimumOrderAmount\ValidationMessage $minimumAmountMessage
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
     public function __construct(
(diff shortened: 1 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9714) <br/>All committers have signed the CLA.

app/code/Magento/Quote/Model/ShippingAddressManagement.php
```diff
@@ -117,10 +117,6 @@ public function assign($cartId, \Magento\Quote\Api\Data\AddressInterface $addres
         $address->setSaveInAddressBook($saveInAddressBook);
         $address->setCollectShippingRates(true);
 
-        if (!$quote->validateMinimumAmount($quote->getIsMultiShipping())) {
```
ishakhsuvarov, line 120: The only other check for Minimum Amount occurs in the Checkout Index controller and it is not sufficient, since checkout flow may be performed using Web API without calling the controller.
Please consider moving this check to Place Order call.

ishakhsuvarov (changes requested):

app/code/Magento/Quote/Model/QuoteManagement.php
```diff
@@ -179,7 +186,8 @@ public function __construct(
         \Magento\Checkout\Model\Session $checkoutSession,
         \Magento\Customer\Model\Session $customerSession,
         \Magento\Customer\Api\AccountManagementInterface $accountManagement,
-        \Magento\Quote\Model\QuoteFactory $quoteFactory
+        \Magento\Quote\Model\QuoteFactory $quoteFactory,
+        \Magento\Quote\Model\Quote\Validator\MinimumOrderAmount\ValidationMessage $minimumAmountErrorMessage
```
ishakhsuvarov: Due to backwards compatibility policy we can not introduce a new required constructor argument as it may break any code which extends from this class.
Please change this argument to be optional and use ObjectManager to get the dependency, it will be refactored later when such changes are allowed.
Example:
```php
public function __construct(
    \Magento\Quote\Model\OldDependency $old,
    \Magento\Quote\Model\NewDependency $new = null
) {
    $this->old = $old;
    $this->new = $new ?: \Magento\Framework\App\ObjectManager::getInstance()
            ->get(\Magento\Quote\Model\NewDependency::class);
}
```

app/code/Magento/Quote/Test/Unit/Model/QuoteManagementTest.php
```diff
@@ -122,6 +122,11 @@ class QuoteManagementTest extends \PHPUnit_Framework_TestCase
     protected $quoteMock;
 
     /**
+     * @var \Magento\Quote\Model\Quote\Validator\MinimumOrderAmount\ValidationMessage
+     */
+    protected $minimumAmountErrorMessage;
```
ishakhsuvarov: We do not encourage inheritance based API's even in unit tests.
Please change this to `private`.

ishakhsuvarov (changes requested):

ishakhsuvarov:
This could potentially fix magento/magento2#8601

bka:
@ishakhsuvarov: addressed your comments, maybe you can have a look at it again

ishakhsuvarov (approved):

magento-team:
@storbahn thank you for your contribution to Magento 2 project

magento-team:
@storbahn thank you for your contribution to Magento 2 project

vkublytskyi:
@storbahn This issue still valid for Magento 2.1. Would you be interested in backporting of this solution to `2.1-develop`?

PR #9732, opened 2017-05-22
Map modal widget file paths to aliases

ntoombs19:
### Description

In this [devdoc](http://devdocs.magento.com/guides/v2.0/javascript-dev-guide/javascript/custom_js.html), the outlined method of overriding or extending the modal widgets could not be done without duplicating the entire code of the respective widget. The proposed solution would allow the modal widgets to be extended and/or overridden rather than only overridden.

### Manual testing scenarios

The proposed solution should be easily verified as working by viewing any modal on a clean install. One example of where a modal can be opened is on the Customer Account -> Add Address page.

### Contribution checklist


omiroshnichenko:
Hi, @ntoombs19. If you want to extend modal widget globally you can use mixin functionality that described [here](http://devdocs.magento.com/guides/v2.1/javascript-dev-guide/javascript/js_mixins.html). If you want to override you can use require.js map in require.js config. Here is a small example how to make extend via mixin:
```javascript
//====== Your extension ========
define([
    'jquery',
    'jquery/ui'
], function ($) {
    'use strict';

    /**
     * @param {Function} original - Original modal widget.
     */
    return function (original) {
        $.widget('mage.modal', original, {
            _create: function () {
                console.log('extended create');
                this._super();
            }
        });

        return $.mage.modal;
    };
});
//====== requirejs-config.js ========
var config = {
    config: {
        mixins: {
            'Magento_Ui/js/modal/modal': {
                'Magento_Ui/js/modal/modal-mixin': true
            }
        }
    }
};
```
So, no reason to make map on modal widget.

okorshenko:
Hi @ntoombs19 
Please, look at @omiroshnichenko comment. Should we close this Pull Request?
Thank you

ishakhsuvarov:
@ntoombs19 I am closing this PR for now, as it seems, that solution provided by @omiroshnichenko  is valid. 
Please comment and reopen if there is anything to add.

Thank you for collaboration!

Issue #7959, opened 2016-12-26
JS error on product page Cannot read property 'oldPrice' of undefined


katesuykovskaya:
### Preconditions

1. M 2.1.3

### Steps to reproduce

1. Create conf. product on 2 swatch attributes (for ex. color and size) with assigned simple products
2. upload gallery to first simple product (will be needed to test gallery reload)
3. go to product page, click on 1st swatch

### Expected result

1. gallery of first simple product should be loaded

### Actual result

1. JS error
Trace:
swatch-renderer.js:694 Uncaught TypeError: Cannot read property 'oldPrice' of undefined
    at $.(anonymous function).(anonymous function)._UpdatePrice (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/Magento_Swatches/js/swatch-renderer.js:694:23)
    at $.(anonymous function).(anonymous function)._UpdatePrice (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/jquery/jquery-ui.js:402:25)
    at $.(anonymous function).(anonymous function)._OnClick (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/Magento_Swatches/js/swatch-renderer.js:532:25)
    at $.(anonymous function).(anonymous function)._OnClick (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/jquery/jquery-ui.js:402:25)
    at HTMLDivElement.<anonymous> (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/Magento_Swatches/js/swatch-renderer.js:482:32)
    at HTMLDivElement.dispatch (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/jquery.js:4624:9)
    at HTMLDivElement.elemData.handle (http://magento213.loc/pub/static/version1482569607/frontend/Magento/luma/en_US/jquery.js:4292:28)
![image](https://cloud.githubusercontent.com/assets/10142848/21478695/7cbda606-cb56-11e6-8eb2-dac3aecfb070.png)
2. Notes:
2.1. if I click on next swatch (color), error is not occurred (in other words when all options/swatches are chosen)
2.2. there was no such error on 2.1.2

ashsmith:
I'm also getting this after upgrading a client from 2.1.2 to 2.1.3.

[...]

PR #9776, opened 2017-05-29
#7959 - Fix for JS error on Swatch Renderer for undefined oldPrice

dreamworkers:
This PR backports commit [269c215](https://github.com/magento/magento2/commit/269c2154db82e2c3a19f8ba9a36e9b249a745998) (originally for -develop) for 2.1-develop and issue #7959.

### Description
Check and verify if 'result' is defined or not before oldPrice is fetched from it.

1. magento/magento2#7959: JS error on product page Cannot read property 'oldPrice' of undefined

### Manual testing scenarios
1. Follow the list of steps at the issue that is resolved here.

### DIFF
--- a/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
+++ b/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
@@ -711,7 +711,7 @@ define([
                 }
             );
 
-            if (result.oldPrice.amount !== result.finalPrice.amount) {
+            if (typeof result != 'undefined' && result.oldPrice.amount !== result.finalPrice.amount) {
                 $(this.options.slyOldPriceSelector).show();
             } else {
                 $(this.options.slyOldPriceSelector).hide();


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=9776) <br/>All committers have signed the CLA.

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
```diff
@@ -711,7 +711,7 @@ define([
                 }
             );
 
-            if (result.oldPrice.amount !== result.finalPrice.amount) {
+            if (typeof result != 'undefined' && result.oldPrice.amount !== result.finalPrice.amount) {
```
zetlen, line 714: Two things about this:'
1. Always use strict `!==` comparison rather than loose `!=` comparison, since the coercion rules of the latter can be surprising.
2. Under what circumstances would `result` not exist? Can we eliminate that circumstance?

zetlen (changes requested):

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
korostii: I wonder why these questions weren't raised when the original fix 269c215 was implemented yet they slow down backporting.
Maybe @aholovan could provide the insights you're looking for?

P.S. To quote some guy reporting an issue here on GitHub, "It's unbelievable the effort you guys go to, to avoid fixing the issue."

OZZlE: typeof always returns a string so it's not necessary or standard practice to use strict comparison

magento-team:
@dreamworkers thank you for your contribution. Your Pull Request has been successfully merged

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
aholovan: Thanks @OZZlE , you are right about unnecessary strict comparison.

@zetlen, 'result' is undefined before all of the swatches are selected. It was the issue as with oldPrice and also with tierPrices. (how to reproduce is described in #8923 pull request)

zetlen: @aholovan Thanks for the background. Making the `result` always present is a good idea, but it seems to be out of scope for this PR.

@korostii I look at a fairly random sample of pull requests; my schedule doesn't allow me to review every single one. I didn't notice that this was a backport, so that's on me!

As for strict comparison, it really is a standard practice in all cases. Linters in standard configuration will complain. But it's all right; in a later project, we'll use an automated tool to change all of these loose comparisons to strict ones. So I will approve this!

PR #9786, opened 2017-05-29
[#7291] Change the default contact form email template to HTML

VincentMarmiesse:
Hello,

I don't know if there is a reason for that, but having the contact form email in plain text is not really nice and it makes it difficult to override.

### Description
So just a few changes here:
- move the `submitted_form.html` template from `adminhtml` to `frontend`
- include header and footer in this template
- change the template's declaration into the `email_templates.xml` file to set `type` to `html` and `area` to `frontend`

1. #7291

### Manual testing scenarios
1. Go to the contact form and send a mail
2. Check the mail has the default footer and header

I'm open to discussion on this point if it's done on purpose.

Thanks.


PieterCappelle:
👍🏻

vherasymenko:
@VincentMarmiesse Hi. After checkout your PR i try send mail using contact form.
On attachment displayed result of received message. There are no correct message

Maybe I need do something else steps than written in your PR ??

Issue #6746, opened 2016-09-25
Magento 2.1.1 Problem with change currency


Omar-Oleh:
**Preconditions**
Magento 2.1.1
Created two currencies (United Arab Emirates Dirham and Qatari Rial)
Cache is enabled

**Steps to reproduce**
Open category page on frontend
Change currency

**Expected result**
Currency is changed on product listing 

**Actual result**
Currency do not changed. Problem with FPC(I have refreshed FPC, the currency is work fine. But if I changed currency one more, it is do not working)

SerhiyShkolyarenko:
@Omar-Oleh have you created currency rates?
What is base currency for your store?

SerhiyShkolyarenko:
@Omar-Oleh reproduced. Answers to that questions are not necessary. Thank you for reporting issue!

SerhiyShkolyarenko:
Our internal issue is MAGETWO-59089.

eldonbite:
We're experiencing the same when switching between GBP and JPY currencies:

![image](https://cloud.githubusercontent.com/assets/3441674/19182657/f246e0ac-8ca6-11e6-9c14-3b8c3d52a1d3.png)

![image](https://cloud.githubusercontent.com/assets/3441674/19182743/6b8e072e-8ca7-11e6-9ce2-cbada5501184.png)

Some workarounds that I know of to prevent this issue from occurring is to:
1. Disable the block cache (BLOCK_HTML)
2. Edit Magento\Framework\Pricing\Render\PriceBox and either:
   - set the DEFAULT_LIFETIME to 0, or
   - just force getCacheLifetime to return 0

I tried writing an extension to override this behaviour, to no avail (I might be doing something wrong LOL). But it still doesn't solve the root cause of the problem.

SerhiyShkolyarenko:
Fix is delivered to mainline. Closing the issue.

Ctucker9233:
@SerhiyShkolyarenko is there a commit reference?

SerhiyShkolyarenko:
There we six commits for this issue: 8ec49df, 027491e, be4d557, 0d0e683, b1066bf, e908e6c

eldonbite:
Hi @SerhiyShkolyarenko,

Noticed another issue when selecting product variants using the swatches.
1. GBP:
   ![image](https://cloud.githubusercontent.com/assets/3441674/19993278/53c6b85a-a27f-11e6-867b-902e3df6d2ca.png)
2. Select a product variant while in GBP:
[...]

PR #9841, opened 2017-06-03
Backport of MAGETWO-59089 for Magento 2.1: Magento 2.1.1 Problem with change currency

hostep:
### Description
This problem was discovered in scope of issue https://github.com/magento/magento2/issues/9562
But the fix was already on develop and was fixing another issue: https://github.com/magento/magento2/issues/6746

See the steps below to reproduce the issue we were having which lead me to create this PR.

Since this PR fixes multiple non-related issues, I think you should test this very well. It might even fix other reported issues as well, but it's really hard to tell...

1. Fixes https://github.com/magento/magento2/issues/9562: ItemZone on product detail is not set correctly when chaning products via related/upsell products list
2. Fixes https://github.com/magento/magento2/issues/6746: Magento 2.1.1 Problem with change currency

Reference to the commit on the develop branch: a9510d3166e4016d2cc2977d7b785f0676d52afd

### Manual testing scenarios
1. Setup a new Magento 2.1.7 installation
2. Create 2 products: Product1 and Product2, make sure they have a different price and that they are visible on the frontend
3. Edit Product2 and assign Product1 as an upsell product to it
4. Reindex and flush caches
5. Visit Product2 on the frontend and save the generated html: `curl --insecure https://mydomain/product2.html > correct.html`
6. Flush the cache again
7. Visit Product1 on the frontend using a browser, or curl or wget, or whatever, so its data gets cached
8. Visit Product2 on the frontend and save the generated html: `curl --insecure https://mydomain/product2.html > wrong.html`
9. Compare the `correct.html` against the `wrong.html`:
```diff
$ diff -u correct.html wrong.html
--- correct.html	2017-06-03 13:51:02.000000000 +0200
+++ wrong.html	2017-06-03 13:51:38.000000000 +0200
@@ -622,13 +622,14 @@


 <span class="price-container price-final_price tax weee"
-        >
+         itemprop="offers" itemscope itemtype="http://schema.org/Offer">
         <span  id="product-price-1"                data-price-amount="123"
         data-price-type="finalPrice"
         class="price-wrapper "
-        >
+         itemprop="price">
         <span class="price">€123.00</span>    </span>
-        </span>
+                <meta itemprop="priceCurrency" content="EUR" />
+    </span>

 </div>
```

The html files should be identical, but they aren't, because price data of Product1 is cached and then being used in the upsell list when visiting Product2.

In our case, this problem is causing some of the Google indexing services to pick the wrong price for Product2, instead it displays the price from Product1 for Product2 in some of its services (like Google adwords), which makes our client and marketeers really nervous as you can imagine :)

### DIFF
--- a/lib/internal/Magento/Framework/Pricing/Render/PriceBox.php
+++ b/lib/internal/Magento/Framework/Pricing/Render/PriceBox.php
@@ -86,7 +86,7 @@ class PriceBox extends Template implements PriceBoxRenderInterface, IdentityInte
      */
     protected function getCacheLifetime()
     {
-        return parent::hasCacheLifetime() ? parent::getCacheLifetime() : self::DEFAULT_LIFETIME;
+        return parent::hasCacheLifetime() ? parent::getCacheLifetime() : null;
     }
     
     /**
--- a/lib/internal/Magento/Framework/Pricing/Test/Unit/Render/PriceBoxTest.php
+++ b/lib/internal/Magento/Framework/Pricing/Test/Unit/Render/PriceBoxTest.php
@@ -233,4 +233,17 @@ class PriceBoxTest extends \PHPUnit_Framework_TestCase
     {
         $this->assertEquals($this->rendererPool, $this->model->getRendererPool());
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


magento-team:

PR #9842, opened 2017-06-03
Fix broken Parallel deploy for redis users

southerncomputer:
Change default_job_amount to 1 - fixes setup:static-content:deploy redis issue

Redis and parallel deploy static content do not mix due to faulty implementation

### Description
Disable Parallel deploy for now until better fix can come along

Fixes many redis errors since 2.1.6 during setup:static-content:deploy 

### Manual testing scenarios
1. php bin/magento setup:static-content:deploy en_US
2. find random redis errors such as closed connection 
3. website is now broken!

This is due to parallel deploy jobs not properly initializing a separate redis session when pcntl fork happens.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Deploy/Console/DeployStaticOptions.php
+++ b/app/code/Magento/Deploy/Console/DeployStaticOptions.php
@@ -119,7 +119,7 @@ class DeployStaticOptions
     /**
      * Default jobs amount
      */
-    const DEFAULT_JOBS_AMOUNT = 0;
+    const DEFAULT_JOBS_AMOUNT = 1;
 
     /**
      * Key for languages parameter


southerncomputer:
This value appears to have been fixed in develop. the value should change from 4 to 1 in 2.1.x branch users only!

hostep:
@southerncomputer: not sure if this is helpful, but we remove the Redis & Varnish configuration from the `env.php` file during deploys, and put those configurations back just before we put the new deployed version live. If you remove the configurations, it will fall back to normal file based caching, which will probably also fix your issue.

Not sure if this applies to the way you are deploying, we use something similar as how Capistrano manages deploys, with preparing a new release in a seperate directory and when the whole deployment process nears its end, we then switch the docroot from the old release to the new release with symlinks. The Classy Llama guys wrote [a nice blogpost](https://www.classyllama.com/blog/deploying-magento-2-using-capistrano) about this deployment flow.
We use something very similar to that workflow, but before starting to execute all magento's deployment steps, we rewrite the `env.php` file so it no longer contains Redis & Varnish configuration. We do this so that during a deploy (which can take a couple of minutes) we don't contaminate the caches which are currently being used by the live environment.

Some steps of our deployment script should you be curious (Python code):
```python
# before this piece of code we setup the file structure for a new release

# move the symlinked env.php file aside and make a copy, then manipulate it so Redis and Varnish connections are removed
run('mv app/etc/env.php app/etc/env.orig.php')
run('cp app/etc/env.orig.php app/etc/env.php')
run('php -f tools/rewrite_env_php_file.php -- -f app/etc/env.php')

run('composer install --prefer-dist --no-dev')

run('php bin/magento deploy:mode:set --skip-compilation production')
run('php bin/magento setup:di:compile')
run('touch pub/static/deployed_version.txt')  # workaround for new bug/feature introduced in M2.1.1, keep in here until https://github.com/magento/magento2/issues/6426 is fixed

# The two commands below no longer work properly since Magento 2.1.3, so this is replaced with compiling the static assets for each locale separately
# Can be restored whenever https://github.com/magento/magento2/issues/7862 gets fixed
# run('php bin/magento setup:static-content:deploy %s %s' % (' '.join(['--theme %s' % t for t in conf.adminhtml_themes]), ' '.join(conf.adminhtml_languages)))
# run('php bin/magento setup:static-content:deploy %s %s' % (' '.join(['--theme %s' % t for t in conf.frontend_themes]),  ' '.join(conf.frontend_languages)))
for adminhtml_lang in conf.adminhtml_languages:
    run('php bin/magento setup:static-content:deploy %s %s' % (' '.join(['--theme %s' % t for t in conf.adminhtml_themes]), adminhtml_lang))

for frontend_lang in conf.frontend_languages:
    run('php bin/magento setup:static-content:deploy %s %s' % (' '.join(['--theme %s' % t for t in conf.frontend_themes]),  frontend_lang))

# remove the manipulated env.php file and put the original back
run('rm app/etc/env.php')
run('mv app/etc/env.orig.php app/etc/env.php')

# after this piece of code we perform some other steps and then switch the docroot to the new release
```

The `rewrite_env_php_file.php` file is the script which removes Redis & Varnish configuration, and that script looks something like this:
```php
<?php

$argOptions = getopt("f:");

if ($argOptions === false || !array_key_exists('f', $argOptions))
{
    echo "There is at least one missing option!\n";
    exit(1);
}

$envPhpFile = $argOptions['f'];
$realEnvPhpFile = realpath($envPhpFile);

if ($realEnvPhpFile === false || !file_exists($realEnvPhpFile))
{
    echo "File $envPhpFile doesn't exist!\n";
    exit(1);
}

$envData = include $realEnvPhpFile;

if (!is_array($envData))
{
    echo "File $realEnvPhpFile doesn't return an array!\n";
    exit(1);
}

// remove cache entry entirely, so we fall back to file based caching
unset($envData['cache']);

// remove session entry entirely, so we fall back to file based sessions
unset($envData['session']);

// remove http_cache_hosts entry entirely, so we don't send Varnish purges
unset($envData['http_cache_hosts']);

//

// from Magento\Framework\App\DeploymentConfig\Writer\PhpFormatter
$newEnvData = "<?php\nreturn " . var_export($envData, true) . ";\n";

//

$writeResult = file_put_contents($realEnvPhpFile, $newEnvData);

if ($writeResult === false)
{
    echo "Couldn't write the new env data back to $realEnvPhpFile!\n";
    exit(1);
}

echo "Success!\n";
exit(0);
```

Not sure if this helps in your particular workflow, but maybe it can help somebody else who runs into a similar problem.

southerncomputer:
Nice Thanks! I'll bookmark your method!

PR #9858, opened 2017-06-06
Fix memory leak for images resize command.

Koc:
### Description
We have large catalog with many products and 2Gb memory limit is not enough to resize all of the images. I've removed storing loaded products inside identity map and it helps a lot.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Api/ProductRepositoryInterface.php
+++ b/app/code/Magento/Catalog/Api/ProductRepositoryInterface.php
@@ -72,4 +72,11 @@ interface ProductRepositoryInterface
      * @return \Magento\Catalog\Api\Data\ProductSearchResultsInterface
      */
     public function getList(\Magento\Framework\Api\SearchCriteriaInterface $searchCriteria);
+
+    /**
+     * Clean internal product cache
+     *
+     * @return void
+     */
+    public function cleanCache();
 }
--- a/app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
+++ b/app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
@@ -77,8 +77,7 @@ class ImagesResizeCommand extends Command
         $productCollection = $this->productCollectionFactory->create();
         $productIds = $productCollection->getAllIds();
         if (!count($productIds)) {
-            $output->writeln("<info>No product images to resize</info>");
-            // we must have an exit code higher than zero to indicate something was wrong
+            $output->writeln('<info>No product images to resize</info>');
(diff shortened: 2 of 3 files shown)


Koc:
I can also add progressbar to indicate resizing progress if this acceptable

mslabko:
Hi @Koc ,
For mainline this issue already fixed - https://github.com/magento/magento2/blob/develop/app/code/Magento/Catalog/Model/ProductRepository.php#L311-L312 
I think it make sense to port existing solution to 2.1, does it work for you?

Koc:
@mslabko I haven't tested this yet but looks like it should works. Agree, it is better backport it to 2.1. But who  should do it: am I in this PR or somewho from Magento's team can?

ishakhsuvarov:
@Koc I am pretty sure you can work on this, if you wish

app/code/Magento/Catalog/Api/ProductRepositoryInterface.php
```diff
@@ -72,4 +72,11 @@ public function deleteById($sku);
      * @return \Magento\Catalog\Api\Data\ProductSearchResultsInterface
      */
     public function getList(\Magento\Framework\Api\SearchCriteriaInterface $searchCriteria);
+
+    /**
+     * Clean internal product cache
+     *
+     * @return void
+     */
+    public function cleanCache();
```
okorshenko, line 81: it is not allowed change @api interfaces in patch releases. Unfortunately, we can not accept this change.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -95,15 +94,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 $imageCache = $this->imageCacheFactory->create();
                 $imageCache->generate($product);
 
-                $output->write(".");
+                $this->productRepository->cleanCache();
```
okorshenko, line 97: this method is out of RepositoryInterface, so we can not call it directly. As a workaround, we can check if method_exists and call it. In this case the fix will be BC

okorshenko (changes requested):

app/code/Magento/Catalog/Api/ProductRepositoryInterface.php
Koc: I will try backport another implementation from develop brahcn.

okorshenko:
Closing this PR due to inactivity. provided changes do not follow Backward compatibility policy. 
Thank you @Koc for your effort

PR #9859, opened 2017-06-07
Removed unused $_customerSession property

edenreich:
replaced it with a property that being assigned but not declared, $httpContext


magento-team:
@edenreich thank you for your contribution to Magento 2 project

PR #9872, opened 2017-06-07
Fixed issue causing static test failure to report success on Travis

davidalger:
### Fixed Issues

There was a regression introduced in e68d69c00de96afc3 as a result of moving the test script from .travis.yml to the script.sh file. The script was not configured to exit on command failure, which caused the static tests to succeed when `phpunit` fails but the subsequent `grunt static` command met success.

### Work Done

I've moved test script back into .travis.yml both to resolve the failure, but also to preserve proper verboseness of tests when they fail. Experimented heavily here when I reworked the entire travis configuration a year or two ago, and it's best for these to be in .travis.yml if possible, so I've moved them back there and made sure all the tests are still run correctly.

The js static tests have also been moved into their own job vs being mashed into the same job running phpunit static tests for the sake of travis test suite simplicity.

### Contribution checklist


davidalger:
Here is example static build where there were failures, but the test shows green: https://travis-ci.org/magento/magento2/jobs/239757912. This resolves that issue.

PR #9873, opened 2017-06-07
Fixes layered navigation options being cached using the wrong store id.

hostep:
### Description

This is a "forward" port of MAGETWO-69543.
See https://github.com/magento/magento2/pull/9704 for more details.

The reason why this was first fixed on the `2.1-develop` branch, is that the feature of caching the layered navigation options only landed on the `develop` branch a few days after https://github.com/magento/magento2/pull/9704 was created.

1. https://github.com/magento/magento2/issues/9679: Translation for layered navigation attribute option not working

### Manual testing scenarios
See https://github.com/magento/magento2/pull/9704

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php
+++ b/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php
@@ -13,7 +13,7 @@ namespace Magento\Eav\Model\Entity\Attribute\Frontend;
 
 use Magento\Framework\App\CacheInterface;
 use Magento\Framework\Serialize\Serializer\Json as Serializer;
-use Magento\Store\Api\StoreResolverInterface;
+use Magento\Store\Model\StoreManagerInterface;
 use Magento\Framework\App\ObjectManager;
 use Magento\Eav\Model\Cache\Type as CacheType;
 use Magento\Eav\Model\Entity\Attribute;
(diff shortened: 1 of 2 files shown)


hostep:
Just pushed a new commit which should fix the unit tests.
It would be nice if someone with more knowledge about writing tests then myself could add a new test which tests the caching of options with multiple store id's, to make sure this bug doesn't resurface later on.

hostep:
@magento-team or @okorshenko: why did you switch the order of the `StoreManagerInterface` and `Serializer` parameters around in the constructor? This is introducing a backwards incompatible change, if you upgrade from 2.1.x to 2.2.
Since the `Serializer` parameter doesn't exist in 2.1.x
[2.1-develop branch](https://github.com/magento/magento2/blob/7294054/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php#L70) vs [develop branch](https://github.com/magento/magento2/blob/58b4c7c/app/code/Magento/Eav/Model/Entity/Attribute/Frontend/AbstractFrontend.php#L81)

I did this on purpose so there would be no backwards incompatibility being introduced, but you switched it around ... The `Serializer` should be the last one I believe.

Thanks!

hostep:
@okorshenko or @ishakhsuvarov, can you please check my previous comment? Thanks!

okorshenko:
Hi @hostep thank you for pointing on this. we will review the code one more time

okorshenko:
Your comment make sense. We will fix it or you can create PR and we will accept ASAP. The reason why we did that is here: http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

hostep:
@okorshenko: here you go: https://github.com/magento/magento2/pull/9922 :)

PR #9904, opened 2017-06-11
Fixed pointless exception in logs every time a category with image is saved

woutersamaey:
Fixed a pointless exception in the logs every time a category is saved that has an image, even when nothing was changed.
When no image is uploaded, there is no need to try to process it.

Currently there is a try catch which writes an error to the log, but this is also triggered when no image is changed or uploaded. The log thus contains errors that aren't problematic.

### Description
Fixed superflous exception from logs every time a category (with image) is saved, even when no changes were made to the category,

1. Fixed a superflous error message from getting in the logs. This alarms the developer/shopkeeper when there is actually no problem at all.

### Manual testing scenarios
1. Save a category that has an image. No need to change anything really.
2. The exception log should not contain a "critical" exception about a problem writing the file.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Category/Attribute/Backend/Image.php
+++ b/app/code/Magento/Catalog/Model/Category/Attribute/Backend/Image.php
@@ -134,11 +134,14 @@ class Image extends \Magento\Eav\Model\Entity\Attribute\Backend\AbstractBackend
     {
         $value = $object->getData($this->additionalData . $this->getAttribute()->getName());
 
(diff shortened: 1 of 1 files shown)


okorshenko:
HI @woutersamaey 
Thank you for your fix. Could you please fix broken tests so that we can proceed with Pull Requests acceptance?

Thank you

Ctucker9233:
@magento-team What version of 2.2 will have this fix? It's not in 2.2.7.

PR #9911, opened 2017-06-12
Bring multithreading to the image resizer command

AntonEvers:
### Description

Multi-threaded image resizing and limit and offset control

We had the following scenario: A customer could only provide his full catalog a day before the launch of his online store. All image resizes had to be generated over the course of one night.
With a catalog of 100.000+ products and 3+ images per product this was not possible with the current `catalog:images:resize` command.

The customer did however have 4 24 core web servers at his disposal which all shared a NFS `pub/media` directory. With the addition of multi-threading and limit and offset control to the `catalog:images:resize` command it is now possible to run a 24 thread process which resizes 1/4 of the catalog images on each web server. This way catalog images are resized up to 96 times faster than before in this scenario.

This also prevents the web shop from being way too slow and way to heavy in it's requests in the first few days because images have to be generated on the fly.

The code responsible for the php forks is heavily based on the code used for static content deployment.

1. none yet

### Manual testing scenarios

1. resize the sample data images without job configuration: ~30 mins. on a 2,6 GHz Intel Core i7
2. resize the sample data images with 8 jobs: ~15 minutes 2,6 GHz Intel Core i7

### Contribution checklist


app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,83 +22,129 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
-        $this->imageCacheFactory = $imageCacheFactory;
+        $this->queueFactory = $queueFactory;
+        $this->options = $options;
         parent::__construct();
     }
 
     /**
      * {@inheritdoc}
      */
-    protected function configure()
```
ishakhsuvarov: Contract of the parent class implies that this method is protected.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,83 +22,129 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
-            ->setDescription('Creates resized product images');
+            ->setDescription('Creates resized product images')
+            ->setDefinition($this->options->getOptionsList());
     }
 
     /**
      * {@inheritdoc}
      */
-    protected function execute(
+    public function execute(
```
ishakhsuvarov: Contract of the parent class implies that this method is protected.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,83 +22,129 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
         $output->write("\n");
         $output->writeln("<info>Product images resized successfully</info>");
+        return Cli::RETURN_SUCCESS;
+    }
+
+    /**
+     * @param array $options
+     * @return int
+     */
+    public function getProcessesAmount(array $options)
```
ishakhsuvarov: Looks like this method can be `private`.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,83 +22,129 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
+        return isset($options[Options::JOBS_AMOUNT])
+            ? (int)$options[Options::JOBS_AMOUNT]
+            : Options::DEFAULT_JOBS_AMOUNT;
+    }
+
+    /**
+     * @param array $options
+     * @return int
+     */
+    public function getLimit(array $options)
```
ishakhsuvarov: Looks like this method can be `private`.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,83 +22,129 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
+        return isset($options[Options::PRODUCT_LIMIT])
+            ? (int)$options[Options::PRODUCT_LIMIT]
+            : Options::DEFAULT_PRODUCT_LIMIT;
+    }
+
+    /**
+     * @param array $options
+     * @return int
+     */
+    public function getOffset(array $options)
```
ishakhsuvarov: Looks like this method can be `private`.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,83 +22,129 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
+        $limit = $this->getLimit($options);
+        $offset = $this->getOffset($options);
+
+        if ($limit) {
+            $productIds = array_slice($productIds, $offset, $limit);
+        }
+
+        $maxOffset = $productCount - 1;
+        if ($offset > $maxOffset) {
+            $output->writeln("<error>Offset may not be higher than $maxOffset</error>");
```
ishakhsuvarov: Since it is an error, maybe `return Cli::RETURN_FAILURE` is needed here?

app/code/Magento/Catalog/Console/ImageResizeOptions.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        return $this->getBasicOptions();
+    }
+
+    /**
+     * Basic options
+     *
+     * @return array
+     */
+    private function getBasicOptions()
```
ishakhsuvarov: Do you really need a separate method which is just proxied to another public method?

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -55,7 +55,7 @@ public function __construct(
      *
      * @return array
      */
-    protected function getData()
+    public function getData()
```
ishakhsuvarov: Generally changing `protected` to `public` isn't very good, it was designed that way on purpose. You may try looking for a better public method or in the worst case extracting logic of this method to a separate class.

app/code/Magento/Catalog/Model/Product/Image/Process/Queue.php
```diff
@@ -0,0 +1,323 @@
+<?php
```
ishakhsuvarov: Magento does not encourage inheritance-based APIs, therefore `protected` class properties should be made `private` when it's possible.

ishakhsuvarov: Please review method visibility in this class, it looks like a lot of them need to be `private`.

app/code/Magento/Catalog/Model/Product/Image/Process/Queue.php
```diff
@@ -0,0 +1,323 @@
...
+            // process child process
+            $this->inProgress = [];
+            $this->imageCache->generate($product);
+            $this->output->write('.');
+            exit();
+        }
+
+        try {
+            /** @var \Magento\Catalog\Model\Product $product */
+            $product = $this->productRepository->getById($productId);
```
ishakhsuvarov, line 231: Looks like these actions were already performed 10 lines above. Is there any reason for this duplication?

app/code/Magento/Catalog/Model/Product/Image/Process/QueueFactory.php
```diff
@@ -0,0 +1,60 @@
+<?php
```
ishakhsuvarov: Why not use auto-generated factory?

ishakhsuvarov (changes requested):
Please review the CR comments.
More in-depth review may be required once fixed.

ishakhsuvarov:
@ajpevers Thanks! This is really a great addition

app/code/Magento/Catalog/Model/Product/Image/Process/Queue.php
AntonEvers: The lines above are for forked php processes. If there is no forking, these lines will be executed. Forked processes will not get here because of the `exit();`. But from the duplicate code perspective I will make a change here nonetheless. Although it adds a few more `if`'s.

AntonEvers:
@ishakhsuvarov I've processed the requested changes. Also I have fixed unit tests that were still failing.

AntonEvers:
@ishakhsuvarov side note, there are no unit tests for the forked version of the command `--jobs=n` . Is that even unit testable? I started writing unit tests this week so I'm still very much learning at that point.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,31 +22,31 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
      * @param \Magento\Catalog\Model\ResourceModel\Product\CollectionFactory $productCollectionFactory
-     * @param \Magento\Catalog\Api\ProductRepositoryInterface $productRepository
-     * @param \Magento\Catalog\Model\Product\Image\CacheFactory $imageCacheFactory
+     * @param QueueFactory $queueFactory
+     * @param Options $options
      */
     public function __construct(
         \Magento\Framework\App\State $appState,
         \Magento\Catalog\Model\ResourceModel\Product\CollectionFactory $productCollectionFactory,
-        \Magento\Catalog\Api\ProductRepositoryInterface $productRepository,
```
ishakhsuvarov: To maximize backwards compatibility removing non-optional constructor parameter is not allowed.
If this is not needed anymore please remove the type hint and mark it as a deprecated.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -18,31 +22,31 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
-        \Magento\Catalog\Model\Product\Image\CacheFactory $imageCacheFactory
+        QueueFactory $queueFactory,
+        Options $options
     ) {
         $this->appState = $appState;
         $this->productCollectionFactory = $productCollectionFactory;
-        $this->productRepository = $productRepository;
-        $this->imageCacheFactory = $imageCacheFactory;
+        $this->queueFactory = $queueFactory;
```
ishakhsuvarov: Due to backwards compatibility police it prohibited to add new non-optional arguments to the constructor. Please make it optional and nullable and get the required dependency with the ObjectManager static call if value is null.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -62,39 +67,86 @@ protected function execute(
...
-
-                /** @var \Magento\Catalog\Model\Product\Image\Cache $imageCache */
-                $imageCache = $this->imageCacheFactory->create();
-                $imageCache->generate($product);
-
-                $output->write(".");
-            }
+            $queue->setProducts($productIds);
+            $output->writeln("<info>Resizing product images</info>");
+            $queue->process();
```
ishakhsuvarov, line 127: I would suggest to keep the non-threaded execution logic in place and use the multi-threaded implementation only in case if it is requested.
This may improve future debugging experience and makes sure, that none of the existing features are affected if there are any issues.

app/code/Magento/Catalog/Console/Command/ImagesResizeCommand.php
```diff
@@ -62,39 +67,86 @@ protected function execute(
...
+            return Cli::RETURN_FAILURE;
         }
 
         $output->write("\n");
         $output->writeln("<info>Product images resized successfully</info>");
+        return Cli::RETURN_SUCCESS;
+    }
+
+    /**
+     * @param array $options
```
ishakhsuvarov, line 140: It would be nice to specify information about the variables in the docblock for each method.

app/code/Magento/Catalog/Console/ImageResizeOptions.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    const PRODUCT_LIMIT = 'product-limit';
+
+    /**
+     * Default jobs amount
+     */
+    const DEFAULT_JOBS_AMOUNT = 0;
+
+    /**
+     * Default jobs amount
```
ishakhsuvarov: This docBlock looks like an artifact of copy-pasting. Please review and update descriptions accordingly.

app/code/Magento/Catalog/Model/Product/Image/Process/Cache.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Catalog\Model\Product\Image\Process;
+
+class Cache extends \Magento\Catalog\Model\Product\Image\Cache
```
ishakhsuvarov, line 8: Magento does not recognize inheritance as a good practice at the moment. 
Please look into the possibility to extract required business logic into the different class and refactor the old code to use it as well.

app/code/Magento/Catalog/Model/Product/Image/Process/Queue.php
```diff
@@ -0,0 +1,322 @@
...
+        CacheFactory $imageCacheFactory,
+        OutputInterface $output,
+        $maxProcesses = self::DEFAULT_MAX_PROCESSES_AMOUNT
+    ) {
+        $this->productRepository = $productRepository;
+        $this->resourceConnection = $resourceConnection;
+        $this->output = $output;
+        $this->imageCacheFactory = $imageCacheFactory;
+        $this->maxProcesses = $maxProcesses;
+        $this->imageCache = $this->imageCacheFactory->create();
```
ishakhsuvarov: I would suggest moving this out of the class constructor to the method where it is used. It is not considered a good practice to execute anything else then argument validation and assignment in the constructor.

app/code/Magento/Catalog/Model/Product/Image/Process/Queue.php
```diff
@@ -0,0 +1,322 @@
...
+    {
+        return function_exists('pcntl_fork') && $this->maxProcesses > 1;
+    }
+
+    /**
+     * @param int $productId
+     * @return bool true on success for main process and exit for child process
+     * @SuppressWarnings(PHPMD.ExitExpression)
+     */
+    private function execute(int $productId)
```
ishakhsuvarov, line 211: It would be much better from the developer experience perspective to completely split logic for multi-threaded and single-threaded processing. Ideally, old single-threaded code should be intact as much as possible.

app/code/Magento/Catalog/Model/Product/Image/Process/Queue.php
```diff
@@ -0,0 +1,322 @@
...
+            $this->resourceConnection->closeConnection();
+        }
+    }
+
+    /**
+     * @return bool
+     */
+    private function isCanBeParalleled()
+    {
+        return function_exists('pcntl_fork') && $this->maxProcesses > 1;
```
ishakhsuvarov, line 203: Please look into usage of `popen`, if possible, as we have encountered issues with parallelization with `pcntl_fork` in the past (for instance Redis session management may not really work as expected in this case).

ishakhsuvarov (changes requested):

ishakhsuvarov:
Hi @ajpevers 
I've updated the code review.

Please feel free to reach me in comments or directly if you have any questions or would like 
 to get assistance.

Thank you!

ishakhsuvarov:
Hi @ajpevers 
Would you like to continue working on this issue? Any comments regarding the code review?
Thanks

AntonEvers:
Yes I do. I need to find some extra time to attend to the comments.

ishakhsuvarov:
@ajpevers I am closing this PR for now due to inactivity. Please reopen once you are ready to continue.
Thank you!

PR #9917, opened 2017-06-12
Add cron expression check when setting cron expression into cron schedule model

adrian-martinez-interactiv4:
Add cron expression check when setting cron expression into cron schedule model. 

### Description
\Magento\Cron\Model\Schedule::setCronExpr() method only checks if cron expression has between 5 and 6 components. \Magento\Cron\Model\Schedule::trySchedule() allows to check a cron expression against a timestamp value, but there is no way to check if a cron expression is valid by itself, without a timestamp.

\Magento\Cron\Model\Schedule::checkCronExpr() has been added to add this missing functionality, and now it's used by default at \Magento\Cron\Model\Schedule::setCronExpr() method.

Adapted unit tests to new check when setting cron expression into schedule model.

### Fixed Issues
- Expressions like '2-1 * * * *' were accepted as valid expressions before. New method check does not allow to set a from-to value with the $from value higher than $to value.

### Manual testing scenarios
`php vendor/phpunit/phpunit/phpunit -c dev/tests/unit/phpunit.xml.dist app/code/Magento/Cron/Test/Unit/`

### Contribution checklist
 - [X]  All new or changed code is covered with unit/integration tests (if applicable)


okorshenko:
Hi @adrian-martinez-interactiv4 
Thank you for your contribution. The idea of this Pull Request is good, but unfortunately we can not accept this PR in current state. I have few comments about implementation approach:
1. Do not add new public methods to interface. This is @api class and we should not change it interface.
2. Create new class:` \Magento\Cron\Model\ResourceModel\Schedule\ExpressionValidator` with the method `validate` and move all validation logic to this class. Inject validator to Schedule model to perform validation.
3. In current implementation you have duplicated logic between methods `matchCronExpression` and `checkCronExpression `. We would like to avoid duplication in algorithms. 
4. The interface of the class is not intuitive: what is the difference between `checkCronExpr ` and `checkCronExpression`? Without digging into implementation details it is hard to understand what you should use.

Also, please, check this documentation: http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/ This document will help you with refactoring.

Please, let me know if you have any questions.
Thank you.

okorshenko:
Hi @adrian-martinez-interactiv4 
Could you please review my previous comment?

okorshenko:
Closing this PR due to inactivity. @adrian-martinez-interactiv4 feel free to reopen this PR once ready. thank you

adrian-martinez-interactiv4:
Moved to https://github.com/magento/magento2/pull/11146

PR #9925, opened 2017-06-13
Fix #9924, prefill prefix and suffix in checkout shipping address

### LINKED ISSUE #9924: Prefix and suffix are not prefilled in the quote shipping address
### Preconditions

1. enable customer prefix and suffix in the customer config
2. create an account with a prefix and a suffix

### Steps to reproduce

1. start an order, using an account with a prefix and a suffix
2. visit the shipping address step in the checkout

### Expected result

1. shipping address should be prefilled with prefix, firstname, middlename, lastname, suffix

### Actual result

1. shipping address is prefilled with firstname, lastname

the middle name is already added in #8878 (issue #8879)

AntonEvers:
### Description

Prefill prefix and suffix in the checkout if they are present on a logged in customer.

1. magento/magento2#9924 : Prefix and suffix are not prefilled in the quote shipping address

### Manual testing scenarios

1. start an order, using an existing account with a prefix and a suffix
2. visit the shipping address step in the checkout
3. prefix and suffix should be prefilled in the shipping address form for customers without an address
4. prefix and suffix should be prefilled in the NEW shipping address form for customers with addresses

### Contribution checklist


app/code/Magento/Checkout/Block/Checkout/AttributeMerger.php
```diff
@@ -388,4 +387,36 @@ protected function orderCountryOptions(array $countryOptions)
         }
         return array_merge($headOptions, $tailOptions);
     }
+
+    /**
+     * @return null|string
+     */
+    protected function getDefaultPrefix()
```
ishakhsuvarov: Magento does not encourage usage of `protected` methods.

app/code/Magento/Checkout/Block/Checkout/AttributeMerger.php
```diff
@@ -316,20 +316,19 @@ protected function getMultilineFieldConfig($attributeCode, array $attributeConfi
     protected function getDefaultValue($attributeCode)
     {
         switch ($attributeCode) {
```
ishakhsuvarov: Wouldn't it be better if you return `null` straight away, at the beginning of the method in case where there is no customer? I guess it should reduce complexity that way, since you would not have to check for customer every time.

ishakhsuvarov (changes requested):

app/code/Magento/Checkout/Block/Checkout/AttributeMerger.php
AntonEvers: Hahaha, brilliant... I guess programming for 12 hours straight is not paying off ;)

app/code/Magento/Checkout/Block/Checkout/AttributeMerger.php
AntonEvers: Good to know. Because of plugins right?

ishakhsuvarov: Because we do not encourage inheritance-based API. So composition in favor of inheritance. It's easier to maintain and reuse.

app/code/Magento/Checkout/Block/Checkout/AttributeMerger.php
AntonEvers: Correction, `$this->directoryHelper->getDefaultCountry()` doesn't need a customer. Let me draft up a commit.

AntonEvers:
@ishakhsuvarov how about this?

app/code/Magento/Checkout/Block/Checkout/AttributeMerger.php
AntonEvers: Ah yes of course. That makes sense.

PR #9937, opened 2017-06-14
$quote->save() is deprecated, use CartRepositoryInterface instead

ghost:
### Description

Create a plugin for Magento\Quote\Api\CartRepositoryInterface::save(), then wonder why it doesn't get triggered after totals have been collected. This PR fixes two occurrences where a direct `$quote->save()` was used.

### Manual testing scenarios

1. ... Create a plugin for Magento\Quote\Api\CartRepositoryInterface::save()
2. ... Make a REST call to V1/guest-carts/:cartId:/set-payment-information
3. ... The plugin should get executed twice as the quote gets saved twice

### Contribution checklist

### DIFF
--- a/app/code/Magento/Quote/Model/PaymentMethodManagement.php
+++ b/app/code/Magento/Quote/Model/PaymentMethodManagement.php
@@ -80,7 +80,9 @@ class PaymentMethodManagement implements \Magento\Quote\Api\PaymentMethodManagem
             throw new InvalidTransitionException(__('The requested Payment Method is not available.'));
         }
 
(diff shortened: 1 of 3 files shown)


fooman:
@ishakhsuvarov any idea on what is causing the issue in the B2B module? It currently sounds to me that if their behaviour depends on `$quote->save()` instead of `$repository->save($quote)` it's something that would need changing there.

okorshenko:
Hi @fooman 
Thank you for your contribution, but unfortunately, this fix leads to other issues out of Open Source edition. I will send this ticket to the core team backlog and we will process it on our side. Thank you

PR #9955, opened 2017-06-15
Fix Issue: “Asymmetric transaction rollback error” if commit callbacks throw exceptions

waynetheisinger:
If a callback in commit throws an excecption, the calling plugin needs to be able to distinguish this from a unsuccessful commit, so that it doesn't attempt a rollback on an already commited transaction. (Which would cause an  “Asymmetric transaction rollback" error)

1. magento/magento2#6497: “Asymmetric transaction rollback error” if commit callbacks throw exceptions

### Manual testing scenarios

Let a commit callback throw an exception

### Example:

1. create a new search engine

        <type name="Magento\Search\Model\AdapterFactory">
            <arguments>
                <argument name="adapters" xsi:type="array">
                    <item name="dummy" xsi:type="string">Magento\Framework\Search\Adapter\Mysql\Adapter</item>
                </argument>
            </arguments>
        </type>

2. Do not register a corresponding indexer handler for `Magento\CatalogSearch\Model\Indexer\IndexerHandlerFactory`
3. The commit callback from the fulltext indexer plugin will cause an exception "There is no such indexer handler: dummy"
4. Trigger this commit (for example, save a product)

### Expected result

Exception from the commit callback is thrown and can be catched later (for ex. by PHPUnit)

### Actual result

`rollback()` is called after `commit()` already has been called for the same transaction, resulting in an “Asymmetric transaction rollback error” exception when Magento tries to commit or rollback the last transaction on the stack.

### Contribution checklist


maghamed (approved):
Make sense for me.

Thanks for contribution

maghamed:
I left a comment regarding this PR in the GitHub issue #6497 

https://github.com/magento/magento2/issues/6497#issuecomment-312687130

waynetheisinger:
@maghamed I have replied to your comment regarding this PR in the GitHub issue #6497 

[#6497 (my reply)](https://github.com/magento/magento2/issues/6497#issuecomment-312689832)

waynetheisinger:
@maghamed - I've reverted the original changes and added the logging - so this is ready for another look.

PR #10031, opened 2017-06-23
Allow option disabling for optgroup binding

Bart-art:
The optgroup Knockout binding formats its options before usage in the optionsAfterRender of the form/element/select.html knockout template. In this process all extra data of the options will be stripped and the optionsAfterRenderer cannot access the "disabled" value of any option. Because of this the delimiter option in the country_id select in checkout is choosable by the customer which is not the intended behaviour.

### DIFF
--- a/app/code/Magento/Ui/view/base/web/js/lib/knockout/bindings/optgroup.js
+++ b/app/code/Magento/Ui/view/base/web/js/lib/knockout/bindings/optgroup.js
@@ -302,9 +302,8 @@ define([
                         space = '\u2007\u2007\u2007';
 
                     obj[optionTitle] = applyToObject(option, optionsText + 'title', value);
-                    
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10031) <br/>All committers have signed the CLA.

ishakhsuvarov:
Hi @Bart-art 
Thank you for your contribution. Could you please check JS static test failures on Travis CI?
Thanks!

Bart-art:
Hi @ishakhsuvarov,
the JS static test failures are fixed.

ishakhsuvarov:
@Bart-art Thank you.
We will proceed with review and merge. Could you describe any actual case for manual testing?

PR #10089, opened 2017-06-28
Add fulltext/keyword search field for product admin grid

josefbehr:
### Description
This adds a keyword search field (like on sales_order_grid) to catalog_product_grid (product_listing)

### Manual testing scenarios
1. Log into the backend
2. Go to Catalog -> Products
3. Notice the key word search field
4. Enter arbitrary search terms occurring in product attributes
5. Apply and filter grid to only show matching products

### Fixes issues
1. magento/magento2#5785

### Contribution checklist


maghamed (dismissed):
Grid Search is not a responsibility of CatalogSearch module, as CatalogSearch is responsible just for Front-End search. 
Grid Search is the responsibility of Catalog module, so fix should be placed there

Thus, you can change directly product_listing.xml

josefbehr:
@maghamed 
Can do, but what about the dependency on the Collection from CatalogSearch

maghamed:
@josefbehr  I looked through the CatalogSearch code and found that there is already code responsible for Backend search

Magento\CatalogSearch\Model\Search\Catalog

so, let's leave your fix in CatalogSearch, because you right, for now, we have the dependency on this Collection

app/code/Magento/CatalogSearch/Ui/DataProvider/Product/AddFulltextFilterToCollection.php
```diff
@@ -0,0 +1,45 @@
...
+ * Class AddFulltextFilterToCollection
+ */
+class AddFulltextFilterToCollection implements AddFilterToCollectionInterface
+{
+    /*
+     * Search Collection
+     *
+     * @var SearchCollection
+     */
+    protected $searchCollection;
```
maghamed: please use private instead of protected

app/code/Magento/CatalogSearch/Ui/DataProvider/Product/AddFulltextFilterToCollection.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function __construct(SearchCollection $searchCollection)
+    {
+        $this->searchCollection = $searchCollection;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @SuppressWarnings("unused")
```
maghamed: please change to UnusedFormalParam

maghamed:
BTW, could you please test that general Admin Search works correctly.
For now, in the upper right corner, we have a search box. When Admin starts typing there a search phrase - dropdown box appears which provides an ability to search in "Products", "Customers", "Orders", "Pages".

screen shot 2017-07-07 at 07 46 31


Search in Products doesn't work because for now there is no Search on Product grid.
Let's check that searching from the Admin Search will make a correct redirect to the Product grid with applying Search phrase as a filter.

screen shot 2017-07-07 at 07 47 15

josefbehr:
@maghamed 
Admin Search does (still) not appear to be working for Products. But neither does it so for Orders. What would I have to change to make it workable?

maghamed:
hey @josefbehr. For me search for Orders works as expected 
http://b2b.dev.magento.com/builds/20170616035016/backend/sales/order/index/?search=Katia

I thought that Search for Products will work after your changes. If not, need to double check that.

All that global search does is redirects on Order/Customer/Page/Product grid with **?search** GET parameter and whether Grid supports search - it applies the filter.

Ctucker9233:
@magento-team Can this be backported to 2.2?

PR #10130, opened 2017-07-03
Fix shipping address can use for billing

tufahu:
### Description
Fix bug on checkout process, when billing address is empty at quote and magento tries to use shipping address as billing address. Currently it ignores the property of address canUseForBilling.


ishakhsuvarov:
Hi @tufahu 
Could you please specify more precise steps to reproduce the issue you are fixing with this PR?
Thanks

tufahu:
Hi @ishakhsuvarov! 

The base functionality has no problem with this implementation, because base address classes have canUseForBilling is hardcoded with value true.

refs:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Customer/view/frontend/web/js/model/customer/address.js#L101
https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/model/new-customer-address.js#L94

So every time magento, handle all address can be used for billing too.

I've written a magento shipping module (shipping to pickup points) where I add a new address at checkout process to quote. These are only shipping adresses (gasoil stations etc), what i want is to customer cannot select address as billing too, and customer have to set a different billing address.

The current implementation ignores this. If i set canUseForBilling false on address, magento add the choice for customer to tick [x] billing and shipping is the same which is wrong.

tufahu:
And if you read the default code part

>             if (!quote.billingAddress()) {
>                 selectBillingAddressAction(quote.shippingAddress());
>             }
> 

If there is no billing address -> set shipping Address as Billing Address
I guess it is not 100%. Seems better with my patch.

>             if (!quote.billingAddress() && quote.shippingAddress().canUseForBilling()) {
>                 selectBillingAddressAction(quote.shippingAddress());
>             }
> 

If there is no billing address -> set shipping Address as Billing Address if this Shipping Address Can be used for Billing

ishakhsuvarov:
> If there is no billing address -> set shipping Address as Billing Address if this Shipping Address Can be used for Billing

@tufahu What happens with the billing address controls in this case? Is it handled properly by default? I'm afraid we may get into the case where customer attempts to place an order with no billing address at all and will get an exception from the API call.
 
I'll check out the branch and do some testing to get the better grip of the idea.
Thanks!

tufahu:
Sorry i've missed to answer to this thread. It is handled properly by default, without any problems. Thanks!

PR #10181, opened 2017-07-07
Added try-catch block in getSwatchAttributeImage

josefbehr:
### Description
getSwatchAttributeImage calls generateSwatchVariations($file), which throws an Exception when the file $file does not exist in the file system (see line 189: $image = $this->imageFactory->create($absoluteImagePath), which calls Image->__construct() and then  Image->open, throwing an Exception on !file_exists($file)). This leads to the rest of the page not being rendered, when just one swatch image is missing. Not great. This fixes it and returns an empty file url (could also return the intended one, or a placeholder).

### Manual testing scenarios
Delete one or more of the images that are used for swatches (alternatively, copy a database from a productive system to a local machine without getting the files, too), then try to visit a category or product page using swatches.

### Contribution checklist


t-richards:
It looks like this fix was included in the 2.3.0 release: https://devdocs.magento.com/guides/v2.3/release-notes/ReleaseNotes2.3.0OpenSource.html

PR #10208, opened 2017-07-11
Fix getImage method when sending product alert email

Jeroenll:
### Description
When sending product alert emails from cron jobs it will do so from the backend area. An exception is thrown `Unable to resolve the source file for 'frontend/_view/nl_NL/Magento_Catalog/images/product/placeholder/.jpg'` because the product image can't be loaded.

Although Magento will mark the subscription send, in fact it's not. To solve this I created an emulation to the product image is loaded in the fronted area.

1. None that I could find.

### Manual testing scenarios
1. Add yourself to the mailing list for product alerts (e.g. out of stock products).
2. Run the cron job which is responsible for sending the email if a product is back in stock.
3. Magento will mark the subscription as send but in fact it's not.

### Contribution checklist

### DIFF
--- a/app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
+++ b/app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
@@ -58,25 +58,26 @@ abstract class AbstractEmail extends \Magento\Framework\View\Element\Template
 
     /**
      * @param \Magento\Framework\View\Element\Template\Context $context
-     * @param \Magento\Framework\Filter\Input\MaliciousCode    $maliciousCode
-     * @param PriceCurrencyInterface                           $priceCurrency
-     * @param \Magento\Catalog\Block\Product\ImageBuilder      $imageBuilder
-     * @param \Magento\Store\Model\App\Emulation               $appEmulation
-     * @param array                                            $data
+     * @param \Magento\Framework\Filter\Input\MaliciousCode $maliciousCode
+     * @param PriceCurrencyInterface $priceCurrency
+     * @param \Magento\Catalog\Block\Product\ImageBuilder $imageBuilder
+     * @param array $data
+     * @param \Magento\Store\Model\App\Emulation $appEmulation
      */
     public function __construct(
         \Magento\Framework\View\Element\Template\Context $context,
         \Magento\Framework\Filter\Input\MaliciousCode $maliciousCode,
         PriceCurrencyInterface $priceCurrency,
         \Magento\Catalog\Block\Product\ImageBuilder $imageBuilder,
-        \Magento\Store\Model\App\Emulation $appEmulation,
-        array $data = []
+        array $data = [],
+        \Magento\Store\Model\App\Emulation $appEmulation = null
     ) {
         $this->imageBuilder = $imageBuilder;
         $this->priceCurrency = $priceCurrency;
         $this->_maliciousCode = $maliciousCode;
-        $this->appEmulation = $appEmulation;
         $this->storeManager = $context->getStoreManager();
+        $this->appEmulation = $appEmulation
+            ?: \Magento\Framework\App\ObjectManager::getInstance()->get(\Magento\Store\Model\App\Emulation::class);
 
         parent::__construct($context, $data);
     }
@@ -229,11 +230,11 @@ abstract class AbstractEmail extends \Magento\Framework\View\Element\Template
     /**
      * Retrieve product image.
      *
-     * @param       $product
-     * @param       $imageId
+     * @param \Magento\Catalog\Model\Product $product
+     * @param string $imageId
      * @param array $attributes
-     *
      * @return \Magento\Catalog\Block\Product\Image
+     * @throws \Exception
      */
     public function getImage($product, $imageId, $attributes = [])
     {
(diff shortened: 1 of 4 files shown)


app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
```diff
@@ -212,18 +227,29 @@ public function getProductPriceHtml(
...
     public function getImage($product, $imageId, $attributes = [])
     {
-        return $this->imageBuilder->setProduct($product)
+        $this->appEmulation->startEnvironmentEmulation(
+            $this->storeManager->getStore()->getId(),
+            \Magento\Framework\App\Area::AREA_FRONTEND,
+            true
+        );
+
+        $image = $this->imageBuilder->setProduct($product)
```
ihor-sviziev: Would be great to wrap this part with try-catch block in order to prevent issues when in this part some exception happens, similar to this:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Payment/Helper/Data.php#L218

Jeroenll: Will do!

Jeroenll:
See unit tests are failing, will update these shortly.

app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
```diff
@@ -212,18 +227,29 @@ public function getProductPriceHtml(
     }
 
     /**
-     * Retrieve product image
+     * Retrieve product image.
      *
-     * @param \Magento\Catalog\Model\Product $product
-     * @param string $imageId
+     * @param       $product
```
ishakhsuvarov: Looks like you removed types from the DocBlock by mistake. Please revert it

app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
```diff
@@ -47,22 +47,37 @@
...
+     * @param \Magento\Catalog\Block\Product\ImageBuilder      $imageBuilder
+     * @param \Magento\Store\Model\App\Emulation               $appEmulation
+     * @param array                                            $data
      */
     public function __construct(
         \Magento\Framework\View\Element\Template\Context $context,
         \Magento\Framework\Filter\Input\MaliciousCode $maliciousCode,
         PriceCurrencyInterface $priceCurrency,
         \Magento\Catalog\Block\Product\ImageBuilder $imageBuilder,
+        \Magento\Store\Model\App\Emulation $appEmulation,
```
ishakhsuvarov: * New constructor argument may only be added as the last on the list due to backwards compatibility policy
* Additional argument must be nullable for the same reason.

Please follow [Backwards Compatible Development Guide](http://devdocs.magento.com/guides/v2.1/ext-best-practices/extension-coding/backwards-compatible-development/index.html#adding-a-constructor-parameter) for examples and more information

app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
```diff
@@ -47,22 +47,37 @@
...
+    /**
      * @param \Magento\Framework\View\Element\Template\Context $context
-     * @param \Magento\Framework\Filter\Input\MaliciousCode $maliciousCode
-     * @param PriceCurrencyInterface $priceCurrency
-     * @param \Magento\Catalog\Block\Product\ImageBuilder $imageBuilder
-     * @param array $data
+     * @param \Magento\Framework\Filter\Input\MaliciousCode    $maliciousCode
+     * @param PriceCurrencyInterface                           $priceCurrency
+     * @param \Magento\Catalog\Block\Product\ImageBuilder      $imageBuilder
+     * @param \Magento\Store\Model\App\Emulation               $appEmulation
```
ishakhsuvarov: Please use same DocBlock notation as all over Magento to increase consistency

ishakhsuvarov (changes requested):
Please check Travis CI build results and fix errors identified by tests, also please consider covering new code with test.

Jeroenll:
@ishakhsuvarov Updated the PR. I'm having a hard time fixing the unit tests, any idea why `$this->_storeManager->getStore()` is `null`? Same goes for `static` tests.

ishakhsuvarov:
@JeroenVanLeusden I am going to look into it shortly

Jeroenll:
@ishakhsuvarov Static tests still fail because the coupling between objects have a value of 14 and should be reduced to 12. What is the preferred way to achieve this?

ishakhsuvarov:
@JeroenVanLeusden I am currently looking into ways to refactor the code to reduce that number. It may be needed to extract a part of this class to a different one, in case we do not find a way to reduce number of dependencies in other way.

I've updated the unit test. The problem occurred due to not initialized mock object for store manager.

Jeroenll:
@ishakhsuvarov I saw the tests, really useful for next time! If there is anything I can do please let me know.

ishakhsuvarov (approved):

ishakhsuvarov:
@JeroenVanLeusden I've fixed the Coupling issue by extracting image building logic into the separate class with the single responsibility.
Could you please check, that your initial issue is fixed after the refactoring?
Thanks!

Jeroenll:
@ishakhsuvarov Just tested and can confirm this patch fixes the issue!

ishakhsuvarov:
@JeroenVanLeusden Great!
We will proceed with the merge procedures then.

daniel-ifrim (commented):
In constructor, array $data = [], shouldn't come after ImageProvider $imageProvider = null ?

Jeroenll:
It does according the [backward compatible guide](http://devdocs.magento.com/guides/v2.1/contributor-guide/backward-compatible-development/index.html).

PR #10213, opened 2017-07-11
Show values that are duplicate in attribute error msg

PieterCappelle:
When you add values to some attribute it's possible you have created duplicated values. If you want to save that attribute at Stores > Attributes > Products you will see the error messagee "The value of Admin must be unique." but that error message is not telling you which value is not unique.

This PR shows which values are duplicated so you easily update/remove the values so you can pass the error. Let me know if I need to modify something because this is first PR.

### Reproduce
1. Create attribute (select)
2. Add values
3. Add value Blue
4. Add second value Blue
5. Submit.

### Contribution checklist

schermafbeelding 2017-07-11 om 21 42 27


app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
```diff
@@ -25,21 +31,27 @@ define([
...
 
+                    if (!isValid) {
+                        msg = _config.message.replace('%1', duplicates.join(', '));
+                    }
+
                     return isValid;
                 },
 
-                _config.message
+                messager
```
omiroshnichenko, line 54: @PieterCappelle Why you can't use `msg` variable? Why you declare `messager` function?

app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
```diff
@@ -10,11 +10,17 @@ define([
     'use strict';
 
     return function (config) {
```
omiroshnichenko, line 12: @PieterCappelle It would be nice if you will cover this changes with JS tests.

PieterCappelle: Do you have any guide to create this, never did this before. Sorry.

app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
PieterCappelle: I can't use the msg variable because its not in the scope of the function inside the `jQuery.validator.addMethod`.

PieterCappelle:
Replied to all your comments @omiroshnichenko. Let me know if I can help.

app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
omiroshnichenko: @PieterCappelle `msg` variable declared in the scope of [main function](https://github.com/PieterCappelle/magento2/blob/7911d0e55564f74c62ac3a8c1cc34c1823bdfc51/app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js#L13) at the same scope declared `messager` function. You can use it.

app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
omiroshnichenko: @PieterCappelle  Here is a [doc](http://devdocs.magento.com/guides/v2.2/test/js/jasmine.html) and also you can see in dev/tests/js/jasmine/tests already created tests. If you will have some additional question contact me directly.

app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
PieterCappelle: @omiroshnichenko No, you will be in a javascript closure - don't know the exact javascript scope name but I think it's a closure - . You have to use custom function to display the msg variable. I tried multiple times like code below and it's not working. 

```
/**
 */

define([
    'jquery',
    'mage/backend/validation'
], function (jQuery) {
    'use strict';

    return function (config) {
        var msg = '',
            _config = jQuery.extend({
                element: null,
                message: '',
                uniqueClass: 'required-unique'
            }, config);

        if (typeof _config.element === 'string') {
            jQuery.validator.addMethod(
                _config.element,

                function (value, element) {
                    var inputs = jQuery(element)
                            .closest('table')
                            .find('.' + _config.uniqueClass + ':visible'),
                        valuesHash = {},
                        isValid = true,
                        duplicates = [];

                    inputs.each(function (el) {
                        var inputValue = inputs[el].value;

                        if (typeof valuesHash[inputValue] !== 'undefined') {
                            isValid = false;
                            duplicates.push(inputValue);
                        }
                        valuesHash[inputValue] = el;
                    });

                    if (!isValid) {
                        msg = _config.message.replace('%1', duplicates.join(', '));
                    }

                    return isValid;
                },

                msg
            );
        }
    };
});
```

app/code/Magento/Catalog/view/adminhtml/web/catalog/product/attribute/unique-validate.js
PieterCappelle: @omiroshnichenko I tried this but it's to complex for me at this moment. I don't even get how I need to make the file structure. Do I need to follow the exact path as in the module directory? Other modules use custom paths and are ignore the path of the module.

omiroshnichenko (approved):

Ctucker9233:
@magento-team This PR isn't included in 2.2 Should it be backported?

PR #10225, opened 2017-07-12
Updated EAV Attribute code max lenght

okorshenko:
### Description
 - EAV attribute code will be transformed into column name for flat tables. MySQL does not support column name with more than 64 symbols

### Manual testing scenarios

1. Install magento 
2. enable flat mode for catalog
3. make sure that catalog works on the frontend

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/Entity/Attribute.php
+++ b/app/code/Magento/Eav/Model/Entity/Attribute.php
@@ -23,7 +23,7 @@ class Attribute extends \Magento\Eav\Model\Entity\Attribute\AbstractAttribute im
     /**
      * Attribute code max length
      */
-    const ATTRIBUTE_CODE_MAX_LENGTH = 255;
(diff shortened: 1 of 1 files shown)


orlangur:
@okorshenko why do we need this change at all and not simply change back to old `30` value?

okorshenko:
Hi @orlangur The value 30 was set based on some plans about Oracle support (which allows only 32 symbols). So Magento officially does not support Oracle DB. In MySQL limitation is 64.

orlangur:
Ok, I thought it was business requirement (like from times when attribute_code was displayed in UI) and if something went wrong it's easier to revert if there is no clear motivation for length increase.

Nolwennig:
When this update will integrate the Magento core?

okovalov-commer:
When will this be released?

okovalov-commer:
Hi guys, any updates here ? when it gets promoted live? thank you

okorshenko:
This PR will be available with 2.3 release

PR #10227, opened 2017-07-13
Fix return-type annotation(in relation to PR#9266)

evgk:
Since a local fix was chosen in PR#9266, there is still an interface mismatch. Proposed fix addresses it.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Pricing/SaleableInterface.php
+++ b/lib/internal/Magento/Framework/Pricing/SaleableInterface.php
@@ -37,7 +37,7 @@ interface SaleableInterface
     /**
      * Returns quantity of saleable item
      *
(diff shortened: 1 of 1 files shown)


ihor-sviziev:
I think implementation should return correct type, interface shouldn't be changed.

ishakhsuvarov:
@evgk I agree with @ihor-sviziev. Why would Qty be represented by a string?

okorshenko:
Hi @evgk thank you for the fix. Unfortunately, we can not accept this change despite it defines real return type from the method. The issue is more complex. We don't have in place the proper conversion of MySQL types to PHP types. We are working on this right now and this will be addressed in 2.3.

PR #10246, opened 2017-07-14
Fixing issue with getSize function not recalculating after adding filters.

KevinDonaldson:
Should fix several issues with `getSize()` returning incorrect values after filters have been applied to a collection.

### Description
- Adds additional guard to check if items on collection is equal to the totalRecords on collection. 

### Fixed Issues 
1. https://github.com/magento/magento2/issues/7730

### Manual testing scenarios
1. Load a collection.
2. Apply a filter to the collection.
3. Get size of collection using `getSize()` function and validate that value is correct.

### DIFF
--- a/lib/internal/Magento/Framework/Data/Collection/AbstractDb.php
+++ b/lib/internal/Magento/Framework/Data/Collection/AbstractDb.php
@@ -215,11 +215,14 @@ abstract class AbstractDb extends \Magento\Framework\Data\Collection
      */
     public function getSize()
     {
-        if ($this->_totalRecords === null) {
+        $itemCount = count($this->_items);
+
+        if ($this->_totalRecords === null || ($itemCount > 0 && $itemCount !== $this->_totalRecords)) {
             $sql = $this->getSelectCountSql();
-            $this->_totalRecords = $this->getConnection()->fetchOne($sql, $this->_bindParams);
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10246) <br/>All committers have signed the CLA.

okorshenko:
Hi @KevinDonaldson Could you please cover your fix with the integration test? This is a very low-level class that must be covered with the tests. Thank you

ihor-sviziev:
@KevinDonaldson @okobchenko it's incorrect behavior because getSize method will return total records in DB, while in $this->_items we will have amount of products that was **limited**. Please do not merge these changes.
I think in order to fix this issue we need to have some "hash" of current filters and if it's different (added new filters) - total records should be added , **but** if we have any results in $this->_items - we don't need to calculate new total records value again and use the last one.

Also there are might be negative effect because getSize result might be already used in UI, as example in pagination, so as for me - you need to find another way to add your filters instead of using before load plugin

okorshenko:
Hi @KevinDonaldson Thank you for your contribution. Please consider the previous comment
Closing this PR.

KevinDonaldson:
Thanks for the feedback.

drew7721:
Is there a new issue created for this? Has this been fixed? From what I see the PR was rejected but the issue was also closed...

PR #10248, opened 2017-07-15
Fix for #4530 $product->getRatingSummary() always returns null […]

decius7bc:
This PR fixes the `\Magento\Review\Block\Product\ReviewRenderer` such that the reviews summary is shown although `$displayIfNoReviews` is set to `false` and reviews are in fact available.

### Description
Always retrieve relevant data before checking whether there are active reviews if `$displayIfNoReviews` is set to `false`.

### Fixed Issue
magento/magento2#4530: $product->getRatingSummary() always returns null on products with ratings

### Manual testing scenario
Change the last parameter `$displayIfNoReviews` in the `\Magento\Review\Block\Product\ReviewRenderer::getReviewsSummaryHtml` method call to `false` such that the call in the template `app/code/Magento/Catalog/view/frontend/templates/product/view/review.phtml` looks like this

```php
<?php echo $block->getReviewsSummaryHtml($block->getProduct(), false, false)?>
```

In both cases (No reviews available/at least one review is available) the reviews summary won't be shown. After applying this PR the reviews summary will be shown as expected only if there is at least one review available. 

### Contribution checklist

### DIFF
--- a/app/code/Magento/Review/Block/Product/ReviewRenderer.php
+++ b/app/code/Magento/Review/Block/Product/ReviewRenderer.php
@@ -57,6 +57,10 @@ class ReviewRenderer extends \Magento\Framework\View\Element\Template implements
         $templateType = self::DEFAULT_VIEW,
         $displayIfNoReviews = false
     ) {
+        if (!$product->getRatingSummary()) {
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10248) <br/>All committers have signed the CLA.

PR #10250, opened 2017-07-16
update phpserver to support versioned static urls

Flyingmana:
the phpserver implementation did not support versioned static urls before, which is solved by this change

I decided to just do a redirect, as it should be enough for dev users, as they usually use no-cache or hard reloads to flush any css files used anyway.

You can reproduce this issue by following the normal setup flow (install via cli) and generate the frontend files.
Then use the php builtin server as a webserver and then load the home page.
Result without this patch will be a bunch of 404 responses for css files.
After this patch, they can get resolved correctly.

### DIFF
--- a/phpserver/router.php
+++ b/phpserver/router.php
@@ -61,6 +61,13 @@ if (php_sapi_name() === 'cli-server') {
     }
 
     $debug($route);
+    
+    if (strpos($route, ‘static/version’) === 0) {
+        $redirectRoute = preg_replace(‘#static/version[0–9]*/#’,’static/’, $route, 1);
+        $debug(“redirect static version string to:” . $redirectRoute);
+        header(‘Location: /’.$redirectRoute);
+        exit;
+    }
 
     if (strpos($route, 'media/') === 0 ||
         strpos($route, 'opt/') === 0 ||
--- a/phpserver/router.php
+++ b/phpserver/router.php
@@ -62,10 +62,10 @@ if (php_sapi_name() === 'cli-server') {
 
     $debug($route);
     
-    if (strpos($route, ‘static/version’) === 0) {
-        $redirectRoute = preg_replace(‘#static/version[0–9]*/#’,’static/’, $route, 1);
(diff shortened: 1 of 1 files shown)


okorshenko:
Hi @Flyingmana 
Could you please provide step-by-step instruction how we can reproduce the issue?
Thank you

Also, please fix the code. It throws a fatal error due to a syntax error. thank you

Flyingmana:
Iam really sorry, I did copy&paste it over different places before submitting it as a PR, seems one of them did replace the quotes...

How detailed do you need the step by step? And which step in the initial description was not detailed enough?

ishakhsuvarov:
Hi @Flyingmana 
I've been trying to manually test according to steps you provided and did not get any success due to  `Too Many Redirects` error all the time.

I've looked at the code and tried to improve the Regular Expression, fixing the coding style errors on the way. Now it works as expected for me. Please check if it works fine for you as well.
Thanks!

Flyingmana:
Will test it out in the evening, but the changes look ok for me.

Was the endless redirect related to a version string which contained something else then numbers?
Did the '/static' part of the replace make problems?

ishakhsuvarov:
@Flyingmana During my testing it did not really replace anything on each iteration, I've also ran some tests with the expression in question and didn't see it work as expected

yarcowang:
Em... I've just did a post in Linkedin [Has anyone tried PHP builtin web server?](https://www.linkedin.com/groups/2083430/2083430-6339797968891277316). Is this patch resolving my problem? (Well, according to the README.md under phpserver, I thought it could work and I did some wrong somewhere... first time touching magento actually)
And for now, I'm using nginx+php-fpm, but I prefer php builtin server for development environment.

Updated:
~~It seems I need to visit `http://localhost:30080/index.php/`, but not `http://localhost:30080/`... with index.php, it works...~~ Still didn't work, maybe I see the cache?

Updated:
The patch could work. Confirmed. And you should visit `/index.php` instead of `/`, or you will see an empty page. (to me)

Issue #10252, opened 2017-07-16
Default values for system config of type image/file is not working


AntonEvers:
This is a clone of https://github.com/magento/magento2/pull/5085 that was closed because of a deleted repository.

### Preconditions

1. Install Magento from `develop` branch 

### Steps to reproduce

1. Add config field with type `image`.
2. Add default config value for that field.
3. Save the config field without uploading a new file.

### Expected result

1. New record created in the database table `core_config_data` with default value.

### Actual result

1. New record created in the database table `core_config_data` with value NULL.

If you set a default value for a config field type image/file and you save it. It won't set the default value in the database. Instead it will save NULL and so overwrite the default config.

With this change the default value will be set as value.

orlangur:
Seems like this can be closed now. Thanks for the fix!

AntonEvers:
@orlangur you are right. Thanks for handling this ticket fast.

magento-team:
Internal ticket to track issue progress: MAGETWO-70764

PR #5085, opened 2016-06-17
Fix overwrite default value image/file with NULL

joost-florijn-kega:
## Steps to reproduce
1. Install Magento from `develop` branch.
2. Add config field with type `image`.
3. Add default config value for that field.
4. Save the config field without uploading a new file.
## Expected result
1. New record created in the database table `core_config_data` with default value.
## Actual result
1. New record created in the database table `core_config_data` with value NULL.

If you set a default value for a config field type image/file and you save it. It won't set the default value in the database. Instead it will save NULL and so overwrite the default config.

With this change the default value will be set as value.

PR #10253, opened 2017-07-16
Fix overwrite default value image/file with NULL

AntonEvers:
If you set a default value for a config field type image/file and you save it. It won't set the default value in the database. Instead it will save NULL and so overwrite the default config.

With this change the default value will be set as value.

### Description

This is a replication of https://github.com/magento/magento2/pull/5085

1. magento/magento2#10252: Default values for system config of type image/file is not working 

### Manual testing scenarios

1. Install Magento from develop branch.
2. Add config field with type image.
3. Add default config value for that field.
4. Save the config field without uploading a new file.

Before the fix:
- New record created in the database table core_config_data with value NULL.

After the fix:
- New record created in the database table core_config_data with default value.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Config/Model/Config/Backend/File.php
+++ b/app/code/Magento/Config/Model/Config/Backend/File.php
@@ -110,6 +110,8 @@ class File extends \Magento\Framework\App\Config\Value
         } else {
             if (is_array($value) && !empty($value['delete'])) {
                 $this->setValue('');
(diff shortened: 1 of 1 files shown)


miguelbalparda:
Hi @ajpevers I'll be reviewing this PR. Thanks for your contribution!

miguelbalparda:
Thanks for your contribution @ajpevers!

PR #6268, opened 2016-08-22
+: getItems from order if already provided.

basselalaraaj:
fixes issue #6267

PR #10254, opened 2017-07-16
getItems from order if already provided.

AntonEvers:
fixes issue #6267

This is a replication of PR https://github.com/magento/magento2/pull/6268 that was closed because of a removed repository

1. magento/magento2#6267: Order getitemscollection Missing items when loading

### Manual testing scenarios

1. See https://github.com/magento/magento2/issues/6267

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -1244,25 +1244,29 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     /**
      * @param array $filterByTypes
      * @param bool $nonChildrenOnly
-     * @return ImportCollection
+     * @return \Magento\Sales\Api\Data\OrderItemInterface[]
      */
     public function getItemsCollection($filterByTypes = [], $nonChildrenOnly = false)
     {
-        $collection = $this->_orderItemCollectionFactory->create()->setOrderFilter($this);
+        if (!$this->hasData(OrderInterface::ITEMS)) {
+            $collection = $this->_orderItemCollectionFactory->create()->setOrderFilter($this);
 
-        if ($filterByTypes) {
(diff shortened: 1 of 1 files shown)


okorshenko:
Hi @ajpevers 
Could you please check failed tests?

AntonEvers:
@okorshenko will do.

AntonEvers:
@okobchenko before I fix the tests, please correct me if I am wrong... I think 
- the outcome of this method should not be cached on the key `OrderInterface::ITEMS` because that is already reserved for the `getItems()` method. 
- the outcome of this method should not be cached without taking the method parameters into account.
- there is no use in caching this method at all.

In #6267 @basselalaraaj mentions calling `getItemsCollection()` after calling `setItems()` and getting no results. I think this is what you should expect. If you want to get the items after setting them you should call `getItems()` instead of `getItemsCollection()`.

This already caches the result of `getItemsCollection()` on the `OrderInterface::ITEMS` key:

```php
    /**
     * @return \Magento\Sales\Api\Data\OrderItemInterface[]
     */
    public function getItems()
    {
        if ($this->getData(OrderInterface::ITEMS) == null) {
            $this->setData(
                OrderInterface::ITEMS,
                $this->getItemsCollection()->getItems()
            );
        }
        return $this->getData(OrderInterface::ITEMS);
    }
```

What is your view on this?

okorshenko:
Hi @ajpevers 

1. The outcome of the `getItemsCollection()` method should not be cached on the key OrderInterface::ITEMS because that is already reserved for the getItems() method.

2. calling `getItemsCollection()` after calling `setItems()` should not return same items.
if you need same items, you can set them to the collection. Setting items on `getItemsCollection` method is no expected behavior. 
I don't think that we should accept this pull request. 
Closing PR. Thank you for pointing this out. I missed that on the first look

PR #10259, opened 2017-07-17
Remove zend json from customer data

dmanners:
### Description
As both Zend_Json and \Magento\Framework\Json\EncoderInterface are now deprecated I have migrated the usage across the the accepted \Magento\Framework\Serialize\Serializer\Json class.

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Block/CustomerScopeData.php
+++ b/app/code/Magento/Customer/Block/CustomerScopeData.php
@@ -25,15 +25,16 @@ class CustomerScopeData extends \Magento\Framework\View\Element\Template
     private $serializer;
 
     /**
+     * CustomerScopeData constructor.
      * @param \Magento\Framework\View\Element\Template\Context $context
-     * @param $jsonEncoder
+     * @param \Magento\Framework\Json\EncoderInterface $jsonEncoder
      * @param array $data
      * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
-     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     * @throws \RuntimeException
      */
     public function __construct(
         \Magento\Framework\View\Element\Template\Context $context,
-        $jsonEncoder,
+        \Magento\Framework\Json\EncoderInterface $jsonEncoder,
         array $data = [],
         \Magento\Framework\Serialize\Serializer\Json $serializer = null
     ) {
--- a/app/code/Magento/Customer/Block/CustomerScopeData.php
+++ b/app/code/Magento/Customer/Block/CustomerScopeData.php
@@ -50,4 +50,30 @@ class CustomerScopeData extends \Magento\Framework\View\Element\Template
     {
         return (int)$this->_storeManager->getStore()->getWebsiteId();
     }
+
+    /**
+     * Get the invalidation rules json encoded
+     *
+     * @return string
+     */
+    public function getInvalidationRulesJson()
+    {
+        return $this->jsonEncoder->encode(
+            [
+                '*' => [
+                    'Magento_Customer/js/invalidation-processor' => [
+                        'invalidationRules' => [
+                            'website-rule' => [
+                                'Magento_Customer/js/invalidation-rules/website-rule' => [
+                                    'scopeConfig' => [
+                                        'websiteId' => $this->getWebsiteId(),
+                                    ]
+                                ]
+                            ]
+                        ]
+                    ]
+                ],
+            ]
+        );
+    }
 }
--- a/app/code/Magento/Customer/Block/CustomerScopeData.php
+++ b/app/code/Magento/Customer/Block/CustomerScopeData.php
@@ -25,12 +25,12 @@ class CustomerScopeData extends \Magento\Framework\View\Element\Template
     private $serializer;
 
     /**
-     * CustomerScopeData constructor.
      * @param \Magento\Framework\View\Element\Template\Context $context
      * @param \Magento\Framework\Json\EncoderInterface $jsonEncoder
      * @param array $data
      * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
      * @throws \RuntimeException
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
     public function __construct(
         \Magento\Framework\View\Element\Template\Context $context,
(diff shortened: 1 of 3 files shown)


dmanners:
Not 100% sure if this is the "best way" ™️ for replacing the deprecated class that is being injected in. Happy for feedback and change requests.

ihor-sviziev:
@dmanners I see that app/code/Magento/Customer/Block/CustomerScopeData.php file didn't exists in 2.1 releases, so I think it's ok that we can change file before release.

dmanners:
Ah @ihor-sviziev thanks for the information. I had not noticed it before.

okorshenko:
Hi @dmanners 
Unfortunately, we can not add new methods to existing @api classes and change constructor signatures. develop branch is designed for 2.3 but this class was added in 2.2. so this will be BiC

dmanners:
Hi @okorshenko, thanks for the feedback. I have a few follow up questions then.

1) If this is a new class in 2.2 can I simply pr against `2.2.0-preview` or is it too later to get changes into that branch?
2) Is there a plan/blog post out there on the "perfect ™️" approach for changing these sort of classes? This might be a simple example but I am sure this is not the first and wont be the last time someone wants to make a change to an `@api` tagged class.

Happy to here any options for making this change.

app/code/Magento/Customer/Block/CustomerScopeData.php
```diff
@@ -50,4 +50,31 @@ public function getWebsiteId()
...
+
+    /**
+     * Get the invalidation rules json encoded
+     *
+     * @return bool|string
+     * @throws \InvalidArgumentException
+     */
+    public function getSerializedInvalidationRules()
+    {
+        return $this->serializer->serialize(
```
orlangur: I don't think serialization in block is a good idea. It is better to just return raw data and JSON (or whatever else) encode in template.

Due to such `get...Json` methods in Magento 1 some extensions had to decode string and then re-encode just to modify array.

dmanners: Thanks for the feedback, generally I agree with you here. I considered putting in one function to cover the serialization of the data. And one to cover the getting of the data. This would then allow 3rd party extensions to manipulate the data. Currently with the data stuck in the template it could not be changed easily anyway. I will have a look around for other examples of code like this.

orlangur:
> So all blocks should be `@api` annotated. This test checks that all blocks declared in layout files are public

Hm, this is something new I was not aware of...

@okorshenko as all blocks became `@api` annotated are you sure that we cannot add new public methods to them? Unlike adding some methods to interface it sounds a bit weird to me.

@dmanners,
> approach for changing these sort of classes?

http://devdocs.magento.com/guides/v2.1/contributor-guide/backward-compatible-development/index.html more or less describes approach, if some cases are not covered yet we can extend it with time.

app/code/Magento/Customer/Block/CustomerScopeData.php
```diff
@@ -50,4 +54,37 @@ public function getWebsiteId()
...
+
+    /**
+     * Get the invalidation rules json encoded
+     *
+     * @return bool|string
+     * @throws \InvalidArgumentException
+     */
+    public function getSerializedInvalidationRules()
+    {
+        return $this->serializer->serialize($this->getInvalidationRules());
```
orlangur, line 89: Still... What are the advantages compared to `echo json_encode($block->getInvalidationRules();` in template? As to me encoding is purely responsibility of template.

Another thing which just came to my mind: do we have a recommendation to leave JS component declaration in templates? Thus they may be overridden more easily, without touching PHP classes. At least I was not able to find similar occurrence of `/js/` or `'*'` in `*.php`.

okorshenko:
Hi @dmanners and @orlangur 
Sorry for confusing. 
We can not add public methods to @api classes in patch releases, but we can do that in minor. So 2.3 is a minor release and we can add this method. The issue was with constructor
Here is SVC report

 - MAJOR - Magento\Customer\Block\CustomerScopeData::__construct - [public] Method parameter changed
- MINOR - Magento\Customer\Block\CustomerScopeData::getSerializedInvalidationRules  - [public] Method has been added

So the issue was with the constructor.

app/code/Magento/Customer/Block/CustomerScopeData.php
```diff
@@ -20,23 +20,27 @@ class CustomerScopeData extends \Magento\Framework\View\Element\Template
...
+     * @param $jsonEncoder
      * @param array $data
+     * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
     public function __construct(
         \Magento\Framework\View\Element\Template\Context $context,
-        \Magento\Framework\Json\EncoderInterface $jsonEncoder,
-        array $data = []
+        $jsonEncoder,
```
ihor-sviziev: Looks like you removed interface by mistake.

app/code/Magento/Customer/Block/CustomerScopeData.php
dmanners: Not sure about the leaving of the js declaration in the templates. Though to me this feels like config and to have to change a full template to change the config feels odd.

With regards to the `json_encode` in the template I dont agree here. I feel that we should either be using json_encode everywhere or wrap it's usage across the system, which would include templates. This way if a mass change is needed it is easier to manage and roll out.

app/code/Magento/Customer/Block/CustomerScopeData.php
dmanners: yup, will add that back in.

app/code/Magento/Customer/Block/CustomerScopeData.php
orlangur: Have no idea which kind of mass change could it be :D

Ok, I see your point, you want to encapsulate serializer logic in block class and don't make template aware of it (while it does `escapeHtml` and other similar things). There is nothing to discuss here, let's wait for the feedback from Oleksii and then maybe put a common recommendation for such situation somewhere to keep the code consistent.

dmanners: Yeah something like `escapeHtml` would be a nice option here also thought about that. It's above this PR but I think some "common" way to do this through templates would be amazing.

Ctucker9233:
@magento-team What version of 2.2 will have this included? As of 2.2.7 it doesn't appear to be in there.

orlangur:
@Ctucker9233 you can check corresponding 2.2 commits: http://prntscr.com/mz7vk2

PR #10273, opened 2017-07-18
Ignore Errors With Module Versions Requiring an Upgrade

paulpartington-cti:
### Description
At the moment, it's not possible to do a deployment without maintenance mode if a module version is being updated. If a page is loaded and the module versions are different, an exception is shown to the customer.

Although I see the purpose behind this in terms of preventing other errors, it does mean that we have to perform disruptive deployments if we want to deploy a module update. 

Very often, this is a temporary issue until the frontend servers have all got the latest version of the code. I believe flexibility should be given to allow developers to control whether this should throw an exception. 

This change introduces a new value in env.php called 'ignore_module_version_exception'. If this is set to '1' then the errors are supressed and not displayed on the frontend. If not set or set to anything other than '1', then the errors are shown as usual.

I was unable to implement this with an integration test due to the contents of env.php being validated during the tests to make sure they're not changed.

I'm very much open to other solutions rather than use env.php. It may be useful to create a console command to set this value if it is used.

### Manual testing scenarios
1. Install a clean version of Magento
2. Ensure that ignore_module_version_exception is added to env.php and set to '0' (this should be set at the same level in the data as 'MAGE_MODE'
3. Load the page
4. Locate a module in the code e.g. module-catalog and increase the version number
5. Clear the cache
6. Reload the page
7. You should now see the exception
8. Update 'ignore_module_version_exception' to '1'
9. Clear the cache
10. Reload the page
11. No exception should be visible

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Module/Plugin/DbStatusValidator.php
+++ b/lib/internal/Magento/Framework/Module/Plugin/DbStatusValidator.php
@@ -7,6 +7,7 @@ namespace Magento\Framework\Module\Plugin;
 
 use Magento\Framework\Cache\FrontendInterface as FrontendCacheInterface;
 use Magento\Framework\Module\DbVersionInfo;
+use Magento\Framework\App\DeploymentConfig;
 use Magento\Framework\App\FrontController;
 use Magento\Framework\App\RequestInterface;
 use Magento\Framework\Exception\LocalizedException;
@@ -17,6 +18,8 @@ use Magento\Framework\Phrase;
  */
 class DbStatusValidator
 {
+    const PARAM_IGNORE_MODULE_VERSION_EXCEPTION = 'ignore_module_version_exception';
+
     /**
      * @var FrontendCacheInterface
      */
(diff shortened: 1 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10273) <br/>All committers have signed the CLA.

orlangur:
I don't think this is a viable solution.

> If a page is loaded and the module versions are different, an exception is shown to the customer.

Exception should never be shown to customer in production mode. It should be only logged and the customer should see some general error page.

> Very often, this is a temporary issue until the frontend servers have all got the latest version of the code

This is pretty unreliable deployment strategy. Frontend server should not serve requests until it is fully ready. Any database change may potentially break the old code.

Technically you can do the same by overriding `$this->dbVersionInfo->getDbVersionErrors()` implementation for your needs if you don't want to fix deployment process but I don't think such kind of flags is suitable for core code as it should encourage best practices and not some random hacks.

paulpartington-cti:
@orlangur thank you for the prompt feedback.

> Exception should never be shown to customer in production mode. It should be only logged and the customer should see some general error page.

Fair point, I hadn't seen that you can override that in pub/errors/local.xml.

That said, just because the version of a module is updated it doesn't mean that the code is going to break. With how packages are now versioned, a minor increment can simply be a template change or some non-breaking update but currently you have to go into maintenance mode for however long the deployment is occurring.

As I said, I'd be happy to look at something other than environment flags but I do think some flexibility in throwing an error in this instance should be possible.

orlangur:
> That said, just because the version of a module is updated it doesn't mean that the code is going to break.

Surely. But it _can_ break. It's always better to assume the worst case and current implementation works like a charm in both cases. I see no need in supporting two flows, strict one and the one which can cause site breakage.

Why do you think such functionality is needed in core if you can simply create an around plugin for `$this->dbVersionInfo->getDbVersionErrors()` which does nothing? Do you think this feature will be used frequently?

paulpartington-cti:
@orlangur I believe that it should be up to the developer whether they put the site into maintenance mode during a deployment as it's not always necessary. Of course, some times it is required and I'd always lean towards using it if there's any doubt. However, arranging downtime to do a deployment tends to lead to out of hours deployments where there is limited support.

I'll close this and make it as a plugin like you suggested. Thanks for taking the time to review it.

PR #10281, opened 2017-07-18
Update DefaultStock.php

flancer64:
Please, don't change stock id on update if you want to make catalog inventory great again ;)

### DIFF
--- a/app/code/Magento/CatalogInventory/Model/ResourceModel/Indexer/Stock/DefaultStock.php
+++ b/app/code/Magento/CatalogInventory/Model/ResourceModel/Indexer/Stock/DefaultStock.php
@@ -296,7 +296,7 @@ class DefaultStock extends AbstractIndexer implements StockInterface
             $data[] = [
                 'product_id' => (int)$row['entity_id'],
                 'website_id' => (int)$row['website_id'],
-                'stock_id' => Stock::DEFAULT_STOCK_ID,
+                'stock_id' => (int)$row['stock_id'],
                 'qty' => (double)$row['qty'],
                 'stock_status' => (int)$row['status'],
             ];


maghamed:
Hi @flancer64 

Currently, we change stock_id to default stock value for StockStatuses because Magento 2 out of the box is a single stock system. And we don't support multiple stocks yet, it's a point of implementation MSI (Multi Source Inventory) project which is being developed now.

But we don't want to make artificial limitations for the community if somebody wants to implement own Multi Stock extension based on current Catalog Inventory interfaces and implementation.

So, potentially we could accept current PR if we would be sure that nothing will go wrong and we will not break existing installations.

Also, in the current codebase we either have hardcoded Stock::DEFAULT_STOCK_ID filter for 
Stock Status Resource model
https://github.com/magento/magento2/blob/develop/app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php#L222-L233
```php
    public function addStockDataToCollection($collection, $isFilterInStock)
    {
        $websiteId = $this->getStockConfiguration()->getDefaultScopeId();
        $joinCondition = $this->getConnection()->quoteInto(
            'e.entity_id = stock_status_index.product_id' . ' AND stock_status_index.website_id = ?',
            $websiteId
        );

        $joinCondition .= $this->getConnection()->quoteInto(
            ' AND stock_status_index.stock_id = ?',
            Stock::DEFAULT_STOCK_ID
        );
```

which is used for applying "In Stock" filter to Stock Status Collection
```php
    public function beforeLoad(Collection $productCollection, $printQuery = false, $logQuery = false)
    {
        $this->stockHelper->addIsInStockFilterToCollection($productCollection);
        return [$printQuery, $logQuery];
    }
```

So, the question to you is whether your system operates normally without changes in StockStatus Resource model? Is it require to you to change just indexation process?

maghamed:
The safest way for us is to introduce new extension point. If it's needed just for this indexer - we could add a new boolean attribute (like $isSingleStock) to DefaultIndexer class and provide a possibility to customize it through DI.xml.

If it's needed, for example, in your extension in several places. It's better to introduce dedicated extension point like this one.
```php
interface SingleStockResolverInterface
{
    /**
     * Detect whether single stock mode is used
     * @return boolean
     */
    public function isSingleStockMode();
}

class SingleStockResolver implements SingleStockResolverInterface
{
    /**
     * @param boolean $isSingleStockMode  Configured through DI.xml, define whether Single or Multi stock used
     * default value is TRUE 
     */
    public function __construct($isSingleStockMode)
    {
        $this->isSingleStockMode = $isSingleStockMode;
    }
    
    /**
     * @inheritdoc
     */
    public function isSingleStockMode()
    {
        return $this->isSingleStockMode;
    }
}
```
And inject SingleStockResolverInterface resolver to DefaultStock.php class and all other places where decision point needed.

As for me I don't like the idea to introduce SingleStockResolverInterface extension point if in fact Magento 2 doesn't provide Multi Stock Functionality yet.

flancer64:
Thanks, @maghamed for the advice.

I will try to make the changes you offer and will answer your questions after that.

ishakhsuvarov:
Hi @flancer64 
Do you need any assistance with this PR?

flancer64:
No, thanks. I still did not make a changes in own code. I need some more time for this. Sorry.

ishakhsuvarov:
@flancer64 
Thanks for the update, take your time.

ishakhsuvarov:
@flancer64 Closing this PR for now due to inactivity. Please reopen when you have time to continue on it. 
Thanks for collaboration!

PR #10302, opened 2017-07-21
Treat message class as newable object to avoid reuse

scottsb:
### Description
Currently the `TransportBuilder` class constructor requests the `Message` class as a singleton. However, since this is a data object, it should be treated as a newable object. Otherwise there is the risk that it will be passed a "dirty" copy of the object when the class is instantiated.

### Manual testing scenarios
The scenario that we ran into this involves inheritance, which I know is no longer recommended by the architecture guidelines, but it is the simplest to explain. The issue could also surface in the case that another class requests a `Message` object and modifies its properties before the `TransportBuilder` class is instantiated.

1. Inject the core `TransportBuilder` class into a new class `CoreSender` and add a method `send` which sends an email using the injected class.
2. Create a new class that inherits from the core `TransportBuilder`. Call it `TransportBuilderChild`. No need to override any methods: it can be an empty class.
3. Inject the new `TransportBuilderChild` class into a new class `ExtendedSender` and add a method `send` which sends an email using the injected class, but to a different email address than in step 1.
4. Call `CoreSender::send()` after `ExtendedSender::send()`.

**Behavior Before PR:** The latter `send()` call ends up sending the email to *both* the recipient from step 1 and step 2. This is because `TransportBuilder` object reuses the dirty copy of `Message` from `TransportBuilderChild`.

**Behavior After PR:** The two `send()` calls send their respective messages only to their specified recipients. The `TransportBuilder` object gets a freshly instantiated instance of `Message`.

### DIFF
--- a/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
+++ b/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
@@ -10,7 +10,7 @@ namespace Magento\Framework\Mail\Template;
 
 use Magento\Framework\App\TemplateTypesInterface;
 use Magento\Framework\Exception\LocalizedException;
-use Magento\Framework\Mail\MessageInterface;
+use Magento\Framework\Mail\MessageInterfaceFactory;
 use Magento\Framework\Mail\TransportInterfaceFactory;
 use Magento\Framework\ObjectManagerInterface;
 use Magento\Framework\Phrase;
@@ -70,6 +70,13 @@ class TransportBuilder
     protected $objectManager;
 
     /**
+     * Message Factory
+     *
+     * @var \Magento\Framework\Mail\MessageInterfaceFactory
+     */
+    protected $messageInterfaceFactory;
+
+    /**
      * Message
      *
      * @var \Magento\Framework\Mail\Message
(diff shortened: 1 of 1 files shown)


okorshenko:
HI @scottsb could you please fix backward compatibility issues in your PR so that we can proceed with PR review.

http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/

orlangur:
I would recommend to mark `message` argument as `shared="false"` in `di.xml` which would lead to the same result without any changes in PHP classes.

scottsb:
@okorshenko: would @orlangur's suggestion meet the backwards-compliant requirement? If so, I'll switch the PR to do that instead.

scottsb:
The one drawback to the `shared="false"` approach is that it doesn't let me get rid of the direct use of the object manager in the `reset()` method, but then that wasn't necessary to fix the bug.

okorshenko:
Hi @orlangur  and @scottsb 
`shared="false"` will fix this particular case, but does solve the problem in general. 
We need to create message object as an internal data storage. As an option we can use factory here, but since we have `Object Manager` in this class, there is no need to add new dependency to the `@api` class.

Also, calling the method `$factory->create` in the constructor is not recommended way. 

We need to use delayed object creation. Introduce new **protected** method `getMessage` and **private** property `messageData` and update all client code to use this method instead of direct usage of property `message`. Only method `reset` will set null to the property `messageData` directly.
Also, we need to deprecate the constructor argument `message` and property `message` with the explanation

scottsb:
@okorshenko Thanks for the detailed guidance. I can definitely implement this, but I wanted to run one other option past you. What we did as a workaround ourselves was a variant of @orlangur's idea but more generalized. We added this to a `di.xml` in our own module:

`<type name="Magento\Framework\Mail\MessageInterface" shared="false" />`

Which will prevent the message class from being reused by the object manager anywhere (rather than just on this one argument as the original suggestion). Would that work?

okorshenko:
Hi @scottsb 
If some singleton object will request non-shared object in the constructor, Object Manager will create the new instance of non-shared object. But if this singleton will be requested by multiple classes, singleton won't get a new non-shared object for each client. As a result, you will get the same behavior. Only for the child classes new non-shared object will be created. 
So it will be more reliable implementation if you will use suggested approach. https://github.com/magento/magento2/pull/10302#issuecomment-318424929

Thank you

okorshenko:
Closing this PR due to inactivity. @scottsb feel free to reopen this PR once ready. These changes will be useful. Thank you

PR #10306, opened 2017-07-24
Remove zend json from data object

dmanners:
### Description
The DataObject class has a method `toJson` that uses `Zend_Json`. Since this at EOF we should replace the usage. Unfortunately in this case that is not so "simple" after some discussion with @okorshenko it was decided to create a new static class for this single case. This should not be used in other places but for this case it is "ok".

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/DataObject.php
+++ b/lib/internal/Magento/Framework/DataObject.php
@@ -325,19 +325,21 @@ class DataObject implements \ArrayAccess
      * Convert object data to JSON
      *
      * @param array $keys array of required keys
-     * @return string
+     * @return bool|string
+     * @throws \InvalidArgumentException
      */
     public function toJson(array $keys = [])
     {
         $data = $this->toArray($keys);
-        return \Zend_Json::encode($data);
+        return \Magento\Framework\Serialize\JsonConverter::convert($data);
     }
 
     /**
      * The "__" style wrapper for toJson
      *
-     * @param  array $keys
-     * @return string
+     * @param array $keys
+     * @return bool|string
+     * @throws \InvalidArgumentException
      */
     public function convertToJson(array $keys = [])
     {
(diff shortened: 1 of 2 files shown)


ihor-sviziev:
@dmanners Why do we need new object with static method? We could add to constructor optional parameter (with fallback to object manager)

dmanners:
@ihor-sviziev I am more than happy to find another option but unfortunately going the route of adding an optional dependency causes big breaks in the test suite. I am not sure exactly how the object manager changes in the test suite so there could be an option to mock the fallback during the test suite.

orlangur:
Why do we need to add new behavior with `@throws \InvalidArgumentException` instead of just calling `json_encode`?

What do you think about deprecating `toJson` method also to decrease amount of `DataObject` class responsibilities?

okorshenko:
Hi @orlangur removing the method from DataObject interface will be BiC that we would like to avoid. The standard approach for adding new dependencies to DataObject does not work well due to modularity. So we suggested the way that @dmanners implemented.

dmanners:
@orlangur normally I would like to keep away from the direct calls to `json_encode` but in the case of this class with the combination of the deprecation of the `toJson` I think I could live with myself. This class itself needs a lot of love or at least the usages of this class need some love and care so highlighting that is not a bad thing in my books. What do you think @okorshenko

orlangur:
@okorshenko I did not say a word regarding method removal, just deprecation. As to me static class introduction with comment to use it in only one place is much uglier than simply using `json_encode` in this place (the only difference compared to serializer https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/Serialize/Serializer/Json.php#L24 is not throwing an exception which corresponds to old method behavior).

korostii:
Hi @okorshenko
As @orlangur says, there's nothing wrong with both keeping the method present and avoiding new dependencies.
However, what's the point of using `\Magento\Framework\Serialize\JsonConverter` instead?

After all, the new `@throws` declaration a [BiC change](http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#modifying-the-types-of-thrown-exceptions) by itself. 
How is it any better than using `json_encode` directly?

PR #10318, opened 2017-07-25
Remove not used `getItemsData` functionality in Checkout

jalogut:
These `Quote Items` are never used, so there is no need to load them. Removing this code will same some resources avoiding loading quote items and its attributes that will never be used. 

`Quote Items` used in checkout are all coming from `totalsData` as shown here:

* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/Model/DefaultConfigProvider.php#L279
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/Model/DefaultConfigProvider.php#L571
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/view/summary/cart-items.js#L27
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/model/totals.js#L15
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/model/quote.js#L36

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/model/quote.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/model/quote.js
@@ -57,13 +57,6 @@ define([
         },
 
         /**
-         * @return {*}
-         */
-        getItems: function () {
-            return window.checkoutConfig.quoteItemData;
-        },
-
-        /**
          *
          * @return {*}
          */
--- a/app/code/Magento/Checkout/Model/DefaultConfigProvider.php
+++ b/app/code/Magento/Checkout/Model/DefaultConfigProvider.php
@@ -257,7 +257,6 @@ class DefaultConfigProvider implements ConfigProviderInterface
         $output['formKey'] = $this->formKey->getFormKey();
         $output['customerData'] = $this->getCustomerData();
         $output['quoteData'] = $this->getQuoteData();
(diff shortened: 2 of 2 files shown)


app/code/Magento/Checkout/view/frontend/web/js/model/quote.js
```diff
@@ -57,13 +57,6 @@ define([
         },
 
         /**
-         * @return {*}
-         */
-        getItems: function () {
```
ishakhsuvarov, line 62: According to the [Backwards Compatible Development Guide](http://devdocs.magento.com/guides/v2.1/ext-best-practices/extension-coding/backwards-compatible-development/index.html#js) this method can not be removed, since it may break some existing extension relying on it or the data it provides.

ishakhsuvarov:
@jalogut While this code may look like unused – it still may be used by third party extensions or customizations, so we may not remove it right away. The most you can do it mark it as "deprecated" and add a "see" annotation with the info about what should be used instead.

Unfortunately, we can not accept this Pull Request in it's current state. Please let us know if you have anything to add or close if you agree.
Thanks!

jalogut:
Hi @ishakhsuvarov you are completely right. I just created a new PR

* https://github.com/magento/magento2/pull/10327

ishakhsuvarov:
@jalogut Thanks

PR #10320, opened 2017-07-25
Remove zend json from the test suite

dmanners:
### Description
Since Zend_Json is at EOL we should replace the usage of `\Magento\Framework\Serialize\Serializer\Json` when required. In this PR I have removed the usage of `Zend_Json` from the test suite.

### Manual testing scenarios

1. Running the test suite causes no failures

### Contribution checklist

### DIFF
--- a/dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
+++ b/dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
@@ -268,14 +268,10 @@ abstract class AbstractController extends \PHPUnit_Framework_TestCase
     {
         /** @var $cookieManager CookieManagerInterface */
         $cookieManager = $this->_objectManager->get(CookieManagerInterface::class);
-        try {
-            $messages = \Zend_Json::decode(
-                $cookieManager->getCookie(MessagePlugin::MESSAGES_COOKIES_NAME, \Zend_Json::encode([]))
-            );
-            if (!is_array($messages)) {
-                $messages = [];
-            }
-        } catch (\Zend_Json_Exception $e) {
+        $messages = json_decode(
+            $cookieManager->getCookie(MessagePlugin::MESSAGES_COOKIES_NAME, json_encode([]))
+        );
+        if (!is_array($messages)) {
             $messages = [];
         }
 
--- a/dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
+++ b/dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
@@ -15,6 +15,7 @@ use Magento\Theme\Controller\Result\MessagePlugin;
 
 /**
  * @SuppressWarnings(PHPMD.NumberOfChildren)
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 abstract class AbstractController extends \PHPUnit_Framework_TestCase
 {
--- a/dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
+++ b/dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
@@ -268,10 +268,21 @@ abstract class AbstractController extends \PHPUnit_Framework_TestCase
     {
         /** @var $cookieManager CookieManagerInterface */
         $cookieManager = $this->_objectManager->get(CookieManagerInterface::class);
-        $messages = json_decode(
-            $cookieManager->getCookie(MessagePlugin::MESSAGES_COOKIES_NAME, json_encode([]))
-        );
-        if (!is_array($messages)) {
+
+        /** @var $jsonSerializer \Magento\Framework\Serialize\Serializer\Json */
+        $jsonSerializer = $this->_objectManager->get(\Magento\Framework\Serialize\Serializer\Json::class);
+        try {
+            $messages = $jsonSerializer->unserialize(
+                $cookieManager->getCookie(
+                    MessagePlugin::MESSAGES_COOKIES_NAME,
+                    $jsonSerializer->serialize([])
+                )
+            );
+
+            if (!is_array($messages)) {
+                $messages = [];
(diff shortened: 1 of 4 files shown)


dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
```diff
@@ -268,14 +269,21 @@ protected function getCookieMessages($messageType = null)
     {
         /** @var $cookieManager CookieManagerInterface */
         $cookieManager = $this->_objectManager->get(CookieManagerInterface::class);
+
+        /** @var $jsonSerializer \Magento\Framework\Serialize\Serializer\Json */
+        $jsonSerializer = $this->_objectManager->get(\Magento\Framework\Serialize\Serializer\Json::class);
```
dmanners, line 273: Now I was not sure if I should use the class `\Magento\Framework\Serialize\Serializer\Json` here or just `json_*` but I thought since this was using the cookie interface it should also use the serializer. Happy to make a change if it is thought best to do something else.

dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
```diff
@@ -15,6 +15,7 @@
 
 /**
  * @SuppressWarnings(PHPMD.NumberOfChildren)
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
```
ishakhsuvarov: Adding SuppressWarnings, specially for Coupling, is generally considered a bad practice.
Did you add this as a result of tests execution? As it looks like that you have added same amount of dependencies as removed. 
Usually in such cases refactoring has to be done to decouple objects

dmanners: @ishakhsuvarov I added this as the static test limit got hit because of adding in https://github.com/magento/magento2/pull/10320/files#diff-22fb93f06b955235735842cbdd6c2ab3R274 which I am still not sure about since this is inside the test framework itself.

ishakhsuvarov: But at the same time you have decoupled this class from `\Zend_Json` and `\Zend_Json_Exception`. So to me it looks like that same coupling value should be in place after your changes. 
Am I missing something?

dmanners: I am not sure myself. https://travis-ci.org/dmanners/magento2/jobs/257261073 that was the failure. I wonder if PHPMD runs from a diff of classes changed?

ishakhsuvarov: Yes, it is executed from a diff, looks like it was already broken before. 
I will look into the possible ways to reduce coupling for this class.

dmanners: Thanks, let me know if I can help out here.

ishakhsuvarov: I've added some changes, which reduced coupling, in a way, which seems reasonable to me. Going to proceed with the merge now. Please let me know if it's ok from your perspective.

dmanners: Yeah those changes seem good to me still not 100% sure why that change fixes the static check, but I can live with not knowing that currently 👍

dev/tests/integration/framework/tests/unit/testsuite/Magento/Test/TestCase/ControllerAbstractTest.php
```diff
@@ -25,12 +25,25 @@ class ControllerAbstractTest extends \Magento\TestFramework\TestCase\AbstractCon
...
     protected function setUp()
     {
         $testObjectManager = new \Magento\Framework\TestFramework\Unit\Helper\ObjectManager($this);
 
         $this->messageManager = $this->getMock(\Magento\Framework\Message\Manager::class, [], [], '', false);
         $this->cookieManagerMock = $this->getMock(CookieManagerInterface::class, [], [], '', false);
+        $this->serializerMock = $this->getMockBuilder(\Magento\Framework\Serialize\Serializer\Json::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $this->serializerMock->expects($this->any())->method('unserialize')->willReturnCallback(
```
dmanners, line 42: Oh I love this approach for mocking methods

ishakhsuvarov: I tend to think that using real dependency in this case would be a better idea, however we usually stick to mocking all the stuff in the unit tests.

dmanners: yeah when it comes to mocking or not mocking I am not so sure what is "best" but the callback here is a nice idea specifically for this mock.

dev/tests/integration/framework/Magento/TestFramework/TestCase/AbstractController.php
orlangur: As we suppress exception anyway here I would simplify implementation and got rid of try/catch at all. But as it is just testing code it does not really matter.

dev/tests/integration/framework/tests/unit/testsuite/Magento/Test/TestCase/ControllerAbstractTest.php
orlangur: Such "mock" is practically equivalent to a real class 😉 I prefer tiny tests with small fixtures having mocks in form of returning plain values. Copy-pasting such callbacks to dozens of tests somewhat increases maintenance efforts.

orlangur (approved):

PR #10322, opened 2017-07-25
Fix Broken DHL Shipping Estimate

gmroughton:
I may be doing something wrong, but it seems like after installing the DHL module and configuring the settings to connect our DHL account we were no longer able to get any shipping quotes on sale orders.  The root of the problem is that $request->getDestStreet returns an array for both Primary Address Fields.  In the existing code, the substr function fails when passing info to incov_substr (see error below) and the second function never returns a value.  I have changed it to index the array properly and only use the second value if it is set.

Error Given:
Warning: iconv_substr() expects parameter 1 to be string, array given in /var/www/magento2/vendor/magento/framework/Stdlib/StringUtils.php on line 183

2. ...

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10322) <br/>All committers have signed the CLA.

YevSent:
Hi, @gmroughton, please provide more details, maybe you are using some additional extensions for checkout or shipping. As I understand you got this issue on the develop branch but I've checked the behavior of `\Magento\Quote\Model\Quote\Address\RateRequest::getDestStreet` and it returns a string, not an array. Even if the address contains few streets they will be returned as a single string separated by `\n`.

ishakhsuvarov:
Hi @gmroughton 
Do you need any assistance with this PR? Let us know if you have questions.
Thanks

ishakhsuvarov:
@gmroughton Closing this PR for now due to inactivity. If you would like to continue please reopen and provide your opinion on the comment above.
Thank you!

PR #10327, opened 2017-07-26
Deprecate not used `getItemsData` functionality in Checkout

jalogut:
These `Quote Items` are never used, so there is no need to load them. Removing this code will same some resources avoiding loading quote items and its attributes that will never be used. 

`Quote Items` used in checkout are all coming from `totalsData` as shown here:

* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/Model/DefaultConfigProvider.php#L279
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/Model/DefaultConfigProvider.php#L571
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/view/summary/cart-items.js#L27
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/model/totals.js#L15
* https://github.com/magento/magento2/blob/develop/app/code/Magento/Checkout/view/frontend/web/js/model/quote.js#L36

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/model/quote.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/model/quote.js
@@ -77,7 +77,7 @@ define([
          * @return {*}
          */
         getItems: function () {
-            return window.checkoutConfig.quoteItemData;
+            return totals.getItems();
         },
 
         /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/Checkout/view/frontend/web/js/model/quote.js
```diff
@@ -77,7 +77,7 @@ define([
          * @return {*}
          */
         getItems: function () {
-            return window.checkoutConfig.quoteItemData;
+            return totals.getItems();
```
ishakhsuvarov, line 80: This seems to be a correct solution, however did you make sure that data returned from this method is exactly the same in all cases?

jalogut:
Hi @ishakhsuvarov , the data returned is not the same. Anyway, let's forget about this pull request. I ended up using `window.checkoutConfig.quoteItemData` because of inconsistent changing data of `totals.getItems()` during the checkout steps. 

I close the PR because after all this method will be needed by people extending the checkout.

ishakhsuvarov:
@jalogut Thank you for collaboration anyways!

Issue #3640, opened 2016-03-03
CartItemInterface cannot add extension_attributes


degaray:
I am trying to extend CartItemInterface with custom attributes. It goes well, in most parts, but it crashes when calling `/carts/mine/payment-information`. This is the error I get:

```
  "message": "Recoverable Error: Argument 1 passed to Magento\Quote\Model\Cart\Totals\Item::setExtensionAttributes() must be an instance of Magento\Quote\Api\Data\TotalsItemExtensionInterface, instance of Magento\Quote\Api\Data\CartItemExtension given, called in /vagrant/data/magento2/lib/internal/Magento/Framework/Api/DataObjectHelper.php on line 125 and defined in /vagrant/data/magento2/app/code/Magento/Quote/Model/Cart/Totals/Item.php on line 470",
  "trace": "#0 /vagrant/data/magento2/app/code/Magento/Quote/Model/Cart/Totals/Item.php(470): 

```

mitul69:
hey @degaray 

Have you added dependency in your di.xml file?

ishakhsuvarov:
@degaray Thank you for reporting. We already have a similar item in our backlog (internal ticket `MAGETWO-50025`) to investigate this.

degaray:
Thanks. Do you have a workaround?

ishakhsuvarov:
@degaray I'm afraid we don't have one currently. We will update this ticket as soon as some more information is available.
Thank you.

degaray:
Ok, here is what I have found so far. If you create an `extension_attribute` for `Magento\Quote\Api\Data\CartItemInterface` it works well. However when you call `/carts/mine/payment-information` Magento tries to convert the CartItemInterface into a `Magento\Quote\Model\Cart\Totals\TotalsItemInterface`. The problem is that at certain point in the conversion specifically in `Magento\Framework\Api\DataObjectHelper::_setDataValues (125)` it tries to run TotalItem's setExtensionAttribute, and it tries to set it with CartItem's `extension_attributes`. Since CartItem's `extension_attributes` has a different type than TotalItem's it gives the exception.
My best guess is you should convert the `extension_attributes` type in `Magento\Quote\Model\Cart\CartTotalRepository::get` before populateWithArray.

ishakhsuvarov:
@degaray Closing this ticket, as the fix seems to be delivered with the linked commit. Please feel free to reopen if you have any additional questions.

magento-team:
[...]

PR #10328, opened 2017-07-26
Fix missing extension_attributes on checkout item

jalogut:
Do not remove extension attributes when converting the quote Item to dataObject. Otherwise it is not possible to display custom attributes on checkout.

### DIFF
--- a/app/code/Magento/Quote/Model/Cart/Totals/ItemConverter.php
+++ b/app/code/Magento/Quote/Model/Cart/Totals/ItemConverter.php
@@ -79,7 +79,6 @@ class ItemConverter
         $this->eventManager->dispatch('items_additional_data', ['item' => $item]);
         $items = $item->toArray();
         $items['options'] = $this->getFormattedOptionValue($item);
-        unset($items[ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY]);
 
         $itemsData = $this->totalsItemFactory->create();
         $this->dataObjectHelper->populateWithArray(


app/code/Magento/Quote/Model/Cart/Totals/ItemConverter.php
```diff
@@ -79,7 +79,6 @@ public function modelToDataObject($item)
         $this->eventManager->dispatch('items_additional_data', ['item' => $item]);
         $items = $item->toArray();
         $items['options'] = $this->getFormattedOptionValue($item);
-        unset($items[ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY]);
```
ishakhsuvarov, line 82: These are removed on purpose, as it causes a fatal error due to types mismatch on checkout when extension attribute is populated. There is more detailed explanation of what is happening [in this comment to issue 3640](https://github.com/magento/magento2/issues/3640#issuecomment-192291078).
We are currently looking into a solution which may be safely used in this case. I will update the ticket when we come up with one.

ishakhsuvarov:
@jalogut Unfortunately, we can not accept this PR for now, since this change will cause fatal errors in some scenarios. You may also refer to the Travis CI [build](https://travis-ci.org/magento/magento2/builds/257634328) for this Pull Request and observe this error.

We are going to investigate this issue additionally and looks for possible solution, which would satisfy all cases. For now I would suggest same method as mentioned by @antonkril on [twitter](https://twitter.com/AntonKril/status/889907764939685890): add a plugin on `ItemConverter`

Thank you for collaboration.

ihor-sviziev:
@ishakhsuvarov I think it would be good idea to add some comment there why it was done in this way and how we can make workaround. I think this issue already had not only @jalogut and it will happen for another people. What do you think?

ishakhsuvarov:
@ihor-sviziev Thank you for the feedback. We will consider adding this comment.

PR #8331, opened 2017-01-30
Remove Zend1 Json from Magento Captcha module

dmanners:
Zend Framework 1 has been at end of life since middle of 2016. Zend Framework 2 comes with a json module that works in the same way as the Zend Framework 1 version.

In this pull request I have updated the captcha module to use the zend framework 2 versions of Zend_Json. I have also updated the code to use json decode via object since that is the new default in Zend Framework 2.

PR #10335, opened 2017-07-26
Admin Config Search

maximgubar:
### Description
This feature will provide a possibility to find the config fields in Magento in a fast and easy way (e.g. live search).

### Manual testing scenarios
1. Open Magento admin in your browser.
2. Type the keyword in global search input field.
3. If there is a match in the config elements - results should appear in 'autocomplete results' area.
4. Each result is clickable and should redirect to you to proper section/group of configuration UI.

### Contribution checklist


app/code/Magento/Backend/Model/Search/Config.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->setResults($this->resultBuilder->getAll());
+        return $this;
+    }
+
+    /**
+     * @param ElementIterator $structureElementIterator
+     * @param string $needle
+     * @param string $pathLabel
+     */
+    public function findInStructure(ElementIterator $structureElementIterator, $needle, $pathLabel = '')
```
maghamed: why does this method declared as public? 
We use it just inside this class

app/code/Magento/Backend/Model/Search/Config.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        foreach ($structureElementIterator as $structureElement) {
+            if (!($structureElement instanceof ElementInterface)) {
+                continue;
+            }
```
maghamed: why do we have this check? 

Magento\Config\Model\Config\Structure\Element\Iterator returns  \Magento\Config\Model\Config\Structure\ElementInterface by contract, do we have any reasons not to trust the declared contract and make additional check? 
```php
    /**
     * Return the current element
     *
     * @return \Magento\Config\Model\Config\Structure\ElementInterface
     */
    public function current()
    {
        return $this->_flyweight;
    }
```

app/code/Magento/Backend/Model/Search/Config.php
maghamed: don't think "needle" is a good name.
Unless we call $structureElementIterator as haystack

It's the user's search term, so let's call it this way

app/code/Magento/Backend/Model/Search/Config.php
```diff
@@ -0,0 +1,81 @@
...
+            if (stripos((string)$structureElement->getLabel(), $needle) !== false) {
```
maghamed: it's better to use 
mb_stripos than stripos
Magento requires        ` "ext-mbstring": "*",` to be installed

app/code/Magento/Backend/Model/Search/Config.php
```diff
@@ -0,0 +1,81 @@
...
+     * Load search results
+     *
+     * @return $this
+     */
+    public function load()
+    {
+        if (!$this->hasQuery()) {
+            $this->setResults($this->resultBuilder->getAll());
+            return $this;
+        }
```
maghamed: if we will move a check for the empty query into the findInStructure function, we could end-up with 
```php
    public function load()
     {
         $this->findInStructure($this->configStructure->getTabs(), $this->getQuery());
         $this->setResults($this->resultBuilder->getAll());
         return $this;
     }
```

app/code/Magento/Backend/Model/Search/Config/Result/Builder.php
```diff
@@ -0,0 +1,97 @@
...
+    const STRUCTURE_ELEMENT_TYPE_FIELD   = 'field';
+
+    /**
+     * @var array
+     */
+    private $supportedElementTypes = [
+        self::STRUCTURE_ELEMENT_TYPE_SECTION,
+        self::STRUCTURE_ELEMENT_TYPE_GROUP,
+        self::STRUCTURE_ELEMENT_TYPE_FIELD,
+    ];
```
maghamed: it's better not to hard code supported elements, but accept an additional parameter, which is configured through the DI.

So, if some new supported component would be added in future we could easily add this new type with DI configuration.

app/code/Magento/Backend/Model/Search/Config/Result/Builder.php
```diff
@@ -0,0 +1,97 @@
...
+                    'section' => $elementPathParts[0],
+                    'group'   => $elementPathParts[1],
+                ];
+                break;
+            case self::STRUCTURE_ELEMENT_TYPE_FIELD:
+                $urlParams = [
+                    'section' => $elementPathParts[0],
+                    'group'   => $elementPathParts[1],
+                    'field'   => $structureElement->getId(),
+                ];
```
maghamed: it's better to introduce an interface and provide implementation for each of the type.
To have an ability to add new supported type in future. 
For now, we hardcoded existing structure elements

app/code/Magento/Config/Block/System/Config/Edit.php
```diff
@@ -41,16 +44,24 @@ class Edit extends \Magento\Backend\Block\Widget
...
      * @param \Magento\Config\Model\Config\Structure $configStructure
      * @param array $data
+     * @param Json|null $jsonSerializer
      */
     public function __construct(
         \Magento\Backend\Block\Template\Context $context,
         \Magento\Config\Model\Config\Structure $configStructure,
-        array $data = []
+        array $data = [],
+        Json $jsonSerializer = null
```
maghamed, line 62: it's better to have a dependency on the interface instead?
`Magento\Framework\Serialize\SerializerInterface`

maghamed (changes requested):

app/code/Magento/Config/Block/System/Config/Edit.php
korostii: As mentioned on https://github.com/magento/magento2/pull/8331#discussion_r99838100 and referenced on [#10342 (comment)](https://github.com/magento/magento2/pull/10342/files#r129688126) , `SerializerInterface` is not guaranteed to return a _JSON_ serializer in the future. And `\Magento\Config\Block\System\Config` uses it in a [method explicitly named `getConfigSearchParamsJson`](https://github.com/liip-forks/magento2/blob/94b423bf985dbd210a5eb13a43ff4dd92a8cc3ca/app/code/Magento/Config/Block/System/Config/Edit.php#L124)

app/code/Magento/Config/Model/Config/Structure/ElementNewInterface.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Config\Model\Config\Structure;
+
+/**
+ * @api
+ */
+interface ElementNewInterface extends ElementInterface
```
maghamed: Don't think that naming is very good in this case ElementInterface -> ElementNewInterface

app/code/Magento/Backend/Model/Search/Config/Result/Builder.php
```diff
@@ -0,0 +1,81 @@
...
+use Magento\Config\Model\Config\Structure\ElementNewInterface;
+
+/**
+ * Config SearchResult Builder
+ */
+class Builder
+{
+    const STRUCTURE_ELEMENT_TYPE_SECTION = 'section';
+    const STRUCTURE_ELEMENT_TYPE_GROUP   = 'group';
+    const STRUCTURE_ELEMENT_TYPE_FIELD   = 'field';
```
maghamed: why do we need these constants if they not used

app/code/Magento/Backend/Model/Search/Config.php
```diff
@@ -0,0 +1,76 @@
...
+use Magento\Config\Model\Config\Structure;
+use Magento\Config\Model\Config\Structure\Element\AbstractComposite;
+use Magento\Config\Model\Config\Structure\Element\Iterator as ElementIterator;
+
+/**
+ * Search Config Model
+ *
+ * @method string|null getQuery()
+ * @method bool hasQuery()
+ * @method Config setResults(array $results)
```
maghamed: it's better to have all these methods as a part of Object contract, not relying on `__call` usage

app/code/Magento/Backend/Model/Search/Config.php
```diff
@@ -0,0 +1,78 @@
...
+use Magento\Config\Model\Config\Structure\Element\Iterator as ElementIterator;
+
+/**
+ * Search Config Model
+ *
+ * @method Config setQuery(string $query)
+ * @method string|null getQuery()
+ * @method bool hasQuery()
+ * @method Config setResults(array $results)
+ * @method array getResults()
```
maghamed: it's better to have these methods in class contract, not relying on `__call` method

app/code/Magento/Config/Model/Config/Structure/ElementNewInterface.php
maximgubar: I'll move it up and name it StructureElementInterface, wdyt ?

maghamed (approved):

ishakhsuvarov:
Hi @maximgubar 
Please check the `Magento\Backend\Model\Search\ConfigTest` failures. Looks like search result in the test is always empty. Let us know if you need any assistance.
Thank you.

ishakhsuvarov:
@maximgubar Thank you for the update

vrann:
@maximgubar 

- [x] add description to the pull request
- [x] please fix static test:
```
Magento\Test\Php\XssPhtmlTemplateTest::testXssSensitiveOutput
Passed: 1172, Failed: 1, Incomplete: 0, Skipped: 0.
Data set: /opt/bamboo/agents/agent1/xml-data/build-dir/M2-AT2472-SLE1/magento2/app/code/Magento/Config/view/adminhtml/templates/system/config/edit.phtml
Potentially XSS vulnerability. Please verify that output is escaped at lines 391
Failed asserting that a string is empty.
/opt/bamboo/agents/agent1/xml-data/build-dir/M2-AT2472-SLE1/magento2/dev/tests/static/testsuite/Magento/Test/Php/XssPhtmlTemplateTest.php:48
/opt/bamboo/agents/agent1/xml-data/build-dir/M2-AT2472-SLE1/magento2/lib/internal/Magento/Framework/App/Utility/AggregateInvoker.php:56
/opt/bamboo/agents/agent1/xml-data/build-dir/M2-AT2472-SLE1/magento2/dev/tests/static/testsuite/Magento/Test/Php/XssPhtmlTemplateTest.php:51
(4 more lines...)
```

vrann:
added changes per UX request:
1. `.highlighted` should be added to tr around the input
2. `.highlighted` color should be #DFF7FF

PR #10340, opened 2017-07-26
Remove zend json from package info

dmanners:
### Description
Replace Zend_Json usage in Framework/Module/PackageInfo.php  …
 - inject the new \Magento\Framework\Serialize\Serializer\Json class,
 - update the load method to use the Serializer,
 - update the load method to catch the InvalidArgumentException exception

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Module/PackageInfo.php
+++ b/lib/internal/Magento/Framework/Module/PackageInfo.php
@@ -64,10 +64,10 @@ class PackageInfo
     private $serializer;
 
     /**
-     * Constructor
-     *
      * @param Dir\Reader $reader
      * @param ComponentRegistrar $componentRegistrar
+     * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @throws \RuntimeException
      */
     public function __construct(
         Dir\Reader $reader,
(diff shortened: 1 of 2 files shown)


lib/internal/Magento/Framework/Module/PackageInfo.php
```diff
@@ -85,9 +95,9 @@ private function load()
                 $key = $moduleDir . '/composer.json';
                 if (isset($jsonData[$key]) && $jsonData[$key]) {
                     try {
-                        $packageData = \Zend_Json::decode($jsonData[$key]);
-                    } catch (\Zend_Json_Exception $e) {
-                        throw new \Zend_Json_Exception(
+                        $packageData = $this->serializer->unserialize($jsonData[$key]);
+                    } catch (\InvalidArgumentException $e) {
```
dmanners, line 99: I am not a big fan of catching this and then throwing the same exception. I was thinking about adding a new exception for this case here. What are people's thoughts on this?

ishakhsuvarov: Do you mean specific exception for Json or Serialization?
I assume this specific fix is fine, since there is no major change.

PR #10341, opened 2017-07-26
Remove zend json from migration

dmanners:
### Description
Remove the Zend_Json call from the Setup/Migration.php  …
 - inject the new \Magento\Framework\Serialize\Serializer\Json class,
 - update _jsonDecode method parameter default from Zend_Json constant,
 - update _jsonDecode to call the new Serializer method version

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Module/Setup/Migration.php
+++ b/lib/internal/Magento/Framework/Module/Setup/Migration.php
@@ -133,6 +133,11 @@ class Migration
     private $setup;
 
     /**
+     * @var \Magento\Framework\Serialize\Serializer\Json
+     */
+    private $serializer;
+
+    /**
      * @param ModuleDataSetupInterface $setup
      * @param Filesystem $filesystem
      * @param MigrationData $migrationData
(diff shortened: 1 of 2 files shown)


lib/internal/Magento/Framework/Module/Setup/Migration.php
```diff
@@ -694,10 +704,12 @@ public function getCompositeModules()
      *
      * @param string $encodedValue
      * @param int $objectDecodeType
-     * @return mixed
+     * @return array|bool|float|int|mixed|null|string
```
dmanners, line 707: Again not happy that the return and the throws have changed been added but not really sure if this is ok/if there is a better option in this case.

ishakhsuvarov: I assume same return type as `SerializerInterface` suggests may be used
Since it's `Json` implementation currently has inheritdoc.

okorshenko: We will take care about this on our side.

PR #10342, opened 2017-07-26
Remove zend json from json controller

dmanners:
### Description
Removing the usage of `Zend_Json` in the Json Controller. Deprecating the method `setData` to encourage the use of the more type specific methods in this class after talking through options with @okorshenko 

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Controller/Result/Json.php
+++ b/lib/internal/Magento/Framework/Controller/Result/Json.php
@@ -59,7 +59,11 @@ class Json extends AbstractResult
      */
     public function setData($data, $cycleCheck = false, $options = [])
     {
-        $this->json = $this->serializer->serialize($data);
+        if (is_object($data) && method_exists($data, 'toJson')) {
+            $this->json = $data->toJson();
+        } else {
+            $this->json = $this->serializer->serialize($data);
+        }
         return $this;
     }
 
--- a/lib/internal/Magento/Framework/Controller/Result/Json.php
+++ b/lib/internal/Magento/Framework/Controller/Result/Json.php
@@ -29,11 +29,22 @@ class Json extends AbstractResult
     protected $json;
 
     /**
-     * @param \Magento\Framework\Translate\InlineInterface $translateInline
+     * @var \Magento\Framework\Serialize\Serializer\Json
      */
-    public function __construct(InlineInterface $translateInline)
-    {
+    private $serializer;
+
+    /**
+     * @param InlineInterface $translateInline
+     * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @throws \RuntimeException
+     */
+    public function __construct(
+        InlineInterface $translateInline,
+        \Magento\Framework\Serialize\Serializer\Json $serializer = null
+    ) {
         $this->translateInline = $translateInline;
+        $this->serializer = $serializer ?: \Magento\Framework\App\ObjectManager::getInstance()
+            ->get(\Magento\Framework\Serialize\Serializer\Json::class);
     }
 
     /**
@@ -43,10 +54,11 @@ class Json extends AbstractResult
      * @param boolean $cycleCheck Optional; whether or not to check for object recursion; off by default
      * @param array $options Additional options used during encoding
      * @return $this
+     * @throws \InvalidArgumentException
      */
     public function setData($data, $cycleCheck = false, $options = [])
     {
-        $this->json = \Zend_Json::encode($data, $cycleCheck, $options);
+        $this->json = $this->serializer->serialize($data);
         return $this;
     }
 
--- a/lib/internal/Magento/Framework/Controller/Result/Json.php
+++ b/lib/internal/Magento/Framework/Controller/Result/Json.php
@@ -55,6 +55,7 @@ class Json extends AbstractResult
      * @param array $options Additional options used during encoding
      * @return $this
      * @throws \InvalidArgumentException
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      */
     public function setData($data, $cycleCheck = false, $options = [])
     {
--- a/lib/internal/Magento/Framework/Controller/Result/Json.php
+++ b/lib/internal/Magento/Framework/Controller/Result/Json.php
@@ -50,20 +50,45 @@ class Json extends AbstractResult
     /**
      * Set json data
      *
-     * @param mixed $data
-     * @param boolean $cycleCheck Optional; whether or not to check for object recursion; off by default
-     * @param array $options Additional options used during encoding
-     * @return $this
+     * @param array|string|\Magento\Framework\DataObject $data
+     * @param bool $cycleCheck
+     * @param array $options
+     * @return Json
      * @throws \InvalidArgumentException
+     * @throws \Magento\Framework\Exception\LocalizedException
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     * @deprecated
      */
     public function setData($data, $cycleCheck = false, $options = [])
     {
-        if (is_object($data) && method_exists($data, 'toJson')) {
-            $this->json = $data->toJson();
-        } else {
-            $this->json = $this->serializer->serialize($data);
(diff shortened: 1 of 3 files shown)


dmanners:
This PR I expect to come up with plenty of talk/opinions. I am happy to hear what people think would be other options if there are some.

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
+    private $serializer;
+
+    /**
+     * @param InlineInterface $translateInline
+     * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @throws \RuntimeException
+     */
+    public function __construct(
+        InlineInterface $translateInline,
+        \Magento\Framework\Serialize\Serializer\Json $serializer = null
```
orlangur, line 43: Why do we rely on particular implementation and not interface by the way?
https://github.com/magento/magento2/blob/develop/app/etc/di.xml#L162

dmanners: https://github.com/magento/magento2/pull/8331#issuecomment-278022525 was the original comment from the team on the matter,

orlangur: Thanks, I expected something like this. Not sure why such interface was introduced at all if we keep creating json-only code.

Needs to be mentioned somewhere to avoid confusion like https://github.com/magento/magento2/pull/10335#discussion_r129590760

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
+            new \Magento\Framework\Phrase('Invalid argument type')
+        );
+    }
+
+    /**
+     * @param array $data
+     * @return $this
+     * @throws \InvalidArgumentException
+     */
+    public function setArrayData(array $data)
```
ishakhsuvarov: Is there any reason for this method being public?

okorshenko:
@dmanners  This pull request has been reverted due to broken functionality. We need to review this change one more time. Thank you

dmanners:
@okorshenko sure. Do you have a specific use/test case that this broke for so that I can have something to review when thinking about possible options here.

Thanks

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
-     * @param array $options Additional options used during encoding
-     * @return $this
+     * @param array|string|\Magento\Framework\DataObject $data
+     * @param bool $cycleCheck
+     * @param array $options
+     * @return Json
+     * @throws \InvalidArgumentException
+     * @throws \Magento\Framework\Exception\LocalizedException
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     * @deprecated
```
orlangur, line 60: I disagree with deprecating this method. `setData` is for setting data which will be encoded and `setJsonData` is for setting raw data.

Two arguments which has no effect should be deprecated though.

Please change method description accordingly.

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
      */
     public function setData($data, $cycleCheck = false, $options = [])
     {
-        $this->json = \Zend_Json::encode($data, $cycleCheck, $options);
```
orlangur, line 49: This method is just setter, no need to call another setter, Magento is already slow enough.

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
+        if ($data instanceof \Magento\Framework\DataObject) {
```
orlangur, line 64: Why need this explicitly?

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
+            return $this->setArrayData($data->toArray());
+        }
+
+        if (is_array($data)) {
```
orlangur, line 68: Please remove this `if`.

lib/internal/Magento/Framework/Controller/Result/Json.php
```diff
@@ -29,24 +29,63 @@ class Json extends AbstractResult
...
+            return $this->setArrayData($data);
+        }
+
+        if (is_string($data)) {
```
orlangur, line 72: Passing string parameter did encode `$data` before, please remove this `if`.

lib/internal/Magento/Framework/Controller/Result/Json.php
orlangur, line 86: Please remove this method as it does not bring any value.

orlangur:
@dmanners I didn't do code review previous time but now did, please check my suggestions and prepare a new PR (merged PR cannot be reopened AFAIK).

lib/internal/Magento/Framework/Controller/Result/Json.php
ishakhsuvarov: This method was introduced to deprecate and remove `setData` method, which has too much responsibilities.

lib/internal/Magento/Framework/Controller/Result/Json.php
ishakhsuvarov: This is needed to maintain compatibility with the old implementation which took care of this inside.

lib/internal/Magento/Framework/Controller/Result/Json.php
ishakhsuvarov: Please clarify

lib/internal/Magento/Framework/Controller/Result/Json.php
ishakhsuvarov: Please clarify

lib/internal/Magento/Framework/Controller/Result/Json.php
orlangur: I see...

```
public static function encode($valueToEncode, $cycleCheck = false, $options = array())
    {
        if (is_object($valueToEncode)) {
            if (method_exists($valueToEncode, 'toJson')) {
                return $valueToEncode->toJson();
            } elseif (method_exists($valueToEncode, 'toArray')) {
                return self::encode($valueToEncode->toArray(), $cycleCheck, $options);
            }
        }
```

Why it is not maintained for any object then?

lib/internal/Magento/Framework/Controller/Result/Json.php
orlangur: `setData` should be as old one, thus `setData([])` just encodes array with no necessity for additional logic.

lib/internal/Magento/Framework/Controller/Result/Json.php
orlangur: `setData('someString')` was setting `json_encode('someString')` and not just `'someString'`.

lib/internal/Magento/Framework/Controller/Result/Json.php
orlangur: This is what I strongly disagree with. Its responsibility is quite clear - encode value and set it. There is no need to think which value you're passing.

`setData` - pass value and it will be encoded
`setJsonData` - pass string and it will be set as is

I do agree that `setData($object)` could be discouraged and corresponding `if` could be marked as deprecated for further removal and `setData` method simplification. But introduction of `setArrayData` method looks like a mess to me.

dmanners:
@orlangur thanks for the feedback, I appreciate the thoughts. I need to re-think these changes anyway so will take your ideas on board and see what I can put together.

PR #10354, opened 2017-07-27
Add missing block name to allow block customisation

ihor-sviziev:
### Description
Add ability to customize order renderers

### Related Issues (if relevant)
1. https://github.com/magento/magento2/pull/10352: Add missing block name to allow block customisation

### Contribution checklist

### DIFF
--- a/app/code/Magento/Bundle/view/frontend/layout/checkout_onepage_review_item_renderers.xml
+++ b/app/code/Magento/Bundle/view/frontend/layout/checkout_onepage_review_item_renderers.xml
@@ -8,7 +8,7 @@
 <page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
     <body>
         <referenceBlock name="checkout.onepage.review.item.renderers">
-            <block class="Magento\Bundle\Block\Checkout\Cart\Item\Renderer" as="bundle" template="Magento_Checkout::onepage/review/item.phtml"/>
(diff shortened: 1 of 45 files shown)


ishakhsuvarov:
@ihor-sviziev Did you automate this process in any way? Would be nice to fix all of these cases around the codebase.

ihor-sviziev:
No, I did them manually, for all sales XML files

ishakhsuvarov:
@ihor-sviziev We are currently looking into possibility to add names to all blocks where it is missing, presumably using some kind of automated tool, and enforcing this behavior with the static test to avoid such issues in the future. Do you have any interest in this?
Thanks!

maghamed:
Hey @ihor-sviziev 
The problem that for now we have 386 unnamed blocks in CE edition
and 180 in EE one.

You can see that applying next Regexp, for example in your PHP Storm
`<block(?=\s|>)(?!(?:[^>=]|=(['"])(?:(?!\1).)*\1)*?\sname=['"])[^>]*>`

What we want to do is to:
1. Give names for all the existing unnamed blocks, using basic pattern like 
    `parent.name + current_block.as`
   if the new name does not represent child block correctly - it could be changed to more appropriate one.

2. Ideally to make all these changes in an automated way (script) - to make it possible to apply changes on all Magento Editions at once.
It could be done in a different way: Pattern Matching, XPath etc.
The one which is more suitable for you

3. Provide a static test which will check that there are no Unnamed Blocks in Magento code base anymore after the changes applied.

The problem that accepting partial fixes don't make us closer to the final goal. Because someone who will work on this task later still needs to provide an implementation for all the steps mentioned above.

That's a general procedure we do in the core for such tasks.

Do you have some more time and will to invest into this issue?

ihor-sviziev:
@maghamed do we have any example for similar case? I can try to do that

PR #10370, opened 2017-07-28
Added a custom variable helper

enrico69:
Add a Variable Helper

### Description
In the module Magento/Variable, add a Helper to store various methods which could be needed by
developers. One method is currently given : _customVariableHasValue()_.

This method return a boolean. It checks if a given variable exists and contain the expected value.
It is possible to be more accurate: the type of content (HTML or full text) and to give a store.

### Manual testing scenarios
Quick test example.
In the back-office, go to System->Custom Variables. Create a new variable named 'foo' and set the full
text value to '2'.

Test your variable in a controller:

>         $variableModel = $this->_objectManager->get('Magento\Variable\Helper\Variable');
>         if ($variableModel->customVariableHasValue('foo', '2')) {
>             echo 'OK';
>         } else {
>             echo 'fail';
>         }

### Contribution checklist

### DIFF
--- a/app/code/Magento/Variable/Helper/Variable.php
+++ b/app/code/Magento/Variable/Helper/Variable.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Copyright © 2013-2017 Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Variable\Helper;
+
+use Magento\Store\Model\StoreManagerInterface;
+use Magento\Variable\Model\Variable as VariableModel;
+
+class Variable
+{
+    /**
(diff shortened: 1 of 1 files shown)


okorshenko:
Hi @enrico69 thank you for your contribution. In Magento2 project, we are not creating and not allowing new Helper classes. Helper in this case is anti-pattern. In some cases we may indtoduce services that are not used by core code directly, but provides some functionality for 3rd extenstions. But we are not adding such code to the modules.
Closing this PR.

enrico69:
@okorshenko OK. 
When you say 'is anti-pattern', then what would be the correct or more elegant way to factorize some code to easily do the same job?

What if I put it in the Variable Model?

PR #10374, opened 2017-07-29
Change Customers tab label to Best Customers, on dashboard

thanasisgalanis:
On the admin Dashboard below the sales chart, I think that the 4th tab according to its content "Best Customers" is more relevance label as "Customers".

Renaming this label also give us the ability for better and more relevance translation, as the "Customers" string used a lot in Magento code.

### DIFF
--- a/app/code/Magento/Backend/Block/Dashboard/Grids.php
+++ b/app/code/Magento/Backend/Block/Dashboard/Grids.php
@@ -77,7 +77,7 @@ class Grids extends \Magento\Backend\Block\Widget\Tabs
         $this->addTab(
             'customers',
             [
-                'label' => __('Customers'),
+                'label' => __('Best Customers'),
                 'url' => $this->getUrl('adminhtml/*/customersMost', ['_current' => true]),
                 'class' => 'ajax'
             ]


okorshenko:
Hi @sakisplus 
Thank you for your contribution. Unfortunately we can not accept this PR as is. Best Customers works for you, but other customers would like to have another labeling here. So we can not cover all cases here. If you are interested in this feature,  I would suggest make the labels configured via layout. In this case everyone will be able to cover their needs.

Closing this PR.

thanasisgalanis:
Hi @okobchenko 
Please allow me to slightly disagree with you and I explain the reason below.

With this logic the 3rd tab should be labeled just "Customers" rather than "New Customers" too! 

In reality the 3rd tab displays the latest customers who signed in (if I am not wrong), so according its content "New Customers" is a relevance label.

The 4th tab displays the highest volume customers! So it should be called "Top Customers", "Best Customers", "Super Customers" or "[ANY] Csutomers" you want, relevance to its content, but NOT just "Customers".

thanasisgalanis:
In addition, if the tab has a specific label to its content, it can easily cover any label need through translation files.

(example in en_US.csv)
"Top Customers","Top Customers"
"Top Customers","Best Customers"
"Top Customers","Super Customers"
"Top Customers","[ANY] Customers"
even
"Top Customers","Customers"

korostii:
Hi @okorshenko,
What if @sakiplus would accompany this change by adding a corresponding `"Best Customers","Customers"` line into [en_US.csv](https://github.com/sakisplus/magento2/blob/develop/app/code/Magento/Backend/i18n/en_US.csv)?

That way, current default behavior will be preserved, **and** at the same time, @sakiplus's target will be fulfilled as well. Would that be something you could merge in?

PR #10397, opened 2017-08-01
Feature/variable-value-exists

enrico69:
Taking into account the remarks about #10370.
The added code has been moved in to the Model.
Also some little PHP CS compliant corrections (variable names and fully qualified names in constructor).

### Description
In the module _Magento/Variable_, added a new method which could be needed by
developers in the _Variable_ Model: _customVariableHasValue()_.

This method return a boolean. It checks if a given variable exists and contain the expected value.
It is possible to be more accurate: the type of content (HTML or full text) and to give a store.

### Manual testing scenarios
Quick test example.
In the back-office, go to System->Custom Variables. Create a new variable named 'foo' and set the full
text value to '2'.

Test your variable in a controller:

>         $variableModel = $this->_objectManager->get('Magento\Variable\Model\Variable');
>         if ($variableModel->customVariableHasValue('foo', '2')) {
>             echo 'OK';
>         } else {
>             echo 'fail';
>         }

### Contribution checklist


app/code/Magento/Variable/Model/Variable.php
```diff
@@ -162,4 +179,39 @@ public function getVariablesOptionArray($withGroup = false)
...
+        $storeId = null
+    ) {
+        $status = false;
+
+        if ($storeId === null) {
+            $storeId = $this->storeManager->getStore()->getId();
+        }
+
+        $this->setStoreId($storeId)
+            ->loadByCode($variableCode);
```
orlangur, line 207: Since `EntityManager` introduction I don't think classes may be responsible for both entity and entity loading (not sure why `loadByCode` is still not deprecated, it should be according to my understanding).

app/code/Magento/Variable/Model/Variable.php
```diff
@@ -162,4 +179,39 @@ public function getVariablesOptionArray($withGroup = false)
...
+     * variable set with the given text value
+     *
+     * @param string $variableCode
+     * @param string $expectedValue
+     * @param string $type
+     * @param int $storeId
+     *
+     * @return boolean
+     */
+    public function customVariableHasValue(
```
orlangur, line 194: And then this method will be detected as unused code and simply removed...

Just FYI, not sure why PHPMD didn't work here, but any method returning `bool` should start with `is` or `has` (any method name should start with verb).

orlangur:
Could you provide a more detailed real-world scenario where such functionality would be useful?

How
```
$variableModel = $this->_objectManager->get('Magento\Variable\Model\Variable');
    if ($variableModel->customVariableHasValue('foo', '2')) {
        echo 'OK';
    } else {
        echo 'fail';
    }
```

is better than

```
$variableModel = $this->_objectManager->get('Magento\Variable\Model\Variable')->loadByCode('foo');
    if ($variableModel->getValue() === '2') {
        echo 'OK';
    } else {
        echo 'fail';
    }
```
?

app/code/Magento/Variable/Model/Variable.php
enrico69: I totally agree, and actually, the first thing I was looking for was a repository... Didn't find it so I went to the Model with the old Active-Record pattern.
Creating a repository for the _Variable_ entity could be a future task...

app/code/Magento/Variable/Model/Variable.php
enrico69: Indeed, lack of focus for me. Very strange that PHP Storm didn't rise a warning. I am going to fix it.

orlangur: No-no, this is no action point, just a FYI. Please read my last comment and let's find out whether we need this feature and where it should be placed.

enrico69:
@orlangur 

The idea is to get the same code more properly and less verbose.

In today coding standards the scenario would be : 

* In a class, I want to check if some variables exist and with a specific value
* Inject with DI a _Variable_ Model in the constructor (since we do not want to create new variable, I think there is no need to use a factory here) or create only one instance with the Object Manager
* Use a simple if ($variableModel->hasValue(...) )  each time I want to make a control
* The new method won't return anything else (null, object or whatever) than a boolean (mixed type to be avoided)

orlangur:
`$variableModel->customVariableHasValue('foo', '2')` is more verbose than `$variableModel->getValue() === '2'` and require knowledge about custom method...

Factory is the right way to go, here is an example from core: https://github.com/magento/magento2/blob/develop/app/code/Magento/Email/Model/Template/Filter.php#L790 Injecting variable model and reusing it with different codes is NOT the right way.

If I understood right, semantically `->hasValue` is more like `->equalsTo`...

> The new method won't return anything else (null, object or whatever) than a boolean (mixed type to be avoided)

It is not an issue as we always compare against concrete value. `$variable->getValue` is always returning string by the way.

To summarize, all described cases for now could be covered with vanilla core code without any issue. It is essential to add something only when some goal was not or hardly achievable to not make the code bloated.

Issue #188, opened 2013-01-13
Standardize Exception Text (with "." or without)


amenk:
Some exception messages are without a "." in the end, some have it.
So when printing an exception message, you can not be sure whether to add a "." or not.
This should be standardized.

Issue #174, opened 2012-12-26
Shopping cart rules with multishipping


ldusan84:
Currently in Magento if you create a shopping cart rule with condition that is satisfied with your current cart and if you checkout with multishipping with 2 addresses with items on those addresses that do not satisfy the condition separately, the rule is not applied. I think this is a bug since shipping should not affect the rule in my opinion and it should be fixed in Magento 2.

verklov:
Hi all, for feature requests, we will now be closing GitHub requests providing the ticket is created in the backlog.
We have created a ticket in our internal tracker for this improvement. We will review and handle it according to the priorities. Thank you for this request!
ID: MAGETWO-23682

piotrekkaminski:
Thank you for your submission.

We recently made some changes to the way we process GitHub submissions to more quickly identify and respond to core code issues.

Feature Requests and Improvements should now be submitted to the new [Magento 2 Feature Requests and Improvements](https://community.magento.com/t5/Magento-2-Feature-Requests-and/idb-p/feature-requests) forum (see details [here](https://community.magento.com/t5/News-Announcements/Improvements-to-GitHub-Management/m-p/44572#M96)).

We are closing this GitHub ticket and have moved your request to the new [forum](https://community.magento.com/t5/Magento-2-Feature-Requests-and/Shopping-cart-rules-with-multishipping/idi-p/46283).

PR #10403, opened 2017-08-02
#10402 fix theme templates overrides in environment emulation cross-area

alessandroniciforo:
Row 141 and 145 are obviously returning a theme id, so we'd better load the theme by id. I'm not sure if execution will fall into row 143 what to expect.

1. magento/magento2#10402: Theme templates overrides don't work in environment emulation cross-area

### DIFF
--- a/lib/internal/Magento/Framework/View/Asset/Repository.php
+++ b/lib/internal/Magento/Framework/View/Asset/Repository.php
@@ -148,7 +148,7 @@ class Repository
 
         if ($themeId) {
             $params['themeModel'] = $this->getThemeProvider()->getThemeById($themeId);
-            if (!$params['themeModel']) {
+            if (!$params['themeModel'] || $params['themeModel']->getArea() != $area) {
                 throw new \UnexpectedValueException("Could not find theme '$themeId' for area '$area'");
             }
         } elseif ($themePath) {
--- a/lib/internal/Magento/Framework/View/Asset/Repository.php
+++ b/lib/internal/Magento/Framework/View/Asset/Repository.php
@@ -142,7 +142,7 @@ class Repository
             $params['themeModel'] = $this->getThemeByPath($params['theme'], $area);
         } elseif (empty($params['themeModel']) && $area !== $this->getDefaultParameter('area')) {
             $themeId = $this->design->getConfigurationDesignTheme($area);
-            $params['themeModel'] = $this->getThemeById($themeId, $area);
+            $params['themeModel'] = $this->getTheme($themeId, $area);
         }
 
         if (empty($params['themeModel'])) {
@@ -159,6 +159,22 @@ class Repository
             $params['locale'] = $this->getDefaultParameter('locale');
         }
         return $this;
+    }
+
+    /**
+     * Get theme either by id or path based on whether the passed $theme is a numeric value or not
+     *
+     * @param string|int $theme
+     * @param $area
+     * @return \Magento\Framework\View\Design\ThemeInterface
+     */
+    private function getTheme($theme, $area)
+    {
+        if (is_numeric($theme)) {
+            return $this->getThemeById($theme, $area);
+        }
+
+        return $this->getThemeByPath($theme, $area);
     }
 
     /**
--- a/lib/internal/Magento/Framework/View/Asset/Repository.php
+++ b/lib/internal/Magento/Framework/View/Asset/Repository.php
@@ -135,28 +135,28 @@ class Repository
         }
 
         // Set themeModel
-        $themeId = null;
-        $themePath = null;
         $area = $params['area'];
         if (!empty($params['themeId'])) {
             $themeId = $params['themeId'];
-        } elseif (isset($params['theme'])) {
-            $themePath = $params['theme'];
-        } elseif (empty($params['themeModel']) && $area !== $this->getDefaultParameter('area')) {
-            $themeId = $this->design->getConfigurationDesignTheme($area);
-        }
-
-        if ($themeId) {
             $params['themeModel'] = $this->getThemeProvider()->getThemeById($themeId);
             if (!$params['themeModel'] || $params['themeModel']->getArea() != $area) {
                 throw new \UnexpectedValueException("Could not find theme '$themeId' for area '$area'");
             }
-        } elseif ($themePath) {
+        } elseif (isset($params['theme'])) {
+            $themePath = $params['theme'];
             $params['themeModel'] = $this->getThemeProvider()->getThemeByFullPath($area . '/' . $themePath);
             if (!$params['themeModel']) {
                 throw new \UnexpectedValueException("Could not find theme '$themePath' for area '$area'");
             }
-        } elseif (empty($params['themeModel'])) {
+        } elseif (empty($params['themeModel']) && $area !== $this->getDefaultParameter('area')) {
+            $themeId = $this->design->getConfigurationDesignTheme($area);
+            $params['themeModel'] = $this->getThemeProvider()->getThemeById($themeId);
+            if (!$params['themeModel'] || $params['themeModel']->getArea() != $area) {
+                throw new \UnexpectedValueException("Could not find theme '$themeId' for area '$area'");
+            }
+        }
+
+        if (empty($params['themeModel'])) {
             $params['themeModel'] = $this->getDefaultParameter('themeModel');
         }
 
--- a/lib/internal/Magento/Framework/View/Asset/Repository.php
+++ b/lib/internal/Magento/Framework/View/Asset/Repository.php
@@ -146,7 +146,7 @@ class Repository
         }
 
         if ($theme) {
-            $params['themeModel'] = $this->getThemeProvider()->getThemeByFullPath($area . '/' . $theme);
+            $params['themeModel'] = $this->getThemeProvider()->getThemeById($theme);
             if (!$params['themeModel']) {
                 throw new \UnexpectedValueException("Could not find theme '$theme' for area '$area'");
             }
--- a/lib/internal/Magento/Framework/View/Asset/Repository.php
+++ b/lib/internal/Magento/Framework/View/Asset/Repository.php
@@ -135,20 +135,26 @@ class Repository
         }
 
         // Set themeModel
-        $theme = null;
+        $themeId = null;
+        $themePath = null;
         $area = $params['area'];
         if (!empty($params['themeId'])) {
-            $theme = $params['themeId'];
+            $themeId = $params['themeId'];
         } elseif (isset($params['theme'])) {
-            $theme = $params['theme'];
+            $themePath = $params['theme'];
         } elseif (empty($params['themeModel']) && $area !== $this->getDefaultParameter('area')) {
-            $theme = $this->design->getConfigurationDesignTheme($area);
+            $themeId = $this->design->getConfigurationDesignTheme($area);
         }
 
-        if ($theme) {
-            $params['themeModel'] = $this->getThemeProvider()->getThemeById($theme);
+        if ($themeId) {
+            $params['themeModel'] = $this->getThemeProvider()->getThemeById($themeId);
             if (!$params['themeModel']) {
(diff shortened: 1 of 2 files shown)


orlangur:
> I'm not sure if execution will fall into row 143 what to expect.

In such case `$theme` will not contain `themeId` and suggested code will fail :) While original one worked...

orlangur:
Note that `magento:2.1-develop` target branch should be used for backports of pull requests already merged to `develop` branch only.

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -146,7 +146,7 @@ public function updateDesignParams(array &$params)
         }
 
         if ($theme) {
-            $params['themeModel'] = $this->getThemeProvider()->getThemeByFullPath($area . '/' . $theme);
+            $params['themeModel'] = $this->getThemeProvider()->getThemeById($theme);
```
orlangur: Please rewrite method in such a way that if we go through line 141 `getThemeById` call is used and `getThemeByFullPath` otherwise.

Also, such case with emulated cross-area needs to be covered with integration test.

orlangur (changes requested):

lib/internal/Magento/Framework/View/Asset/Repository.php
alessandroniciforo: Actually, also for line 145 `getThemeById` should be called as `getConfigurationDesignTheme` returns the id. I don't have time to write integration test at this moment, unfortunately.
Should I open another PR on develop branch?

orlangur: I changed base branch, please force push into the same branch applying your commit on top of `develop`.

Yeah, you are right about line 145.

Just finish your implementation without integration test then.

alessandroniciforo:
Pushed the suggested change and the fix to the unit test.

orlangur:
Please try to simplify modified method
```
1) Magento\Test\Php\LiveCodeTest::testCodeMess
PHP Code Mess has found error(s):
/home/travis/build/magento/magento2/lib/internal/Magento/Framework/View/Asset/Repository.php:130	The method updateDesignParams() has an NPath complexity of 240. The configured NPath complexity threshold is 200.
Failed asserting that 2 matches expected 0.
```

Maybe we can leave only one exception throwing and decrease amount of `if`s?

alessandroniciforo:
Just moving the exception throwing was not enough to go below 200. On top of that, I found that in case we load the theme by id we should ensure that the theme exists in the requested area (this is implicit in the load by full path), so I added that check and refactored a bit.
Hopefully it looks better now.

alessandroniciforo:
The issue unveiled by the integration tests run is that `Magento\Theme\Model\View::getConfigurationDesignTheme()` can either return:

- theme id: if execution pass into rows [#174](https://github.com/magento/magento2/blob/develop/app/code/Magento/Theme/Model/View/Design.php#L174) or [#179](https://github.com/magento/magento2/blob/develop/app/code/Magento/Theme/Model/View/Design.php#L179) AND if the design configuration has been saved on the backend, thus inserting the config line into `core_config_data`
- theme path: in all the other cases. There's a fallback at row [#188](https://github.com/magento/magento2/blob/develop/app/code/Magento/Theme/Model/View/Design.php#L188) so that at worst it takes the theme passed in its constructor, injected in its [di.xml](https://github.com/magento/magento2/blob/develop/app/code/Magento/Theme/etc/di.xml#L62), which is a theme path

So my (ugly) solution is that, in case we rely on `Magento\Theme\Model\View::getConfigurationDesignTheme()`, we should detect whether the returned value is an id or a path.

A more appropriate solution would be to modify the implementation of `getConfigurationDesignTheme` to have a mean of knowing whether the returned value is an id or a path, but I'm not confident enough to refactor that part too.

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -135,22 +135,17 @@ public function updateDesignParams(array &$params)
...
         }
 
-        if ($theme) {
-            $params['themeModel'] = $this->getThemeProvider()->getThemeByFullPath($area . '/' . $theme);
-            if (!$params['themeModel']) {
-                throw new \UnexpectedValueException("Could not find theme '$theme' for area '$area'");
-            }
-        } elseif (empty($params['themeModel'])) {
+        if (empty($params['themeModel'])) {
             $params['themeModel'] = $this->getDefaultParameter('themeModel');
```
orlangur: For better readability please do `$themeModel = ...` instead of `$params['themeModel'] = ...` and only in the very end do `$params['themeModel'] = $themeModel ?: $this->getDefaultParameter('themeModel')`.

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -167,6 +162,52 @@ public function updateDesignParams(array &$params)
...
     /**
+     * Get theme either by id or path based on whether the passed $theme is a numeric value or not
+     *
+     * @param string|int $theme
+     * @param $area
+     * @return \Magento\Framework\View\Design\ThemeInterface
+     */
+    private function getTheme($theme, $area)
+    {
+        if (is_numeric($theme)) {
```
orlangur, line 170: Yeah, looking into `\Magento\Theme\Model\View\Design::getConfigurationDesignTheme` this seems to be the best option to me (although we won't support string theme identifiers consisting of digits only).

Can we eliminate this method and still pass PHPMD? `getThemeById` and `getThemeByPath` seem much more understandable to me at the place of call but I would like to avoid method named like `getThemeByIdOrPath`.

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -167,6 +162,52 @@ public function updateDesignParams(array &$params)
...
+     * @param string $themePath
+     * @param string $area
+     * @return \Magento\Framework\View\Design\ThemeInterface
+     * @throws \UnexpectedValueException
+     */
+    private function getThemeByPath($themePath, $area)
+    {
+        $themeModel = $this->getThemeProvider()->getThemeByFullPath($area . '/' . $themePath);
+        if (!$themeModel) {
+            throw new \UnexpectedValueException("Could not find theme '$themePath' for area '$area'");
```
orlangur: Please reword `Could not find theme by full path ...`

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -167,6 +162,52 @@ public function updateDesignParams(array &$params)
...
+     * @param string $themeId
+     * @param string $area
+     * @return \Magento\Framework\View\Design\ThemeInterface
+     * @throws \UnexpectedValueException
+     */
+    private function getThemeById($themeId, $area)
+    {
+        $themeModel = $this->getThemeProvider()->getThemeById($themeId);
+        if (!$themeModel || $themeModel->getArea() != $area) {
+            throw new \UnexpectedValueException("Could not find theme '$themeId' for area '$area'");
```
orlangur: Please reword `Could not find theme by id ...`

lib/internal/Magento/Framework/View/Test/Unit/Asset/RepositoryTest.php
```diff
@@ -157,36 +157,218 @@ public function testUpdateDesignParamsWrongTheme()
...
+            ->method('getArea')
+            ->willReturn('frontend');
+
+        $this->themeProvider
+            ->expects($this->never())
+            ->method('getThemeById');
+        $this->themeProvider
+            ->expects($this->any())
+            ->method('getThemeByFullPath')
+            ->with('frontend/Magento/luma')
```
orlangur: Please avoid Luma theme mentioning in framework, just use some fictional theme path (fix both occurrences).

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -135,22 +135,17 @@ public function updateDesignParams(array &$params)
         }
```
orlangur, line 134: As this method was refactored significantly, please remove `@SuppressWarnings(PHPMD.CyclomaticComplexity)`, probably it is already passing this check.

lib/internal/Magento/Framework/View/Asset/Repository.php
```diff
@@ -167,6 +162,52 @@ public function updateDesignParams(array &$params)
...
+    {
+        $themeModel = $this->getThemeProvider()->getThemeByFullPath($area . '/' . $themePath);
+        if (!$themeModel) {
+            throw new \UnexpectedValueException("Could not find theme '$themePath' for area '$area'");
+        }
+        return $themeModel;
+    }
+
+    /**
+     * @param string $themeId
```
orlangur: Change to `int`.

orlangur (changes requested):

orlangur:
Thanks, your findings look great to me, especially that there was `theme_id` in some tests not containing number and that there are actually some integration and functional tests enforcing this logic.

Would be great if you add case to check new logic into some existing integration test but it's not mandatory. Please check my comments to slightly polish up the code.

lib/internal/Magento/Framework/View/Asset/Repository.php
alessandroniciforo: Moving this in the method we wouldn't pass the cyclomatic complexity

orlangur:
@alessandroniciforo there is one real failure left: https://travis-ci.org/magento/magento2/jobs/261023165
```
There was 1 failure:
1) Magento\Deploy\DeployTest::testDeploy
Failed asserting that 3266.36328125 is less than 1177.6.
/home/travis/build/magento/magento2/dev/tests/integration/testsuite/Magento/Deploy/DeployTest.php:232
/home/travis/build/magento/magento2/dev/tests/integration/testsuite/Magento/Deploy/DeployTest.php:151
                                                                        
FAILURES!                                                               
Tests: 1291, Assertions: 5716, Failures: 1, Incomplete: 43, Skipped: 11.
```

Could you please look into it?

ishakhsuvarov:
@alessandroniciforo @orlangur Is there any progress on this Pull Request?

orlangur:
@ishakhsuvarov I planned to look into this failure if Alessandro will not do it as fix seems quite valuable to me. Didn't have time yet, will have during this weekend.

Feel free to investigate failure yourself (maybe it occurred already in some other circumstances already?) or close this PR for now, I'll reopen when ready.

okorshenko:
Hi All 
We are closing this PR due to inactivity. If you would like to proceed with this PR, please reopen it any time

diamondavocado:
@orlangur Could you please re-open this PR? It was very close to being finished and it would solve issues https://github.com/magento/magento2/issues/10402 and https://github.com/magento/magento2/issues/8884.

orlangur:
@diamondavocado sure, not sure how it could help though since you are not an author :)

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10403) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=10403) before we can accept your contribution.<br/>

alessandroniciforo:
Hey guys,
sorry but I couldn't understand the reason for the failure of the integration test and I've since moved away from Magento.
Looking at the PR now, it seems that the issue has been solved already on 2.3-develop by this commit https://github.com/magento/magento2/commit/d3e32b4e5db80f7c04734f63cfeec83155613222.
I don't know if it makes sense to merge it to 2.1 (which I don't know if it's still maintained) or to 2.2

sidolov:
Hi @alessandroniciforo , thank you for your contribution! Unfortunately, we are no longer accept contributions to 2.1-develop and 2.2-develop branches. Looks like the issue was fixed in the scope of d3e32b4e5db80f7c04734f63cfeec83155613222, as you mentioned before. 
I'm closing this PR as not relevant. 
Thank you!

PR #10409, opened 2017-08-02
Get scope config values from stores scope by default

VincentMarmiesse:
Hello,

The `getValue()` and `isSetFlag()` methods from the `ScopeConfigInterface` should search in the `stores` scope by default, except in some very specific cases.
It's very confusing to have to use `$this->scopeConfig->getValue('path/to/config', 'stores')` everytime we want to access a config value.

It should be like in M1 where we just had to do `Mage::getStoreConfig('path/to/config')` and if we want to look for a specific scope we can with the additional parameters.

I'm open to discussion though.

### DIFF
--- a/app/code/Magento/Backup/Cron/SystemBackup.php
+++ b/app/code/Magento/Backup/Cron/SystemBackup.php
@@ -101,11 +101,11 @@ class SystemBackup
      */
     public function execute()
     {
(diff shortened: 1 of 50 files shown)


okorshenko:
Hi @VincentMarmiesse 
These changes are backward incompatible. Please look at our BC policy: http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#modifying-the-default-values-of-optional-arguments-in-public-and-protected-methods

PR #10426, opened 2017-08-03
Improves cms page options in the adminhtml configuration by sorting them alphabetically and adding the store code to the label.

hostep:
### Description

This is a first quick proposition for improving the options in the dropdown for selecting a cms page in the configuration in the adminhtml.

Changed:
- Alphabetised the cms pages
- Added the store code at the end between round brackets

Before:
screen shot 2017-08-03 at 22 00 35

After:
screen shot 2017-08-03 at 22 00 57

Possible further improvements I had in mind:
- Only add the store code when there is more then one store view
- Instead of the store code, use the name of the store view
- Maybe also include the name of the website and/or store, since a store view's name isn't unique

This is only a first attempt, feel free to give me feedback if something could be improved.
If this shouldn't be accepted to be added to Magento's core code, please let me know, it's not a big deal.

None

### Manual testing scenarios
1. Add a couple of storeviews
2. Create a cms page for each storeview and use the exact same name
3. Go to Stores => Configuration => General => Web => Default Pages
4. Expect to easily find the cms page you want instead of seeing the exact same page names

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cms/Model/ResourceModel/Page/Collection.php
+++ b/app/code/Magento/Cms/Model/ResourceModel/Page/Collection.php
@@ -51,7 +51,8 @@ class Collection extends AbstractCollection
             $identifier = $item->getData('identifier');
 
             $data['value'] = $identifier;
-            $data['label'] = $item->getData('title') . ' (' . $item->getData('store_code') . ')';
+            $data['label'] = $item->getData('title');
+            $data['store'] = $item->getData('store_code');
 
             if (in_array($identifier, $existingIdentifiers)) {
                 $data['value'] .= '|' . $item->getData('page_id');
@@ -62,11 +63,41 @@ class Collection extends AbstractCollection
             $res[] = $data;
         }
 
+        // sort cms pages by their title
         usort($res, function ($optionA, $optionB) {
             return strcmp($optionA['label'], $optionB['label']);
         });
 
+        $res = $this->groupOptionsByStore($res);
+
+        // TODO: transform the storecode to the website/store/storeview name here
+        // we have to do this here, because website/store/storeview combination isn't strictly unique,
+        // whereas the storecode is
+        // if we do the transformation earlier, the 'groupOptionsByStore' method would behave incorrectly
+        // TODO: also sort the array by the website/store/storeview name
+
         return $res;
+    }
+
+    /**
+     * Changes the options, so they are grouped by store
+     *
+     * @param array $options
+     * @return array
+     */
+    private function groupOptionsByStore(array $options)
+    {
+        $grouped = [];
+
+        foreach ($options as $option) {
+            $store = $option['store'];
+            if (!array_key_exists($store, $grouped)) {
+                $grouped[$store] = ['label' => $store, 'value' => []];
+            }
+            $grouped[$store]['value'][] = ['label' => $option['label'], 'value' => $option['value']];
+        }
+
+        return ['values' => ['label' => '', 'value' => $grouped]];
     }
 
     /**
--- a/app/code/Magento/Cms/Model/ResourceModel/Page/Collection.php
+++ b/app/code/Magento/Cms/Model/ResourceModel/Page/Collection.php
@@ -26,6 +26,12 @@ class Collection extends AbstractCollection
     protected $_previewFlag;
 
     /**
+     *
+     * @var array|null
+     */
+    private $storeLabelsCache;
+
+    /**
      * Define resource model
      *
      * @return void
@@ -52,7 +58,7 @@ class Collection extends AbstractCollection
 
             $data['value'] = $identifier;
             $data['label'] = $item->getData('title');
-            $data['store'] = $item->getData('store_code');
+            $data['store_ids'] = $item->getData('store_id');
 
             if (in_array($identifier, $existingIdentifiers)) {
                 $data['value'] .= '|' . $item->getData('page_id');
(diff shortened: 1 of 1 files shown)


okorshenko:
Waiting for PO approval

dmanners:
I like this idea as it can be really confusing with multiple pages with the same name. My only concern is that an admin user may get confused by the store codes. I am not 100% how many admin users understand what a store code relates to. Other options could be store name or url but I am not sure how well that these would work either.

hostep:
I agree with @dmanners: especially the store code 'admin' is confusing, it would be better if it was called 'default' for example.

That's why I'm looking for feedback for improving the current proposal :)

dmanners:
@hostep yeah I think that Name (and then default for the admin) could be a good one, but then looking at some of the stores we have some pointless names!

hostep:
In our case, the name of the store view wouldn't even be enough, look at this example from a client of ours (4x 'English' as store view name):
screen shot 2017-08-04 at 09 12 17

Maybe a combination of Website/Store/Store view name is better?

dmanners:
Another option could be to limit this drop down based on the store view you have selected. Then if an admin is looking at a specific store they would only see pages that store is set-up for. It does not get around the admin problem and assumes a user sets-up pages "well" but we cannot hold everyone's hands through life.

mzeis:
If a specific store view is selected, you could show only the pages for that store view. The same goes for the website scope when only one store view is available below.

If the global/default scope or website scope is selected and there is more than one store view, you could use optgroups (https://www.w3schools.com/tags/tag_optgroup.asp) to group the pages:

* Main Website > Main Website Store > English
  - Homepage
  - Page A
  - Page B
* North America Website > Canada Store > English
  - Homepage
  - Page B
  - Page C
* North America Website > USA Store > English
  - Homepage
  - Page A
  - Page C

piotrkwiecinski:
I really like the idea presented by @mzeis using option groups.

hostep:
I decided to test out @mzeis proposed solution, since I like it.
It still uses the store code for now, I'll update it later on to the website/store/storeview when I find some more time.

In case anyone is interested, the only config option in Magento which was already using this optgroup approach is Stores => Configuration => Advanced => Admin => Startup Page => Startup Page, I used that one to figure out how the structure of the returned array should look. There is an empty optgroup at the top which I can't seem to get rid of, but that's not a big deal I think.

It currently looks like this:
screen shot 2017-08-05 at 12 04 49

dmanners:
Looks good @hostep. I have restarted the build as travis had an issue.

I assume the PO still needs the final word on this but to me this makes things a lot easier to see what cms page is for what store. Still might consider showing the store information rather than the code but that should not be too much of a change.

okorshenko:
Hi All, 
We suggest to show what storeview CMS page belongs using Website/Store/Storview. Show website/store name if there are more than 1 website/store. Utilize **single** select dropdown component, similar to what we use on Product Page in Admin panel
screen shot 2017-08-04 at 5 26 41 pm 1

hostep:
Ok thanks for the feedback @okorshenko. I'm a little bit hesitant for implementing this, as I'm not sure if Magento's configuration section can already render such components? I can't seem to find an example in the configuration section currently? If you know if such a thing exists, or maybe a third party open source module which does something like that, that would probably help in figuring out how to implement this.

Would it be ok to go for the optgroup solution at first and then later on optimize it further with this component solution? Unless it's actually really simple to implement...
I haven't started looking at the code of these components yet, so maybe I'm a bit pessimistic about this.

hostep:
New update, still using optgroups, but with website/store/storeview labels instead of the storeview code:
screen shot 2017-08-10 at 21 06 20

I also realised that pages can be assigned to multiple storeviews, which didn't work properly in my previous solution. This is solved now, by outputting it multiple times underneath every storeview it is assigned to (example: 'Page is assigned to all storeviews').

Above result is based on these test pages:
screen shot 2017-08-10 at 21 10 01

okorshenko:
Hi @hostep thank you for latest fixes.
What about UI component implementation?
You are right. We don't have similar implementation yet, but this is a good time to start. I prepared a code example for you how to add advanced select to Store Configuration Form: https://github.com/okorshenko/magento2/pull/4/files . You can take it as a patch to start your implementation. In this implementation it is still multiselect, but we would like to have it as a single select. So should not be a huge problem. 
Please let me know if you need some help with this. Thank you

belbiy:
Hi @hostep! For me, this solution seems complicated from user experience perspective:
image
We understand the need for identifying the page in the list, but we see that the similar page names are more of an edge-case. And you can still have pages with the same name assigned to same Store View, without any distinction. This solution will work for your case but will be a chore for most users - a lot of page duplicates.

The other problem is that this list still shows merchant pages that not really available for selected scope. It allows making an error by configuring the homepage that is not available on selected Store View. 

What I propose instead is using page ID instead of Scope to identify them. And it's only needed in case of pages having the same name. That will help us avoid duplicating page names, and keep it clean for all merchants:

default-cms-page

We should also filter this list using current scope selected in configuration. This will prevent the merchant from configuring a home page from a different scope and getting this error:

image003

And it's also a good idea to do this for other options for Default Pages:  CMS No Route Page, CMS No Cookies Page.

hostep:
Thanks for the feedback @belbiy!

At first I didn't like your proposal for using the ID (because that's not clear enough in my opinion), but when we limit the options based on the current chosen scope, then you are right and it won't happen very frequently that two pages will have the same title in that filtered list, so using the ID isn't a bad option in those cases.

I'll see what I can do, might take me a couple of days/weeks until I can continue with this, depends on the energy I have left after a hard working day :)

dmanners:
hi @hostep have you had any chance to look into this as yet?

hostep:
Hi @dmanners 

I'm really sorry, but I have the feeling I won't have time anymore this year to look into this.
Sorry to all the people who provided excellent feedback here.

If someone else wants to pick this up, feel free to do so.

Should I close this PR?

dmanners:
@hostep no worries about not being able to finish the PR. I will check what we would do to allow someone else to pick this up and I will get back to you about it.

dmanners:
@hostep hope you dont mind but I have converted this PR into an issue so someone can pick it up and continue with it. I think this would be a great improvement in usability so would love to see it progresses. Feel free to pick it up again if you have time. Thanks for your help.

PR #10432, opened 2017-08-04
Test cron expressions against localized time

AntonEvers:
### Description

As mentioned in #4237: 
@AlexanderRakov I do expect it now you mention it, having the current code change in place. But I would not expect it if I were a shop owner. So now that the times are correct in the cron_schedule table, I need to add a change that converts the crontab expression to the correct unix time stamp (UTC) for the matching time in the admin store view time zone.

1. magento/magento2#4237: Cron times in the database have a double timezone correction

### Manual testing scenarios

1. Set the global time zone to something other than GMT.
2. Schedule the sitemap_generate cron job at 00:00:00.
3. Configure cron to schedule ahead for 1440
4. Flush caches
5. run `bin/magento cron:run`
6. Run `SELECT `scheduled_at` FROM `cron_schedule` WHERE `job_code` = "sitemap_generate";`
7. This should be yyyy-mm-dd 00:00:00 (it was yyyy-mm-dd 00:06:00 in timezone GMT+6 before this PR)

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cron/Model/Schedule.php
+++ b/app/code/Magento/Cron/Model/Schedule.php
@@ -45,9 +45,9 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
     const STATUS_ERROR = 'error';
 
     /**
-     * @var \Magento\Framework\Stdlib\DateTime\DateTime
+     * @var \Magento\Framework\Stdlib\DateTime\TimezoneInterface
      */
-    private $dateTime;
+    private $timeZone;
 
     /**
      * @param \Magento\Framework\Model\Context $context
@@ -55,7 +55,7 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
      * @param \Magento\Framework\Model\ResourceModel\AbstractResource $resource
      * @param \Magento\Framework\Data\Collection\AbstractDb $resourceCollection
      * @param array $data
-     * @param \Magento\Framework\Stdlib\DateTime\DateTime|null $dateTime
+     * @param \Magento\Framework\Stdlib\DateTime\TimezoneInterface|null $timeZone
      */
     public function __construct(
         \Magento\Framework\Model\Context $context,
@@ -63,10 +63,10 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
         array $data = [],
-        \Magento\Framework\Stdlib\DateTime\DateTime $dateTime = null
+        \Magento\Framework\Stdlib\DateTime\TimezoneInterface $timeZone = null
     ) {
         parent::__construct($context, $registry, $resource, $resourceCollection, $data);
-        $this->dateTime = $dateTime;
+        $this->timeZone = $timeZone;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ishakhsuvarov:
Hi @ajpevers
Could you please take a look at the failed unit tests?
Thanks

AntonEvers:
Checks passing. @ishakhsuvarov is this also a valid way to maintain backward compatibility or do you want me to move the object manager fallback back to the constructor and change the unit test?

ishakhsuvarov:
> do you want me to move the object manager fallback back to the constructor

This would be a correct solution according to current guidelines.

app/code/Magento/Cron/Model/Schedule.php
```diff
@@ -94,14 +102,16 @@ public function setCronExpr($expr)
...
+            ->get(\Magento\Framework\Stdlib\DateTime\DateTime::class);
         $time = $this->getScheduledAt();
         $e = $this->getCronExprArr();
 
         if (!$e || !$time) {
             return false;
         }
         if (!is_numeric($time)) {
-            $time = strtotime($time);
+            $time = strtotime($time) + $this->dateTime->getGmtOffset();
```
okorshenko: We should not calculate the time for another timezone using inline code. In the future, it will be hard to track all such places if there some bugs. Please, use` \Magento\Framework\Stdlib\DateTime\TimezoneInterface` to convert to date `$time` to Default timezone.

okorshenko (changes requested):

AntonEvers:
@okorshenko for a few hours I've been struggling with the unit tests that should be rewritten for my latest change. I'm quite new to this and came a long way, but here I'm stuck. Can you connect me with someone that can help me fix the tests? (It would be nice if we can configure the test with a timezone offset while we're at it.)

app/code/Magento/Cron/Model/Schedule.php
```diff
@@ -45,20 +45,28 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
...
+     * @param \Magento\Framework\Stdlib\DateTime\TimezoneInterface|null $timeZone
      */
     public function __construct(
         \Magento\Framework\Model\Context $context,
         \Magento\Framework\Registry $registry,
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
-        array $data = []
+        array $data = [],
+        \Magento\Framework\Stdlib\DateTime\TimezoneInterface $timeZone = null
```
sshymko, line 66: This has to be a required argument (to avoid the object manager hack below). Move it up to go before the optional ones.

app/code/Magento/Cron/Model/Schedule.php
```diff
@@ -241,4 +256,15 @@ public function tryLockJob()
...
         return false;
     }
+
+    /**
+     * @deprecated
+     * @return \Magento\Framework\Stdlib\DateTime\TimezoneInterface
+     */
+    private function getDateTime()
+    {
+        $this->timeZone = $this->timeZone ?: \Magento\Framework\App\ObjectManager::getInstance()
```
sshymko, line 266: Static access to the object manager is highly discouraged. The interface primarily exists for the unserialization use case. It should not be used outside of that. The object manager will not be necessary if to inject the dependency as required on constructor. See the constructor comment above.

sshymko (changes requested):

app/code/Magento/Cron/Model/Schedule.php
ishakhsuvarov: Adding new required argument to the class constructor will break backwards compatibility, for those third-party extensions which could have extended from it.
Please refer to the [Backwards Compatible Development Guide](http://devdocs.magento.com/guides/v2.1/ext-best-practices/extension-coding/backwards-compatible-development/index.html) for more information (this case is covered in `Adding a constructor parameter` section)

sshymko: @ishakhsuvarov 
Good to know. The BC policy seems very decremental to the codebase :'(

PR #10442, opened 2017-08-04
magento/magento2#10045 Rename item resolver to item provider

piotrkwiecinski:
### Description

Follow up PR to address naming as it'll be released for 2.3 there should be no BC here.

1. magento/magento2#10045: Refactoring: Adding items to the Sitemap

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sitemap/Model/ItemProvider/Composite.php
+++ b/app/code/Magento/Sitemap/Model/ItemProvider/Composite.php
@@ -4,25 +4,25 @@
  * See COPYING.txt for license details.
  */
 
-namespace Magento\Sitemap\Model\ItemResolver;
+namespace Magento\Sitemap\Model\ItemProvider;
 
-class Composite implements ItemResolverInterface
+class Composite implements ItemProviderInterface
 {
     /**
      * Item resolvers
      *
-     * @var ItemResolverInterface[]
+     * @var ItemProviderInterface[]
      */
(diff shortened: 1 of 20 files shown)


piotrkwiecinski:
FYI @antonkril

antonkril:
Thank you.

app/code/Magento/Sitemap/Model/ItemProvider/Composite.php
```diff
@@ -32,7 +32,7 @@ public function getItems($storeId)
     {
         $items = [];
 
-        foreach ($this->itemResolvers as $resolver) {
+        foreach ($this->itemProviders as $resolver) {
             foreach ($resolver->getItems($storeId) as $item) {
                 $items[] = $item;
```
Koc, line 37: Do we really collect all items in array? Why no use generators for lower memory usage?

piotrkwiecinski: @Koc goal of this pull request was to improve way to customise for 3rd party extensions to add how items are added without need to extend or override sitemap class itself. I haven't changed way how data is extracted from database itself and passed around. This is way I came up with while working on it in my free personal time and it was accepted as good enough solution. If you are able to propose more efficient way feel free to create and issue with your ideas as a way to start discussion or even better PR with an implementation.

Issue #10436, opened 2017-08-04
Customer group CUST_GROUP_ALL not selected in adminhtml after attribute save


mashb1t:
### Preconditions
1. Magento 2.1.7  EE
2. PHP 7.0.20 

### Steps to reproduce
1. Call method` \Magento\Customer\Model\Customer\Source\Group::toOptionArray`

### Expected result
1. All `$customerGroups['value']` should be of type string

### Actual result
1. First customer group value (`\Magento\Customer\Api\Data\GroupInterface::CUST_GROUP_ALL`) is integer

### Solution
1. Cast `GroupInterface::CUST_GROUP_ALL` to string 

https://github.com/magento/magento2/blob/354fc11a1a2011f6ff4ce9a352ef59f22953b88f/app/code/Magento/Customer/Model/Customer/Source/Group.php#L59

When using the result of `\Magento\Customer\Model\Customer\Source\Group::toOptionArray` as options in an (multiselect) attribute source model, the respective customer group `CUST_GROUP_ALL` is not getting selected in the Magento Admin-Panel despite the value of the group being set.

magento-team:
Internal ticket to track issue progress: MAGETWO-71432

PR #10475, opened 2017-08-09
Add cast to string for CUST_GROUP_ALL

mashb1t:
Add cast to string for GroupInterface::CUST_GROUP_ALL in customer group source model and adjust test

Fixes https://github.com/magento/magento2/issues/10436

### DIFF
--- a/app/code/Magento/Customer/Model/Customer/Source/Group.php
+++ b/app/code/Magento/Customer/Model/Customer/Source/Group.php
@@ -56,7 +56,7 @@ class Group implements GroupSourceInterface
         $customerGroups = [];
         $customerGroups[] = [
             'label' => __('ALL GROUPS'),
-            'value' => GroupInterface::CUST_GROUP_ALL,
+            'value' => (string)GroupInterface::CUST_GROUP_ALL,
         ];
 
         /** @var GroupSearchResultsInterface $groups */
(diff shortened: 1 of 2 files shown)


ishakhsuvarov:
@mash1t Thank you for your PR. Could you please be more precise on the steps to reproduce the issue? I understand the concept, but have to verify the current behavior first.

orlangur:
@ishakhsuvarov 
> When using the result of \Magento\Customer\Model\Customer\Source\Group::toOptionArray as options in an (multiselect) attribute source model, the respective customer group CUST_GROUP_ALL is not getting selected in the Magento Admin-Panel despite the value of the group being set.

from issue claimed as fixed looks pretty much like StR. The question is do you observe the same behavior or not.

mashb1t:
@ishakhsuvarov Currently i use a source model for a multiselect product attribute using the return value of `app/code/Magento/Customer/Model/Customer/Source/Group.php::toOptionArray` in its method `getAllOptions` as options.
The problem occurs when editing and saving the product in the backend with the attribute being set to value 32000 (`int`) => "ALL GROUPS".
- After saving the product the new value of the attribute value for the product is updated in the database (attribute type being `text`) but does not get preselected in the backend, because it is of type `int` instead of `string`.
- Every other customer group selected and saved will be preselected in the backend.

I guess that knockout only preselects values of the same data type.

PR #10390, opened 2017-08-01
Translate order getCreatedAtFormatted() to store locale

Jeroenll:
### Description
Extend `getDefaultLocale()` to allow fetching scoped values. Use this in `getCreatedAtFormatted()` so `created_at` date of order will be translated in emails to locale being used in that store view.

### Fixed Issues (if relevant)
1. None that I could find.

### Manual testing scenarios
1. Set default locale of your store to any language but English.
2. Create order and send confirmation email.

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

PR #10491, opened 2017-08-11
Translate order getCreatedAtFormatted() to store locale

Jeroenll:
### Description
Updated version of #10390, added new interface method en deprecated the current one.

Add `getDefaultStoreLocale()` to allow fetching scoped values. Use this in `getCreatedAtFormatted()` so `created_at` date of order will be translated in emails to locale being used in that store view.

1. None that I could find.

### Manual testing scenarios
1. Set default locale of your store to any language but English.
2. Create order and send confirmation email.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -7,6 +7,8 @@ namespace Magento\Sales\Model;
 
 use Magento\Directory\Model\Currency;
 use Magento\Framework\Api\AttributeValueFactory;
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\Locale\ResolverInterface;
 use Magento\Framework\Pricing\PriceCurrencyInterface;
 use Magento\Sales\Api\Data\OrderInterface;
 use Magento\Sales\Api\Data\OrderStatusHistoryInterface;
@@ -212,6 +214,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $_currencyFactory;
 
     /**
+     * @var \Magento\Eav\Model\Config
+     */
+    private $_eavConfig;
+
+    /**
      * @var \Magento\Sales\Model\Order\Status\HistoryFactory
      */
     protected $_orderHistoryFactory;
@@ -267,6 +274,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $timezone;
 
     /**
+     * @var ResolverInterface
+     */
+    private $localeResolver;
+
+    /**
      * @param \Magento\Framework\Model\Context $context
      * @param \Magento\Framework\Registry $registry
      * @param \Magento\Framework\Api\ExtensionAttributesFactory $extensionFactory
@@ -294,6 +306,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
      * @param \Magento\Framework\Model\ResourceModel\AbstractResource $resource
      * @param \Magento\Framework\Data\Collection\AbstractDb $resourceCollection
      * @param array $data
+     * @param ResolverInterface $localeResolver
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
     public function __construct(
@@ -323,7 +336,8 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
         \Magento\Catalog\Model\ResourceModel\Product\CollectionFactory $productListFactory,
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
-        array $data = []
+        array $data = [],
+        ResolverInterface $localeResolver = null
     ) {
         $this->_storeManager = $storeManager;
         $this->_orderConfig = $orderConfig;
@@ -345,6 +359,8 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
         $this->_trackCollectionFactory = $trackCollectionFactory;
         $this->salesOrderCollectionFactory = $salesOrderCollectionFactory;
         $this->priceCurrency = $priceCurrency;
+        $this->localeResolver = $localeResolver ?: ObjectManager::getInstance()->get(ResolverInterface::class);
+
         parent::__construct(
             $context,
             $registry,
@@ -1829,7 +1845,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
             new \DateTime($this->getCreatedAt()),
             $format,
             $format,
-            null,
+            $this->localeResolver->getDefaultLocale(),
             $this->timezone->getConfigTimezone('store', $this->getStore())
         );
     }
--- a/app/code/Magento/Sales/Test/Unit/Model/OrderTest.php
+++ b/app/code/Magento/Sales/Test/Unit/Model/OrderTest.php
@@ -7,6 +7,8 @@ namespace Magento\Sales\Test\Unit\Model;
 
 use Magento\Catalog\Api\Data\ProductInterface;
 use Magento\Catalog\Model\ResourceModel\Product\CollectionFactory as ProductCollectionFactory;
+use Magento\Framework\Locale\ResolverInterface;
+use Magento\Framework\Stdlib\DateTime\TimezoneInterface;
 use Magento\Sales\Api\Data\OrderInterface;
 use Magento\Sales\Model\Order;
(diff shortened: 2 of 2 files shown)


orlangur:
You cannot add new methods into existing `@api` interfaces as it will break every existing interface implementation. Please check http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#introduction-of-a-method-to-a-class-or-interface

Jeroenll:
@orlangur creating a new implementation and use that in `\Magento\Framework\Locale\Resolver` would be the way to go? If so, I will update the PR when I have the time.

orlangur:
I think `\Magento\Framework\Locale\ResolverInterface` should not be touched (changed/deprecated) for the sake of the fix even if we didn't have BC constraints.

Any idea why `\Magento\Framework\Locale\Resolver` and `\Magento\Backend\Model\Locale\Resolver` does not cover this case with order model properly? From implementation it looks like correct locale should be already there.

So, I wouldn't recommend to introduce any `StoreAwareLocaleResolverInterface` unless really necessary.

Jeroenll:
@orlangur After some more debugging I found the real issue.

`\Magento\Framework\Locale\Resolver::__construct()` will set the locale using `$this->setLocale($locale);` where `$locale` is `null` since no argument is set in the `di.xml`. 

Then in the `Magento\Backend\Model\Locale\Resolver::setLocale()` will get `$local` as `null`. In there it will load the session locale which is also null and the user locale which is `en_US` (Magento backend locale). So there it will set the locale to `en_US` although our store locale is `nl_NL`.  The changes I've made aren't neccesary because the `getDefaultLocale()` will fetch the correct value. So this issue is only relevant is backend and frontend locales don't match.

In my opinion the best solution would be to emulate the store locale in `\Magento\Sales\Model\OrdergetCreatedAtFormatted()` and change it back after the date is formatted. If you agree on this one I will implement this.

orlangur:
Do we need to really emulate locale or it's enough to pass proper locale in
```
-            null,
 +            $this->localeResolver->getDefaultStoreLocale($this->getStore()),
```
?

Jeroenll:
@orlangur Yes, if you have the proper locale you can insert it there. Using `$this->localeResolver->getDefaultLocale()` there will solve this if I don't miss anything?

orlangur:
Not sure, please try it.

Please go this way, if `getDefaultLocale` will not work do something like `$locale = $this->scopeConfig->getValue($this->getDefaultLocalePath(), $this->scopeType, $store);` in place (just a fix in order model, no need in additional interfaces/classes).

Please force push into the same branch with a single commit when it is ready (i.e. works & builds are green).

Jeroenll:
@orlangur Tested and works. Tests are green, only one job exceeded the maximum time limit for jobs, and has been terminated.

app/code/Magento/Sales/Model/Order.php
```diff
@@ -214,6 +216,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $_currencyFactory;
 
     /**
+     * @var \Magento\Eav\Model\Config
+     */
+    protected $_eavConfig;
```
orlangur: Thanks for noticing this 👍

app/code/Magento/Sales/Model/Order.php
```diff
@@ -269,6 +276,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $timezone;
 
     /**
+     * @var ResolverInterface
+     */
+    protected $localeResolver;
```
orlangur: Please make it private (this is a rule for any new class properties/methods).

app/code/Magento/Sales/Model/Order.php
```diff
@@ -1831,7 +1847,7 @@ public function getCreatedAtFormatted($format)
             new \DateTime($this->getCreatedAt()),
             $format,
             $format,
-            null,
+            $this->localeResolver->getDefaultLocale(),
```
orlangur, line 1848: Seems like this public method can be easily covered with unit test, however, as there is no logic and we just call method of another class with a bunch of arguments I do not insist.

orlangur (changes requested):

app/code/Magento/Sales/Model/Order.php
orlangur: No need to make it private I believe, not sure it will even pass the tests this way.

orlangur:
@JeroenVanLeusden cool job! Please do an amend commit + force push (only protected -> private for one field is a necessary field, other is at your discretion) and I will proceed with merging.

app/code/Magento/Sales/Model/Order.php
Jeroenll: @orlangur It was dynamically declared (read: public), making it private may break things in 3rd party code?

app/code/Magento/Sales/Model/Order.php
Jeroenll: It indeed calls public methods, making a test would be duplicate I guess?

orlangur: Nope, mock this dependency, the only thing to be asserted that we are passing concrete locale and not `null`.

Jeroenll: Will do that later today and notify you when ready!

orlangur: Thanks for keeping me updated, waiting for a ping.

app/code/Magento/Sales/Model/Order.php
orlangur: I would give it a try and if no core tests are broken leave it `private`.

I believe if any extension relied on this property as public it's totally weird, inheritance from core classes is not recommended and property is not marked as `@api`.

Jeroenll:
@orlangur Want to test if the dependency is not `null` like
```
$localeResolver = $this->getMockBuilder(\Magento\Framework\Locale\ResolverInterface::class)
    ->getMock();
$this->assertNotNull($localeResolver);
```

Or if the `getDefaultLocale()` returns a `string`?

orlangur:
@JeroenVanLeusden NotNull is bad practice, some good-looking locale string like `'nl_NL'` please.

Jeroenll:
@orlangur Still getting used to Magento's tests please bare with me :)
```
$localeResolver = $this->getMockBuilder(\Magento\Framework\Locale\ResolverInterface::class)
    ->getMock();

$localeResolver->expects($this->once())
     ->method('getDefaultLocale')
    ->willReturn('en_US');
```

Taken this looking from other tests but this doesn't work, method isn't being called. Also would like to test with some different languages then just `en_US`, what would be the way to approach this?

orlangur:
@JeroenVanLeusden just make sure that you passed `localeResolver` into object under test.

As I mentioned, better use some non-default locale like `'nl_NL'`. No need to test multiple cases I believe.

There is no need to assert that `getDefaultLocale` method was called, such checks are fragile, just check that correct locale is passed into called method of other mock dependency (timeWhatever->someMethod(..., locale instead of null, ...).

orlangur:
@JeroenVanLeusden feel free to push your test even if it does not work yet, I'll do necessary changes, push and then you'll squash all changes into single commit again.

Jeroenll:
@orlangur Pushed my changes so far. `getDefaultLocale()` returns `null` so that will need some work. Waiting for you ping to squash all changes :)

Jeroenll:
@orlangur Any progress with the tests? Would love to see this merged.

orlangur:
@JeroenVanLeusden, yeah, I'll have some time to work on this today, stay tuned.

orlangur:
So, here is an example of appropriate test: https://github.com/orlangur/magento2/commit/9e58ec25211c53e740ee20c13f5a311aaf44a279

Main idea behind unit tests is that we call only public methods of object under test and all other classes should be replaced with mocks (so, for example, `setDefaultLocale` of `localeResolver` does not work as it does nothing).

Please rebase against `develop` branch, adopt test to PHPUnit 6 to minimize merging efforts (`getMock` -> `createMock`, maybe something else) and then amend all changes into your first commit.

Jeroenll:
@orlangur Getting an `Error: Call to undefined method getMock()` on almost all calls. Any idea where this might be coming from?

orlangur:
@JeroenVanLeusden, I mentioned this, it is called now `createMock`. More details in https://thephp.cc/news/2017/02/migrating-to-phpunit-6

Jeroenll:
@orlangur My bad, all good now!

PR #10500, opened 2017-08-11
Improvements to the phpserver router

aredridel:
- handles versioned static assets
- our own mime handling for those, since php's built in web server's mime table isn't accessible from userland
- works with a magento instance set up with use-rewrites=1

Is this a useful amendment? I'm happy to fix this up with suggestions, too, as I intend to use it for my daily development.

### DIFF
--- a/phpserver/router.php
+++ b/phpserver/router.php
@@ -76,10 +76,12 @@ if (php_sapi_name() === 'cli-server') {
 
         $debug("file: $file");
 
-        if (file_exists($origFile)) {
+        if (file_exists($origFile) || file_exists($file)) {
+            if (file_exists($origFile)) {
+                $file = $origFile;
+            }
+
             $debug('file exists');
-            return false;
-        } else if (file_exists($file)) {
             $mimeTypes = [
                 'css' => 'text/css',
                 'js'  => 'application/javascript',
@@ -93,9 +95,8 @@ if (php_sapi_name() === 'cli-server') {
                 'html' => 'text/html',
             ];
 
-            $type = isset($mimeTypes[$ext]) && $mimeTypes[$ext];
-            if ($type) {
-                header("Content-Type: $type");
+            if (isset($mimeTypes[$ext])) {
+                header("Content-Type: $mimeTypes[$ext]");
             }
             readfile($file);
             return;
--- a/dev/tests/integration/testsuite/Magento/Phpserver/PhpserverTest.php
+++ b/dev/tests/integration/testsuite/Magento/Phpserver/PhpserverTest.php
@@ -15,7 +15,7 @@ namespace Magento\Phpserver;
  *
  * @magentoAppArea frontend
  */
-class PhpserverTest extends \PHPUnit_Framework_TestCase
+class PhpserverTest extends \PHPUnit\Framework\TestCase
 {
 
     const BASE_URL = '127.0.0.1:8082';
--- a/dev/tests/integration/testsuite/Magento/Phpserver/PhpserverTest.php
+++ b/dev/tests/integration/testsuite/Magento/Phpserver/PhpserverTest.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Phpserver;
+
+/**
+ * @magentoAppIsolation enabled
+ *
+ * @magentoConfigFixture current_store web/secure/base_url http://127.0.0.1:8082/
+ * @magentoConfigFixture current_store web/unsecure/base_link_url http://127.0.0.1:8082/
+ * @magentoConfigFixture current_store web/secure/base_link_url http://127.0.0.1:8082/
+ * @magentoConfigFixture current_store web/secure/use_in_frontend 0
+ *
+ * @magentoAppArea frontend
+ */
+class PhpserverTest extends \PHPUnit_Framework_TestCase
+{
+
+    const BASE_URL = '127.0.0.1:8082';
+
+    private static $serverPid;
+
+    private $httpClient;
+
+    /**
+     * Instantiate phpserver in the pub folder
+     */
+    public static function setUpBeforeClass()
+    {
+        $return = [];
+
+        $baseDir = __DIR__.'/../../../../../../';
+        $command = sprintf(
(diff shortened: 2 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10500) <br/>All committers have signed the CLA.

fooman:
Thank you @aredridel for your contribution. I haven't used the phpserver myself but anything that makes Magento easier to develop with sounds very useful to me. Travis picked up some minor style nitpicks - would you mind taking a look at [those](https://travis-ci.org/magento/magento2/jobs/263504693).

@Flyingmana I believe you provided the initial implementation. Any chance you are able to take a look?

phpserver/router.php
```diff
@@ -28,74 +28,93 @@
...
         return false;    // serve the requested resource as-is.
     }
 
     $path = pathinfo($_SERVER["SCRIPT_FILENAME"]);
     $url   = pathinfo(substr($_SERVER["REQUEST_URI"], 1));
     $route = parse_url(substr($_SERVER["REQUEST_URI"], 1))["path"];
-
-    $debug($path);
-    $debug($route);
+    $ext = @pathinfo($route)['extension'];
```
Flyingmana: are there cases when pathinfo is throwing an error, or is it because the array access for 'extension' is sometimes not working?

phpserver/router.php
```diff
@@ -28,74 +28,93 @@
...
+                'png' => 'image/png',
+                'gif' => 'image/gif',
+                'svg' => 'image/svg+xml',
+                'map' => 'application/json',
+                'woff' => 'application/x-woff',
+                'woff2' => 'application/font-woff2',
+                'html' => 'text/html',
+            ];
+
+            $type = @$mimeTypes[$ext];
```
Flyingmana: I believe its still better to check via isset() instead of just surpressing the error here. 
maybe outdated reference: https://stackoverflow.com/questions/12733666/which-is-better-in-php-suppress-warnings-with-or-run-extra-checks-with-isse

phpserver/router.php
```diff
@@ -28,74 +28,93 @@
...
+            if ($type) {
+                header("Content-Type: $type");
+            }
+            return readfile($file);
```
Flyingmana: is the return of readfile intended? I dont know how php handles the returned number of bytes

Flyingmana (commented):
That is a great addition and I like it a lot, that you found solutions to this problems.

At the moment I dont have something to test these, but from what I see the changes make sense and will be very useful for other users of the phpserver

phpserver/router.php
aredridel: It's because there's not always an extension.

phpserver/router.php
aredridel: Yeah.  I prefer the shorter but not a problem to go the other way.

phpserver/router.php
aredridel: Habit -- I always return the return value from things. Too long as a rubyist.

phpserver/router.php
aredridel: Updated to not use @

phpserver/router.php
aredridel: updated to use isset

phpserver/router.php
aredridel: Separated the return

fooman:
@aredridel thanks for making those changes. This all looks good to me and I have put this in the queue to be processed for merging into develop.

phpserver/router.php
```diff
@@ -28,74 +28,92 @@
...
+                'png' => 'image/png',
+                'gif' => 'image/gif',
+                'svg' => 'image/svg+xml',
+                'map' => 'application/json',
+                'woff' => 'application/x-woff',
+                'woff2' => 'application/font-woff2',
+                'html' => 'text/html',
+            ];
+
+            $type = isset($mimeTypes[$ext]) && $mimeTypes[$ext];
```
hostep: This will return a boolean and not the mime type, I think this needs to become: `$type = $mimeTypes[$ext] ?? '';` or something like that.

Haven't verified the rest of the code, this just popped out to me :)

ishakhsuvarov: Thanks for pointing out!

fooman: @aredridel apologies I had missed the above. Would you mind taking a look? This might be syntax borrowed from another language -
`isset($mimeTypes[$ext])? $mimeTypes[$ext]:'';` might be more appropriate. Should anything happen if we encounter a file that is not in the list of file extensions?

fooman:
Hi @aredridel just checking in if there is anything I can help with to get your pull request merged?

phpserver/router.php
```diff
@@ -28,74 +28,92 @@
...
 
-        $file = $magentoPackagePubDir . '/' . $route;
-        $debug($file);
-        if (file_exists($file)) {
+        $debug("file: $file");
+
+        if (file_exists($origFile)) {
             $debug('file exists');
             return false;
+        } else if (file_exists($file)) {
```
fuunnx: Sadly it does not take into account other locales than en_US. e.g: the luma theme breaks with 404 errors when the store in in fr_FR.

I did a quick fix on my dev env, testing both fr_FR and en_US static files. 

Great PR by the way

fooman:
Apologies @aredridel in not getting this moved forward so far. As Magento had so far zero test coverage in this area I have added a couple of updates to your pull request to add some tests. Please let me know if anything in the latest changes looks out of place to you.

@fuunnx I am keen to hear what additional changes you needed to make to enable a multi-locale environment (I am not sure if this is reliant on a particular setting - for example to add the store code to the url). If you are interested feel free to create a separate pull request once this one has merged.

fuunnx:
@fooman My locale was `fr_FR` so I hardcoded the following change:
[In this code](https://github.com/magento/magento2/blob/046585f02443bb6ead56302d24da7dc6d990db03/phpserver/router.php#L75-L102) if neither `$origFile` nor `$file` exist then check for alternative file by replacing `fr_FR` in the file path by `en_US`. Then it works :)

But I don't know the fallback logic magento use internally for this case so I'm not comfortable making a PR =)

aredridel:
Oh nice! I've been intending to come back to this but haven't had time lately!

Issue #7471, opened 2016-11-17
No multi currency support in google analytics module


sam-g-roberts:
There is currently no support for multi currency in the magento google analytics module. It's a 1 minute change and should really be included as default as most stores will use Google analytics and if they have multiple currencies not having multicurrency support will skew their data.

In the google analytics module block Ga.php the below needs to be changed from

```
            $result[] = sprintf(
                "ga('ec:setAction', 'purchase', {
                    'id': '%s',
                    'affiliation': '%s',
                    'revenue': '%s',
                    'tax': '%s',
                    'shipping': '%s'
                });",
                $order->getIncrementId(),
                $this->escapeJsQuote($this->_storeManager->getStore()->getFrontendName()),
                $order->getBaseGrandTotal(),
                $order->getBaseTaxAmount(),
                $order->getBaseShippingAmount()
            );
```

to

```
            $result[] = sprintf(
                "ga('ec:setAction', 'purchase', {
                    'id': '%s',
                    'affiliation': '%s',
                    'revenue': '%s',
                    'tax': '%s',
                    'shipping': '%s',
                    'currency': '%s'
                });",
                $order->getIncrementId(),
                $this->escapeJsQuote($this->_storeManager->getStore()->getFrontendName()),
                $order->getBaseGrandTotal(),
                $order->getBaseTaxAmount(),
                $order->getBaseShippingAmount(),
                $order->getOrderCurrencyCode()
            );
```

veloraven:
@sam-g-roberts thank you for your feedback.
Please, format this issue according to the [Issue reporting guidelines](https://github.com/magento/magento2/wiki/Issue-reporting-guidelines): with steps to reproduce, actual result and expected result.
Please, also identify which version of Magento you are running.

sam-g-roberts:
Magento version is github develop version

You haven't included the google analytics multi currency paramater when sending off a conversion to google analytics.

In this file https://github.com/magento/magento2/blob/develop/app/code/Magento/GoogleAnalytics/Block/Ga.php

lines 130-143 need to be replaced with the below code

```
$result[] = sprintf(
                "ga('ec:setAction', 'purchase', {
                    'id': '%s',
                    'affiliation': '%s',
                    'revenue': '%s',
                    'tax': '%s',
                    'shipping': '%s',
                    'currency': '%s'
                });",
                $order->getIncrementId(),
                $this->escapeJsQuote($this->_storeManager->getStore()->getFrontendName()),
[...]

PR #10508, opened 2017-08-12
Send different base currency in Google analytics

DominicWatts:
We've encountered a scenario where payment providers require different base currencies configured for different websites in a multisite and as a result need to send different base currency in Google analytics

1. https://github.com/magento/magento2/issues/6709
2. https://github.com/magento/magento2/issues/7471

### Manual testing scenarios
1. Add to basket
2. Complete checkout
3. Observe results within tracking

### Contribution checklist

### DIFF
--- a/app/code/Magento/GoogleAnalytics/Block/Ga.php
+++ b/app/code/Magento/GoogleAnalytics/Block/Ga.php
@@ -120,7 +120,7 @@ class Ga extends \Magento\Framework\View\Element\Template
         $result[] = "ga('require', 'ec', 'ec.js');";
 
         foreach ($collection as $order) {
-            $result[] = "ga('set', 'currencyCode', '" . $order->getBaseCurrencyCode() . "');";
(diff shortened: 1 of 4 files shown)


DominicWatts:
FYI https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#multicurrency

app/code/Magento/GoogleAnalytics/Block/Ga.php
```diff
@@ -120,6 +120,7 @@ public function getOrdersTrackingCode()
         $result[] = "ga('require', 'ec', 'ec.js');";
 
         foreach ($collection as $order) {
+            $result[] = "ga('set', 'currencyCode', '" . $order->getBaseCurrencyCode() . "');";
```
ishakhsuvarov: Looks like `BaseCurrencyCode` is not exactly the currency code needed here. With the multi-store setup it will always be the same, regardless of different display currency.

ishakhsuvarov (changes requested):

Issue #10142, opened 2017-07-04
Incorrect order behavior using PayPal Express Checkout


farior:
### Preconditions
1. PHP 7.0.10
2. Mysql 5.7.15
3. Fresh Magento CE 2.1.7 installation
4. Sample Data deployed
5. PayPal Express Checkout sandbox configured

### Steps to reproduce
1. Go to Admin > Products > Catalog, find "Hero Hoodie" product and ensure all configurations have **100** qty
1. Go to home page and click on "Hero Hoodie" product
2. Choose **Black** color, **XS** size, input qty=**100** and Add product to cart
3. Continue through checkout process, choose PayPal and place order
4. Click Pay Now in PayPal window

### Expected result
1. User is redirected to checkout success page
2. Order is successfully created with invoice
3. Order confirmation email is sent
4. Stock qty is correctly decreased to 0

### Actual result
**For the first try:**
1. User is redirected to /checkout/cart page
2. A bunch of errors presented (We can't place order, We can't initialize Express Checkout, Some products are out of stock)

![pp_error](https://user-images.githubusercontent.com/10270387/27832839-e9947882-60d8-11e7-9ecb-646d116dc2c9.jpg)

3. Order confirmation is not sent
4. PayPal payment is successful, but order doesn't contain invoice
5. "Hero Hoodie" Product with configurations (Black, XS) now 200 at stock

![3c67d34edf](https://user-images.githubusercontent.com/10270387/27832933-44aad8c4-60d9-11e7-91ab-0ea00bcb38c7.jpg)

**Next attempts with another size:**
1. User is redirected to /paypal/express/review/
2. "We can't place the order." error displayed
3. Order confirmation is not sent
4. PayPal payment is successful, but order is not created
5. Stock qty is correctly decreased to 0

YevSent:
Hi, @farior, please, enable debug mode for PayPal Express Checkout and attach `debug.log` and `exception.log` to this ticket.

farior:
@joni-jones  Here are debug.log and system.log (renamed with txt extension to allow github to attach it)

exception.log was empty

[debug.txt](https://github.com/magento/magento2/files/1124168/debug.txt)
[system.txt](https://github.com/magento/magento2/files/1124170/system.txt)

vishveskrish:
Hi @joni-jones 

We are facing the same issue on frontend site. When customers try to place an order with equal numbers of quantity persist in the system.
@farior did you found any patch for that?

Let us know the solution ASAP!

[...]

Issue #9116, opened 2017-04-03
PayPal Error -  'Unable To Place Order' Message - M2.1.5


col-be:
### Preconditions
1.  Install Magento 2.1.5 in standard environment with php 7.0.15
2. Setup PayPal Express in Live mode with Payment Action set to 'Sale', PayPal button displayed on shopping cart, Skip Order Review Step set to 'Yes'.  

### Steps to reproduce
1. Add item with custom options to shopping cart.
2. Proceed to shopping cart page.
3. Proceed to checkout.
4. Complete Shipping and Billing fields and select shipping method.
5. Proceed to payment screen, select Pay With Paypal as payment option.
6. Select Continue to Paypal.
7. Redirected to Paypal.
8. Log in to PayPal.
9. Confirm order with 'Pay Now' button.

### Expected result

1. User redirected to order success screen with order confirmation and order ID.
2. User receives order confirmation email.
3. Order recorded in Magento Admin.
4. User's PayPal account debited.
5. Stock Quantity is reduced.

### Actual result

1. User presented with message indicating that product is out of stock and that we are 'Unable to place the order.'
2. Order recorded in admin with correct order ID.
3. Stock quantity is reduced.
4. User's PayPal account is debited.
5. Order confirmation email is NOT sent.
6. Customer sends angry email confused as to what has actually happened.

This was apparently present in Magento 2.1, highlighted here in issue [5327](https://github.com/magento/magento2/issues/5327) but fixed in [MAGETWO-56699](https://github.com/magento/magento2/commit/b9d96707a7f0d9d7f599812440f20344f1d45b91).

But we are still seeing it in 2.1.5.  In our installation, the relevant files (e.g. app/code/Magento/Catalog/Model/Product.php) include the changes made in commit 56699.

We have just upgraded from Magento 1.7.0.2, and feeling bugged out that we are still having to deal with issues in M2.

Any ideas?

YevSent:
Hi, @col-be, please provide additional details, like transaction details log, browser console log and order status.
I've checked on Magento 2.1.5 version with a configurable and simple product and order was placed successfully.

korostii:
There seems to exist an inconsistency in processing exceptions by Magento2 if it happens at a certain point during the creation an order. 
Namely, the user message says that the order was not created - which, in this case, is wrong.
This was also mentioned in [#8978](https://github.com/magento/magento2/issues/8978#issuecomment-291261988)

col-be:
[...]

PR #10511, opened 2017-08-12
PayPal checkout fails if twice the required stock is not available

driskell:
### Description
When checking out using PayPal payment method in checkout, you will be thrown back to the "Review" screen with "We can't place the order" message if one or more items in the cart are ordering half of the available stock. For example, "RED BOOT" is in the cart with quantity 1, and there is only 2 left in stock.

#10512 is a continuation of this as there is another section of code that triggers same error with slightly different conditions - when ordering the last item in stock. #10511 and #10512 together have resolved and stabilised all of our PayPal transactions to date.

Part of #9116 
Possibly #10142 

### Manual testing scenarios
1. Create configurable product with variations RED and BLUE color
2. Set RED variation to stock level 2
3. Add RED to cart and proceed to checkout and then at payment method stage select PayPal and proceed to pay.

### Expected
1. Successful checkout
2. Stock of RED becomes 1

### Actual
1. We can't place the order - PayPal review step shown
2. Order is actually recorded unbeknownst to the user
3. Stock of RED becomes 3

### Rationale
The quote object is saved via repository TWICE during PayPal checkout.
Saving a quote via repository triggers CartItemPersister of all items.
In the first save, the product options of the items are not present since no changes were made to the items. Thus the CartItemOptionsProcessor returns just the quantity and no buyRequestData. (This could be an issue itself but it seems to work OK.) During this first save the CartItemPersister at the end, however, then adds the product options to the item. As with any quote save, the stock is also validated.
Then, during the second save via repository, because product options now exist, updateItems is called for every configurable product in the cart.
Because the buyRequestData passed to updateItem does not contain the ID of the item like it would if it was triggered from a cart edit, it results in the reset quantity flag being ignored, thus, the quantity of the item within the cart is increased to 2 and a stock check for 2 is performed.
PayPal checkout has a quote save early on and then it has one in Quote->submit AFTER order placement. So at this point, the RED stock is now 1, but then a stock check is performed for 2 items, and throws an exception post order placement, causing the redirect to the review screen with error message. In fact, this then triggers a return to stock of the items in the quote for some reason, resulting in 3 RED items now in stock...

### Contribution checklist
 [x] Pull request has a meaningful description of its purpose
 [x] All commits are accompanied by meaningful commit messages
 [ ] All new or changed code is covered with unit/integration tests (if applicable)
 [ ] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Quote/Model/Quote/Item/CartItemPersister.php
+++ b/app/code/Magento/Quote/Model/Quote/Item/CartItemPersister.php
@@ -69,6 +69,7 @@ class CartItemPersister
                 $buyRequestData = $this->cartItemOptionProcessor->getBuyRequest($productType, $item);
                 if (is_object($buyRequestData)) {
                     /** Update item product options */
+                    $buyRequestData->setId($itemId);
                     $item = $quote->updateItem($itemId, $buyRequestData);
                 } else {
                     if ($item->getQty() !== $currentItem->getQty()) {


YevSent:
Hi, @driskell. We are already working on the fix related to issue with PayPal Express Checkout and configurable products. As I see, proposed changes do not solve the problem with out of stock products on payment review page (our fix wider). Did you test proposed changes on the place order flow with a bundle and configurable products?

driskell:
We tested thoroughly this in combination with #10512 with configurable products only.

Together they solved the error for ordering down to the last item in stock. We did not test ordering out of stock item.

driskell:
@joni-jones Is your wider fix visible in the repository? I'd be super happy to test it and compare it with our discoveries. It'll also help us keep in-line with upstream.

YevSent:
@driskell, now our fix only in the internal repository. We are preparing PR to deliver it.

ishakhsuvarov:
Hi @driskell 
Did you have a change to check the fix provided by @magento-team ?

ishakhsuvarov:
@driskell Closing this PR for now. Please reopen if issue is not fixed with the solution proposed in `2.2.0-preview`.
Thank you for collaboration.

ebaschiera:
Hi @ishakhsuvarov ,
what are the fixes proposed in `2.2.0-preview`? Can you provide a link to a commit please?
This issue feels like a dead end to me.

Thank you.

PR #10512, opened 2017-08-12
PayPal checkout causes quote validation after order placement

driskell:
### Description
A continuation of #10511 but generally a different change so raising separately. They are closely related, however.

Part of #9116
Possibly #10142

### Manual testing scenarios
See #10511

### Rationale
Even after #10511 is committed there is still the issue you can't order the last item in stock. So if quantity is 1 and you order that last item you have the same issue. This is a slightly different cause and fix even though the issue is generally the same.
This commit just stops the quote being validated and stock checked AFTER the order was placed, which is the issue. If order is placed for last item in stock, the stock is now 0. Thus validating the quote stock again (which saving via repository does) will throw an exception that stock is not available.
However, save is REQUIRED since we have to persist the inactive state of the quote.
Fix here is to save by resource so it does no validation, since the order is placed and the quote is already validated earlier on via a prior save.

### Contribution checklist
[x] Pull request has a meaningful description of its purpose
[x] All commits are accompanied by meaningful commit messages
[ ] All new or changed code is covered with unit/integration tests (if applicable)
[ ] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Quote/Model/QuoteManagement.php
+++ b/app/code/Magento/Quote/Model/QuoteManagement.php
@@ -508,7 +508,7 @@ class QuoteManagement implements \Magento\Quote\Api\CartManagementInterface
                     'quote' => $quote
                 ]
             );
-            $this->quoteRepository->save($quote);
+            $quote->getResource()->save($quote);
         } catch (\Exception $e) {
             if (!empty($this->addressesToSync)) {
                 foreach ($this->addressesToSync as $addressId) {


fooman:
Thanks @driskell for providing a fix for the issue you encountered. Great detective work!

I have a different Pull Request [here](https://github.com/magento/magento2/pull/9937) which goes the other direction - using the Repository even more as direct use of the resource is being deprecated.

Do you have any other ideas on how we could solve this while still allowing save to be called multiple times on the repository? Maybe the inventory check should just not run on a quote which is about to be inactivated?

ishakhsuvarov:
@fooman @driskell Indeed, increasing API usage is a suggested way to go. 
Currently similar issue is in progress by magento team (https://github.com/magento/magento2/pull/10511#issuecomment-322695713), I would suggest checking out the fix when it's available, since it may be covering both #10511 and #10512. It should be available in `2.2-preview` soon, I will additionally notify in both threads.

ishakhsuvarov:
Hi @driskell 
Fix implemented by the @magento-team is already delivered into branch `2.2.0-preview`. Please check if it solves the issue.

ishakhsuvarov:
@driskell Closing this PR for now. Please reopen if issue is not fixed with the solution proposed in `2.2.0-preview`.
Thank you for collaboration.

Issue #7931, opened 2016-12-22
Unable to send the cookie. Maximum number of cookies would be exceeded


gputignano:
It happens on a fresh magento v2.1.3 installation via composer.
Apache/2.4.23 (Win32) OpenSSL/1.0.2h PHP/5.6.28
Versione del client del database: libmysql - mysqlnd 5.0.11-dev - 20120503
Estensione PHP: mysqli Documentazione
Versione PHP: 5.6.28

When I try to access admin area I get:

```
There has been an error processing your request
Exception printing is disabled by default for security reasons.
Error log record number: 1761570144
```

In var/report I get:

```
a:4:{i:0;s:71:"Unable to send the cookie. Maximum number of cookies would be exceeded.";i:1;s:5183:"#0 C:\xampp\htdocs\magento2\vendor\magento\framework\Stdlib\Cookie\PhpCookieManager.php(116): Magento\Framework\Stdlib\Cookie\PhpCookieManager->checkAbilityToSendCookie('admin', 'luov8matk9kh36a...')
#1 C:\xampp\htdocs\magento2\vendor\magento\framework\Stdlib\Cookie\PhpCookieManager.php(98): Magento\Framework\Stdlib\Cookie\PhpCookieManager->setCookie('admin', 'luov8matk9kh36a...', Array)
#2 C:\xampp\htdocs\magento2\vendor\magento\module-backend\Model\Auth\Session.php(177): Magento\Framework\Stdlib\Cookie\PhpCookieManager->setPublicCookie('admin', 'luov8matk9kh36a...', Object(Magento\Framework\Stdlib\Cookie\PublicCookieMetadata))
#3 C:\xampp\htdocs\magento2\vendor\magento\framework\Interception\Interceptor.php(74): Magento\Backend\Model\Auth\Session->prolong()
#4 C:\xampp\htdocs\magento2\vendor\magento\framework\Interception\Chain\Chain.php(70): Magento\Backend\Model\Auth\Session\Interceptor->___callParent('prolong', Array)
#5 C:\xampp\htdocs\magento2\vendor\magento\framework\Interception\Interceptor.php(138): Magento\Framework\Interception\Chain\Chain->invokeNext('Magento\\Backend...', 'prolong', Object(Magento\Backend\Model\Auth\Session\Interceptor), Array, 'security_admin_...')
[...]

PR #10516, opened 2017-08-12
Update Cookie Manager library to make max cookie number configurable

bartoszherba:
# Configurable max cookie number

### Description
Replace max_cookie_number constant with a value configurable from admin panel.
Default value is set to 50.

### Manual testing scenarios
1. To change maximum cookies number in admin panel go to: Stores -> Configuration -> Web -> Default Cookies Settings -> Cookie Max Number and set required value

2. Use following script in console to bulk create cookies:
for (let i = 0; i < 50; i++) {
 	let t = Math.random().toString(36).substring(7); 
        document.cookie = t+"="+t;   
}

3. Test if exception has been thrown.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cookie/etc/adminhtml/system.xml
+++ b/app/code/Magento/Cookie/etc/adminhtml/system.xml
@@ -22,14 +22,18 @@
                     <label>Cookie Domain</label>
                     <backend_model>Magento\Cookie\Model\Config\Backend\Domain</backend_model>
                 </field>
-                <field id="cookie_httponly" translate="label" type="select" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="1">
+                <field id="cookie_max" translate="label" type="text" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="0">
+                    <label>Cookie Max Number</label>
+                    <validate>validate-digits</validate>
+                </field>
+                <field id="cookie_httponly" translate="label" type="select" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="1">
                     <label>Use HTTP Only</label>
                     <comment>
                         <![CDATA[<strong style="color:red">Warning</strong>:  Do not set to "No". User security could be compromised.]]>
                     </comment>
                     <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                 </field>
-                <field id="cookie_restriction" translate="label" type="select" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="0" canRestore="1">
+                <field id="cookie_restriction" translate="label" type="select" sortOrder="60" showInDefault="1" showInWebsite="1" showInStore="0" canRestore="1">
                     <label>Cookie Restriction Mode</label>
                     <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     <backend_model>Magento\Cookie\Model\Config\Backend\Cookie</backend_model>
--- a/lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php
+++ b/lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php
@@ -7,9 +7,11 @@
 namespace Magento\Framework\Stdlib\Cookie;
 
 use Magento\Framework\App\ObjectManager;
+use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Exception\InputException;
 use Magento\Framework\Stdlib\CookieManagerInterface;
 use Magento\Framework\Phrase;
+use Magento\Store\Model\ScopeInterface;
 use Magento\Framework\HTTP\Header as HttpHeader;
 use Psr\Log\LoggerInterface;
 
@@ -20,26 +22,32 @@ use Psr\Log\LoggerInterface;
  * sensitive data so that extra protection can be added to the contents of the cookie as well as how the browser
  * stores the cookie.
  *
+ * RFC 2109 - Page 15
+ * http://www.ietf.org/rfc/rfc6265.txt
+ *
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 class PhpCookieManager implements CookieManagerInterface
 {
-    /**#@+
-     * Constants for Cookie manager.
-     * RFC 2109 - Page 15
-     * http://www.ietf.org/rfc/rfc6265.txt
+    /**
+     * Max cookie size
      */
-    const MAX_NUM_COOKIES = 50;
     const MAX_COOKIE_SIZE = 4096;
-    const EXPIRE_NOW_TIME = 1;
-    const EXPIRE_AT_END_OF_SESSION_TIME = 0;
-    /**#@-*/
 
-    /**#@+
+    /**
+     * Expire now time
+     */
+    const EXPIRE_NOW_TIME = 1;
+
+    /**
+     * Expire at the end of session
+     */
+    const EXPIRE_AT_END_OF_SESSION_TIME = 0;
+
+    /**
      * Constant for metadata array key
      */
     const KEY_EXPIRE_TIME = 'expiry';
-    /**#@-*/
 
     /**
      * @var CookieScopeInterface
@@ -50,6 +58,10 @@ class PhpCookieManager implements CookieManagerInterface
      * @var CookieReaderInterface
      */
     private $reader;
+    /**
+     * @var ScopeConfigInterface
+     */
+    private $scopeConfig;
 
     /**
      * Logger for warning details.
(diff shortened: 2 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10516) <br/>All committers have signed the CLA.

orlangur:
> At any time, the user agent MAY "remove excess cookies" from the
   cookie store if the number of cookies sharing a domain field exceeds
   some implementation-defined upper bound (such as 50 cookies).

 > At any time, the user agent MAY "remove excess cookies" from the
   cookie store if the cookie store exceeds some predetermined upper
   bound (such as 3000 cookies).

Why did you remove reference to RFC and for which practical purpose do you need more than 50 cookies?

bartoszherba:
@orlangur 
1. I removed rfc by mistake while refactoring consts docblocks according to M2 code style guide. My mistake.

2. I had few projects with live chats and some additional plugins/modules that was using a lot of cookies and therefore I had to increase this limit. Since PhpCookieManager is a Lib class and there is no possible way to extend it and change the max cookie limit I always was overriding it by copy-paste most of the code and set preference. I do not want to mess like that with any lib/core classes so either Manager should be extendable or configurable. According the M2 team to core libs extensibility the latest approach is best.

So, in short, this change is required by projects development practice. Browsers can manage more than 50 cookies and it should be configurable because overriding lib classes yield hard to maintain app.

One of the feature request I found is https://github.com/magento/magento2/issues/7931

orlangur:
Thanks for the reply.

> refactoring consts docblocks according to M2 code style guide

Where is that from? `/**#@+` docblocks are perfectly valid for constant grouping AFAIK.

I have no idea why it was implemented as `PhpCookieManager::MAX_NUM_COOKIES` currently and not `self::MAX_NUM_COOKIES`. I would prefer it as `static::MAX_NUM_COOKIES` rather than Admin UI field so that it can be easily overridden  by inheritance (which is a discouraged approach in M2, but using cookies that much instead of local storage is an outdated approach as well). 

Let's see what other guys think (Cc: @akaplya maybe you remember decisions behind https://github.com/magento/magento2/commit/c945dbe42e5947e29fb977fd44300730e5c46bb7).

bartoszherba:
@orlangur 
regarding const please refer to http://devdocs.magento.com/guides/v2.0/coding-standards/docblock-standard-general.html, the consts section

As for the "static approach", notice that methods are checking that cookie can be set are private and therefore will be run from parent scope. Therefore static will not get overridden const value. And, well, const should be a const, changing its value like this smells like a bad approach for me plus I totally agree with limiting inheritance possibilities which leads to a better architecture but this, on the other hand requires, that classes are following SOLID rules. Cookie manager in current shape is nothing like that. Eventually, we can make max_cookie_number as a constructor's configurable variable (DI approach) but then it still requires a developer to produce at least a module with relevant di.xml
From my point of view in context of projects, customers needs and experience, making max cookie configurable is desired.

orlangur:
Thanks, I somehow missed when http://devdocs.magento.com/guides/v2.2/coding-standards/docblock-standard-general.html#DocBlock-templates was changed in https://github.com/magento/devdocs/commit/7e31ce6a12a5a366440ab1efec731ccd190f54c1.

driskell:
I think we're missing one possible issue here. I've had this occur where a new Magento 2 site went live and the previous site did not care much for the cookie limit. Furthermore, when there are many subdomains active on the domain storing cookies at the domain root it can further increase cookie counts. It does not give a good impression to launch a new site and it simply fail as soon as you change the hostname.

Would be great to get this merged.

orlangur:
@driskell my main concern here about changing `50` to any value (no matter via Admin UI or some other approach) is that I have no idea which is the real limit modern browsers support starting from which cookies will start to be dropped off. Hard limit encourage developers to not bloat the cookies by using local storage or something.

driskell:
@orlangur I can agree with that. I have been thinking about this a little and I wonder if the issue is not the prevention attempt, it's the action taken to prevent.

I am thinking that in any scenario this occurs the problem is not being prevented. If you have 50 cookies already the horse has already bolted.

So I can understand this only as a developer deterrence during development. Which makes me think this should be a developer only mode feature.

Thinking further if we want to prevent issues on Production mode due to limit than maybe not failing is a better option. Instead log some warning maybe even flagging a cache entry to trigger an admin message, "Some users are experiencing too many cookies - please investigate your cookie usage."
It could then possibly have a whitelist of critical cookies to set and refuse setting non-essential ones to try keep under the limit (one would hope the essential session cookie is sent first so always gets a slot.)
I actually think this kind of message would be far better than waking up the ops team at 3am with a sudden burst of 503 errors they can do absolutely nothing about without extensive development work to work out what cookies are on that persons machine and not other peoples (we are finding that due to age of the domain the number of cookies wildly varies machine to machine in our offices alone.)

orlangur:
@driskell this PR is about max cookie number only, does current logic https://github.com/bartoszherba/magento2/blob/885d2e38bc1945556fc25c2ded0521d9238f7da5/lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php#L238 cause 503 in production?

Or do you refer to the maximum cookie size exception? It is not avoidable I believe: https://www.google.com/search?q=maximum+cookie+size

driskell:
Ahhhhh!

https://github.com/magento/magento2/commit/1598c0fa816be99cef010721792755fc7506eced

Perfect! I am using 2.1, great to see it fixes in 2.2 :) Hopefully OP will like this too.

(Just edited - noticed the link wasn't on the magento2 upstream repo - it's definitely in 2.2!)

orlangur:
@driskell nice catch!

@bartoszherba so, "Test if exception has been thrown" does not occur on `develop` in fact, are you satisfied with warning logged?

app/code/Magento/Cookie/etc/adminhtml/system.xml
```diff
@@ -22,14 +22,18 @@
                     <label>Cookie Domain</label>
                     <backend_model>Magento\Cookie\Model\Config\Backend\Domain</backend_model>
                 </field>
-                <field id="cookie_httponly" translate="label" type="select" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="1">
+                <field id="cookie_max" translate="label" type="text" sortOrder="40" showInDefault="1" showInWebsite="1" showInStore="0">
```
ishakhsuvarov, line 25: This setting is most likely to be set just once, so there is not really a reason to have it in the admin. Consider implementing this as a configuration file.

lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php
```diff
@@ -20,26 +22,32 @@
...
 class PhpCookieManager implements CookieManagerInterface
 {
-    /**#@+
-     * Constants for Cookie manager.
-     * RFC 2109 - Page 15
-     * http://www.ietf.org/rfc/rfc6265.txt
+    /**
+     * Max cookie size
      */
-    const MAX_NUM_COOKIES = 50;
```
ishakhsuvarov, line 32: Removing a constant is a backwards incompatible change, since some third-party customization could rely on it. Please revert it and add a deprecated annotation.

lib/internal/Magento/Framework/Stdlib/Cookie/PhpCookieManager.php
```diff
@@ -70,17 +82,20 @@ class PhpCookieManager implements CookieManagerInterface
...
      * @param HttpHeader $httpHeader
+     * @param ScopeConfigInterface $scopeConfig
      */
     public function __construct(
         CookieScopeInterface $scope,
         CookieReaderInterface $reader,
         LoggerInterface $logger = null,
-        HttpHeader $httpHeader = null
+        HttpHeader $httpHeader = null,
+        ScopeConfigInterface $scopeConfig
```
ishakhsuvarov, line 92: Adding a required constructor parameter is a backwards incompatible change, which may affect classes extending this one. Please check code example in [Backwards Compatible Development Guide](http://devdocs.magento.com/guides/v2.1/ext-best-practices/extension-coding/backwards-compatible-development/index.html#adding-a-constructor-parameter) and fix accordingly.

ishakhsuvarov (changes requested):

ishakhsuvarov:
@bartoszherba Please let us know if you would like to continue with this Pull Requests, since looking at the conversation here it looks like the initial problem is solved.

bartoszherba:
Sorry for delay.
Considering all arguments and changes already made I am happy with the current solution and do not feel like continuing this PR is still required.

orlangur:
@bartoszherba thanks for collaboration and picking up the best solution for practical purposes!

ishakhsuvarov:
@bartoszherba Thanks for collaboration!

PR #10521, opened 2017-08-13
Set correct Content-Type header for ajax responses in the Magento_UI module.

hostep:
### Description
We started noticing issues when New Relic's Browser profiling was enabled, which inserts a javascript block into `text/html` responses whenever a `<head>` tag was found in such a response.

This has the possibility to break certain ajax responses in Magento's backend, since they have Content-Type: `text/html` instead of `application/json`.

I'm not sure if this fix is the correct fix, and if there aren't other violations in the code for this same problem. Feel free to let me know if that should be the case and I'll try to add them to the PR.

None that I could find

### Manual testing scenarios
1. Tested on Magento CE 2.1.7 and develop branch (rev: 45896ba)
2. Make sure you have a New Relic integration setup with your server and that the Browser Profiling is enabled.
3. Edit or create a new CMS Page and add the following to the 'Layout Update XML' field:
```xml
<head>
    <meta name="robots" content="NOINDEX,NOFOLLOW"/>
</head>
```
4. Save the page and return to the cms pages list
5. The cms page grid no longer loads and an error: "Something went wrong" appears
screen shot 2017-08-13 at 17 02 23
6. Inspecting the response, it contains invalid json, because New Relic inserted its javascript in there:

```
...
"layout_update_xml":"<head><script type="text/javascript">window.NREUM||(NREUM={}) ... </script>\r\n    <meta name=\"robots\" content=\"NOINDEX,NOFOLLOW\"\/>\r\n<\/head>"
...
```

Changing the Content-Type to `application/json` no longer makes New Relic insert its javascript into the response and the grid loads perfectly fine.

### Contribution checklist


app/code/Magento/Ui/Controller/Adminhtml/Index/Render.php
```diff
@@ -30,6 +30,7 @@ public function execute()
         $component = $this->factory->create($this->_request->getParam('namespace'));
         $this->prepareComponent($component);
         $this->_response->appendBody((string) $component->render());
+        $this->_response->setHeader('Content-Type', 'application/json', true);
```
ishakhsuvarov: Main problem with hard-coding `Content-Type` here is that the `$component->render()` has multiple types of renderers. 

You may refer to `\Magento\Ui\Component\AbstractComponent::render` which in place selects the render engine from the classes which implement `\Magento\Framework\View\Element\UiComponent\ContentType\ContentTypeInterface`.

This list is not limited to Json, but also has an Html and Xml implementations out of the box.
I suggest looking into different approach to solve this issue, such as: 
* Possibly limiting the `Content-Type` header to the correct renderer
* Working with the `Accept` header provided by client to ensure the correct response.

ishakhsuvarov (dismissed):

app/code/Magento/Ui/Controller/Adminhtml/Index/Render.php
hostep: Good point, I wasn't sure if all responses were in json format (didn't take my time to look it through, my mistake).
I'll have a look at a better solution.

hostep:
I force pushed a new commit, this is probably a better way to handle it.
I only tested json responses, couldn't quickly find xml or html responses, so that's not tested.

Comments are welcome.

ishakhsuvarov:
Hi @hostep 
After discussion with @omiroshnichenko and @okorshenko we found that we could add some more flexibility to the proposed solution. I've pushed the proposed changes back to this PR, if you don't mind. 

Please let me know what you think of this. 
Thank you.

hostep:
@ishakhsuvarov: nice, looks good.

So it looks like it is now based on the `Accept` header? That should probably work fine, I just tested (without this patch) the CMS Pages Grid, and the request includes `application/json` in the `Accept` header, so it looks like this will fix the bug I described in the initial post.

Good job, and thank you! :)

ishakhsuvarov:
`Accept` header seems logical to follow in this case, since not following it violates standards anyway. I've tested some cases, including the one you have described in the issue and it seems to work fine.

PR #10537, opened 2017-08-15
Align Magento\Framework\Mail\Message with MessageInterface

navarr:
### Description

MessageInterface allows for an array, but implementation does not.  These changes align the implementation with the interface in such a way that aligns with other methods (such as the addTo method).

This allows for all the following to be possible without a bc break:

* `->setFrom('support@magento.com')`
* `->setFrom(['support@magento.com'])`
* `->setFrom(['Magento Support' => 'support@magento.com'])`

There is no public issue for the bug of the interface not matching the implementation.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Mail/Message.php
+++ b/lib/internal/Magento/Framework/Mail/Message.php
@@ -34,6 +34,23 @@ class Message extends \Zend_Mail implements MessageInterface
     {
         return $this->messageType == self::TYPE_TEXT ? $this->setBodyText($body) : $this->setBodyHtml($body);
     }
+    
+    /**
+     * Set from address
+     *
+     * @param string|array $email
+     * @return $this
+     */
+    public function setFrom($email)
+    {
+        $name = null;
+        if (is_array($email)) {
+            $name = array_keys($email)[0];
+            $name = !is_int($name) ? $name : null;
+            $email = array_values($email)[0];
+        }
+        return parent::setFrom($email, $name);
+    }
 
     /**
      * Set message body


lib/internal/Magento/Framework/Mail/Message.php
```diff
@@ -34,6 +34,23 @@ public function setBody($body)
...
+     * @param string|array $email
+     * @return $this
+     */
+    public function setFrom($email)
+    {
+        $name = null;
+        if (is_array($email)) {
+            $name = array_keys($email)[0];
+            $name = !is_int($name) ? $name : null;
+            $email = array_values($email)[0];
```
orlangur, line 50: This is `O(n)` while can be `O(1)` quite easily.

Something like
```
foreach ($email as $key => $value) {
    $name = is_int($key) ? null : $key;
    $email = $value;
    break;
}
```
or maybe even 
```
foreach ($email as $name => $email) {
    break;
}
$name = !is_int($name) ? $name : null;
```
should do the trick 😉 

Unit test maybe?

orlangur (changes requested):

orlangur:
The main issue with this PR is that we shall not directly change `2.1-develop`, only backport some fixes from `develop`. @navarr, what do you think?

navarr:
@orlangur If that's policy there's not much that can be done - however, 2.2 moved from ZF1 to ZF2 - where this problem likely doesn't occur, meaning that 2.1 will never be able to send emails with a from name via the service contracts - which is sad.

orlangur:
Yeah, I wrote my comment after I checked `develop` branch.

Thanks for your efforts, but I believe we should not misuse the process introduced, there are a lot of bugs still present in `2.1.x` and it's better to concentrate on backporting rather than on improvements.

`2.2` to be released soon anyway.

lib/internal/Magento/Framework/Mail/Message.php
kandy: Or
reset($email);
list($name , $email) = each($email);
;)

navarr:
With that in mind I'll retract my PR.

PR #10563, opened 2017-08-17
Modified Bundle.js because of breaking Encoding in Production Mode.

ulftietze:
### Description

This pull request prevents wrong or too much encoding of UTF-8 Files.

1. magento/magento2#10562: Merging / Bundling JS Files Production Mode breaks Regular Expressions
2. maybe magento/magento2#6733: Bad ukrainian character encoding in minicart when Enabled JavaScript Bundling in production mode

### Manual testing scenarios

1. Reproduce the issue #10562


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10563) <br/>All committers have signed the CLA.

ulftietze:
Hey @ishakhsuvarov,

could you restart the travis-ci integration test? Just one of them failes with a timeout ^^

Thanks, Ulf

ishakhsuvarov:
@ulftietze Restarted.
Please consider covering changes with tests.

ulftietze:
Hey @ishakhsuvarov,

same problem as last time, the Job gives us a timeout. Do you know, why this is happening?

I tried to modify the test for this class, but unfortunatly there are only a test which tests perhaps ~20% of the Bundle.php Class.
Also it's an deprecated class and i did just a small change for this class. May we can agree, that a full unit test (which i could do but with a ton of mocking) is a little bit overhead.

Thanks, Ulf

ishakhsuvarov:
@ulftietze We are looking into functional test timeouts. I've restarted the job again.
Could you please provide some manual test cases which are fixed with this PR?

kkkonrad:
can you check validation messages in customer registration page? looks like on this page validation messages not work when store is on Production mode+enabled all JS optimizaton, i test it on 2.1.3, 2.1.8 and 2.2.0-rc and looks like never get simple translation "This is required field."

ulftietze:
Moin @kkkonrad,

i tried to reproduce your issue with my fix but i just can't :/ I think, that this could be a conflict with another corebug i think. This could possible be #7862.

Greets, Ulf

ulftietze:
Moin @ishakhsuvarov,

i don't know, how explict you are expecting to get the tests. But i did a very short script, where you explictly can see the difference. Or at least the problem :)

```
<?php

$jsToParse = 'if (!(value.match(/^([a-zA-ZßäöüÄÖÜ\s.]{3,})+$/))) {
                return this.customValidationError($t(\"Please provide a correct string.\"));
           }';

print_r(mb_convert_encoding($jsToParse, "UTF-8") . PHP_EOL);
print_r(utf8_encode($jsToParse) . PHP_EOL);
```
This script has the following output:
```
// Correct:
if (!(value.match(/^([a-zA-ZßäöüÄÖÜ\s.]{3,})+$/))) {
                return this.customValidationError($t(\"Please provide a correct string.\"));
           }
// Wrong:
if (!(value.match(/^([a-zA-ZÃÃ¤Ã¶Ã¼ÃÃÃ\s.]{3,})+$/))) {
                return this.customValidationError($t(\"Please provide a correct string.\"));
           }
```

Greets, Ulf

ishakhsuvarov:
@ulftietze Thanks for providing details.
I could not see any effect from the fix on the latest develop branch.

I assume this happens because `\Magento\Framework\View\Asset\Bundle` is deprecated and unused. Moving the fix to the `\Magento\Deploy\Package\Bundle\RequireJs` however, seems to solve the issue. Could you please re-check this?

PR #10570, opened 2017-08-17
add subnet mask IP support for maintenance mode

AntonEvers:
### Description

Maintenance mode should accept IP ranges instead of separate IP's so that for instance you can add the whole range of your payment service provider or CDN etc...

1. No issues so far
### Manual testing scenarios

1. run `bin/magento maintenance:enable`
2. run `bin/magento maintenance:allow-ips 192.168.33.0/16` (for instance)
3. get access to the site

Mostly I am curious what you think about this kind of functionality and the use of darsyn/ip for this. I did take IPv6 into account. 

### Contribution checklist


ishakhsuvarov:
@ajpevers I like the idea, however I'm not really a fan of adding additional dependency to solve a single problem, since it requires an adapter on magento side, to make sure it can be replaceable and testable.
Did you consider implementing cidr check with plain php and maybe existing dependencies?
Also please check the merge conflict in `composer.lock`.

AntonEvers:
Sure, it's a proof of concept. I'll see if I can get it to work with existing dependencies. Doesn't look like very exotic functionality so I expect it would be somewhere in ZF or Symfony or something. As long as it works with IPv6 as well. I'll come back on this.

AntonEvers:
Hi @ishakhsuvarov I took a good hard look but there is no existing code that does the cidr check for IPv6. So I included `symfony/http-foundation` to take care of this in favor of `darsyn/ip`. It's more elegant an fits in with the already existing Symfony components dependencies.

The IpValidator could be built using the existing Zend_Validate_Ip.

What do you think, should we just copy the one file (https://github.com/symfony/http-foundation/blob/master/IpUtils.php) and risk missing bug fixes, or should we include the package as part of the dependencies?

ishakhsuvarov:
@ajpevers I don't think just copying a single file is a good idea, however adding a dependency requires a wrapper on the Magento side, which would allow to use any different dependency in place without breaking existing API.
Also I would suggest to avoid using outdated Zend 1 components, since it is end-of-life. Actually there are some PR's around which replace those usages to a newer versions.

AntonEvers:
Okay, I'll work it out. Can you point out an existing wrapper so that I can follow the example?

ishakhsuvarov:
@ajpevers I guess more common name would be Adapter, you may refer to the wiki for the basic [description](https://en.wikipedia.org/wiki/Adapter_pattern).
Some of the existing which first came to mind – would be `\Magento\Framework\Serialize\SerializerInterface` which is simply a wrapper for php's built-in `json_encode`, `serialize` calls, or some other implementations provided by third-parties. Or `\Magento\Framework\Code\Minifier\AdapterInterface` which is an adapter for different implementations of minifiers.

AntonEvers:
Great, that is clear enough

orlangur:
I don't think adapter is suitable here, looks more like a https://en.wikipedia.org/wiki/Facade_pattern to me.

So, the idea is that Symfony's `IpUtils` must be encapsulated in singe Magento Framework component and not referred anywhere outside of this component.

For example, it can be `\Magento\Framework\HTTP\IpChecker` class with whatever methods are best suitable for Magento (I don't think we need separate method for IP v4/v6 exposed, just a single method `IpChecker->check(...)` seems enough).

ishakhsuvarov:
Hi @ajpevers
Would you like to proceed with this PR? Any help required from our side?

Thanks

AntonEvers:
Hi @ishakhsuvarov,

The thing I need most is spare time :) can you help me with that?
I'm going to finish this today.

Cheers,

Anton

vrann:
@ajpevers can you please merge it with the latest mainline?

AntonEvers:
@vrann done

lib/internal/Magento/Framework/App/Test/Unit/MaintenanceModeTest.php
```diff
@@ -58,7 +58,7 @@ public function testisOnWithIP()
         ];
         $this->flagDir->expects($this->exactly(2))->method('isExist')
             ->will(($this->returnValueMap($mapisExist)));
-        $this->assertFalse($this->model->isOn());
+        $this->assertTrue($this->model->isOn());
```
orlangur, line 61: Why need this change?

setup/src/Magento/Setup/Validator/IpValidator.php
```diff
@@ -66,14 +66,32 @@ public function validateIps(array $ips, $noneAllowed)
...
+        foreach ($ips as $range) {
+            if ($range === 'none') {
+                $this->none[] = $range;
+                continue;
             }
+
+            $cidr = 32;
+            $ip = $range;
+            if (strpos($range, '/') !== false) {
+                list($ip, $cidr) = explode('/', $range);
```
orlangur: cidr? Is there some existing logic in ZF2 or Symfony which can be reused?

setup/src/Magento/Setup/Validator/IpValidator.php
```diff
@@ -66,14 +66,32 @@ public function validateIps(array $ips, $noneAllowed)
...
+            }
+
+            $ipValidator = new \Zend_Validate_Ip(['allowipv6' => true, 'allowipv4' => true]);
+            if (!$ipValidator->isValid($ip)) {
+                $this->invalidIps[] = $range;
+                continue;
+            }
+
+            $ipv4Validator = new \Zend_Validate_Ip(['allowipv6' => false, 'allowipv4' => true]);
```
orlangur: We must not add code using ZF1 classes. Looks like there is ZF2 equivalent https://framework.zend.com/manual/2.1/en/modules/zend.validator.ip.html

setup/src/Magento/Setup/Validator/IpValidator.php
```diff
@@ -66,14 +66,32 @@ public function validateIps(array $ips, $noneAllowed)
...
+            $max = $ipv4Validator->isValid($ip) ? 32 : 128;
```
orlangur: Please choose a more meaningful name. `$maxBits`?

orlangur (changes requested):

setup/src/Magento/Setup/Validator/IpValidator.php
ajpevers: See https://github.com/magento/magento2/pull/10570#issuecomment-323649544

lib/internal/Magento/Framework/App/Test/Unit/MaintenanceModeTest.php
AntonEvers: This test was wrong, it basically stated:
If there is a maintenance flag file AND if there is an empty maintenance op file maintenance mode should be off.

This is not true. If the ip file is present but empty maintenance mode should be on, not off.

orlangur: Oh, nice catch 👍 Just still don't see corresponding changes in implementation which could invert this assertion.

setup/src/Magento/Setup/Validator/IpValidator.php
AntonEvers: Maybe there is, but I havent found it. And it seems overkill to include another class as replacement for a list = explode.

lib/internal/Magento/Framework/App/Test/Unit/MaintenanceModeTest.php
AntonEvers: Wait... this cannot be right... Ha! https://github.com/magento/magento2/pull/10570/files#diff-4b45df6c223e8d98dcc949e45bfc4f1bR73 This is the wrong way around! Sorry.

setup/src/Magento/Setup/Validator/IpValidator.php
orlangur: I mean this logic in general - splitting by slash etc.

orlangur: Still not sure what "cidr" means, https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation looks like after splitting by slash there should be `$networkMask` instead.

AntonEvers: Ah that's what you mean. Yes you are right. I'll rename it.

AntonEvers: I'll call it `$subnetMask` as that is the name I heard most for this.

okorshenko:
Hi @orlangur 
can we accept this PR or are you waiting for additional changes?

orlangur:
@okorshenko waiting, otherwise I would resolve JIRA issue.

lib/internal/Magento/Framework/App/Test/Unit/MaintenanceModeTest.php
orlangur: Just to clarify: this needs to be changed or currently logic is correct? How this test worked with incorrect previous logic though...

orlangur: Ping @ajpevers :)

AntonEvers: Hi @orlangur, sorry for the silence. I need to free some time to test why this worked without the PR and why it needs a change with the PR. I have it on my task list. Keep you posted.

okorshenko:
Hi @ajpevers @orlangur do we have any progress on this? Do you need some help?

orlangur:
Last comment from author was 15 days ago https://github.com/magento/magento2/pull/10570#discussion_r150762199

Basically, the changes seem pretty good to me, just that unit test needs to be investigated.

orlangur:
@ajpevers let's close this for now and reopen when you or me have some new information regarding odd unit test behavior.

lib/internal/Magento/Framework/App/Test/Unit/MaintenanceModeTest.php
AntonEvers: @orlangur I revisited this PR and it appears that this test is testing `!in_array("", [""])` which results in `false`.
So if your IP is `""` (the default value of the first and only `isON()` parameter) and the list of whitelisted IP's is `[""]` then you are marked as whitelisted and maintenance is not on for you. This is wrong.
If maintenance mode is on and the ip whitelist file is present but empty, maintenance mode should be on for everyone and not off. So this test is incorrect. I will fix this by adding an `array_filter` around the whitelisted IP's so that you can never compare against `[""]`, but in that case against `[]`, which will never have a false positive on `in_array` comparisons.

PR #10590, opened 2017-08-18
Removed instantiating new driverpool class when it's already available.

dverkade:
Removed instantiating new driverpool class when it's already available. This change will make Magento a little bit faster since it will not instantiate a new class, but use the class already available. Will also make Magento use slightly less memory.


Ctucker9233:
@magento-team Will this get backported to 2.2?

PR #10593, opened 2017-08-20
Fix issue 10032

will-b:
### Description

This fixes the issue where the link to the download of a backup in the admin gives the wrong file.
The wrong file seems to be the latest backup that was generated.

### Fixed Issues
- #10032

### Manual testing scenarios
1. Create multiple backups via the admin panel.
2. Check that the downloads provided in the backup grid have the correct links

### Contribution checklist

### DIFF
--- a/app/code/Magento/Backup/Model/BackupFactory.php
+++ b/app/code/Magento/Backup/Model/BackupFactory.php
@@ -38,23 +38,20 @@ class BackupFactory
      */
     public function create($timestamp, $type)
     {
-        $backupId = $timestamp . '_' . $type;
         $fsCollection = $this->_objectManager->get(\Magento\Backup\Model\Fs\Collection::class);
         $backupInstance = $this->_objectManager->get(\Magento\Backup\Model\Backup::class);
+
         foreach ($fsCollection as $backup) {
-            if ($backup->getId() == $backupId) {
-                $backupInstance->setType(
-                    $backup->getType()
-                )->setTime(
-                    $backup->getTime()
-                )->setName(
-                    $backup->getName()
-                )->setPath(
-                    $backup->getPath()
-                );
+            if ($backup->getTime() === (int) $timestamp && $backup->getType() === $type) {
+                $backupInstance->setData(['id' => $backup->getId()])
+                    ->setType($backup->getType())
+                    ->setTime($backup->getTime())
+                    ->setName($backup->getName())
+                    ->setPath($backup->getPath());
                 break;
             }
         }
+
         return $backupInstance;
     }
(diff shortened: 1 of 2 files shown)


app/code/Magento/Backup/Test/Unit/Model/BackupFactoryTest.php
```diff
@@ -113,6 +113,13 @@ public function testCreate()
         )->will(
             $this->returnSelf()
         );
+        $this->_backupModel->expects(
+            $this->once()
```
orlangur: Please reformat similar to change in `app/code/Magento/Backup/Model/BackupFactory.php`, current ugly formatting is derived from poor formatting tool.

app/code/Magento/Backup/Model/BackupFactory.php
```diff
@@ -41,20 +41,18 @@ public function create($timestamp, $type)
...
-                $backupInstance->setType(
-                    $backup->getType()
-                )->setTime(
-                    $backup->getTime()
-                )->setName(
-                    $backup->getName()
-                )->setPath(
-                    $backup->getPath()
-                );
+            if (strpos($backup->getId(), $backupId) !== false) {
```
orlangur: Can we replace this condition with more accurate one? In which case previous `$backup->getId() == $backupId` condition didn't work?

For instance, if the string contained additional slash, we should search for `'/' . $backupId` and so on.

orlangur (changes requested):

app/code/Magento/Backup/Model/BackupFactory.php
will-b: Hi,

The reason for using strpos was that the id returned by the collection has the name appended to the time stamp and type.
$backup->getId() ==> 1503256588_nomediasecond (notice no underscore between nomedia and 'second' which is the name of the backup)

Could be a bug in the _generateRow method of Magento\Backup\Model\Fs\Collection.

The files created in the backup folder have the name convention of {{time_stamp}}_{{type}}_{{name}}.tgz

The id returned by the collection does not have the second underscore.

In the mass delete controller action only the time stamp and type is used to identify the backup.
Magento\Backup\Controller\Adminhtml\Index\MassDelete:39
```
 list($time, $type) = explode('_', $id);
 $backupModel = $this->_backupModelFactory->create($time, $type)->deleteFile();
```

I updated the comparison to check for the type and time stamp only like in the mass delete action.

Hope that is ok.

will-b:
@orlangur updated

orlangur (approved):

PR #10607, opened 2017-08-21
FIX quote is_active flag handling

phoenix128:
Wrong "is_active" handling in quote table

### Description
This is not a per se bugfix, but a better `is_active` flag handling suggestion.
We should not check the last active cart, but we should check the last cart and verify if it is active or not.

If, for unexpected reasons, one user has more than one active quotes and then the user completes an order, a previous cart is restored instead of having an empty new cart.

This does not happen in standard scenarios, but it can happen with programmatically created quotes.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Quote/Model/ResourceModel/Quote.php
+++ b/app/code/Magento/Quote/Model/ResourceModel/Quote.php
@@ -91,9 +91,6 @@ class Quote extends AbstractDb
             'customer_id',
             $customerId,
             $quote
-        )->where(
-            'is_active = ?',
-            1
         )->order(
             'updated_at ' . \Magento\Framework\DB\Select::SQL_DESC
         )->limit(
@@ -101,8 +98,7 @@ class Quote extends AbstractDb
         );
 
         $data = $connection->fetchRow($select);
-
-        if ($data) {
+        if ($data && $data['is_active']) {
             $quote->setData($data);
         }
 


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10607) <br/>All committers have signed the CLA.

ishakhsuvarov:
Hi @phoenix128
Please sign the CLA with the email you used for the commit so that we can proceed with processing.
Thanks!

phoenix128:
Hello @ishakhsuvarov ,
just did it. Thanks.

orlangur:
Can this be intentional for some functionality like abandoned shopping carts or persistent shopping cart?

phoenix128:
Good point, but I do not think so.
If you have more than one active quote, as explained in my description, you will have one big side effect: if you complete one order putting your current quote in "non active" status, you will automatically fallback to the last active cart. So you will have a filled cart even after a completed order and this is definitely an unexpected and unwanted behaviour.

I think `is_active` was intended to avoid a double order creation over the same quote. In a standard scenario, the Magento base code is fine, but I experienced major issues while creating quotes programmatically.

You can try by yourself manually modifying the `quote` table for a test putting two quotes belonging to the same user with `is_active = 1`, then place an order.

When you will come back to thank you page, you will see your cart filled with items from a previous cart.

okorshenko:
HI @phoenix128 
Could you please provide code snippet to demonstrate the issue? Also, it is required to cover your fix with the integration test. The test should cover the case you described. But let's start from code snippet to see if the logic is correct. We should not modify the core code if the issue in the 3rd code
Thank you

orlangur:
Closing due to no activity, @phoenix128 please ping us back when you have a code shippet mentioned by Oleksii so that we can continue with this PR processing.

Issue #5262, opened 2016-06-26
[2.1.0] composer install --no-dev should not install development dependencies


hostep:
## Steps to reproduce
1. Install Magento 2.1.0 using composer but use the `--no-dev` flag
2. Look at the `dev/` directory in the filesystem
## Expected result
1. The `dev/` directory should be empty or non-existing
## Actual result
1. The `dev/` directory exists and contains a bunch of development stuff
## Discussion

In an effort to try to find the best options of deploying a Magento 2 installation to a production server, I keep discovering issues which aren't ideal.
As a best practise, we use `composer install --no-dev --prefer-dist --optimize-autoloader` on a production server when deploying. I would expect that no development dependencies gets installed because of this, but this is not true.

This isn't only about the `/dev` directory, but also for example about the `lib/web/css/docs/` directory.
The last one for example causes the `setup:static-content:deploy` command to compile a less file which has absolutely no purpose of existing on a production server while it takes multiple seconds to compile this file, per locale, per theme.
In a very quick test, I discovered that when removing the above directories before executing `setup:di:compile` and `setup:static-content:deploy` makes those two run about 58 seconds faster (on my local machine) then without removing those directories.

Here is a list of files/directories which I believe shouldn't get deployed to a production server:
- .htaccess.sample
- .php_cs
- .travis.yml
- CHANGELOG.md
- CONTRIBUTING.md
- COPYING.txt
- Gruntfile.js.sample
- ISSUE_TEMPLATE.md
- LICENSE.txt
- LICENSE_AFL.txt
- nginx.conf.sample
- package.json.sample
- php.ini.sample
- dev/
- lib/web/css/docs/
- phpserver/
- ...

(the license files, I'm not sure about)

I think most of these files are getting into place because of the [mapping](https://github.com/magento/magento2/blob/216ad8865244877170d8bbf174a5f75d2f48cd43/dev/tests/integration/testsuite/Magento/Framework/Composer/_files/testFromCreateProject/composer.lock#L706) in the `composer.json` file of the `magento2-base` module.
It would be great if those mappings can get split of into a mapping for development purposes and one for all environments.
[...]

PR #10608, opened 2017-08-21
Option to enable template hints on frontend with a configurable parameter

DominicWatts:
The recent addition to allow template hints to be toggled via the CLI is a step in the right direction as doing so via the admin is time consuming.  However frontend development can be streamlined further with this additional feature.  

### Description
Within admin user is given option to enable template hints on frontend with a configurable URL parameter.  So when enabled can pass ?templatehints=[parameter_value] to enable template hints providing they're enabled and the additional new option is enabled.  Also for security the parameter value is configurable.

We often find ourselves in two situations
  -  Dev 1 working on site wants template hints enabled, Dev 2 doesn't
  -  Dev 1 on 1 screen wants template hints, on Dev 1 on screen 2 doesn't

This update easily allows for this.

N/A

### Manual testing scenarios
1. Within Magento backend under Advanced | Developer 
2. Enabled Template Path Hints for Storefront - Yes
2. Enable Hints for Storefront with URL Parameter - Yes
3. Configure Parameter Value - magento
4. On frontend Storeview enable template hints with parameter ?templatehints=magento

### Contribution checklist

### DIFF
--- a/app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
+++ b/app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
@@ -14,6 +14,7 @@ use Magento\Framework\View\TemplateEngineFactory;
 use Magento\Framework\View\TemplateEngineInterface;
 use Magento\Store\Model\ScopeInterface;
 use Magento\Store\Model\StoreManagerInterface;
+use Magento\Framework\App\Request\Http;
 
 /**
  * Class \Magento\Developer\Model\TemplateEngine\Plugin\DebugHints
@@ -58,24 +59,51 @@ class DebugHints
     protected $debugHintsPath;
 
     /**
+     * XPath of configuration of the debug hints show with parameter
+     *
+     *     dev/debug/template_hints_storefront_show_with_parameter
+     *
+     * @var string
+     */
+    protected $debugHintsShowWithParameter;
+
+    /**
+     * XPath of configuration of the debug hints URL parameter
+     *
+     *     dev/debug/template_hints_parameter_value
+     *
+     * @var string
+     */
+    protected $debugHintsParameter;
+
+    /**
      * @param ScopeConfigInterface $scopeConfig
      * @param StoreManagerInterface $storeManager
      * @param DevHelper $devHelper
      * @param DebugHintsFactory $debugHintsFactory
      * @param string $debugHintsPath
+     * @param Http $http
+     * @param string $debugHintsShowWithParameter
+     * @param string $debugHintsParameter
      */
     public function __construct(
         ScopeConfigInterface $scopeConfig,
         StoreManagerInterface $storeManager,
         DevHelper $devHelper,
         DebugHintsFactory $debugHintsFactory,
-        $debugHintsPath
+        $debugHintsPath,
+        Http $http,
+        $debugHintsShowWithParameter,
+        $debugHintsParameter
     ) {
         $this->scopeConfig = $scopeConfig;
         $this->storeManager = $storeManager;
         $this->devHelper = $devHelper;
         $this->debugHintsFactory = $debugHintsFactory;
         $this->debugHintsPath = $debugHintsPath;
+        $this->http = $http;
+        $this->debugHintsShowWithParameter = $debugHintsShowWithParameter;
+        $this->debugHintsParameter = $debugHintsParameter;
     }
 
     /**
@@ -94,15 +122,30 @@ class DebugHints
         $storeCode = $this->storeManager->getStore()->getCode();
         if ($this->scopeConfig->getValue($this->debugHintsPath, ScopeInterface::SCOPE_STORE, $storeCode)
             && $this->devHelper->isDevAllowed()) {
-            $showBlockHints = $this->scopeConfig->getValue(
-                self::XML_PATH_DEBUG_TEMPLATE_HINTS_BLOCKS,
+            $debugHintsShowWithParameter = $this->scopeConfig->getValue(
+                $this->debugHintsShowWithParameter,
                 ScopeInterface::SCOPE_STORE,
                 $storeCode
             );
-            return $this->debugHintsFactory->create([
-                'subject' => $invocationResult,
-                'showBlockHints' => $showBlockHints,
-            ]);
+            $debugHintsParameter = $this->scopeConfig->getValue(
+                $this->debugHintsParameter,
+                ScopeInterface::SCOPE_STORE,
+                $storeCode
+            );
+            $debugHintsParameterInUrl = $this->http->getParam('templatehints');
+            if ((!$debugHintsShowWithParameter) ||
+                ($debugHintsShowWithParameter &&
+                 $debugHintsParameter == $debugHintsParameterInUrl)) {
+                $showBlockHints = $this->scopeConfig->getValue(
+                    self::XML_PATH_DEBUG_TEMPLATE_HINTS_BLOCKS,
+                    ScopeInterface::SCOPE_STORE,
+                    $storeCode
+                );
+                return $this->debugHintsFactory->create([
+                    'subject' => $invocationResult,
+                    'showBlockHints' => $showBlockHints,
+                ]);
+            }
         }
         return $invocationResult;
     }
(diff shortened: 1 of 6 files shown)


orlangur:
Seems overcomplicated to me. Then it's easy to overlook turning this configuration off and it will go to production.

> We often find ourselves in two situations

Why each developer does not use own local environment for development?

Probably if we stick to `xdebug_session` cookie or something it could become something worth merging.

XigenDominic:
I respectfully disagree.  When a site is already live and minor frontend updates are being made it makes no sense for each developer to have their own local environment with potentially dummy date.  Instead we often use a single client facing shared staging environment with near live product data.  Also with the introduction of the parameter code that gives you a bit of protection if the config is left enabled.

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
```diff
@@ -58,24 +59,51 @@ class DebugHints
...
+     * @param string $debugHintsParameter
      */
     public function __construct(
         ScopeConfigInterface $scopeConfig,
         StoreManagerInterface $storeManager,
         DevHelper $devHelper,
         DebugHintsFactory $debugHintsFactory,
-        $debugHintsPath
+        $debugHintsPath,
+        Http $http,
```
korostii: This is a backwards-incompatible change

korostii:
> Why each developer does not use own local environment for development?

Even on a local environment it's a burden to go into admin in order every time you want to switch it on or off. Being able to do so using a simple GET parameter is a handy feature to have.

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
XigenDominic: I understand.  Any suggestion on how to make this backwards compatible?

korostii: Yes, sure. There's an example in the guides:
http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#ui-id-3

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
```diff
@@ -58,24 +59,51 @@ class DebugHints
     protected $debugHintsPath;
 
     /**
+     * XPath of configuration of the debug hints show with parameter
+     *
+     *     dev/debug/template_hints_storefront_show_with_parameter
+     *
+     * @var string
+     */
+    protected $debugHintsShowWithParameter;
```
ishakhsuvarov: Current Magento Technical Vision does not recommend designing classes for inheritance, therefore please declare properties as `private`.

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
```diff
@@ -94,15 +122,30 @@ public function afterCreate(
...
+                $this->debugHintsParameter,
                 ScopeInterface::SCOPE_STORE,
                 $storeCode
             );
-            return $this->debugHintsFactory->create([
-                'subject' => $invocationResult,
-                'showBlockHints' => $showBlockHints,
-            ]);
+            $debugHintsParameterInUrl = $this->http->getParam('templatehints');
+            if ((!$debugHintsShowWithParameter) ||
```
ishakhsuvarov: This condition looks hard to read and understand, increasing possibility of an error in future refactoring. Please attempt to simplify it.

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
orlangur: *for inheritance

I believe extensibility is still a priority, just achieved with other approaches :)

orlangur:
@korostii,
> go into admin in order every time you want to switch it on or off. Being able to do so using a simple GET parameter is a handy feature to have.

Never tried it via CLI, I'm quite used to Admin UI always opened in separate tab just because you'll need to flush caches together with switch change anyway.

Would be less scary if we'd have `Magento_Developer` module disabled by design when deployed to production or something like that.

korostii:
> I'm quite used to Admin UI always opened in separate tab just because you'll need to flush caches

Oh wow. Yeah, you should definitely try CLI, clearing caches feels much faster this way.

> Would be less scary if we'd have Magento_Developer module disabled by design when deployed to production or something like that.

Yes, that makes sense. It's actually quite easy to enforce using composer (by moving relevant modules to `require-dev` section, see also #8971, #5262 ). Oh, and you'll also need to have the `config.php` added to `.gitignore` in order to have different sets of enabled modules between instances.
However there might still be some internal dependencies you'd need to untangle additionally, of course.

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
ishakhsuvarov: You are right. Been thinking of `extends` keyword. Fixed.

piotrekkaminski:
For security reasons we cannot have ability for anyone to just switch on debug info on a site. If such feature is really helpful for developers, it has to be wrapped in IP whitelisting, ie only developers coming from a known, specified IP address can enable it. Then it would not even need the special parameter value, just ?showtemplatehints=1 or similar.

orlangur:
@piotrekkaminski yeah, I was thinking about IP whitelisting as well together with possible filtering by xdebug session. Rejected such option as for scenario provided developers could be sitting in the same office having the same IP.

XigenDominic:
@piotrekkaminski Okay so assuming someone leaves the option enabled and a third party discovers the parameter code and enables template hints is it any more of a risk than say for example someone running your site through magereport.com?  The only serious negative I can think is that template hints would get cached and make your site look terrible/broken.

@piotrekkaminski @orlangur I nested my change within isDevAllowed() so surely the option for IP whitelisting is already achieved?

Ctucker9233:

app/code/Magento/Developer/Model/TemplateEngine/Plugin/DebugHints.php
```diff
@@ -94,15 +124,37 @@ public function afterCreate(
...
-            ]);
+            $debugHintsParameterInUrl = $this->http->getParam('templatehints');
+
+            $showHints = false;
+            if (!$debugHintsWithParam) {
+                $showHints = true;
+            }
+            if ($debugHintsWithParam && $debugHintsParameter == $debugHintsParameterInUrl) {
+                $showHints = true;
+            }
```
sshymko, line 145: Overcomplicated implementation with unnecessarily high cyclomatic complexity.

Equivalent implementation:
```php
$showHints = !$debugHintsWithParam || $debugHintsParameter == $debugHintsParameterInUrl;
```

DominicWatts: I wrote something similar to that at first and the original CI review process told me to change it. But feel free to do whatever needs to be done. I still think this is a useful feature.

sshymko: Just wanted to leave a comment in case someone stumbles upon it.
I am not part of the Magento team (not anymore) and no action is needed on your part :)

sshymko: The feature is indeed useful. Thank you so much for contributing it!
Being the original author of that code, I'm just being sensitive to increase of its complexity :)

orlangur: @sshymko,

> Just wanted to leave a comment in case someone stumbles upon it.

How is your comment ever useful to anybody? Please provide a use case.

Thanks for paying attention to code quality though. And also let me inform you that we are accepting pull requests for more than a year already ;) This is the most natural way of constructive collaboration.

sshymko: I'll try to find time to learn the guidelines and submit the PR.
Until then the comment serves as a reminder to do that.

PR #10619, opened 2017-08-22
Prevent unset Boolean product attributes from displaying in product page

usesi-jlambert:
### Description
Per the discussion here:
https://github.com/magento/magento2/issues/6634

I found when I implemented the proposed fix that was implemented in this file in my own Magento installation, all boolean attributes were returned even if they were not assigned to a product and had no value in the database. So my product pages were full of NULLs and had a large number of non-relevant attributes listed because of the large number of custom attributes we are utilizing for our product catalog. The small change I have made here allows attributes that have been assigned a value to be displayed, but non-relevant/unset attributes are not, therefore I feel it is a much cleaner solution to the problem.

1. magento/magento2#6634: Yes/No attribute value is not shown on a product details page

### Manual testing scenarios
Before Fix:
1. Create a custom boolean attribute but do not set the value for any products
2. Open a product's detail page and the attribute should be visible and display a value of "NULL"

After Fix:
1. Create a custom boolean attribute but do not set the value for any products
2. Open a product's detail page and the attribute should not be displayed in the list of product attributes

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Block/Product/View/Attributes.php
+++ b/app/code/Magento/Catalog/Block/Product/View/Attributes.php
@@ -84,13 +84,15 @@ class Attributes extends \Magento\Framework\View\Element\Template
 
                 if (!$product->hasData($attribute->getAttributeCode())) {
                     $value = __('N/A');
+                } elseif ($value instanceof Phrase) {
+                    $value = (string)$value;
                 } elseif ((string)$value == '') {
                     $value = __('No');
                 } elseif ($attribute->getFrontendInput() == 'price' && is_string($value)) {
                     $value = $this->priceCurrency->convertAndFormat($value);
                 }
 
-                if ($value instanceof Phrase || (is_string($value) && strlen($value))) {
+                if (is_string($value) && strlen($value)) {
                     $data[$attribute->getAttributeCode()] = [
                         'label' => __($attribute->getStoreLabel()),
                         'value' => $value,
--- a/app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
+++ b/app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
@@ -1,0 +1,176 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Catalog\Test\Unit\Block\Product\View;
+
+use \PHPUnit\Framework\TestCase;
+use \Magento\Framework\Phrase;
+use \Magento\Eav\Model\Entity\Attribute\AbstractAttribute;
+use \Magento\Eav\Model\Entity\Attribute\Frontend\AbstractFrontend;
+use \Magento\Catalog\Model\Product;
+use \Magento\Framework\View\Element\Template\Context;
+use \Magento\Framework\Registry;
+use \Magento\Framework\Pricing\PriceCurrencyInterface;
+use \Magento\Catalog\Block\Product\View\Attributes as AttributesBlock;
+
+/**
+ * Test class for \Magento\Catalog\Block\Product\View\Attributes
+ *
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
+class AttributesTest extends TestCase
+{
+    /**
+     * @var \Magento\Framework\Phrase
+     */
+    private $phrase;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Eav\Model\Entity\Attribute\AbstractAttribute
+     */
+    private $attribute;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Eav\Model\Entity\Attribute\Frontend\AbstractFrontend
+     */
+    private $frontendAttribute;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Catalog\Model\Product
+     */
+    private $product;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Framework\View\Element\Template\Context
+     */
+    private $context;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Framework\Registry
+     */
+    private $registry;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Framework\Pricing\PriceCurrencyInterface
+     */
+    private $priceCurrencyInterface;
+
+    /**
+     * @var \Magento\Catalog\Block\Product\View\Attributes
+     */
+    private $attributesBlock;
+
+    protected function setUp()
+    {
+        $this->phrase = new Phrase(__(''));
+
+        $this->attribute = $this
+            ->getMockBuilder(AbstractAttribute::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->attribute
+            ->expects($this->any())
+            ->method('getIsVisibleOnFront')
+            ->willReturn(true);
+
+        $this->attribute
+            ->expects($this->any())
+            ->method('getAttributeCode')
+            ->willReturn('phrase');
+
+        $this->frontendAttribute = $this
+            ->getMockBuilder(AbstractFrontend::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->attribute
+            ->expects($this->any())
+            ->method('getFrontendInput')
+            ->willReturn('phrase');
+
+        $this->attribute
+            ->expects($this->any())
+            ->method('getFrontend')
+            ->willReturn($this->frontendAttribute);
+
+        $this->product = $this
+            ->getMockBuilder(Product::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->product
+            ->expects($this->any())
+            ->method('getAttributes')
+            ->willReturn([$this->attribute]);
+
+        $this->product
+            ->expects($this->any())
+            ->method('hasData')
+            ->willReturn(true);
+
+        $this->context = $this
+            ->getMockBuilder(Context::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->registry = $this
+            ->getMockBuilder(Registry::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->registry
+            ->expects($this->any())
+            ->method('registry')
+            ->willReturn($this->product);
+
+        $this->priceCurrencyInterface = $this
+            ->getMockBuilder(PriceCurrencyInterface::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->attributesBlock = new AttributesBlock(
+            $this->context,
+            $this->registry,
+            $this->priceCurrencyInterface
+        );
+    }
+
+    /**
+     * @return void
+     */
+    public function testGetAttributesBooleanNoValue()
+    {
+        $this->phrase = new Phrase(__(''));
+
+        $this->frontendAttribute
+            ->expects($this->any())
+            ->method('getValue')
+            ->willReturn($this->phrase);
+
+        $attributes = $this->attributesBlock->getAdditionalData();
+
+        $this->assertTrue(empty($attributes['phrase']));
+    }
+
+    /**
+     * @return void
+     */
+    public function testGetAttributesBooleanHasValue()
+    {
+        $this->phrase = new Phrase(__('Yes'));
+
+        $this->frontendAttribute
+            ->expects($this->any())
+            ->method('getValue')
+            ->willReturn($this->phrase);
+
+        $attributes = $this->attributesBlock->getAdditionalData();
+
+        $this->assertNotTrue(empty($attributes['phrase']));
+        $this->assertNotTrue(empty($attributes['phrase']['value']));
+        $this->assertEquals('Yes', $attributes['phrase']['value']);
+    }
+}


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10619) <br/>All committers have signed the CLA.

orlangur:
Thanks, that's exactly what I was suggesting in https://github.com/magento/magento2/pull/8623/files#r102056203.

What do you think about writing a unit test? It's enough to cover just the case you fixed when `$value` is `Phrase` containing empty string.

usesi-jlambert:
@orlangur I have added a test class with two test cases for when a custom attribute of class Phrase is used, but I also modified the proposed fix to change the line:
$value = $value->getText();
to instead be:
$value = $value->render();
because in 99% of cases the default result of "render()" is to return the same value that "getText()" would, but just in case a renderer is used it will return the expected value instead of just the textual representation that was stored in the DB.

orlangur:
Thanks! I see `\PHPUnit_Framework_TestCase` is used - could you please rebase this PR against develop which uses PHPUnit 6 already with `\PHPUnit\Framework\TestCase` class and squash changes into single commit?

orlangur:
Nice catch with `render`! Even better solution would be just cast `Phrase` to string.

```
} elseif ((string)$value == '') {
                    $value = __('No');
                } elseif ($attribute->getFrontendInput() == 'price' && is_string($value)) {
                    $value = $this->priceCurrency->convertAndFormat($value);
                } elseif ($value instanceof Phrase) {
```

As you see here, in case of `Phrase` the `$value` will be rendered twice. Could you rewrite implementation a bit to avoid this? If not, no problem, it can be done as separate improvement.

usesi-jlambert:
@orlangur I just changed the ordering of the ifelse checks to place the "instanceof Phrase" check before the "(string) $value" check, so if it is a Phrase it will not be cast to a string for one check and then again rendered in a later check. Does this satisfy the improvement you had in mind?

orlangur:
@usesi-jlambert yeah, should render only once now.

```
$value = $value->render();
```

What do you think about making it

```
$value = (string)$value;
```
? It will call `render` as well.

usesi-jlambert:
@orlangur Updated with the latest suggestion, can you perform a "squash and merge" through Github or do I need to rebase it in my own repo?

orlangur:
This should be done in your branch and force-pushed as otherwise PR will not be marked as Merged.

But please check unit test is still working under PHPUnit 6 locally first.

usesi-jlambert:
@orlangur Commits merged and updated in the pull request

orlangur:
@usesi-jlambert sorry, I forgot to write a comment regarding static test failing. Would you mind to fix it (and amend commit, as always)? `TestAttributeName` constant needs to be `TEST_ATTRIBUTE_NAME` to conform with PSR-2 standard.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,187 @@
...
+use \Magento\Catalog\Model\Product;
+use \Magento\Framework\View\Element\Template\Context;
+use \Magento\Framework\Registry;
+use \Magento\Framework\Pricing\PriceCurrencyInterface;
+use \Magento\Catalog\Block\Product\View\Attributes as AttributesBlock;
+
+/**
+ * Test class for \Magento\Catalog\Block\Product\View\Attributes
+ *
+ * @SuppressWarnings(PHPMD.LongVariable)
```
orlangur: Please get rid of this. `priceCurrencyInterfaceMock` could be names simply `priceCurrency` (everything besides test object is a mock in unit test).

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,187 @@
...
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
+class AttributesTest extends TestCase
+{
+    const TESTATTRIBUTENAME = 'phrase';
```
orlangur: Please use underscore as word separator: `TEST_ATTRIBUTE_NAME`.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,187 @@
...
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Magento\Framework\Pricing\PriceCurrencyInterface
+     */
+    private $priceCurrencyInterfaceMock;
+
+    /**
+     * @var \Magento\Catalog\Block\Product\View\Attributes
+     */
+    private $attributesBlock;
+
+    protected function setUp() // @codingStandardsIgnoreLine
```
orlangur: PHPCS errors suppression must be avoided. Please get rid of all such ignores.

Use real strings instead of constants or replace real `Phrase` object with a mock.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,187 @@
...
+            ->expects($this::any())
+            ->method('getFrontendInput')
+            ->willReturn(self::TESTATTRIBUTENAME);
+
+        $this->attributeMock
+            ->expects($this::any())
+            ->method('getFrontend')
+            ->willReturn($this->frontendAttributeMock);
+
+        $this->productMock = $this
```
orlangur: I do not insist on changing all mocks creation this time, but generally `createMock` with all statements in one line would be much faster to read.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,187 @@
...
+        );
+
+        $this->frontendAttributeMock
+            ->expects($this::any())
+            ->method('getValue')
+            ->willReturn($this->phraseResource);
+
+        $attributes = $this->attributesBlock->getAdditionalData();
+
+        $this::assertTrue(empty($attributes[self::TESTATTRIBUTENAME]));
```
orlangur: Replace all static calls with `$this->assert*`.

orlangur (changes requested):

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: Lines not exceeding 120 characters are per the MEQP2 standard available here:
https://github.com/magento/marketplace-eqp

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: I modified these to static because I was receiving a warning for calling a static method dynamically, but it works regardless of which way they are called so I will swap them back.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
orlangur: Sorry, I don't understand this comment.

Yeah, 120 characters limit should be honored, just there is no need to break the line for each method call, it could be:
```
$mock->expects()->method()
    ->willReturn()
```

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: Removed Constants and used strings where necessary to remove the Phrase class complaint.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: Removed

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
orlangur: Warning in IDE? Surely it works, just that it is used as `$this->assert*` in most of the tests (~200 VS ~8000 occurrences).

Just found out that `$this->` is a recommended by PHPUnit author way: https://github.com/sebastianbergmann/phpunit/issues/1922, https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.static-vs-non-static-usage-of-assertion-methods

Please eliminate these static calls.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: I structured it this way to avoid exceeding line length, but I also find it easier to read with a method on each line so I can read it like a checklist of items that are completed to get the final result. I can refactor if it is a big deal.

orlangur: It more a matter of taste, as I said, there is no strict requirement regarding this. Please leave as is.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: They have been swapped back in the latest commit

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
orlangur: Ping @usesi-jlambert :)

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * @var \Magento\Catalog\Block\Product\View\Attributes
+     */
+    private $attributesBlock;
+
+    protected function setUp()
+    {
+        $this->phrase = new Phrase(
+            ''
```
orlangur: Can we have each `new Phrase` on a single line?

orlangur (changes requested):

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: Done

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
usesi-jlambert: Mock has been removed from all variable names.

usesi-jlambert:
@orlangur Any thoughts on why the travic-ci build is failing now? I'm at a bit of a loss on that error.

orlangur:
@usesi-jlambert this failure is Travis-specific, simply ignore it, I'll run against internal CI.

Please remove the `@SuppressWarnings(PHPMD.LongVariable)` annotation (I believe no more long variables left, right? AFAIR we removed it from Magento ruleset, if only Codacy is complaining you can ignore it) and squash all changes into one commit.

orlangur:
@usesi-jlambert please amend all changes into one commit, besides that looks good to me, great job 👍

hostep:
Small remark: what is the point of creating the Phrases `N/A` and `No` when they will never be outputted?
I suggest these two lines should get changed to `$value = ''`, it feels like a better solution, less confusing, no?

orlangur:
> creating the Phrases N/A and No when they will never be outputted?

Are you saying that with `Phrase` we will never pass `if (is_string($value) && strlen($value)) {` condition?

hostep:
@orlangur: indeed, a Phrase passes `strlen` but not `is_string`.

orlangur:
@hostep I see... I believe previous PR was exactly about fixing this case and we want to see `'N/A'` and `'No'` outputted sometimes. Do you see any sense in `strlen` here?

I'm inclined to change last condition to `if ($value instanceof Phrase || is_string($value))`.

hostep:
The problem where we are running into, is that we don't want to see `N/A` being outputted. This change was introduced in Magento 2.1.7 [over here](https://github.com/magento/magento2/commit/d977dc3313f294f9ffe7e8a38f4093b894524c70#diff-8bb5f6b700bf55d266640da3251af32cL89) and now suddenly we see product attributes with `N/A` being outputted where we don't want to see those. See this ticket for more info: https://github.com/magento/magento2/issues/9961

This PR by @usesi-jlambert happens to (accidentally?) fix https://github.com/magento/magento2/issues/9961, but the code is really confusing right now as you can imagine :)

Or should the behavior of displaying `N/A` be configurable? I can imagine some shop owners maybe want to display them, but others not? I'm more in favor of not displaying them (just like in Magento 2.1.6 and before), because it doesn't add useful information to the product detail view when an attribute is advertised as `N/A`. But that's just my opinion.

orlangur:
@hostep, `if ($attribute->getIsVisibleOnFront()` is the key, just 
> Edit Attribute page in Admin UI Storefront Properties > Visible on Catalog Pages on Storefront need to be set to No

The fact such attributes with `'N/A'` were not displayed before is a bug even though such behavior could be useful for somebody.

usesi-jlambert:
@orlangur I have squashed the commits so for the original bug can this fix be merged now?

orlangur:
@usesi-jlambert thanks! Are you able to see `'No'` boolean attribute value in current fix?

As pointed out by @hostep it looks like `'N/A'` and `'No'` became unreachable.

usesi-jlambert:
@orlangur I am able to see "Yes" and "No" attribute values for Booleans, and Attributes that are not of type Phrase are not affected by this change.

orlangur (approved):

korostii:
@orlangur, I think you might've misunderstood the use case @hostep was describing. 
There might be an attribute which is intended to visible on product pages in general, but _only for those products_ which happen to have a non-empty value set for this attribute. 
It seems that with the behavior introduced in Magento 2.1.7 the only possible options are now "visible on all products" and "never visible", neither of which fits the expected behavior explained above.

usesi-jlambert:
@korostii With the change that I have proposed, the behavior for attributes will be as follows:

Non-boolean attributes:
    2.1.7: not modifed
    10619: not modified
Boolean attributes:
    2.1.7: All boolean attributes that are set to be visible on store front will be visible, even if they have not been set for the product you are viewing. This causes product attribute lists to be flooded with useless values if you have a large number of custom boolean attributes (like we did). The only way to hide these attributes is by disabling them from being visible on the store front at all times, even if they have  value.
    10619: Only boolean attributes that have a value are displayed, if the attribute is not set then it is not visible. If you wish for an attribute that doesn't have a value to be visible on the storefront, then you need to assign a value to it.

hostep:
> Non-boolean attributes:
> 10619: not modified

@usesi-jlambert: this is an incorrect statement, because you removed `$value instanceof Phrase` from the `if`

So this means the `$value = __('N/A');` won't be outputted because it's a `Phrase`

The question is: is this the correct behaviour or not, do we want to see `N/A` when an non-boolean attribute isn't filled in on a product?
In Magento 2.1.6 it wasn't visible, in 2.1.7 it is visible. The question is: which one is the most correct, or which one do merchants prefer?

usesi-jlambert:
@hostep 
My apologies, I was referring to the old behavior of 2.1.6, not the new behavior introduced in 2.1.7, so you are correct in saying my statement was inaccurate for version 2.1.7.

As for which is the preferred behavior, I created this pull request because I ran into the problem of boolean attributes not displaying properly, and we have close to 50 custom boolean product attributes so that was a problem, but when I made the modification that was eventually merged into 2.1.7 I had a product page filled with useless information for attributes that had no values and no in built way to hide them. I found that the proposed alternate solution was much more desirable from my perspective, and when compared to the 2.1.6 release only changes the result to include information that was omitted before, while the 2.1.7 version's change introduced a large amount of useless information with no good method of hiding it on a product by product basis.

If you wanted to force a product to show an attribute value of N/A when no value was set, then you could create a custom text attribute, and set the default value to N/A. So working around case by case issues, from my perspective, is easier with my version of the fix than it is with the current 2.1.7 fix.

Thoughts?

hostep:
I agree @usesi-jlambert, I also want to go back to the state like it was in Magento 2.1.6.

But, looking at the code, it seems like it was meant to display `N/A` when an attribute is empty.
So I wonder if it was a bug in 2.1.6 (and earlier) that this value wasn't being displayed, because the code seems to imply it was meant to be displayed.

But I agree, and (some of) our clients as well, they also seem to want to go back to how it was in Magento 2.1.6.

But then I think the code has to change a little bit, to make it less confusing, and I'd like to propose to change: `$value = __('N/A');` into `$value = '';`, so it is more obvious that it is the expected behaviour that non-filled in attributes shouldn't show up in the frontend.

@orlangur, or someone from the Magento devs, feedback would be appreciated! :)

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,176 @@
...
+    private $priceCurrencyInterface;
+
+    /**
+     * @var \Magento\Catalog\Block\Product\View\Attributes
+     */
+    private $attributesBlock;
+
+    protected function setUp()
+    {
+        $this->phrase = new Phrase(__(''));
```
vkublytskyi, line 67: Function `__` returns a Phrase instance. Use `new Phrase('')` or `__('')`. Result will be the same. While your code is working it lead to static tests failure:

```
Magento\Test\Integrity\Phrase\ArgumentsTest::testArguments
2 missed phrases were discovered: 

Missed Phrase: File: /opt/bamboo/bamboo-agent-home/xml-data/build-dir/M2-AT2646-INSE/magento2ce/app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php 
Line: 67


Missed Phrase: File: /opt/bamboo/bamboo-agent-home/xml-data/build-dir/M2-AT2646-INSE/magento2ce/app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php 
Line: 146

/opt/bamboo/bamboo-agent-home/xml-data/build-dir/M2-AT2646-INSE/magento2ce/dev/tests/static/testsuite/Magento/Test/Integrity/Phrase/ArgumentsTest.php:66
```

app/code/Magento/Catalog/Block/Product/View/Attributes.php
```diff
@@ -84,13 +84,15 @@ public function getAdditionalData(array $excludeAttr = [])
 
                 if (!$product->hasData($attribute->getAttributeCode())) {
```
vkublytskyi, line 85: Please remove this `if` as this is dead code for a long time.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
```diff
@@ -84,13 +84,15 @@ public function getAdditionalData(array $excludeAttr = [])
...
                     $value = __('N/A');
+                } elseif ($value instanceof Phrase) {
+                    $value = (string)$value;
                 } elseif ((string)$value == '') {
```
vkublytskyi, line 89: Please remove this `elseif` as converting an empty value to `No` is pretty weird logic. If this is a workaround for some bug it should be fixed in another way as this code only hides the real problem.

vkublytskyi (changes requested):

vkublytskyi:
Regarding @hostep comment

> But then I think the code has to change a little bit, to make it less confusing, and I'd like to propose to change: $value = __('N/A'); into $value = '';, so it is more obvious that it is the expected behaviour that non-filled in attributes shouldn't show up in the frontend.

It won't have impact as then it was filtered out by `is_string($value) && strlen($value)`. This check was from the initial commit so I think "display only not empty text values" contract should be preserved.

vkublytskyi:
@usesi-jlambert If you don't have time to work on this PR now, please [allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) and I will finish and merge it.

vkublytskyi:
All required fixes applied in internal repo. PR will be merged soon.

usesi-jlambert:
@vkublytskyi I have modified this PR to allow edits from maintainers.

PR #10634, opened 2017-08-23
Backwards compatibility check for serialized array

ghost:
* If used as backend model with existing serialized config data the change to JSON in 2.2 would result in an exception when viewing the corresponding store config page.
* This fix catches the exception and uses unserialize as fallback. On save the array will be stored using the new Serializer.

### DIFF
--- a/app/code/Magento/Config/Model/Config/Backend/Serialized.php
+++ b/app/code/Magento/Config/Model/Config/Backend/Serialized.php
@@ -52,7 +52,12 @@ class Serialized extends \Magento\Framework\App\Config\Value
     {
         $value = $this->getValue();
         if (!is_array($value)) {
-            $this->setValue(empty($value) ? false : $this->serializer->unserialize($value));
+            try {
+                $this->setValue(empty($value) ? false : $this->serializer->unserialize($value));
+            } catch (\InvalidArgumentException $e) {
+                // backwards compatibility: check for serialized array
+                $this->setValue(empty($value) ? false : unserialize($value));
+            }
         }
     }
 


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10634) <br/>All committers have signed the CLA.

orlangur:
> serialized config data the change to JSON in 2.2 would result in an exception when viewing the corresponding store config page

There is a migration to convert this, isn't it?

PHOENIX-MEDIA:
@orlangur Magento core does not convert configuration values of 3rd party modules. But to not end up with an exception in the backend it would be handy to do the conversion on-the-fly when the configuration section is viewed. This way the value won't be modified (and still may be used as a serialized array) but also keep the configuration page usable - at least till the user saves the configuration with 2.2 which will then safe the values as JSON. Then the code of the 3rd party module would break as it maybe still uses unserialize().

orlangur:
Any 3rd party modules must be updated according to http://devdocs.magento.com/guides/v2.2/release-notes/backward-incompatible-changes.html#database-data-format-changes before using with Magento 2.2.

Proposed solution is unreliable and encourages to use your Magento instance in semi-ready state.

korostii:
Hi @orlangur,

What happened to being community-friendly? 
Is creating pain and misery for both 3rd-party developers and merchants really necessary?
Never mind marking `Serializer` as `@deprecated` and assigning json_encode behavior to a different interface, let's just put a new urgent responsibility on 3rd-party developers instead.

Let me assure you: if 3rd-party modules will start throwing exceptions right after upgrading the core Magento to version 2.2, it's the Magento who will be blamed "unreliable". 

Having all 3rd party modules updated... 
1) will never gonna happen in reality: consider modules, whose developers were all hit by a bus, for example.
2) implies that the merchants will update the modules the same day they upgrade to 2.2 (or earlier). This isn't likely neither. Not unless you'll explicitly and publicly announce that necessity. And I don't see no "Please make sure all the 3rd-party modules you use are updated to a 2.2.0-compatible version prior to upgrading" nowhere near [release notes](http://devdocs.magento.com/guides/v2.2/release-notes/release-notes-2-2-RC2-CE.html). Do you?

This will be, by far, the biggest backwards-incompatible change in the 2.2 release, and you seem to prefer to have it blow up into everyone's face. I mean, why?

What's so "unreliable" about resolving a corner-case and making the change (a little bit more) backwards-compatible?

PR #10635, opened 2017-08-23
Do not show lonely comma on empty region

amenk:
### Description
If the address is empty or partial, a lonley comma was shown

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/template/shipping-information/address-renderer/default.html
+++ b/app/code/Magento/Checkout/view/frontend/web/template/shipping-information/address-renderer/default.html
@@ -8,7 +8,7 @@
     <!-- ko text: address().prefix --><!-- /ko --> <!-- ko text: address().firstname --><!-- /ko --> <!-- ko text: address().middlename --><!-- /ko -->
     <!-- ko text: address().lastname --><!-- /ko --> <!-- ko text: address().suffix --><!-- /ko --><br/>
     <!-- ko text: address().street --><!-- /ko --><br/>
-    <!-- ko text: address().city --><!-- /ko -->, <span data-bind="html: address().region"></span> <!-- ko text: address().postcode --><!-- /ko --><br/>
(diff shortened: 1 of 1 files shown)


app/code/Magento/Checkout/view/frontend/web/template/shipping-information/address-renderer/default.html
```diff
@@ -8,7 +8,7 @@
     <!-- ko text: address().prefix --><!-- /ko --> <!-- ko text: address().firstname --><!-- /ko --> <!-- ko text: address().middlename --><!-- /ko -->
     <!-- ko text: address().lastname --><!-- /ko --> <!-- ko text: address().suffix --><!-- /ko --><br/>
     <!-- ko text: address().street --><!-- /ko --><br/>
-    <!-- ko text: address().city --><!-- /ko -->, <span data-bind="html: address().region"></span> <!-- ko text: address().postcode --><!-- /ko --><br/>
+    <!-- ko text: address().city --><!-- /ko --><!-- ko if: !(address().region.trim() == "") -->, <span data-bind="html: address().region"></span><!-- /ko --> <!-- ko text: address().postcode --><!-- /ko --><br/>
```
omiroshnichenko, line 11: Use strict equality or check on string length.

ishakhsuvarov:
Hi @amenk 
Could you please check and fix according to review comments? Let us know if you wish any assistance with this.
Thanks.

app/code/Magento/Checkout/view/frontend/web/template/shipping-information/address-renderer/default.html
okorshenko: we will fix this on our side. thank you for review

okorshenko:
Hi @amenk 
Thank you for your contribution. Unfortunately, we can not accept this PR as is since it adds inconsistency to the rendering. 
We need to fix not only shipping information block, but address selection on the first step of checkout (when you are logged in and have few address in the address book), billing address rendering,  address book rendering on Customer Account page. 
So if we decided to change address rendering format, we must make it consistent. Otherwise, it will be worse than now

amenk:
Yes feel free to fix all the the renderings,best by referring to a single
template.

okorshenko:
Hi @amenk we will add this to our backlog. Closing this PR, since we need to deliver fixes to all places in one PR.

PR #10640, opened 2017-08-23
Add Austria and Finland to countries that do not require region/state

lazyguru:
### Description
Both Austria and Finland do not require a region/state in the address. Every merchant I have worked with has had to manually change the config setting in the admin. Instead of requiring every merchant in those countries (or that ships to those countries) to do this, the default should be updated. Obviously, this will resolve the issue for new installs only.

### Manual testing scenarios
1. Install Magento 2.2
2. Verify that **Stores->Configuration->General->State Options->State is Required for** does not show Austria or Finland as selected


avoelkl:
I can confirm we disable the region field in AT very often :)

Issue #10112, opened 2017-06-30
Module names sorting issue?


magekey:
Not sure it could be an issue but I found some problem during develop my module.

### Preconditions

1. Release 2.1.7 with Sample Data
2. Ubuntu 16.04, php-fpm 5.6.30, MySql Percona Server 5.6

I did some tests with modules sorting function in \Magento\Framework\Module\ModuleList\Loader.
Here is results of print_r function.

**TEST 1** - No custom modules.

_**Result:**_

> Magento_Checkout - line 558 (Line in browser)
> Magento_Tax - line 618

---

**TEST 2** - added module MageKey_Module - no sequences
```
<module name="MageKey_Module" setup_version="2.1.0"/>
```

_**Result:**_

> MageKey_Module - line 3
> Magento_Checkout - line 568
> Magento_Tax - line 628 

looks correct

---

**TEST 3** - added sequence

```
<module name="MageKey_Module" setup_version="2.1.0">
    <sequence>
        <module name="Magento_Checkout"/>
    </sequence>
</module>
```
_**Result:**_

> Magento_Tax - line 511
> Magento_Checkout - line 571
> MageKey_Module - line 1073

now Magento_Checkout has high priority which is not good.

---
Not sure it can be an issue but I added only Checkout sequence, not sure why Tax has low priority now. As result I got an issue on Checkout cart page.

### Expected result
On Checkout Cart page on summary sidebar to render Grand Total (grand-total) we are using component **Magento_Tax/js/view/checkout/cart/totals/grand-total**. Order Total shows correct info.

### Actual result
When module has been added now we have checkout component **Magento_Checkout/js/view/summary/grand-total**. And Order Total collected without Taxes.

Not sure maybe I missed something.

korostii:
I will duplicate here some considerations which I've also mentioned on #10649.

Magento_Tax module depends on code from Magento_Checkout module and will crash without it.

A couple examples:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Tax/etc/frontend/di.xml#L17
https://github.com/magento/magento2/blob/develop/app/code/Magento/Tax/view/frontend/web/js/view/checkout/cart/totals/tax.js#L12

This should be fixed by adding an making that dependency explicit instead (via adding `Magento_Checkout` into the `Magento_Tax`'s `<sequence>`). 

This seem to be done on 2.2.0 branch already;
[...]

Issue #10331, opened 2017-07-26
[2.1.7] Enabling module brakes sorting of modules in config.php


ihor-sviziev:
We're keeping config.php file in git to make sure that only modules that we have enabled on local environment - will be enabled on production.

### Preconditions
1. Magento 2.1.7
2. Remove app/etc/config.php file from .gitignore

### Steps to reproduce
1. Install magento via web setup wizard
2. Commit file app/etc/config.php to git
3. Install module via composer ```composer require magepal/magento2-gmailsmtpapp:"^1.4"```
4. Enable module via command line ```php -f bin/magento module:enable MagePal_GmailSmtpApp```

### Expected result
1. in app/etc/config.php we see that one line was added

### Actual result
1. app/etc/config.php file has a lot changes

You can see following diff that I god after installing module in my [gist](https://gist.github.com/ihor-sviziev/f0d802e172f8a694c0efe883be616ae3/revisions#diff-9e5e2527d69c009a81b8ecd730f3957e).
It's really hard to understand what's changed, and during git merge we can broke something.
This issue reproducing with a lot another modules, I used specific one just to make it 100% reproducible.

I suggest to sort modules in config.php, it will prevent such issues.

korostii:
You're quite right, but this issue has already been reported: see #8479.

hostep:
Yes, I 100% agree with @ihor-sviziev!

We've been keeping the config.php file outside of our git repo's for Magento 2.0.x and 2.1.x projects, just because of this reordening of the modules. This keeps screwing every developer when switching branches and working on different modules and this causes merge conflicts all the time. Which is really annoying.
We keep it outside of our git repo, because Magento just regenerates it every time you run `bin/magento setup:upgrade`.

But now that Magento 2.2 will also put configuration data inside that file, we will need to include it in our git repo's and it's going to be a total mess :(
~~(This comment was based on incorrect documentation, which has been corrected recently: https://github.com/magento/devdocs/pull/1294)~~
(Update: comment above is still true apparently: https://community.magento.com/t5/Magento-DevBlog/Magento-2-2-Deployment-Improvements/ba-p/73119)

I would like to have this sort order of how the modules are supposed to be loaded to be stored in another file which we can ignore in our git repo's. That would be a better solution then how it is working currently I think.

ihor-sviziev:
@magento-team @antonkril @okorshenko @maghamed is there any ideas why /app/etc/config.php is ignored in .gitignore?

[...]

Issue #8479, opened 2017-02-08
Sequence of module load order should be deterministic


scottsb:
### Preconditions
Magento 2.1.3

### Steps to reproduce
1. Developer enables a new module using CLI command.
2. Magento adds an entry for the module to `config.php`.

### Expected result
The only change to `config.php` is the addition of the module that changed.

### Actual result
In addition to the new line for the new module, many other lines change sequence.

### Discussion
It seems that the current semi-official recommendation is not to version the `config.php` file with the code base and to treat it as an asset along with images, as evidenced by the fact that it is git-ignored in the Magento repository by default and by the fact this advice is upvoted and without alternative:
http://magento.stackexchange.com/a/130109/567

If I understand the intent behind this recommendation, it's that one should be able to pull the latest copy of the code without immediately breaking the site due to an incompatible database schema. My counterpoint is that since themes and existing modules may be modified to depend on certain new modules being enabled, pulling code without including `config.php` is just as likely to break a site.

Ultimately whether `config.php` is versioned or not is probably a decision best made by each individual implementor. However, for us, it definitely makes sense to version it. It allows us to have one atomic unit of enabled code to test and deploy from dev through to production. I know others will agree, given the comments on #800 and the original Stack Exchange question referenced above.

Given that, in order to minimize merge conflicts, it would be best for the sequence of modules in `config.php` to be deterministically generated. It appears that currently it is based on the module load order, but that is not fully deterministic. ~Since the sequence of modules in `config.php` is not itself used to determine any programatic behavior, my suggestion is that the array be sorted alphabetically by key, but I would consider any deterministic ordering to meet the goal.~

korostii:
While I agree that seemingly random order in config.php may lead to unexpected side-effects, and that it might be a good idea to versionize it is well (also in order to track which modules should be enabled), you have to also consider that this file is in fact re-generated by Magento whenever you enable or disable a module based on modules' <sequence> tags (more on that [here](http://devdocs.magento.com/guides/v2.0/extension-dev-guide/build/module-load-order.html) ) thus purely alphabetical is just not an option. (I'm sincerely sorry if I misunderstood your point, please do correct me If I missed something)

scottsb:
@korostii, I know this file is regenerated every time you enable/disable a module. That's *why* I'm suggesting that the order be deterministic. That way when it's regenerated, the only change between the old version and the new version is the relevant line.

[...]

PR #10649, opened 2017-08-24
Properly compute module load order

t-richards:
### Description
This PR replaces the ... _interesting_ ... choice of bubble sort for determining the order in which modules should be loaded.

The `sortBySequence` method now uses topological sorting to correctly linearize the sequence of modules. The topological sort algorithm implemented in the `Magento\Framework\Data\Graph` data structure is [Kahn's algorithm][kahn].

1. magento/magento2#10112: Module names sorting issue?

### Manual testing scenarios
1. Create a module which contains frontend layout XML directives, and has a sequence dependency on `Magento_Customer`
2. Before, with bubble sort, the incorrect load order would cause those frontend layout directives to be processed before `Magento_Catalog`, causing all sorts of trouble.
3. Now, with topological sort, the load order is correct and no such trouble occurs.

### Contribution checklist

>All new or changed code is covered with unit/integration tests

The code is covered by existing unit tests. More tests can be added if desired.

[kahn]: https://en.wikipedia.org/wiki/Topological_sorting#Kahn.27s_algorithm

### DIFF
--- a/lib/internal/Magento/Framework/Data/Graph.php
+++ b/lib/internal/Magento/Framework/Data/Graph.php
@@ -77,6 +77,26 @@ class Graph
     }
 
     /**
+     * Unset a relation between nodes
+     *
+     * @param string|int $fromNode
+     * @param string|int $toNode
+     * @return $this
+     * @throws \InvalidArgumentException
+     */
+    public function removeRelation($fromNode, $toNode)
+    {
+        if ($fromNode == $toNode) {
+            throw new \InvalidArgumentException("Graph node '{$fromNode}' is linked to itself.");
+        }
+        $this->_assertNode($fromNode, true);
+        $this->_assertNode($toNode, true);
+        unset($this->_from[$fromNode][$toNode]);
+        unset($this->_to[$toNode][$fromNode]);
+        return $this;
+    }
+
+    /**
      * Export relations between nodes. Can return inverse relations
      *
      * @param int $mode
@@ -182,6 +202,68 @@ class Graph
     }
 
     /**
+     * Perform a topological sort on the graph. This algorithm is not possible if the graph contains a cycle.
+     *
+     * @return array
+     * @throws \Exception
+     */
+    public function topoSort()
+    {
+        $l = [];
+        $s = $this->_findStartNodes();
+
+        while (!empty($s)) {
+            $n = array_shift($s);
+            array_unshift($l, $n);
+            if (!isset($this->_from[$n])) {
+                continue;
+            }
+            foreach ($this->_from[$n] as $m) {
+                $this->removeRelation($n, $m);
+                if (empty($this->_to[$m])) {
+                    array_push($s, $m);
+                }
+            }
+        }
+
+        if ($this->hasEdges()) {
+            $cycle = $this->findCycle();
+            $message = "Dependency cycle detected: " . implode(" -> ", $cycle);
+            throw new \Exception($message);
+        }
+
+        return $l;
+    }
+
+    public function hasEdges()
+    {
+        foreach ($this->_from as $from) {
+            if (count($from) > 0) {
+                return true;
+            }
+        }
+
+        foreach ($this->_to as $to) {
+            if (count($to) > 0) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Finds all graph nodes with no incoming edge
+     *
+     * @return array
+     */
+    protected function _findStartNodes()
+    {
+        $nodesWithIncomingEdges = array_keys($this->_to);
+        return array_diff($this->_nodes, $nodesWithIncomingEdges);
+    }
+
+        /**
      * Recursive sub-routine of dfs()
      *
      * @param string|int $fromNode
--- a/lib/internal/Magento/Framework/Data/Graph.php
+++ b/lib/internal/Magento/Framework/Data/Graph.php
@@ -263,7 +263,7 @@ class Graph
         return array_diff($this->_nodes, $nodesWithIncomingEdges);
     }
 
-        /**
+    /**
      * Recursive sub-routine of dfs()
      *
      * @param string|int $fromNode
--- a/lib/internal/Magento/Framework/Data/Graph.php
+++ b/lib/internal/Magento/Framework/Data/Graph.php
@@ -210,7 +210,7 @@ class Graph
     public function topoSort()
     {
         $sortedElements = [];
-        $nodeSet = $this->_findStartNodes();
+        $nodeSet = $this->findStartNodes();
 
         while (!empty($nodeSet)) {
             $cur = array_shift($nodeSet);
@@ -264,7 +264,7 @@ class Graph
      *
      * @return array
      */
-    protected function _findStartNodes()
+    public function findStartNodes()
     {
         $incomingEdgeNodes = array_keys($this->_to);
         return array_diff($this->_nodes, $incomingEdgeNodes);
--- a/lib/internal/Magento/Framework/Data/Graph.php
+++ b/lib/internal/Magento/Framework/Data/Graph.php
@@ -213,15 +213,15 @@ class Graph
         $nodeSet = $this->findStartNodes();
 
         while (!empty($nodeSet)) {
-            $cur = array_shift($nodeSet);
-            array_unshift($sortedElements, $cur);
+            $cur = array_pop($nodeSet);
+            $sortedElements[] = $cur;
             if (!isset($this->_from[$cur])) {
                 continue;
             }
             foreach ($this->_from[$cur] as $parent) {
                 $this->removeRelation($cur, $parent);
                 if (empty($this->_to[$parent])) {
-                    array_push($nodeSet, $parent);
+                    $nodeSet[] = $parent;
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


orlangur:
Did you compare an actual execution time of old vs new implementation?

Unfortunately proposed implementation is still `O(n²)` just like bubble sorting and contains a lot of unnecessary array manipulations.

Proper implementation may be based on https://e-maxx-eng.appspot.com/graph/topological-sort.html or simply involve some existing library like https://github.com/marcj/topsort.php.

t-richards:
@orlangur Thank you for your feedback! Yes, I measured some significant differences between the original bubble sort implementation and my topological sort implementation.

Before getting into the micro-benchmark results, I'd like to briefly mention that the goal of this sorting implementation is to order the modules more accurately, and not necessarily improve the speed in which the dependency graph is sorted.

That said, outperforming the time complexity of bubble sort is trivial; just about any other algorithm can do better. Fortunately for us, Kahn's algorithm is _not_, in fact, an `O(n^2)` algorithm. Just like its DFS counterpart, Kahn's algorithm is linear time: `O(|V| + |E|)`, where `V` is the number of graph vertices and `E` is the number of graph edges. Despite the appearance of the implementation (i.e. the nested while/for loops), the performance improvement and technical correctness is plain as day.

I measured the unit tests (where there are some very small graphs, 3 and 5 nodes each), and then I measured a real production Magento 2 store (containing 119 nodes). The results below are the average of three runs for each sorting algorithm.

## Bubble Sort
| Graph size, nodes | # inner loop iterations | time, s  | % decrease in time taken |
|-------------------|-------------------------|----------|--------------------------|
| 3                 | 5                       | 0.000008 | 0%                       |
| 5                 | 14                      | 0.000020  | 0%                       |
| 119               | 7259                    | 0.011596 | 0%                       |

## Toposort
| Graph size, nodes | # inner loop iterations | time, s  | % decrease in time taken |
|-------------------|-------------------------|----------|--------------------------|
| 3                 | 2                       | 0.000006 | 25%                      |
| 5                 | 4                       | 0.000011 | 45%                      |
| 119               | 188                     | 0.000625 | 94.61%                   |

orlangur:
I know the complexity of Kahn's algorithm, the problem is your implementation is `O(n²)` which needs to be fixed :)

Thanks for the benchmark, I do agree that the main point here is to get rid of buggy ordering, just wanted to be sure new implementation is not slower. Could you please provide a gist with patch for such measurement?

> with bubble sort, the incorrect load order would cause those frontend layout directives to be processed before `Magento_Catalog`, causing all sorts of trouble

Automated test case needs to be added which reveals such bug in old implementation. Ideally if it will be the unit test.

What do you think about https://github.com/marcj/topsort.php? Not necessary to fully include the library, for example, we can borrow just `ArraySort` implementation.

t-richards:
>the problem is your implementation is O(n²) which needs to be fixed :)

I'm not sure I agree with this, what part of the implementation is `n^2`? The inner loop only traverses the current node's edges to visit its parents, and does _not_ re-visit every single node in the process.

>Automated test case needs to be added which reveals such bug in old implementation. Ideally if it will be the unit test.

I do agree with this; I'm working on trimming down a large production case that can demonstrate improper sorting, and I will add it as a unit test.

>What do you think about https://github.com/marcj/topsort.php?

I strayed away from adding this dependency, for two reasons:

1. In previous discussions involving topological sort, the Magento team requested the addition of the algorithm to the graph data structure: https://github.com/magento/magento2/pull/65
2. If we were to use the library as-is, it would almost certainly throw `ElementNotFoundException`s everywhere, because of the prevalence of module authors declaring sequence dependencies on modules which are not actually installed.

orlangur:
At least `array_shift` makes it `O(n²)`, I didn't check for other similar issues as just one is enough (see http://php.net/manual/en/function.array-shift.php#112947).

> I strayed away from adding this dependency, for two reasons

Thanks, especially for the old links, I remember reading it before and was surprised there is no topological sorting in Magento Framework yet (I thought it was merged). Will look deeper into it tomorrow. Could you provide a benchmarking gist? Wanna do a quick check of that library speed compared to current PR.

t-richards:
>At least array_shift makes it O(n²)

Thanks for the information about `array_shift`. At that particular point in the algorithm, the order of node removal is not important so this could be replaced with an `array_pop` which is `O(1)`.

>Could you provide a benchmarking gist? Wanna do a quick check of that library speed compared to current PR.

I will provide a link to a gist with some benchmark patches.

korostii:
> Fixed Issues (if relevant)

I believe you can also add the following issues to the list:
#10331 [2.1.7] Enabling module brakes sorting of modules in config.php
#8479 Sequence of module load order should be deterministic

orlangur:
@korostii how using topological sort instead of bubble sort will make order deterministic?

korostii:
@orlangur, 
Well, I cannot guarantee that it does, but if really does solve #10112 it has to retain the original order at least to to some degree, when inserting\removing new entries (as opposed to shuffling it seemingly randomly - that's what usually happened before). 
I haven't tested it myself though, does it not? 

I am sorry if I lead to some confusion on this part. I might've misunderstood the point of changes done here when I skimmed through the PR for the first time.

Here's what I was basing that assumption on: if you take a look at that issue this PR claims to solve, which is #10112, the issue described seems to be the following:
Magento_Checkout and Magento_Tax have an implicit dependency which is not reflected into `<sequence>` tags of their `module.xml` files. Adding new modules and dependencies to the system results into shuffling `config.php` entries, causing Magento_Checkout and Magento_Tax to switch places. In turn, that seems to be breaking some functionality having implicit dependency on their mutual load order.

Now that I've given it some thought: why not make that broken dependency explicit instead (via adding `Magento_Checkout` into the `Magento_Tax`'s `<sequence>`)? That should be enough to solve both #10112 and the manual testing scenario given here (whatever sorting algorithm is used, as long as it enforces these dependencies in the final ordered list)

I now suspect that solving #10112 is just a side-effect of changing the algorithm and it might not work consistently (given different conflicting modules). Could maybe @t-richards shine some explanation into how this all is tied together?

orlangur:
@korostii, see https://en.wikipedia.org/wiki/Sorting_algorithm#Stability. As you can see bubble sort is stable, I believe the order could be random just because modules are randomly read from filesystem.

So, by design we will not get any stability from topological sort usage instead of bubble sort. With something like https://stackoverflow.com/questions/11230881/stable-topological-sort/11236027#11236027 we can get it (if we make the input deterministic, of course), but it will add a `logV` multiplier to a computational complexity.

> why not make that broken dependency explicit instead (via adding `Magento_Checkout` into the `Magento_Tax`'s `<sequence>`)?

Modules in sequence must be specified according to business logic only.

I'm not even sure 8479 is even valid, will share my thoughts there.

korostii:
No, modules in sequence _should_ be specified according to business logic only. In theory.
In practice, Magento_Tax module depends on code from Magento_Checkout module and will crash without it. 

A couple examples:
https://github.com/magento/magento2/blob/develop/app/code/Magento/Tax/etc/frontend/di.xml#L17
https://github.com/magento/magento2/blob/develop/app/code/Magento/Tax/view/frontend/web/js/view/checkout/cart/totals/tax.js#L12

So, the Tax module should either learn to live without these or declare these dependencies outright. 
Since the former solution will require some refactoring and might turn out to be impossible, the latter is a faster fix, obiously.

korostii:
Just noticed: Magento_Tax/Magento_Checkout dependency is already added on 2.2.0 branch already;
https://github.com/magento/magento2/blob/2.2.0-preview/app/code/Magento/Tax/etc/module.xml#L12
But not yet backported to 2.1:
https://github.com/magento/magento2/blob/2.1.10-preview/app/code/Magento/Tax/etc/module.xml

I am now sure this PR cannot fix #10112 on branch `develop` since that issues is already fixed on branch `develop`.

Additionally, it is my understanding that module sorting is only performed during some CLI commands (`module:enable`, `module:disable`, `setup:upgrade`). As such, performance shouldn't be much of an issue here. I think I would prefer current battle-tested code over new one if the benefits are unclear.

@t-richards, could you please elaborate on\explain the benefits of this change? The PR description seems to be lacking on that part right now.

orlangur:
@korostii,
> In practice, `Magento_Tax` module depends on code from `Magento_Checkout` module and will crash without it.

What I meant is that we should not add unreasonable nodes to `<sequence>` just for the sake of unrelated issues solving. Like `Magento_Eav` references `Magento_Catalog` in fact while it should not and specifying it in `<sequence>` would produce a cycle. 

Too bad https://github.com/magento/magento2/commit/eca1579580f4c06be1a290e0b2caabbc7bb37029 happened in scope of some random bugfixing and not solved for the overall system yet. As you can see, `Magento_Tax` extends `Magento_Checkout` with some additional logic from business perspective, on the other hand only a couple references to `Magento_Tax` left in `Magento_Checkout` templates. In opposite situation fix would be incorrect.

> the benefits of this change?

Let me quote the PR author:
> I'm working on trimming down a large production case that can demonstrate improper sorting, and I will add it as a unit test.

So, looks like in some cases sorting did not correspond to `<sequence>` configuration.

t-richards:
@orlangur Thanks for pointing out the stable topological sort, I will do some more research on this and see how it compares.

Despite many attempts, I was not able to produce any test cases which demonstrated an incorrect sort of the modules using the original sorting method. This makes me wonder whether there are module dependency issue(s) of a different kind. For now, I suppose I'll have to live with bubble sort.

If I happen to come up with a better set of test cases, I'll make a new PR.

PR #10670, opened 2017-08-26
Removed deprecated Zend\Stdlib\JsonSerializable references

dverkade:
Removed all other references to deprecated Zend\Stdlib\JsonSerializable which has been deprecated since 20 october 2015. See this commit: zendframework/zend-stdlib@6da50a9#diff-c958cc2b645be4b72d33e0d6dd23c070

Since Magento 2 system requirements are PHP 5.6 or newer we can safely depend on the PHP native JsonSerializable interface since it has been present since PHP 5.4


Ctucker9233:
@magento-team will this get backported to 2.2?

PR #10694, opened 2017-08-29
Fix #10693 : Use Distro Base Url if configuration value has been prev…

### LINKED ISSUE #10693: Not able anymore to use Magento by specifying base_url placeholder
We are running Magento automatical deployments on a **containers architecture**.

URL of the Magento instances are generated "on the fly" (like magento-test-23qsqd456.domain.com) and therefore we process an automatical installation on these containers via something like this (configuration is coming from environment variables) : 

```
    php bin/magento setup:install                  \
        --admin-firstname=$ADMIN_FIRSTNAME         \
        --admin-lastname=$ADMIN_LASTNAME           \
        --admin-email=$ADMIN_EMAIL                 \
        --admin-user=$ADMIN_USER                   \
        --admin-password=$ADMIN_PASSWORD           \
        --backend-frontname=$BACKEND_FRONTNAME     \
        --db-host=$MYSQL_HOSTNAME                  \
        --db-name=$MYSQL_DATABASE                  \
        --db-user=$MYSQL_USER                      \
        --db-password=$MYSQL_PASSWORD              \
        --base-url={{base_url}}                    \
        --use-rewrites=1                           \
        --cleanup-database                         \
        --magento-init-params=MAGE_MODE=production

```

By using these parameters, we were previously able to have a running Magento instance on a dynamically created URL (eg : magento-23qs2d3qsd.test.domain.com)

This was due to the fact that Magento is retrieving the base url via ```getDistroBaseUrl``` as used here : https://github.com/magento/magento2/blob/develop/app/code/Magento/Store/Model/Store.php#L648

But since this commit (https://github.com/magento/magento2/commit/4bfd90d324ca6b743b04db5963bcb3aea44617e8), the value is always explicitely replaced by "http://localhost/" (see here : https://github.com/magento/magento2/blob/develop/app/code/Magento/Store/Model/Config/Placeholder.php#L94) which is irrelevant and leads to Magento redirecting to "http://localhost/" when accessing it.

I note that there is a comment saying "localhost is replaced for cli requests, for http requests method getDistroBaseUrl is used" which is obviously **wrong** because it **does** affect HTTP requests.

### Preconditions

1. Magento 2.1.8 or Magento 2.2
2. PHP 7.0 (but should be the same on 5.6)

### Steps to reproduce

1. Install Magento on a container or a machine having its own domain name (eg test-magento.mytestingserver.com )
2. Use {{base_url}} placeholder when installing
3. Access test-magento.mytestingserver.com

### Expected result

1. Magento should render the page by dynamically using the request URL as base_url

### Actual result

1. [Screenshot, logs]

Redirection to http://localhost/?SID=22u1m71ajo1fup8aq31m4kj641

romainruaud:
…iously defaulted to http://localhost/

### Description

You can see more details on the submitted issue #10693 

1. magento/magento2#10693: Not able anymore to use Magento by specifying base_url placeholder
2. ...

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/Model/Store.php
+++ b/app/code/Magento/Store/Model/Store.php
@@ -643,6 +643,8 @@ class Store extends AbstractExtensibleModel implements
 
             if (false !== strpos($url, self::BASE_URL_PLACEHOLDER)) {
                 $url = str_replace(self::BASE_URL_PLACEHOLDER, $this->_request->getDistroBaseUrl(), $url);
+            } elseif (rtrim($url, '/') . '/' === 'http://localhost/') {
+                $url = $this->_request->getDistroBaseUrl();
             }
 
             $this->_baseUrlCache[$cacheKey] = $this->getUrlModifier()->execute(


okorshenko:
Hi @romainruaud 
Thank you for your contribution. Could you please cover your fix with integration test so that we can proceed with PR acceptance? Thank you

romainruaud:
Hello @okorshenko 

imho it's already covered properly by : https://github.com/magento/magento2/blob/develop/app/code/Magento/Store/Test/Unit/Model/StoreTest.php#L235

(it's the last case of the DataProvider).

But it differs from what is really happening on a web server : 

When testing we expects $request->getDistroBaseUrl to return always "http://distro.com" which is obviously not always the case on server, especially if something is run via the cli.

On a server you will typically have this behavior : 

- set the web/secure/base_url to {{base_url}}
- install Magento, or flush the cache by cli, or whatever cli operation
- configuration get loaded, $request->getDistroBaseUrl returns "http://localhost/" since there is no http request.
- **this get cached**
- when accessing the application via the Webserver, configuration value is retrieven from cache and is now "http://localhost/"
- without my fix, your browser gets redirected to "http://localhost/" => fail
- with my fix, $request->getDistroBaseUrl is called => user is redirected to http://distro.com

Not sure how I could cover this better, maybe separate the testing when using the {{base_url}} placeholder to two other methods ? One with the HTTP request, and another without ?

Regards

orlangur:
@romainruaud,
> imho it's already covered properly by

The rule of thumb is: if you changed logic and no test started to fail, then logic is **NOT** covered by any test. Even if such new lines are executed during some test execution.

So, the test needs to be added for which old implementation would fail and new one is passing.

romainruaud:
Well forget about this PR, it does not fix properly the issue #10693 (issues remains with backend URLs).

To solve this one this is necessary to go deeper into configuration caching when using placeholder values (see the issue for more details).

Please let me know when your fix for internal MAGETWO-71893 will be available.

regards

PR #10698, opened 2017-08-29
Allow browser back/forward button to shipping step

amenk:
- Set visibility (before just everything disappeared when going back to the shipping step)
- The comment was misleading

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/view/shipping.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/view/shipping.js
@@ -126,11 +126,8 @@ define([
             return this;
         },
 
-        /**
-         * Load data from server for shipping step
-         */
         navigate: function () {
-            //load data from server for shipping step
+            this.isVisible(true);
         },
 
         /**


app/code/Magento/Checkout/view/frontend/web/js/view/shipping.js
```diff
@@ -126,11 +126,8 @@ define([
             return this;
         },
 
-        /**
-         * Load data from server for shipping step
-         */
```
omiroshnichenko, line 131: Add doc block.

app/code/Magento/Checkout/view/frontend/web/js/view/shipping.js
```diff
@@ -126,11 +126,8 @@ define([
...
         navigate: function () {
-            //load data from server for shipping step
+            this.isVisible(true);
```
omiroshnichenko, line 130: It will not work because navigate method has a context of shipping component instead of step context. Remove context binding in https://github.com/amenk/magento2/blob/3f2295fc131d879b853e521fee3b989f234196a3/app/code/Magento/Checkout/view/frontend/web/js/view/shipping.js#L89

ishakhsuvarov:
Hi @amenk 
Please check the review comments if you would like to proceed with this PR.
Thanks!

amenk:
Sorry don't have the time in the moment to correct the PR, maybe later .. if anybody want's to take over I am happy.

omiroshnichenko:
@amenk I will finish this issue.

omiroshnichenko (approved):

PR #10730, opened 2017-08-31
[Merged] Don't append a colon after payment method title in checkout.

evktalo:
### Description

The get method for payment methods title adds a colon. It's inconsistent with other checkout step titles. It is hard to remove it when customizing a theme.

### Manual testing scenarios
1. In default site with sample data, add items to cart
2. Proceed to checkout
3. Proceed to payment step. The 'Payment Methods' title should appear without the colon.

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/view/payment/list.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/view/payment/list.js
@@ -193,7 +193,7 @@ define([
                 title = this.defaultGroupTitle;
             }
 
-            return title + ':';
+            return title;
         },
 
         /**
--- a/app/code/Magento/Checkout/view/frontend/web/template/payment-methods/list.html
+++ b/app/code/Magento/Checkout/view/frontend/web/template/payment-methods/list.html
@@ -11,7 +11,7 @@
         <div if="getRegion($group().displayArea)().length"
              class="step-title"
              data-role="title">
-            <!-- ko i18n: getGroupTitle($group) --><!-- /ko -->:
+            <translate args="getGroupTitle($group)"/>:
         </div>
         <each args="data: getRegion($group().displayArea), as: 'method'" render=""/>
     </div>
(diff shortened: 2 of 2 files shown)


app/code/Magento/Checkout/view/frontend/web/js/view/payment/list.js
```diff
@@ -193,7 +193,7 @@ define([
                 title = this.defaultGroupTitle;
             }
 
-            return title + ':';
```
ishakhsuvarov, line 196: I assume colon should be moved to the template then, to preserve consistency with the current look.

ishakhsuvarov:
Hi @evktalo 
Would you like to update PR so that colon is still displayed?
Thank you.

evktalo:
Adding the colon back in a template still requires you to override the template to remove the semicolon. I found this unintuitive when implementing a design, where the title did not have the colon. I was expecting to simply change the title via translations (the case being I was working on a Finnish-only site).

I understand the desire to keep the status quo though. Removing the colon by overriding the template is a little bit easier than extending/overriding the js function, so there is still worth to this PR. :) If you can, consider removing the colon altogether however, as it is still inconsistent with the titles of the other steps.

app/code/Magento/Checkout/view/frontend/web/template/payment-methods/list.html
```diff
@@ -9,9 +9,9 @@
     <div repeat="foreach: paymentGroupsList, item: '$group'"
          class="payment-group">
         <div if="getRegion($group().displayArea)().length"
-             translate="getGroupTitle($group)"
              class="step-title"
              data-role="title">
+            <!-- ko i18n: getGroupTitle($group) --><!-- /ko -->:
```
omiroshnichenko: Please, replace comment with node(e.g `<translate args="getGroupTitle($group)" />:`)

vrann:
PR was merged to the mainline

PR #10740, opened 2017-09-01
add new field to order grid: payment_cctype

ebaschiera:
### Description

My idea is to add the cc_type information to the order grid in admin. This information is taken from `sales_order_payment.cc_type`. The new field is `sales_order_grid.payment_cctype`.

This is very useful for payment methods that save in `cc_type` relevant data that are useful to display and filter in the order grid. I am referring in particular to Adyen HPP, which stores "adyen_hpp" in "payment_method", but the real payment method (for example: banktransfer_iban, mastercard, ideal) is stored into cc_type column.

I added a particular feature: the filter loads the options from the table itself so that only existing cc_type(s) will be available to the filter. This is necessary for those payment methods that store dynamic values into `cc_type`.

### DIFF
--- a/app/code/Magento/Sales/Setup/UpgradeSchema.php
+++ b/app/code/Magento/Sales/Setup/UpgradeSchema.php
@@ -69,6 +69,9 @@ class UpgradeSchema implements UpgradeSchemaInterface
             $this->addColumnBaseGrandTotal($installer);
             $this->addIndexBaseGrandTotal($installer);
         }
+        if (version_compare($context->getVersion(), '2.0.4', '<')) {
+            $this->addColumnPaymentCctype($installer);
+        }
     }
 
     /**
@@ -103,5 +106,41 @@ class UpgradeSchema implements UpgradeSchemaInterface
             $installer->getIdxName('sales_invoice_grid', ['base_grand_total'], '', self::$connectionName),
             ['base_grand_total']
         );
+    }
+    
+    private function addColumnPaymentCctype(SchemaSetupInterface $installer)
+    {
+        /** @var $connection \Magento\Framework\DB\Adapter\Pdo\MySql */
+        $connection = $installer->getConnection();
+        $gridTableName = $installer->getTable('sales_order_grid');
+        $paymentTableName = $installer->getTable('sales_order_payment');
+
+        $cctypeColumns = [
+            'type' => \Magento\Framework\DB\Ddl\Table::TYPE_TEXT,
+            'length' => 32,
+            'nullable' => true,
+            'comment' => 'Payment CC Type',
+        ];
+
+        $connection->addColumn(
+            $gridTableName,
+            'payment_cctype',
+            $cctypeColumns
+        );
+        
+        
+        //Populate "payment_cctype" in sales_order_grid with values in sales_order_payments
+        $connection->query(
+            $connection->updateFromSelect(
+                $connection->select()
+                    ->join(
+                        $paymentTableName,
+                        sprintf('%s.entity_id = %s.parent_id', $gridTableName, $paymentTableName),
+                        'cc_type AS payment_cctype'
+                    ),
+                $gridTableName
+            )
+        );
+        
     }
 }
--- a/app/code/Magento/Payment/Ui/Component/Listing/Column/CcType/Options.php
+++ b/app/code/Magento/Payment/Ui/Component/Listing/Column/CcType/Options.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Magento\Payment\Ui\Component\Listing\Column\CcType;
+
+use Magento\Framework\Data\OptionSourceInterface;
+
+/**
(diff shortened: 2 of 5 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10740) <br/>All committers have signed the CLA.

miguelbalparda:
While interesting, this looks like a feature and not like a bugfix. Maybe create a module instead of including it in the core?

app/code/Magento/Sales/Setup/UpgradeSchema.php
```diff
@@ -104,4 +107,40 @@ private function addIndexBaseGrandTotal(SchemaSetupInterface $installer)
...
+        
+        
+        //Populate "payment_cctype" in sales_order_grid with values in sales_order_payments
+        $connection->query(
+            $connection->updateFromSelect(
+                $connection->select()
+                    ->join(
+                        $paymentTableName,
+                        sprintf('%s.entity_id = %s.parent_id', $gridTableName, $paymentTableName),
+                        'cc_type AS payment_cctype'
```
orlangur, line 139: Why need to denormalize instead of dynamic joins?

ebaschiera:
Hi @miguelbalparda,
yes, this is a feature. Only bug fixes are allowed via PR?
I didn't think about a module before because I thought that this would have been useful if it was in the core. I just wanted to contribute. If this feature will not be welcome, I will proceed with a module, no worries. :)

miguelbalparda:
We usually try to disencourage feature requests here @ebaschiera, feel free to add this [in the forum](https://community.magento.com/t5/Magento-2-Feature-Requests-and/idb-p/feature-requests). It might probably get some more traction if you already have a working module.

app/code/Magento/Sales/Setup/UpgradeSchema.php
ebaschiera: I am not sure what you mean... I looked at the `sales_order_grid.payment_method` column and followed that approach. Those lines take care of copying existing values from `sales_order_payment.cc_type` to `sales_order_grid.payment_cctype`

ebaschiera:
Thanks for pointing that out @miguelbalparda , I wasn't aware of that. Sorry!

app/code/Magento/Sales/Setup/UpgradeSchema.php
orlangur: Why this data needs to be copied at all? How it works when there are multiple invoices associated to order with different payment methods?

orlangur:
@miguelbalparda @ebaschiera I would not say feature contributions to core are restricted but any of them should be evaluated thoroughly. Ideally it should be done prior to implementation just to avoid waste of efforts.

In general `Catalog`, `Checkout`, `Payment`, `Sales` modules are already bloated enough, new features are best to be created as disableable separate modules. In this case I'm not sure, maybe like `payment_method` field it fits naturally into `Sales`.

And the last one, `2.1-develop` target branch should only be used for backports already merged into main `develop` branch. I believe 2.1 and 2.2 already passed the feature freeze stage, current `develop` reflects the future 2.3 release which will contain new features, including ones contributed by Community. Features already approved for development are marked with `'up for grabs'` label: https://github.com/magento/magento2/issues?q=is%3Aopen+is%3Aissue+label%3A"up+for+grabs" Maybe you'll be interested in participation in some of them.

app/code/Magento/Sales/Setup/UpgradeSchema.php
ebaschiera: I understand your point, but I believe that your questions are valid also for `sales_order_grid.payment_method` , that is in the core and that I took inspiration from.
Data need to be copied to populate the field in the existing orders.

ebaschiera:
Thanks @orlangur , I will have a look.

miguelbalparda:
No need to be sorry, please keep doing what you are doing to help. I really appreciate the fact that you took some of your time to contribute to the project. I'll close this for now, it can be reopen if anything changes in the future.

korostii:
Hi @miguelbalparda,
To make this perfectly clear: I assume this PR is closed because it targets `2.1-develop` ?
> We usually try to disencourage feature requests here

Whom do you mean by "we" in this sentence and is it written down somewhere? 
What's motivation behind discouraging the community from contributing new features into the core?
Unless you can support that sentence with something, it looks very odd to me.

What @orlangur says makes much more sense. I'd be glad to see that as an official policy \ standard response on such PRs. These are all perfect recommendations:

> any of them [feature contributions] should be evaluated thoroughly. Ideally it should be done prior to implementation just to avoid waste of efforts.
> new features are best to be created as disableable separate modules
> Features already approved for development are marked with 'up for grabs' label: https://github.com/magento/magento2/issues?q=is%3Aopen+is%3Aissue+label%3A"up+for+grabs"

PR #7919, opened 2016-12-21
Using Dynamic Protocol Concatination

brobie:
Without this fix, when accessing a product page in https, it produces: Mixed Content: This page ______ was loaded over HTTPS, but requested an insecure resource 'http://player.vimeo.com/video/158730450?api=1&player_id=vimeo1587304501482350824734&autoplay=1'. This request has been blocked; the content must be served over HTTPS.

PR #10748, opened 2017-09-02
Backport PR-7919 for Magento 2.1 - Using Dynamic Protocol Concatination

hostep:
(cherry picked from commit 1fbe1f004ed911291e8a1ee390dff409c7a35b4f)

### Description
Backported https://github.com/magento/magento2/pull/7919 for Magento 2.1

1. https://github.com/magento/magento2/issues/7311: Vimeo videos in product gallery do not work over https
2. https://github.com/magento/magento2/issues/8574: Product Gallery Vimeo Videos Don't Play Over HTTPS

I might have missed some though.

### DIFF
--- a/app/code/Magento/ProductVideo/view/adminhtml/web/js/get-video-information.js
+++ b/app/code/Magento/ProductVideo/view/adminhtml/web/js/get-video-information.js
@@ -302,7 +302,7 @@ define([
                     additionalParams += '&autoplay=1';
                 }
 
-                src = window.location.protocol + '//player.vimeo.com/video/' +
+                src = 'https://player.vimeo.com/video/' +
                     this._code + '?api=1&player_id=vimeo' +
                     this._code +
                     timestamp +
@@ -525,7 +525,7 @@ define([
                     );
                 } else if (type === 'vimeo') {
                     $.ajax({
-                        url: window.location.protocol + '//www.vimeo.com/api/v2/video/' + id + '.json',
+                        url: 'https://www.vimeo.com/api/v2/video/' + id + '.json',
                         dataType: 'jsonp',
                         data: {
                             format: 'json'
--- a/app/code/Magento/ProductVideo/view/frontend/web/js/load-player.js
+++ b/app/code/Magento/ProductVideo/view/frontend/web/js/load-player.js
@@ -317,7 +317,7 @@ define(['jquery', 'jquery/ui'], function ($) {
             if (this._loop) {
                 additionalParams += '&loop=1';
             }
-            src = 'http://player.vimeo.com/video/' +
+            src = 'https://player.vimeo.com/video/' +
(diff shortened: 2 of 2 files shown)


orlangur:
Will it work correctly if we omit protocol at all?

hostep:
I think the better fix is to hardcode it to `https` instead of using the current protocol, like is being done for youtube: https://github.com/hostep/magento2/blob/0f37ce0310f86e40e87de33b7ea1e1b42b1eea9a/app/code/Magento/ProductVideo/view/frontend/web/js/load-player.js#L245

I just cherry-picked this like it is currently on the `develop` branch.

If it is decided that using `https` hardcoded is the best solution, then we first have to fix this in the `develop` branch and then include that fix in this PR.

orlangur:
>  then we first have to fix this in the develop branch and then include that fix in this PR

Yeah, I understand this for any changed variant.

Why hardcoding https is the best one instead of using current protocol? Something like "always use https when it is available as more secure protocol"?

hostep:
@orlangur: indeed. And also because vimeo redirects all requests using `http` to `https`, so if you host Magento over `http`, then the video will load a fraction of time faster since only one request is made instead of 2 requests.

But the current fix which I backported is already a step forwards, otherwise some browsers will refuse to play the video when you host Magento over `https` and load a vimeo video over `http`, due to [mixed content](https://developers.google.com/web/fundamentals/security/prevent-mixed-content/what-is-mixed-content) warnings.

Let's wait for feedback from the Magento devs to see what solution they want to go for.

orlangur:
> because vimeo redirects all requests using http to https

Kudos! That was something I was not aware of.

> already a step forwards, otherwise some browsers

Sure thing, the problem is clear, I've faced with similar situation during custom development with other services.

> Let's wait for feedback from the Magento devs to see what solution they want to go for.

Please go forward with proposed always-https implementation to `develop` branch and then backport the best possible fix to `2.1.x`. I will process both of such PRs (this one can be reused with force-push).

hostep:
Created PR for the `develop` branch over here: https://github.com/magento/magento2/pull/10768

orlangur:
Let's go ;)

orlangur:
Ping @hostep :) Please make this PR a copy of one merged into `develop`. `2.2` will remain somewhat like "a temporary slightly less effective fix".

hostep:
@orlangur: ok, amended commit and force pushed it.

orlangur (approved):

Ctucker9233:
Still not included in 2.1.10. What version of 2.1 will have it?

hostep:
@Ctucker9233, it is included in 2.1.10, see for example: https://github.com/magento/magento2/blob/2.1.10/app/code/Magento/ProductVideo/view/frontend/web/js/load-player.js#L320

Ctucker9233:
@hostep Thanks for the clarification!

PR #10760, opened 2017-09-04
phpDocumentor Return tag must be an int

reindert-vetter:
Because count() gives an int, hasShipment returns an int. Now i can't do this:
```if ($view->getOrder()->hasShipments() === true) {```

### DIFF
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -1774,17 +1774,17 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
      */
     public function hasInvoices()
     {
-        return $this->getInvoiceCollection()->count();
+        return (bool)$this->getInvoiceCollection()->count();
     }
 
     /**
      * Check order shipments availability
      *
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by magento-cicd2** [comment]
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10760) <br/>All committers have signed the CLA.

**Comment by orlangur** [comment]
@reindert-vetter code like 
`if ($view->getOrder()->hasShipments() === true) {`
should be always written as
`if ($view->getOrder()->hasShipments()) {`
actually.

I do agree that current implementation of the method could be a bit confusing but I believe we should not change contract here but just change implementation.

Please change all `has*` methods in this class so that they always return boolean value if you wish:
```
public function hasShipments()
    {
        return (bool)$this->getShipmentsCollection()->count();
    }
```
To match implementation with PHPDoc correctly.

reindert-vetter:
Done

orlangur:
Great, thanks for the quick change!

orlangur (approved):

PR #10768, opened 2017-09-04
Always use https for Vimeo video's.

hostep:
### Description
Instead of using the current protocol being used by Magento (http or https), always go for https. Since Vimeo redirects all non-https traffic to https.
See discussion with @orlangur in https://github.com/magento/magento2/pull/10748

This still happens in the fotorama library though: https://github.com/magento/magento2/blob/a232e0753026b081a83f1aa6fd47800da125a8c6/lib/web/fotorama/fotorama.js#L853-L858
Not sure if this needs to be changed as well?

Watch out, I didn't test these changes, so please test before approving if possible :)

Once this is in the `develop` branch and backported to Magento 2.2 by someone, I'll update https://github.com/magento/magento2/pull/10748 so this also gets into Magento 2.1
I don't want to have an awkward situation where this is merged in `develop` and `2.1` but not in `2.2`

None

### Contribution checklist

### DIFF
--- a/app/code/Magento/ProductVideo/view/adminhtml/web/js/get-video-information.js
+++ b/app/code/Magento/ProductVideo/view/adminhtml/web/js/get-video-information.js
@@ -302,7 +302,7 @@ define([
                     additionalParams += '&autoplay=1';
                 }
 
-                src = window.location.protocol + '//player.vimeo.com/video/' +
+                src = 'https://player.vimeo.com/video/' +
                     this._code + '?api=1&player_id=vimeo' +
                     this._code +
                     timestamp +
@@ -525,7 +525,7 @@ define([
                     );
                 } else if (type === 'vimeo') {
                     $.ajax({
-                        url: window.location.protocol + '//www.vimeo.com/api/v2/video/' + id + '.json',
+                        url: 'https://www.vimeo.com/api/v2/video/' + id + '.json',
                         dataType: 'jsonp',
                         data: {
                             format: 'json'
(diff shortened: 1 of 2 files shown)


orlangur (approved):

orlangur:
> Not sure if this needs to be changed as well?

Nope, but you can suggest such improvement to fotorama itself. Then it can appear in Magento someday after uprading the fotorama. Not sure what the process would be if there was a severe bug in third-party library but luckily this is not the case.

> an awkward situation where this is merged in develop and `2.1` but not in `2.2`

`2.2` is currently closed for contributions, however, as to me it is not really necessary to backport as `2.2` contains another version of the fix. What do you think?

hostep:
@orlangur: ok, but then #10748 should be merged in 2.1 as is without other changes.

Currently there is only a bug in 2.1, not in 2.2 or `develop`.
The changes proposed in this PR, are only nice-to-haves.

So summary:
- using hardcoded `http://` for vimeo url's is incorrect and will expose bugs when you run the shop over `https` (this is what happens on 2.1 right now)
- using the same protocol for vimeo as the Magento shop isn't a problem, but has as a consequence that when running in Magento over `http`, it will perform 2 HTTP requests for every Vimeo loaded on the shop (this is what happens on 2.2 and `develop` right now)
- using hardcoded `https://` for vimeo url's in the best solution, as it will work perfectly on Magento shops using both `http` and `https` (this is what is proposed in this PR and will get merged to `develop` when approved)

orlangur:
> then #10748 should be merged in 2.1 as is without other changes

Why? I thought it would be better similar to this PR, so that 2.1 and 2.3+ have one version and 2.2 will have redundant redirect on Vimeo for http, but it's not a big deal.

hostep:
In this particular case it's not really a big problem, but it feels very wrong to me to leave 2.2 behind and have an improvement in 2.1 which is also in 2.3 but not in 2.2...

PR #10770, opened 2017-09-04
Fix: Run php-cs-fixer as part of static tests

localheinz:
### Description

As of the moment, [`friendsofphp/php-cs-fixer`](https://github.com/magento/magento2/blob/c2810e0d6014fd8aa787d24e5d46c6593437f00a/composer.json#L81) is used within this project, but it is only run as part of the [`pre-commit`](https://github.com/magento/magento2/blob/c2810e0d6014fd8aa787d24e5d46c6593437f00a/dev/tools/Magento/Tools/StaticReview/pre-commit) hook.

This PR

* adds a new test that runs `php-cs-fixer` in `dry-run` mode and fails if the exit code is not `0`
* speeds up the build by caching `.php_cs.cache` between Travis builds
* runs `php-cs-fixer`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/2.2#caching.

n/a

### Manual testing scenarios

n/a

### Contribution checklist

### DIFF
--- a/dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php
+++ b/dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php
@@ -199,7 +199,7 @@ class LiveCodeTest extends \PHPUnit\Framework\TestCase
         return Files::init()->readLists(__DIR__ . '/_files/whitelist/common.txt');
     }
 
-    public function testCodeStyle()
+    public function testCodeStyleWithPhpCodeSniffer()
     {
         $reportFile = self::$reportDir . '/phpcs_report.txt';
         $codeSniffer = new CodeSniffer('Magento', $reportFile, new Wrapper());
@@ -207,6 +207,26 @@ class LiveCodeTest extends \PHPUnit\Framework\TestCase
             0,
             $result = $codeSniffer->run($this->getFullWhitelist()),
             "PHP Code Sniffer detected {$result} violation(s): " . PHP_EOL . file_get_contents($reportFile)
+        );
+    }
+
+    /**
+     * @SuppressWarnings(PHPMD.UnusedLocalVariable)
+     */
+    public function testCodeStyleWithPhpCsFixer()
+    {
+        $fixCommand = './vendor/bin/php-cs-fixer fix . --config=.php_cs.dist';
+        $checkCommand = $fixCommand . ' --dry-run 2> /dev/null';
+
+        exec($checkCommand, $output, $exitCode);
+
+        $this->assertEquals(
+            0,
+            $exitCode,
+            sprintf(
+                'php-cs-fixer detected violations, please run "%s"',
+                $fixCommand
+            )
         );
     }
 
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -24,7 +24,10 @@ $finder = PhpCsFixer\Finder::create()
     ->exclude('setup/vendor')
     ->exclude('var');
 
+$cacheDir = getenv('TRAVIS') ? getenv('HOME') . '/.php-cs-fixer' : __DIR__;
+
 return PhpCsFixer\Config::create()
+    ->setCacheFile($cacheDir . '/.php_cs.cache')
     ->setFinder($finder)
     ->setRules([
         '@PSR2' => true,
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ cache:
   directories:
     - $HOME/.composer/cache
     - $HOME/.nvm
+    - $HOME/.php-cs-fixer
     - $HOME/node_modules
     - $HOME/yarn.lock
 before_install: ./dev/travis/before_install.sh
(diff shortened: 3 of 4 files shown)


orlangur:
Thanks, you did it the right way, unlike similar PRs with manual changes 👍 

However, such process will be endless if there is no static test enforcing the rules. Could you please create a test similar to https://github.com/magento/magento2/blob/develop/dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php#L202 which would run `php-cs-fixer` in `dry-run` mode using the https://github.com/magento/magento2/blob/develop/.php_cs.dist ruleset?

Also, among tests currently broken on Travis you can ignore integration tests failure but some unit tests need to be adopted.

localheinz:
Thank you for your feedback, @orlangur! 

Taking a look immediately.

localheinz:
@orlangur 

Not sure if this is what you had in mind, especially since this PR entails a lot of changes to the code after actually running `php-cs-fixer`.

Please let me know if I can improve this!

orlangur:
@localheinz thanks, will give it a try. For now could you please force push this branch without `php-cs-fixer` applied to simplify code review and work on stabilization in another branch (which should consist of `php-cs-fixer` applied and then arbitrary amount of needed manual changes, like the one needed to fix unit test failure).

dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php
```diff
@@ -210,6 +210,26 @@ public function testCodeStyle()
...
 
+    /**
+     * @SuppressWarnings(PHPMD.UnusedLocalVariable)
+     */
+    public function testCodeStyleWithPhpCsFixer()
+    {
+        $fixCommand = './vendor/bin/php-cs-fixer fix . --config=.php_cs.dist';
+        $checkCommand = $fixCommand . ' --dry-run 2> /dev/null';
+
+        exec($checkCommand, $output, $exitCode);
```
localheinz, line 221: Suppressing warnings for unused local variables here because I need to pass in `$output` (never used) to be able to assert on `$exitCode`.

localheinz:
@orlangur 

What do you need me to update here?

orlangur:
Sorry, didn't have time to play with the implementation today, hope to get some till end of the week.

Wanted to do a thorough code review before suggesting any changes so that you can apply them all at once.

Just a quick list of thoughts:

1. `php-cs-fixer` execution must be encapsulated in separate class, like other tools
2. Using `exec` is probably not the best choice, maybe we can just use some PHP classes instead
3. How does the `php-cs-fixer` work if no cache dir specified, for non-Travis env we definitely need something better than project root folder
4. We need to see actual violations instead of passing them to `\dev\null`
5. (probably, could be addressed separately) Maybe it make sense to check only changed files, similar to `testCodeMess`
6. No need to specify config file `--config=.php_cs.dist`

Please concentrate on making builds pass after automatic change in temporary branch (or maybe you alreaady achieved green builds, I didn't check your fork).

keradus:
If you wish, you can consider using Runner (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/src/Runner/Runner.php) or FixCommand (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/src/Console/Command/FixCommand.php)
more boilerplate, yet no need to call console

localheinz:
I think that running `php-cs-fixer` as part of tests is a bad idea, and shouldn't be done in the first place.

As far as I see it, both

* `squizlabs/php_codesniffer`
* `friendsofphp/php-cs-fixer`

are primarily command-line tools, and they should be used as such.

Wrapping a run of either of these tools in a test has the following disavantages:

* it introduces complexity that serves no purpose
* no output is rendered until the tests are finished

If you want to enforce running these tools, then do this

* run them on Travis and fail the build if these runs don't pass
* add composer scripts that make it easy for contributors to run these

I have no interest in building something more complex than necessary.

orlangur:
@localheinz,

> it introduces complexity that serves no purpose

Not true, such unified approach _decreases_ complexity in fact.

> no output is rendered until the tests are finished

How partial `php-cs-fixer` output is useful for you?

> run them on Travis and fail the build if these runs don't pass

Such approach do increases complexity in fact. Note that Travis CI is not the only and not the main CI used, there are at least Bamboo and Jenkins also used internally.

> I have no interest in building something more complex than necessary.

No problems, thanks for your endeavors to make Magento better.

@keradus do you see why `php-cs-fixer` should not be used that way? :)

keradus:
I agree with @localheinz here, running any (not only PHP CS Fixer) SCA tool as part of test is abusing SCA idea. PHPUnit is a tool to execute **unit tests**, not Static Code Analysis. What you want to achieve in this project is having single entry job runner, and you decided to put it on top of PHPUnit instead of using proper job runners like grumphp or even composer scripts. PHPUnit is not for that.

PR #10779, opened 2017-09-05
Fix: Reference phpunit.xsd using relativ path

localheinz:
### Description

Using a relative path to the corresponding `phpunit.xsd` schema has the following advantages

* when using an IDE such as PhpStorm, an external resource does not need to be fetched
* the reference does not need to be modified when version constrains for `phpunit/phpunit` are modified.

n/a

### Manual testing scenarios

n/a

### Contribution checklist

### DIFF
--- a/dev/tests/api-functional/phpunit.xml.dist
+++ b/dev/tests/api-functional/phpunit.xml.dist
@@ -8,7 +8,7 @@
  */
 -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="../../../vendor/phpunit/phpunit/phpunit.xsd"
          colors="true"
          beStrictAboutTestsThatDoNotTestAnything="false"
          bootstrap="./framework/bootstrap.php"
--- a/dev/tests/functional/phpunit.xml.dist
+++ b/dev/tests/functional/phpunit.xml.dist
@@ -6,7 +6,7 @@
  */
 -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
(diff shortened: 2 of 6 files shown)


dev/tests/functional/phpunit.xml.dist
```diff
@@ -6,7 +6,7 @@
  */
 -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
```
localheinz, line 9: [`dev/tests/functional/composer.json`](https://github.com/magento/magento2/blob/c2810e0d6014fd8aa787d24e5d46c6593437f00a/dev/tests/functional/composer.json) suggests that for running functional tests, the dependencies declared therein will be used. Not sure if this is actually the case, but this is the reason I'm using a different location here.

dev/tests/api-functional/phpunit.xml.dist
```diff
@@ -8,7 +8,7 @@
  */
 -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
```
localheinz, line 11: As can be seen in [`composer.json`](https://github.com/magento/magento2/blob/c2810e0d6014fd8aa787d24e5d46c6593437f00a/composer.json#L77), we are using `phpunit/phpunit:~6.2.0`, so the schema referenced here is actually not applicable anymore.

localheinz:
Build failure is unrelated to the changes in this PR.

okorshenko:
Hi @localheinz 
Thank you for your contribution. Unfortunately, we can not accept this pull request as is. `vendor` directory name is configurable and can be different on the other installations. So we can not hard-code it. This is known issue and Magento 2 Core Team already working on this issue. The fix will be delivered soon. 
Thank you

localheinz:
Thank you for pointing out, @okorshenko!

PR #10784, opened 2017-09-06
Fix: Use all columns when running tests

localheinz:
### Description

It has been possible for a while to configure the number of columns `phpunit` should use for rendering progress output.

While [`80`](https://github.com/sebastianbergmann/phpunit/blob/6.2/phpunit.xsd#L205) is the default, we can also set it to [`max`](https://github.com/sebastianbergmann/phpunit/blob/6.2/phpunit.xsd#L102-L113), which means that `phpunit` will use the entire width of the screen if it has been able to detect it.

Since most of the screens we use today have more than `80` columns, this has the advantage that we we can fit more output on a screen, requiring less scrolling if needed.

For reference, see https://phpunit.de/manual/current/en/textui.html#textui.clioptions:

>`--columns`
>Defines the number of columns to use for progress output. If `max` is defined as value, the number of columns will be maximum of the current terminal.

n/a

### Manual testing scenarios

n/a

### Contribution checklist

Somewhat related to #10779.

### DIFF
--- a/dev/tests/api-functional/phpunit.xml.dist
+++ b/dev/tests/api-functional/phpunit.xml.dist
@@ -10,6 +10,7 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
          colors="true"
+         columns="max"
          beStrictAboutTestsThatDoNotTestAnything="false"
          bootstrap="./framework/bootstrap.php"
 >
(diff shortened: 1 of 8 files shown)


orlangur:
I don't see any difference on Travis CI: https://travis-ci.org/magento/magento2/jobs/272362095

Some lines are longer than others but PHPUnit cannot determine the maximum is more than 80 to display dots?

localheinz:
@orlangur 

As I pointed out above:

>While `80` is the default, we can also set it to `max`, which means that `phpunit` will use the entire width of the screen **if it has been able to detect it**.

You can see the effect when running tests locally.

orlangur (approved):

orlangur:
@localheinz thanks, was just checking as it wasn't explicitly mentioned that such option does not work on Travis CI at the moment.

PR #10785, opened 2017-09-06
Fix: Consistently use verbose option when running tests

localheinz:
### Description

The `verbose` option is already configured in [`dev/tests/functional/phpunit.xml.dist`](https://github.com/magento/magento2/blob/c2810e0d6014fd8aa787d24e5d46c6593437f00a/dev/tests/functional/phpunit.xml.dist#L13). 

This PR consistently applies it to all `phpunit` configurations.

For reference, see https://phpunit.de/manual/current/en/textui.html#textui.clioptions:

>`--verbose`
>Output more verbose information, for instance the names of tests that were incomplete or have been skipped.

n/a

### Manual testing scenarios

n/a

### Contribution checklist

Somewhat related to #10779.

### DIFF
--- a/dev/tests/api-functional/phpunit.xml.dist
+++ b/dev/tests/api-functional/phpunit.xml.dist
@@ -12,6 +12,7 @@
          colors="true"
          beStrictAboutTestsThatDoNotTestAnything="false"
          bootstrap="./framework/bootstrap.php"
+         verbose="true"
 >
     <!-- Test suites definition -->
     <testsuites>
--- a/dev/tests/integration/framework/tests/unit/phpunit.xml.dist
+++ b/dev/tests/integration/framework/tests/unit/phpunit.xml.dist
@@ -10,6 +10,7 @@
          colors="true"
          beStrictAboutTestsThatDoNotTestAnything="false"
          bootstrap="./framework/bootstrap.php"
+         verbose="true"
 >
     <!-- Test suites definition -->
     <testsuites>
--- a/dev/tests/integration/phpunit.xml.dist
+++ b/dev/tests/integration/phpunit.xml.dist
@@ -11,6 +11,7 @@
          beStrictAboutTestsThatDoNotTestAnything="false"
          bootstrap="./framework/bootstrap.php"
          stderr="true"
+         verbose="true"
 >
     <!-- Test suites definition -->
     <testsuites>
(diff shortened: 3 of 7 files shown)


orlangur:
I'm not quite sure why it was added in functional tests (I remember there was a combination of `verbose` + `debug` to see failed tests even when the build not finishes correctly) but this is definitely something we don't want to bloat the output for other test types: https://travis-ci.org/magento/magento2/jobs/272363867

Among other possible improvements related to PHPUnit configuration I remember there is an undesired output in integration tests which could be eliminated and then enforced and a bunch of other `strict*` options which can be useful to enforce. Maybe this is something you may be interested to work on.

localheinz:
Here are the reasons why I personally use all of the strict configurations as well as the `verbose` option:

- if a test is skipped, I want to know which one so it can be fixed
- if a test is incomplete, I want to know which one so it can be fixed
- if a test does not adhere to the strictest requirements, I want to know which one, so it can be fixed

Without the `verbose` option, I can see that tests are marked as incomplete, risky, or skipped - but I don't know which one. While it may seem like the output is bloated, it contains just that missing piece of information, and serves as a reminder to fix these issues. 

Without it, these tests hang around forever, unfixed, and developers get used to it. 

Related: https://danluu.com/wat/. 

So how about accepting the change, and then focusing on fixing these issues? Happy to help with that, of course!

orlangur:
I do agree that we should have ~zero incomplete and minimum of skipped tests, just that 99.9% of developers running these tests are not working on incomplete/skipped tests elimination.

On the other hand, anyone working on decreasing of such tests is not blocked in any way as can pass such option for his needs.

Thanks for the good point, I'll raise the idea to report currently incomplete tests as `'up for grabs'` ticket. Most of them are tied to some internal tickets, just maybe too far in roadmap.

orlangur:
> if a test is skipped, I want to know which one so it can be fixed

As far as I remember the PHPUnit documentation `markTestSkipped` is perfectly valid but should only be used for incompatible environment (like MySQL-specific test running under Oracle).

I've consulted with maintainers group, as far as they know there are no ongoing tasks involving unskipping some tests currently. So, feel free to improve any of them, I believe separate ticket reported would make sense in case of coordination, like it was done with ZF1 usages elimination.

I would recommend to start from ones not associated with any `MAGETWO` ticket, like https://github.com/magento/magento2/blob/develop/app/code/Magento/ImportExport/Test/Unit/Model/ImportTest.php#L349

PR #10789, opened 2017-09-06
enable universal use of modal and tab_group

netzkollektiv:
This pull request enables a universal use of modal and tab_group. In detail it allows having tab_group (used in customer edit view) to have child elements which do not need to be validated. Adding a construction like the "related products add"-functionality from catalog product edit via modal was not possible, because the modal inside the tab_group was tried to be validated as a field.

This change request ignores `undefined` elements in validation, but considers `false` elements as `false` and `true` elements as `true`.

### Description
see above

### Manual testing scenarios
1. add similar construction as add related products fucntionality from catalog product edit in customer product edit
2. try to save the customer
3. form will not validate due to `undefined` error


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10789) <br/>All committers have signed the CLA.

ishakhsuvarov:
Hi @netzkollektiv 
Thank you for your contribution. Please sign the CLA with the email used for the commit, so that we could proceed with processing.
Also it would be nice if you could provide some more detailed code example to simplify review and testing.
Thank you!

netzkollektiv:
Hi @ishakhsuvarov

I signed it several times, but I always get:
>Thank you for signing the CLA!
> we are now taking you back to GitHub

and it remains unsigned in github.

Here's a basic example of a adding a new tab including a button and a modal to the customer form:
https://gist.github.com/netzkollektiv/e5bdccc54de3a94cda23aabf69031633

ishakhsuvarov:
@netzkollektiv 
> I signed it several times, but I always get

To fix this you should add an email address used for the commit to your github account (may be private).

Thank you

netzkollektiv:
I added the commit email to the netzkollektiv account (was part of my private acc), but it still does not work. What should I do?

netzkollektiv:
@ishakhsuvarov it needs to be the primary email - interesting.
ok, now you should be good to go, right?

ishakhsuvarov:
@netzkollektiv 
> needs to be the primary email - interesting

Wasn't aware of that, sorry.

Thank you for the update. We will proceed with review and testing.

PR #10792, opened 2017-09-06
Protected getters and setters should be public

VincentMarmiesse:
Hello,

These getters and setters should be public and not protected function, otherwise we can't access to the class params.

### DIFF
--- a/app/code/Magento/Catalog/Helper/Image.php
+++ b/app/code/Magento/Catalog/Helper/Image.php
@@ -600,7 +600,7 @@ class Image extends AbstractHelper
      * @param string $watermark
      * @return $this
      */
-    protected function setWatermark($watermark)
+    public function setWatermark($watermark)
     {
         $this->_watermark = $watermark;
         $this->_getModel()->setWatermarkFile($watermark);
@@ -612,7 +612,7 @@ class Image extends AbstractHelper
      *
      * @return string
      */
-    protected function getWatermark()
+    public function getWatermark()
     {
         return $this->_watermark;
     }
@@ -623,7 +623,7 @@ class Image extends AbstractHelper
      * @param string $position
      * @return $this
      */
-    protected function setWatermarkPosition($position)
+    public function setWatermarkPosition($position)
     {
         $this->_watermarkPosition = $position;
         $this->_getModel()->setWatermarkPosition($position);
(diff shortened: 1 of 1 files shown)


orlangur:
As far as I can see from implementation they are not supposed to be used directly but only via `\Magento\Catalog\Helper\Image::init` method. Please provide a scenario when it is not enough.

VincentMarmiesse:
Hello @orlangur,
I wanted to add a Plugin on the `getLabel` method and retrieve the image's product inside my Plugin.

orlangur:
Necessary product data can be intercepted in `init` method

VincentMarmiesse:
But how can I retrieve the product if I create a Plugin on the `getLabel` function? The `init` method won't help me here right?

orlangur:
For example, such plugin would intercept both `init` and `getLabel` methods and store intercepted product instance internally.

VincentMarmiesse:
Oh ok. But do you agree it's not really intuitive and practical to use? :)

orlangur:
It is quite straightforward if one notice that `init` method is the only way to change image helper state :) Still better than breaking class encapsulation and making its state hardly predictable by exposing unnecessary setters.

Are there any other customization problems which cannot be solved in current implementation?

orlangur:
Another question... Are you trying to customize behavior of some core code or it's a custom implementation?

VincentMarmiesse:
It's a custom implementation, it want to calculate the image label with some product information instead of just displaying it.

orlangur:
In such case you have full control on how and when image helper is called and plugin creation would be a misuse of interception mechanism.

Just encapsulate logic in some class so that you do
```
function ...
{
    $product->setSomeLabel(...);
    $imageHelper->init($product, ...);
}
```

or even not rely on `getLabel` from helper at all.

orlangur:
Let's wait for some alternative opinions a bit. I'm pretty sure we should not expose setters but probably public `getProduct` may appear to be useful for some other customization scenarios.

VincentMarmiesse:
Ok so in your opinion, I should not use a plugin here? I thought it was exactly the perfect scenario to use one.

orlangur:
@VincentMarmiesse yeah, plugin is best suitable when you need to customize some core behavior to your needs. In this case while technically possible it is better to reuse core functionality differently, aggregating image helper in your new class instead of pluginizing it and simply put custom `getLabel` implementation somewhere else.

orlangur:
@VincentMarmiesse thanks for collaboration! Hope suggested implementation approach will work like a charm for you leaving image helper methods with their current visibility (the stricter = the better).

PR #10796, opened 2017-09-06
Fix: Run php-cs-fixer

localheinz:
### Description

As suggested in https://github.com/magento/magento2/pull/10770#issuecomment-327399551, this PR applies the result of running

```
$ ./vendor/bin/php-cs-fixer fix . --config=.php_cs.dist
```

n/a

### Manual testing scenarios

n/a

### Contribution checklist


orlangur:
Let's just continue in https://github.com/magento/magento2/pull/10770, we will decide whether it's suitable to apply `php-cs-fixer` at the moment or merge a ready-to-use test in incomplete state in scope of it.

What I meant is that you need to work on branch stabilization (make all Travis CI builds green except known failure in integration tests) in scope of temporary branch. There is no need to create PR with `php-cs-fixer` applied yet.

Thanks for collaboration!

localheinz:
@orlangur 

I am a bit confused.

1)  In https://github.com/magento/magento2/pull/10770#issuecomment-327126770 you suggest I create a test to assert that rules are followed, which of course will fail if the rules aren't applied.

2) In https://github.com/magento/magento2/pull/10770#issuecomment-327399551 you suggest I split off applying the fixes into a separate branch.

So with #10770 1) is addressed, and with #10796 2) is addressed. But 1) cannot make the build pass with 2). 

So what should be it?

Can you explain?

orlangur:
@localheinz such contribution should consist of three parts:
1) Test to enforce rules
2) Huge commit with automatic changes
3) Necessary manual changes to make builds pass (as you remember, even unused imports removal caused some unit tests failures)

Сould take some time to process as we have B2B and EE codebases also. What I was suggesting is that we definitely need 1) and 3) but probably 2) will be delayed a bit, until a suitable moment. In such case first PR will be merged with ready-to-use, but contain `markTestIncomplete`.

PR #10817, opened 2017-09-08
#10767 Fix Braintree payment prevent multiple charges for the same order

maximbaibakov:
The pull request fixing an issue with Braintree payment gateway (Credit Cards only), when customer can be charged multiple times for the same order in case multiple clicking on Order button

### Description
Currently, when customer click order button, it is trigger preparation of request to communicate the order to braintree. Once the request is ready, it will hover the whole page which will block customer todo any adjustments to the checkout screen.

The preparation sometimes could take around 1-5 seconds depends on customer computer (execution JS). During the time, customer can't see any indication that something is happening and might think that he did not click a button and trying to click the button once more. Also, for sale events it might have worth case, when customers are trying to buy products as far as they can till they sold out. In that both cases multiple charges might appear in braintree, however, only 1 order is created in Magento.

What I'm proposing to change in the PR is to hover the whole page straight after customer clicked Order button, before any other actions are started. 
That will give an indication to a customer that something is happening
That will give browser time to prepare and send requests
That will stop customers from multiple charges
In case, if any errors happened, it will remove the hover from the screen.

We are using the fix for more then a year now, and we have not seen any issues with that so far.

1. magento/magento2#10767: Race condition causing duplicate orders with double-clicks on Braintree "Pay" button

### Manual testing scenarios
1. Add product to cart
2. Go to checkout screen
3. Select Braintree Credit Card as payment method
4. Put your credit card details
5. Try to click Order button as far as you can multiple times until the browser will hover the screen
6. If the fix is working correctly, then you will be charged only once, if not, then you will see multiple charges/transactions in braintree

### Contribution checklist

### DIFF
--- a/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/hosted-fields.js
+++ b/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/hosted-fields.js
@@ -10,8 +10,9 @@ define([
     'Magento_Braintree/js/view/payment/method-renderer/cc-form',
     'Magento_Braintree/js/validator',
     'Magento_Vault/js/view/payment/vault-enabler',
-    'mage/translate'
-], function ($, Component, validator, VaultEnabler, $t) {
+    'mage/translate',
+    'Magento_Checkout/js/model/full-screen-loader'
+], function ($, Component, validator, VaultEnabler, $t, fullScreenLoader) {
     'use strict';
 
     return Component.extend({
@@ -148,6 +149,7 @@ define([
          */
         placeOrderClick: function () {
             if (this.validateCardType()) {
+                fullScreenLoader.startLoader();
                 $(this.getSelector('submit')).trigger('click');
             }
         },
--- a/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/cc-form.js
+++ b/app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/cc-form.js
@@ -79,6 +79,7 @@ define(
                      */
                     onError: function (response) {
                         braintree.showError($t('Payment ' + this.getTitle() + ' can\'t be initialized'));
+                        fullScreenLoader.stopLoader();
                         throw response.message;
                     },
 
--- a/app/code/Magento/Checkout/view/frontend/web/js/view/payment/default.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/view/payment/default.js
@@ -18,7 +18,8 @@ define([
     'Magento_Checkout/js/model/payment/additional-validators',
     'Magento_Ui/js/model/messages',
     'uiLayout',
-    'Magento_Checkout/js/action/redirect-on-success'
+    'Magento_Checkout/js/action/redirect-on-success',
+    'Magento_Checkout/js/model/full-screen-loader'
 ], function (
     ko,
     $,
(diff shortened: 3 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10817) <br/>All committers have signed the CLA.

app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/hosted-fields.js
```diff
@@ -148,6 +149,7 @@ define([
          */
         placeOrderClick: function () {
             if (this.validateCardType()) {
+                fullScreenLoader.startLoader();
```
YevSent, line 152: I think it's a better solution to disable the button after clicking instead showing loader. The possible fix - disabling place order action:
`this.isPlaceOrderActionAllowed(false);` instead `fullScreenLoader.startLoader();`.

Also, I would preferer to create new `isButtonActive` function in `hosted-fields.js` component:
```
isButtonActive: function () {
    return this.isActive() && this.isPlaceOrderActionAllowed();
}
```
And use this function instead `isActive` for the `Place Order` button:
```
<button class="action primary checkout"
    type="submit"
    data-bind="
        click: placeOrderClick,
        attr: {title: $t('Place Order')},
        enable: isButtonActive()
    "
disabled>
    <span data-bind="i18n: 'Place Order'"></span>
</button>
```
Instead:
```
css: {disabled: !isPlaceOrderActionAllowed()},
enable: isActive()
```

YevSent (changes requested):

app/code/Magento/Braintree/view/frontend/web/js/view/payment/method-renderer/hosted-fields.js
maximbaibakov: Hi @joni-jones 

I appreciate your view on that and it completely aligned with my view 1.5 year ago, however, deactivating button still give a client todo some of the actions on page which might affect payment process (we have had such situations).

A few examples on "actions":
- Click somewhere on a page and got redirected somewhere from checkout page (as result the client will be charged, but no order stored in admin)
- Add / cancel coupon
- Add / cancel gift card
- Any other actions on a page

Ideally, once Order button is triggered, the client should be restricted doing anything on the page.

Regards,
Maxim

YevSent: I agree with you but when a customer clicks on `Place Order` button the place order action will be triggered, so loader appears even without your fix. Prosed fix prevents "doubleclicking", your fix masks a problem.

YevSent:
Fixed by the core team. The fix will be available in 2.2.1. Thanks for the contribution.

convenient:
@joni-jones Can you refer to the code fix in 2.2.1 please?

maximbaibakov:
Hi @convenient 

I think you are looking for this https://github.com/magento/magento2/commit/508170225ad966a21a905b16974540ab8fcbd842

Regards,
Maxim

convenient:
Thanks @maximbaibakov. 

I've tried looking for an associated PR with testing notes, is there anything available? I'd be keen to backport this to 2.1 so our client can get it easier.

maximbaibakov:
Hi @convenient 

I do not think so, but you can use the commit for 2.1.x without any changes. 
We are using this for Magento 2.1.3 - 2.1.9 and it works fine.

Regards,
Maxim

PR #10826, opened 2017-09-09
response condition order changes.

RakeshJesadiya:
When Customer login from checkout page, need to redirect customer to account dashborad page from checkout page.

By default Customer Login from checkout page, stays into same page but where need to redirect customer to account dashboard page. we need to create plugin and set response as redirect url into magento customer module.

After create plugin we need to override login.js file cause of response.redirectUrl call condition in elseif.
So if response with redirectUrl is set then first redirect to those action instead of default action.

Check real issue from magento forum,[Forum Question](https://community.magento.com/t5/Programming-Questions/Redirect-to-Account-Dashboard-During-Checkout-Funnel-Login/m-p/74630#M2714)

### DIFF
--- a/app/code/Magento/Customer/view/frontend/web/js/action/login.js
+++ b/app/code/Magento/Customer/view/frontend/web/js/action/login.js
@@ -37,6 +37,7 @@ define([
                         callback(loginData);
                     });
                     customerData.invalidate(['customer']);
+                    
                     if (response.redirectUrl) {
                         window.location.href = response.redirectUrl;
                     } else if (redirectUrl) {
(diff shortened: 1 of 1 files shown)


ishakhsuvarov:
Hi @RakeshJesadiya 
Please take a look at the JavaScript Static tests failure on Travis CI

Thanks

RakeshJesadiya:
Hi @ishakhsuvarov, What to do for Travis CI, I am not known for those. Any guidance would be appericiated.

orlangur:
@RakeshJesadiya error is quite clear: https://travis-ci.org/magento/magento2/jobs/273575190#L817

First the `if` was not aligned properly, now empty line is missing. Please squash your changes into single commit and force push after build is green.

PR #10837, opened 2017-09-11
allow digits in event names

mwr:
At the moment it is not allowed to have event names containing digits, though this was possible in Magento1. 
For example a event name like `mwltr83_my_event` is not allowed within the events.xsd.

### Description
This pull requests updates the events.xsd and adds 0-9 to allowed values in the regex.

void

### Manual testing scenarios
void

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Event/etc/events.xsd
+++ b/lib/internal/Magento/Framework/Event/etc/events.xsd
@@ -60,11 +60,11 @@
     <xs:simpleType name="eventName">
         <xs:annotation>
             <xs:documentation>
-                Event name can contain only [a-zA-Z_].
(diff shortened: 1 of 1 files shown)


miguelbalparda:
Hi @mwr! Checking [this](devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/) it seems this change does not have any backward compatibility issue but I will double check internally and report back. Thanks for your contribution!

fooman:
I am guessing this should bump the feature (minor) version of the framework package to allow for targeting of this feature of the xsd validation as "mwltr83_my_event" would validate on 2.2.1 (if it's included in the next release) but would fail for 2.2.0.

PR #10848, opened 2017-09-12
cviking26_issue-9489 // added query param to url build - description …

cviking26:
There is an error while click on "resend confirmation mail". The mail address will transfered without special chars like "+". This is important for the new gmail feature which provides simply adding aliases.

## Fixed Issues
* https://github.com/magento/magento2/issues/9489

## Preconditions
* Magento 2.1.2
* Enable Double-OptIn Mails for Customer Registration

## Steps to reproduce
* Register as a new Customer with email like "firstname.lastname+PARAM@gmail.com" (customer/account/create/)
* After Redirect to customer/account/login/ click on "Resend Confirmation Link"

## Actual Result
* This redirect to `customer/account/confirmation/email/firstname.lastname+PARAM@gmail.com/` - The Parameter email in url is not urlencoded, so the param is missing in the input field.
Parameter on this page in input field is now "firstname.lastname PARAM@gmail.com" without the plus

## Expected result
* Click on "Resend Confirmation Link"
* Get URL encoded mail in input field on `customer/account/confirmation/email/firstname.lastname%2BPARAM@gmail.com/`
so Parameter in input field is with "+" char e.g. "firstname.lastname+PARAM@gmail.com"

## Solution
Bug can solved simply adding "_query" in htdocs/vendor/magento/module-customer/Model/Url.php:221

`return $this->urlBuilder->getUrl('customer/account/confirmation', array('_query' => ['email' => $email]));`


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10848) <br/>All committers have signed the CLA.

app/code/Magento/Customer/Model/Url.php
```diff
@@ -237,7 +237,7 @@ public function getForgotPasswordUrl()
      */
     public function getEmailConfirmationUrl($email = null)
     {
-        return $this->urlBuilder->getUrl('customer/account/confirmation', ['email' => $email]);
+        return $this->urlBuilder->getUrl('customer/account/confirmation', array('_query' => ['email' => $email]));
```
ishakhsuvarov: Please use short array syntax (`[]` instead of `array()`)

cviking26: @ishakhsuvarov thank's. Adapted and pushed.

ishakhsuvarov:
@cviking26 Thank you for the update. We are going to proceed with review and merge

PR #10888, opened 2017-09-14
Pass the extra params from the no-route config to the request

thomas-kl1:
This modifications allows the no-route handler to retrieve the extra parameters passed through the no-route path config value, and to set them into the no-route request object.

### Description
Today we can set a Default No-route URL in the backend config. It allows us to redirect the 404 errors to the controller action of our choice.
However, we can't pass some parameters, even if we add them in the path. But the ones given in the default web url field are used by Magento in the request object. What I suggest is to allows this same functionality for the no-route request, by passing tan url like the following:

_eg_:
myroute/mycontroller/myaction/myvarname1/myvarvalue1/myvarname2/myvarvalue2/...

In the class \Magento\Framework\App\Router\NoRouteHandler, the process method only retrieve the route name, the controller name and the action name, without regarding the extra value passed.
I suggest to make it possible, so it could be really useful for developers who want to user their own CMS page system (other than the CMS page of Magento).

### Manual testing scenarios
You can check it with the CMS module of Magento:
1. Go to stores > Configuration > General > Web
2. Open Default pages section and edit the Default No-route Url:
cms/page/view/page_id/<your_page_id>
**But it should work also with your own controller action!** 

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/App/Router/NoRouteHandler.php
+++ b/lib/internal/Magento/Framework/App/Router/NoRouteHandler.php
@@ -34,7 +34,7 @@ class NoRouteHandler implements \Magento\Framework\App\Router\NoRouteHandlerInte
         $noRoutePath = $this->_config->getValue('web/default/no_route', 'default');
 
         if ($noRoutePath) {
-            $noRoute = explode('/', $noRoutePath);
+            $noRoute = explode('/', $noRoutePath, 4);
         } else {
             $noRoute = [];
         }
@@ -42,8 +42,17 @@ class NoRouteHandler implements \Magento\Framework\App\Router\NoRouteHandlerInte
         $moduleName = isset($noRoute[0]) ? $noRoute[0] : 'core';
         $actionPath = isset($noRoute[1]) ? $noRoute[1] : 'index';
         $actionName = isset($noRoute[2]) ? $noRoute[2] : 'index';
+        $params = isset($noRoute[3]) ? explode('/', $noRoute[3]) : [];
 
-        $request->setModuleName($moduleName)->setControllerName($actionPath)->setActionName($actionName);
+        $actionParams = [];
+        for ($i = 0, $l = sizeof($params); $i < $l; $i += 2) {
+            $actionParams[$params[$i]] = isset($params[$i + 1]) ? urldecode($params[$i + 1]) : '';
+        }
+
+        $request->setModuleName($moduleName);
+        $request->setControllerName($actionPath);
+        $request->setActionName($actionName);
+        $request->setParams($actionParams);
 
         return true;
     }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10888) <br/>All committers have signed the CLA.

orlangur:
One can easily obtain extra parameters if needed like here: https://github.com/magento/magento2/blob/develop/app/code/Magento/Backend/App/Router/NoRouteHandler.php#L46

> Open Default pages section and edit the `Default No-route Url`:
`cms/page/view/page_id/<your_page_id>`

This is misuse of Magento functionality, no ID or other parameters should be specified in such URL, just specify needed page in `CMS No Route Page` on per-store-view basis.

Could you please provide a valid real-world use case where such change could be useful?

thomas-kl1:
Actually you can do it for the Default Web URL, why not for the Default No-route URL?

There's many extensions which improve the default CMS of Magento by adding their own cms system (just search blog or cms extension for magento 2 in google). As the native Magento CMS, you may want to access to a specific content/page view by specifying an ID or some parameters.
But this path, (and only for the no-route url setting), is splitted in three parts, the route, the action path and the action name, and the remaining parts are never used (throwed to the trash..). So we can't...

My example was wrong because I've taken the default CMS of Magento, but many modules has the same behavior and functionalities to access to the content. So today, as you explained to me, I can only use the Magento default pages, and everyone know that the default CMS of Magento is a bit weak..

I know that you can add your own No-route handler in order to retrieve the parameters passed through the config field value. But why can't we make it easier for everyone? Like I see this, this is a lack of functionality, not even an improvement.

Sorry for my bad english, may I hope you understand my mind!

thomas-kl1:
By the way, I'm speaking about the frontend no-route handler.

orlangur:
Thanks for the explanation, I get your point that `cms/page/view/page_id/<your_page_id>` could be some custom extension with similar functionality.

The problem is that passing some additional parameters using `Default No-route Url` field is just a hack. It should just point to some route and nothing more than that, such route is responsible for everything else, like parsing additional parameters from request.

Thanks for your collaboration and quick feedback! Proposed implementation looks good to me but it doesn't seem to find its use in core.

thomas-kl1:
I don't understand why it's causing some trouble to add some parameters to the request, isn't it a forward execution?

orlangur:
> isn't it a forward execution?

Please reword your question.

thomas-kl1:
So why is it allowed and permit for the Default Web Url (the code clearly retrieve the parameters in order to allow their uses) but not for the Default No-route Url? It's confusing.
You tell me that:

> The problem is that passing some additional parameters using Default No-route Url field is just a hack. 

So in the first case Magento core (Default Web Url) allow the use of this "hack", why?
Of course my code doesn't seems to find its use in the core, but doesn't Magento should make the developments and customization easier, both for merchants and developers?

I'm really interested to know the reason why it's a "hack" and it shouldn't be allowed.
I understand that it's closed, but this modification do not make Magento core less efficient, but offer a larger customization, which is good no?

orlangur:
`Default No-route Url` is not intended to pass anything but `module/controller/action` and existing `NoRoute` route handler perfectly corresponds to this.

Proposed change both:
1) Won't be used in core itself
2) Won't simplify or enlarge customizability

One may already implement any NoRoute handler custom logic by just implementing one interface, like it was done for `\Magento\Framework\App\Router\NoRouteHandler` and `\Magento\Backend\App\Router\NoRouteHandler`.

Any addition to core, especially not used code, make Magento core less efficient at least in terms of maintenance efforts.

PR #10891, opened 2017-09-14
Add Role column to admin user grid

mpchadwick:
### Description

image

Adds a column to the admin user grid to show the users role

This can be useful for example to see all the users with "admin" privileges on a store with many users. Without this, you need to click into each manually

See: https://github.com/magento/magento2/issues/9557

### Contribution checklist


app/code/Magento/User/Model/ResourceModel/User/Collection.php
```diff
@@ -22,4 +22,18 @@ protected function _construct()
...
+    {
+        parent::_initSelect();
+        $this->getSelect()->join(
+            ['user_role' => $this->getTable('authorization_role')],
+            'main_table.user_id = user_role.user_id',
+            []
+        )->join(
+            ['detail_role' => $this->getTable('authorization_role')],
+            'user_role.parent_id = detail_role.role_id',
+            ['role_name']
```
orlangur, line 36: This will affect every user collection instantiated, isn't it?

Something like https://magento.stackexchange.com/a/124204, a separate collection just for grid, looks like a better option. Not sure if there is to way do it in more declarative way.

Just my 2 cents, before proceeding please wait for somebody to actually confirm such feature fits well into core.

mpchadwick: Yes the JOINs will happen on every collection load. This was based on the sales rule grid

https://github.com/magento/magento2/blob/75155ae0d42ce33bc91d7f42c19ee35f1f083d58/app/code/Magento/SalesRule/view/adminhtml/layout/sales_rule_promo_quote_index.xml#L15

https://github.com/magento/magento2/blob/develop/app/code/Magento/SalesRule/Model/ResourceModel/Rule/Quote/Collection.php#L10

https://github.com/magento/magento2/blob/develop/app/code/Magento/SalesRule/Model/ResourceModel/Rule/Collection.php#L279-L288

Not sure what the concern would be. The performance cost should be negligible.

vrann: I searched for the usages of this Collection, it is not used in any performance-critical flows. Join should be fine here.

vrann:
please merge with the latest from `develop`

app/code/Magento/User/Model/ResourceModel/User/Collection.php
orlangur: @vrann glad to know, thanks! Hope you checked EE/B2B codebases as well.

Is it correct that such approach with `_initSelect` is up-to-date? Like, there is no declarative way to declare join for the grid etc.

vrann:
@mpchadwick ^^

mpchadwick:
@vrann Should be merge develop back in tonight hopefully.

app/code/Magento/User/Model/ResourceModel/User/Collection.php
mpchadwick: @vrann @orlangur also note that there are indexes on both the columns used in the JOIN so I really wouldn't be worried about perf.

mpchadwick:
@vrann I've merged `develop` back to this branch.

mpchadwick:
@vrann Just pushed an update to fix visibility on _initSelect. Viewing Locked Users page in admin was resulting in 500 error. Noticed that while looking at the Travis results.

vrann:
@mpchadwick thank you, I also fixed that internally.

Another issue needed to resolve was with the b2b version crashing when filtered by user_id column: now it is ambiguous which table user_id should come from.

vrann:
@mpchadwick thank you, the feature is successfully accepted. It is not indicated as merge because we did the same commit.

PR #10904, opened 2017-09-16
Add escaping of js translation version

ihor-sviziev:
As it was discussed in https://github.com/magento/magento2/pull/10378#discussion_r136984136 - we need to escape js translation version.

### Description
Previously value of $version variable was always sha1, that can be not escaped, but from now it uses raw value from public method, that could be overridden by plugin.

N/A

### Manual testing scenarios
1. Put following code to getTranslationFileVersion method (directly or via plugin):
```php
return "' || alert(1) || '";
```
2. Go to any page

**Actual result:**
1. Alert with message 1 is shown

**Expected result:**
1. Alert mustn't be shown

### Contribution checklist

### DIFF
--- a/app/code/Magento/Translation/view/base/templates/translate.phtml
+++ b/app/code/Magento/Translation/view/base/templates/translate.phtml
@@ -28,7 +28,7 @@
 
                 <?php $version = $block->getTranslationFileVersion(); ?>
 
-                if (versionObj.version !== '<?= /* @noEscape */ $version ?>') {
+                if (versionObj.version !== '<?php $block->escapeJsQuote($version) ?>') {
                     dependencies.push(
                         'text!<?= /* @noEscape */ Magento\Translation\Model\Js\Config::DICTIONARY_FILE_NAME ?>'
                     );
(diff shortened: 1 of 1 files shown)


orlangur:
Sorry, does not sound reasonable to me :) One can put some `alert` into plugin, but why?

Such approach means we shall escape output of EVERY public non-final method as it can be pluginized.

I'll tell you more, `escapeJsQuote` can be pluginized as well. Magento is pretty flexible thus if somebody _really_ wants to shoot in the foot we cannot prevent it.

Cc: @ishakhsuvarov @hostep since you liked the idea, please share your thoughts. I may be a bit out of context.

hostep:
@orlangur: I think if a value is outputted in a javascript string, it should be escaped, so if the value contains single quotes, it won't break the javascript.

This is not really to prevent injecting other javascript code to be executed, but in this case, so that the `$version` can contain single quotes in itself. Suppose someone (for some reason) wants to use a different hashing method that can contain single quotes in the hash, then those single quotes should be escaped in the output, otherwise the javascript code will break.

orlangur:
@hostep ok, thanks, that's a bit closer to reality, just I'm not sure if there are actually some hash functions with single quotes in output. Generally I'm a bigger fan of "escape everything by default" approach, like Twig does, not sure why Magento picked up an opposite approach.

orlangur (approved):

Issue #10673, opened 2017-08-26
[2.2.0-RC2.2] Static content deploy with multiple locales:  js-translation.json files are the same


hostep:
### Preconditions
1. PHP 7.0.22

### Steps to reproduce
1. Install Magento 2.2.0-rc22 using composer
2. Add 2 language packs using composer: `composer require honl/magento2-nl-nl imaginaerum/magento2-language-fr-fr`
3. Create 2 different storeviews in the backend and set the Locale configuration to `Dutch (Netherlands)` and `French (France)` for those 2 storeviews
4. Run `php bin/magento setup:static-content:deploy -f -t Magento/luma nl_NL && php bin/magento setup:static-content:deploy -f -t Magento/luma fr_FR`
5. Run `diff -rwq pub/static/frontend/Magento/luma/nl_NL/ pub/static/frontend/Magento/luma/fr_FR/`
6. Results: `Files pub/static/frontend/Magento/luma/nl_NL/js-translation.json and pub/static/frontend/Magento/luma/fr_FR/js-translation.json differ` => **Good**
7. Remove all generated static files: `rm -R pub/static/* var/view_preprocessed/*`
8. Run `php bin/magento setup:static-content:deploy -f -t Magento/luma nl_NL fr_FR`
9. Run `diff -rwq pub/static/frontend/Magento/luma/nl_NL/ pub/static/frontend/Magento/luma/fr_FR/`
10. Results: ... nothing, all files are equal => **Not good**

### Expected result
1. When generating static content using multiple locales in one go, the generated `js-translation.json` files should be different

### Actual result
1. When generating static content using multiple locales in one go, the generated `js-translation.json` files are the same
This has as result that all javascript translations are in the wrong locale on the frontend.

### Discussion
This is the same issue like I reported on https://github.com/magento/magento2/issues/7862 for Magento 2.1.3, but this was supposed to be fixed by MAGETWO-62660, which is included in Magento 2.2.0-RC2.2, but looks like that didn't really help, or something else might broke it yet again ...

misha-kotov:
Thanks for reporting - internal issue is MAGETWO-72301

dmitry-fedyuk:
[Magento 2.2 and 2.3 bug: `bin/magento setup:static-content:deploy -f <multiple locales>` does not generate a correct `js-translation.json` for the second and next locales (and my fix)](https://mage2.pro/t/topic/4524)

magento-engcom-team:
@hostep, thank you for your report.
[...]

PR #10913, opened 2017-09-17
Fixes translations by re-running the loadData for its correct locale

wiardvanrij:
### Description
When running `setup:static-content:deploy` with multiple languages specified, it will generate a js-translation.json. Even if the locale is set properly for each language, it will translate just based on 
`$this->translator->getData();`  in _Magento\Framework\Phrase\Renderer_

In other words it will use the data retrieved from the `loadData`  function in the _Framework/Translation.php_ 
This uses a constructor which sets the locale; `\Magento\Framework\Locale\ResolverInterface $locale` and obviously followed by `$this->_locale = $locale;`

The problem here is that the loadData only gets ran once on init. It will keep using this for every other locale as 'base'. Therefore a potential locale change should also re-run the loadData.

By forcing a "rebuild" of the loadData with the correct locale, we will parse the data with the correct translations. 

As extra because we are forcing the locale & rebuild we can remove the FallbackContext filler for the locale as we are setting it anyways.

### Fixed Issues 
1. magento/magento2#10673: [2.2.0-RC2.2] Static content deploy with multiple locales: js-translation.json files are the same

### Manual testing scenarios
1 Run `php bin/magento setup:static-content:deploy -f -t Magento/luma nl_NL fr_FR`
2 Run `diff -rwq pub/static/frontend/Magento/luma/nl_NL/ pub/static/frontend/Magento/luma/fr_FR/`
3 js-translation.json are now different from each other with the correct language translations

### DIFF
--- a/app/code/Magento/Translation/Model/Json/PreProcessor.php
+++ b/app/code/Magento/Translation/Model/Json/PreProcessor.php
@@ -81,7 +81,7 @@ class PreProcessor implements PreProcessorInterface
 
             $area = $this->areaList->getArea($areaCode);
             $area->load(\Magento\Framework\App\Area::PART_TRANSLATE);
-    
+
             $this->translate->setLocale($context->getLocale())->loadData($areaCode, true);
 
             $chain->setContent(json_encode($this->dataProvider->getData($themePath)));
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10913) <br/>All committers have signed the CLA.

wiardvanrij:
I do not know why the unit test gives me `Call to a member function loadData() on null`. It makes no sense.. Anyone got an idea?

vrann:
@douweegbertje: This class `\Magento\Translation\Test\Unit\Model\Json\PreProcessorTest` creates a mock of TranslateInterface so it can not possibly have the right logic for setLocale which should return proper object:

```
 $this->translateMock = $this->getMockForAbstractClass(\Magento\Framework\TranslateInterface::class);
```

Please create the mock for `setLocale` method.

PR #10919, opened 2017-09-17
#10803 update OrderService to return correct bool value for cancel me…

freakphp:
…thod

Update OrderService to return correct bool value for cancel method

### Description
Magento/Sales/Model/Service/OrderService::cancel method was returning always true even if cancel method in Magento/Sales/Model/Order didn't do anything cause of canCancel validation.
OrderService cancel method will invoke canCancel to check if order can be canceled and return correct value.

1. magento/magento2#10803: When canceling order with OrderService, the cancel method always saves the order and returns true, even if the order can not be canceled

### Manual testing scenarios

1. Requesting API rest endpoint /V1/orders/:id/cancel for the order which can not be canceled should return false

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Service/OrderService.php
+++ b/app/code/Magento/Sales/Model/Service/OrderService.php
@@ -87,7 +87,8 @@ class OrderService implements OrderManagementInterface
     public function cancel($id)
     {
         $order = $this->orderRepository->get($id);
-        if ((bool)$order->cancel()) {
+        if ((bool)$order->canCancel()) {
+            $order->cancel();
             $this->orderRepository->save($order);
             return true;
         }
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10919) <br/>All committers have signed the CLA.

app/code/Magento/Sales/Model/Service/OrderService.php
```diff
@@ -87,7 +87,8 @@ public function __construct(
     public function cancel($id)
     {
         $order = $this->orderRepository->get($id);
-        if ((bool)$order->cancel()) {
+        if ((bool)$order->canCancel()) {
```
orlangur: Please remove `(bool)` casting, it is useless within `if` anyway.

orlangur (changes requested):

app/code/Magento/Sales/Model/Service/OrderService.php
ishakhsuvarov: fixed

orlangur (approved):

PR #10921, opened 2017-09-17
#10611 Ignore request key check for error page to avoid redirect loop

lbajsarowicz:
### Description
Problems described in the issue was caused by redirect occuring when request made to the application had incorrect URL key. This specific situation took place only when user had no access to any part of application. So finally trying to get access to `adminhtml/index` controller, he got redirected to `adminhtml/noroute/denied` that he also have no access because of incorrect URL key.

1. magento/magento2#10611 Magento admin gets in redirect loop when I login with a role that has no resources assigned

### Manual testing scenarios

1. Create a user role with no resources assigned to it.
2. Create an admin user with that role.
3. Log in with that user.

### Contribution checklist
 - [ - ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ x ] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Backend/Controller/Adminhtml/Noroute/Index.php
+++ b/app/code/Magento/Backend/Controller/Adminhtml/Noroute/Index.php
@@ -39,4 +39,14 @@ class Index extends \Magento\Backend\App\Action
         $resultPage->addHandle('adminhtml_noroute');
         return $resultPage;
     }
+
+    /**
+     * Error page should be public accessible. Do not check keys to avoid redirect loop
+     *
+     * @return bool
+     */
+    protected function _validateSecretKey()
+    {
+        return true;
+    }
 }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10921) <br/>All committers have signed the CLA.

app/code/Magento/Backend/Controller/Adminhtml/Noroute/Index.php
```diff
@@ -39,4 +39,14 @@ public function execute()
         $resultPage->addHandle('adminhtml_noroute');
         return $resultPage;
     }
+
+    /**
+     * Error page should be public accessible. Do not check keys to avoid redirect loop
+     *
+     * @return bool
+     */
+    protected function _validateSecretKey()
```
vkublytskyi, line 48: This can be done in more consistent with other controllers way.

`Magento\Backend\App\AbstractAction` already provides a mechanism for disabling secret key validation: `\Magento\Backend\App\AbstractAction::$_publicActions`. Setting this property will be more correct and consistent fix:

```php
$_publicActions = ['index']
```

However, both solution is fixing the only symptom of an issue, not a root cause. Secret check failed because of invalid value for action (key generated for `denied` and checked for `index`). That should be investigated and fixed as well. Another one issue is why request redirected to `*/*/denied` path which actually does not match any action controller class and fall back to `noroute` controller which does not display own content but has special handling in `Magento\Backend\App\AbstractAction` to render `adminhtml_denied`. In addition `Magento\Backend\Controller\Adminhtml\Denied` action controller is not in proper place and seems should be moved to `Magento\Backend\Controller\Adminhtml\Index` namespace.

As full fix of the issue require deep refactoring of `Magento_Backend` module this fix will be accepted. We will be thankful if you continue your work and fix legacy implementation to achieve more straightforward and predictable routing in admin.

vkublytskyi (approved):

PR #10923, opened 2017-09-17
#10922 add is_active to Magento\Store\Api\Data\StoreInterface

Zefiryn:
add is_active to Magento\Store\Api\Data\StoreInterface to show it in the rest endpoint

### Description

Magento\Store\Api\Data\StoreInterface had no get and set method defined for is_active flag and thus the value did not appear in the V1/store/storeViews endpoint with the store view data

1. magento/magento2#10922: REST endpoint /V1/store/storeViews is missing is_active value in store data

### Manual testing scenarios
1. Perform a rest call to http://store.domain/rest/[store-view-code]/V1/store/storeViews

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/Api/Data/StoreInterface.php
+++ b/app/code/Magento/Store/Api/Data/StoreInterface.php
@@ -67,6 +67,17 @@ interface StoreInterface extends \Magento\Framework\Api\ExtensibleDataInterface
     public function getStoreGroupId();
 
     /**
+     * @param int $isActive
+     * @return $this
+     */
+    public function setIsActive($isActive);
+
+    /**
+     * @return int
+     */
+    public function getIsActive();
+
+    /**
      * @param int $storeGroupId
      * @return $this
      */
--- a/app/code/Magento/Store/Model/Store.php
+++ b/app/code/Magento/Store/Model/Store.php
@@ -27,7 +27,6 @@ use Magento\Store\Api\Data\StoreInterface;
  * @method int getSortOrder()
  * @method int getStoreId()
  * @method Store setSortOrder($value)
- * @method Store setIsActive($value)
  *
  * @SuppressWarnings(PHPMD.TooManyFields)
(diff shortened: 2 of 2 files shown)


app/code/Magento/Store/Api/Data/StoreInterface.php
```diff
@@ -67,6 +67,17 @@ public function setWebsiteId($websiteId);
     public function getStoreGroupId();
 
     /**
+     * @param int $isActive
+     * @return $this
+     */
+    public function setIsActive($isActive);
```
ishakhsuvarov, line 73: Unfortunately, introduction of any new methods to the `@api` interfaces is not possible due to [backwards compatibility policy](http://devdocs.magento.com/guides/v2.1/ext-best-practices/extension-coding/backwards-compatible-development/index.html). This change can potentially break any third-party customization which used this class as SPI (providing their own implementation for existing interface).

Possible workaround would be implementing an extension attribute for this class which would contain the required field. This would not break any existing customization but as a tradeoff would look not as good.

ishakhsuvarov (dismissed):
Please consider updating integration or api-functional for this fix

app/code/Magento/Store/Api/Data/StoreInterface.php
hervebackstage: this change was accepted by Anton even though it gave backward compliance issue. I am happy to give a go with the extension attribute but I'd say any developer would see this change as more beneficial as it is now.
--> what I mean is Magento 2 is complex enough to avoid making minor issue to become a big problem?

vrann:
@Zefiryn now this change can be accepted because it is to the develop branch which will be released as 2.3. This is a MINOR change in API

hatimeria-artur-jewula:
OK, I will prepare update for tests as requested and add this to this PR.

vrann:
@hatimeria-artur-jewula just updated api-functional test for this (were failing on or end), so it is good to go

Issue #9077, opened 2017-03-31
State/Province field doesn't show as required causes checkout error


kjasrotia:
### Preconditions

1. Magento 2.1.4
2. 

### Steps to reproduce

Similar issue as #5279 which was closed by customer but has not actually been fixed in Magento 2.1.4.

1. I have specified "State" field as mandatory for customer address for my Country (India) from Admin.
Stores > Setting Configuration > General > General > Country Options > State Options > State is Required for.

2. From store create a customer account but do not add address.

3. From store Login to account and checkout a product. During checkout, being able to create address without the State as it is not shown as required in the address form (however, when customer has registered account and goes to add a new address from account settings, then 'State' is shown as required - go figure !!)

4. During checkout it gives an "Error occurred on the server. Please try to place the order again" error.   
    When the State is provided in address the checkout goes thru.

This basic issue needs to be fixed ASAP .. losing sales because of this ! Customer obviously does not know issue and abandons the transaction. Specifying "State" as mandatory should make it a required field in ALL the various address forms for customers.

### Expected result

1. The 'State' should be a required field in ALL address forms presented to the customer.

### Actual result

1. Since 'State' is not shown as required, customer is able to create address without the state. However, this causes the "Error occurred on the server. Please try to place the order again" during checkout.  When State is added to address the checkout goes thru.

korostii:
> This basic issue needs to be fixed ASAP

Haha, that's funny. You must be new here.
IF it gets changed right away, there's 2 months on the average until the change reaches a release.
It would speed up the process,, if you could provide a pull request with the fix, otherwise chances are you'll get it fixed earlier if you work around it locally in your codebase, rather than wait for a fix to get released.

Andreshlan:
I have this same problem...Magento 2 is not functional, should still be a beta

ifeltsweet:
Completely agree, Magento 2 is indeed still a beta. Basic functionality doesn't work.

Igloczek:
@Andreshlan @ifeltsweet 
Did you hear about reactions? You just should upvote issues and pull requests, if you think they are important.
Your comments have no value for maintainers or other devs.

Igloczek:
@ishakhsuvarov I manage to fix it - my PR #10924
The issue is already fixed on `develop` branch, so I backported the fix to `2.1-develop`.

ishakhsuvarov:
@Igloczek Thanks!

kjasrotia:
Thank you !

[...]

Issue #5279, opened 2016-06-27
State/Province field doesn't show as required on the add new address page.


dcabrejas:
Steps to reproduce
1. Install Magento 2  ee/ce through composer.
2. Set the default Locale as English (United Kingdom) in the backend at Stores > Setting Configuration > General > General > Locale Options  > Locale.
3. Set State as required for the United Kingdom in the backend at Stores > Setting Configuration > General > General > Country Options > State Options > State is Required for.
4. Create an account in the front-end.
5. Navigate to My Account > Address Book > Add new address.
6. Don't fill any fields, just make sure that the country selected in United Kingdom.
7. Click the save Address Button

Expected result
1. After selecting the United Kingdom as the country, I'd expect a red star to be added next to the State/Province label as this is a required field for the UK.
2. After clicking the Save address button, the validation for the State/Province field would fail with a message stating : This is a required field. This is happening.

Actual result
1. After selecting the UK as the country, the red star is not added next to the label of the state/province filed to indicate that the field is required for the UK. However when you try to save the address without having filled out the state/province field, the validation fails, letting you know that the field is required.

Observations
I have observed that this behaviour only happens when there is no values for the UK state/provinces in the database. This works for the US, I think because the US has predefined states stored in the database and when you select the US as the country, Magento changes the State/Province filed to be a select field and adds the red star next to its label.

Screenshots
1. Working for the US, the red star is placed next to the State/Province label indicating its a required field
![screen shot 2016-06-27 at 10 01 27](https://cloud.githubusercontent.com/assets/8201815/16374500/a36f94f4-3c4e-11e6-8c25-dc530a5b6b7d.png)

2.Upon changing the country to the UK, the red star disappears, indicating the State/Province field is optional.
![screen shot 2016-06-27 at 10 01 39](https://cloud.githubusercontent.com/assets/8201815/16374427/5454bab6-3c4e-11e6-8c6f-4d1cf8d55a84.png)
1. However if we try to save the address without having filled out the State/Province field, the validation for this field doesn't pass as this is configured to be a required field for the UK.
   ![screen shot 2016-06-27 at 10 01 45](https://cloud.githubusercontent.com/assets/8201815/16374447/6f382a7a-3c4e-11e6-89ae-a486db4d5997.png)

antboiko:
Hi @dcabrejasazagra , thanks for reporting this issue. I've created internal ticket MAGETWO-54785.

Thanks,
Anton.

brideo:
[...]

Issue #2946, opened 2016-01-09
Customer country list not correct


gigadesign1:
I was just migrating my data to Magento 2.0 and found this issue.

When there are different configurations for available countries in store views, in the admin section I only get the countries that are available in the _default store view_ of the website, but not the store view the customer is assigned to.

In my case: we have a website for Europe. This website has store views for different countries, for example France. France is not available in other store views of this website.
The default store view for the website is the UK store view.
When I open a customer in the admin section, I cannot select France as country for their address, even though the account is created in the French store view where France is available as a country.

sevos1984:
Internal ticket created MAGETWO-48425

okorshenko:
@gigadesign1 Thank you for reporting the issue. Fix for MAGETWO-48425 merged to develop branch.

slavvka:
The fix has been also ported to 2.1

Ctucker9233:
@slavvka Great! Where can we find it?

slavvka:
In the latest mainline and it will be delivered in the nearest 2.1 patch release. Commits you can see above.

AntonEvers:
@slavvka thank you. Is there already a planned release date of the next patch release?

Ctucker9233:
Yes, is there an estimate of when the next patch will be available?

PR #10924, opened 2017-09-17
[Backport] Make State/Province required field in checkout view based on settings

Igloczek:
It's the backport of changes present on `develop` branch, to fix issues described in #9077 also on 2.1.x.

Related commits and issues:
- https://github.com/magento/magento2/commit/1b4ba9f9f071bfc539602fe2da399ad0a235b232
- https://github.com/magento/magento2/commit/4f2fa65f990edc196ca4314d6e0a5dbd832fbca6
- https://github.com/magento/magento2/issues/5279
- https://github.com/magento/magento2/issues/2946

### DIFF
--- a/app/code/Magento/Ui/view/base/web/js/form/element/region.js
+++ b/app/code/Magento/Ui/view/base/web/js/form/element/region.js
@@ -24,6 +24,7 @@ define([
         update: function (value) {
             var country = registry.get(this.parentName + '.' + 'country_id'),
                 options = country.indexedOptions,
+                isRegionRequired,
                 option;
 
             if (!value) {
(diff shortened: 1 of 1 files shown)


orlangur:
Will this cover https://github.com/magento/magento2/issues/10897 case?

Igloczek:
@orlangur Yup, it will cover this case too.

vrann:
@Igloczek thank you for the backport!
We notice one issue which still has to be resolved: when the State/Province is a required field, it is validated on PHP side, but the frontend does not display it as a required for the countries other than USA (i.e. India).

This is how it looks like on 2.1 after the fix:
susp-03

This is how it looks like on develop
develop

Igloczek:
To be completely honest with you, I'm afraid that it has to be merged in the current state or rejected, because I don't have time to investigate, especially while we have 2.2 up and backport isn't a crucial thing, because users can just make an upgrade.

ishakhsuvarov:
@Igloczek I am closing this now per your comment.
Please reopen if you feel like continuing with this one in the future.

PR #10927, opened 2017-09-17
Fix issue when non-cacheable block added using invalid reference makes page non-cacheable

mcjwsk:
### Description
After applying this pull request all layout blocks added using invalid container or block reference (not existing in the current page context) are ignored while resolving the `cacheable` flag for page cache.
Unfortunately this solution doesn't handle cases when the same scenario is done using a block without name. It's not possible, because we can't identify this kind of blocks in the base layout xml and next in the layout structure (they aren't included in `Magento\Framework\View\Layout\Data\Structure::exportElements()`).

### Fixed Issues
1. magento/magento2#9041: Non-cacheable block added to default handle makes every page non-cacheable

### Manual testing scenarios
1. Create an extension. Add the layout update to the default handle via `Vendor/Module/view/frontend/layout/default.xml`
2. Reference some container which appears on the checkout page and add the layout update with the non-cacheable block:
```
<referenceContainer name="order.success.additional.info">
    <block class="Magento\Framework\View\Element\Template" name="test_block" cacheable="false"/>
</referenceContainer>
```
3. Flush cache.
4. Run the home page.
5. Refresh the page and check was it cached in full page cache.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/View/Layout.php
+++ b/lib/internal/Magento/Framework/View/Layout.php
@@ -1094,8 +1094,20 @@ class Layout extends \Magento\Framework\Simplexml\Config implements \Magento\Fra
     public function isCacheable()
     {
         $this->build();
-        $cacheableXml = !(bool)count($this->getXml()->xpath('//' . Element::TYPE_BLOCK . '[@cacheable="false"]'));
-        return $this->cacheable && $cacheableXml;
+
+        $elements = $this->getXml()->xpath('//' . Element::TYPE_BLOCK . '[@cacheable="false"]');
+        foreach ($elements as $element) {
+            $attributes = $element->attributes();
+            if (empty($attributes['name'])) {
+                return false;
+            }
+            $elementId = (string) $attributes['name'];
+            if ($this->structure->hasElement($elementId)) {
+                return false;
+            }
+        }
+
+        return $this->cacheable;
     }
 
     /**
--- a/lib/internal/Magento/Framework/View/Test/Unit/LayoutTest.php
+++ b/lib/internal/Magento/Framework/View/Test/Unit/LayoutTest.php
@@ -706,11 +706,15 @@ class LayoutTest extends \PHPUnit\Framework\TestCase
 
     /**
      * @param string $xmlString
+     * @param bool $hasNonCacheableElement
      * @param bool $result
      * @dataProvider isCacheableDataProvider
      */
-    public function testIsCacheable($xmlString, $result)
+    public function testIsCacheable($xmlString, $hasNonCacheableElement, $result)
     {
+        $this->structureMock->method('hasElement')
+            ->with($this->equalTo('non_cacheable_block'))
+            ->willReturn($hasNonCacheableElement);
         $xml = simplexml_load_string($xmlString, \Magento\Framework\View\Layout\Element::class);
         $this->assertSame($this->model, $this->model->setXml($xml));
         $this->assertSame($result, $this->model->isCacheable());
@@ -725,11 +729,45 @@ class LayoutTest extends \PHPUnit\Framework\TestCase
             [
                 '<?xml version="1.0"?><layout xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
                 . '<block></block></layout>',
+                false,
                 true,
             ],
             [
                 '<?xml version="1.0"?><layout xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
                 . '<block cacheable="false"></block></layout>',
+                false,
+                false
+            ],
+            [
+                '<?xml version="1.0"?><layout xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
+                . '<referenceBlock name="not_existing_block">'
+                . '<block name="non_cacheable_block" cacheable="false"></block>'
+                . '</referenceBlock></layout>',
+                false,
+                true
+            ],
+            [
+                '<?xml version="1.0"?><layout xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
+                . '<referenceContainer name="not_existing_container">'
+                . '<block name="non_cacheable_block" cacheable="false"></block>'
+                . '</referenceContainer></layout>',
+                false,
+                true
+            ],
+            [
+                '<?xml version="1.0"?><layout xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
+                . '<referenceBlock name="existing_block">'
+                . '<block name="non_cacheable_block" cacheable="false"></block>'
+                . '</referenceBlock></layout>',
+                true,
+                false
+            ],
+            [
+                '<?xml version="1.0"?><layout xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">'
+                . '<referenceContainer name="existing_container">'
+                . '<block name="non_cacheable_block" cacheable="false"></block>'
+                . '</referenceContainer></layout>',
+                true,
                 false
             ],
         ];
--- a/lib/internal/Magento/Framework/View/Test/Unit/LayoutTest.php
+++ b/lib/internal/Magento/Framework/View/Test/Unit/LayoutTest.php
@@ -706,15 +706,15 @@ class LayoutTest extends \PHPUnit\Framework\TestCase
 
     /**
      * @param string $xmlString
-     * @param bool $hasNonCacheableElement
+     * @param bool $nonCacheableElement
      * @param bool $result
      * @dataProvider isCacheableDataProvider
      */
-    public function testIsCacheable($xmlString, $hasNonCacheableElement, $result)
+    public function testIsCacheable($xmlString, $nonCacheableElement, $result)
     {
         $this->structureMock->method('hasElement')
             ->with($this->equalTo('non_cacheable_block'))
-            ->willReturn($hasNonCacheableElement);
+            ->willReturn($nonCacheableElement);
         $xml = simplexml_load_string($xmlString, \Magento\Framework\View\Layout\Element::class);
         $this->assertSame($this->model, $this->model->setXml($xml));
         $this->assertSame($result, $this->model->isCacheable());


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10927) <br/>All committers have signed the CLA.

vrann:
@DariuszMaciejewski there are couple issues with the functional tests, can you verify that the caching works properly there?

```
Magento\CatalogRule\Test\TestCase\ApplyCatalogPriceRulesTest::

test variation ApplyCatalogRuleStaging_ForNewCustomerGroup_0
Wrong product price is displayed.
Expected: 10
Actual: 

Magento\Catalog\Test\TestCase\Product\AddToCartCrossSellTest::

test variation AddToCartCrossSellTestVariation1_0
Wrong products are displayed in cross-sell section.
Failed asserting that two arrays are equal.
--- Expected
+++ Actual

Magento\Checkout\Test\TestCase\OnePageCheckoutOfflinePaymentMethodsTest::

test variation OnePageCheckoutTestVariation4_0
PHPUnit_Extensions_Selenium2TestCase_WebDriverException: Timed out after 300 seconds

/opt/bamboo/builds/M2-AT2513-FTE/instance-0/dev/tests/functional/vendor/magento/mtf/Magento/Mtf/Client/Driver/Selenium/WaitUntil.php:78
/opt/bamboo/builds/M2-AT2513-FTE/instance-0/dev/tests/functional/vendor/magento/mtf/Magento/Mtf/Client/Driver/Selenium/RemoteDriver.php:46
/opt/bamboo/builds/M2-AT2513-FTE/instance-0/dev/tests/functional/vendor/magento/mtf/Magento/Mtf/Client/Driver/Selenium/Driver.php:584
(20 more lines...)Magento\Checkout\Test\TestCase\OnePageCheckoutOfflinePaymentMethodsTest::

test variation OnePageCheckoutTestVariation5_0
PHPUnit_Extensions_Selenium2TestCase_WebDriverException: Timed out after 300 seconds

Magento\CustomerSegment\Test\TestCase\CreateCustomerSegmentEntityWithCustomerConditionsPartTwoTest::

test variation CreateCustomerSegmentEntityTestGroupRetailer_0
PHPUnit_Extensions_Selenium2TestCase_WebDriverException: Timed out after 300 seconds

Magento\Customer\Test\TestCase\UpdateCustomerFrontendEntityTest::

test variation UpdateCustomerFrontendEntityTestVariation4_0
PHPUnit_Extensions_Selenium2TestCase_WebDriverException: Error occurred on attempt to get element. Message: "Timed out after 300 seconds". Locator: "css selector(.column.main) -> css selector(.block-dashboard-info .box-information .action.edit)" . Wait: "1"


Magento\Downloadable\Test\TestCase\CreateDownloadableProductEntityTest::

test variation CreateDownloadableProductEntityTestVariation1_0
Product price in shopping cart is not correct.
Failed asserting that null matches expected '3'.

Magento\Msrp\Test\TestCase\ApplyMapTest::

test variation ApplyMapTestVariation2_0
Incorrect unit price is displayed in Cart
Failed asserting that null matches expected '1'.
```

mcjwsk:
@vrann I checked these errors and I can confirm them.
The issue is that the layout structure is not filled from the beginning, so I'm unable to find some valid elements.
Do you have some idea which class to use instead of? I saw there is a `Magento\Framework\View\Model\Layout\Merge` class, but it contains no existing references data too.
On the other hand, I can try to correct the references inside the layout merging process, but it might be difficult to do.

vrann:
@DariuszMaciejewski can you please clarify what is root cause of the errors? What do you mean by `layout structure is not filled`

mcjwsk:
@vrann yes, sure. 

The method `Magento\Framework\View\Layout::isCacheable()` is called many times while loading a page. When I put `Magento\Framework\View\Layout\Data\Structure::exportElements()` inside the method, I see it returns different results for every call (from an empty array to a correctly filled array with all layout elements).
For example the checkout page doesn't work after my change, because when it checks the `checkout.root` block is cacheable the result of the call `$this->_structure->hasElement('checkout.root')` is `false` (= Magento thinks it is a broken reference).


I made some test and I added a stupid condition like that
```php
if (!$this->_structure->exportElements()) {
    $cacheable = false;
    return $cacheable;
}
```
It "solves" the error, so I guess the root cause is that I don't have an access to the layout structure object containing the same data through all calls.

On the other hand, when you look at the previous code
```php
$cacheableXml = !(bool)count($this->getXml()->xpath('//' . Element::TYPE_BLOCK . '[@cacheable="false"]'));
```
you see the method `$this->getXml()` always use the same layout xml, so it works without issues.

vrann:
@DariuszMaciejewski thank you for the detailed response, now I've got what you mean. You are right that the Layout Structure object is empty, because isCacheable() called from the plugin `\Magento\PageCache\Model\Layout\LayoutPlugin::afterGenerateXml`
After generate XML is good enough to get XML structure, but it is too early to understand the dependencies between blocks.

I think I have a better suggestion, exploring ideas here. What if instead, we add that plugin to be invoked after Elements are generated:  `\Magento\PageCache\Model\Layout\LayoutPlugin::afterGenerateElements`. 

Then, we can use the fact that such blocks which we are interested in (children of default layout, but not children of any other blocks on the page) are added to the "brokenParent" section of the Structure. Check it here:
`\Magento\Framework\View\Layout\ScheduledStructure\Helper::scheduleElement`

Then, ***isCacheable*** can look like this (pseudocode, but it works if you try to create wrong block and run it):

```
 public function isCacheable()
    {
        $this->build();
        $structure = $this->getReaderContext()->getScheduledStructure();
        $cacheableFalse = $this->getXml()->xpath('//' . Element::TYPE_BLOCK . '[@cacheable="false"]');
        if ((bool)count($cacheableFalse) && in_array($cacheableFalse[0]->getBlockName(), array_keys($structure->brokenParent))) {
            return true;
        } else {
             $cacheableXml = false;
        }
        return $this->cacheable && $cacheableXml;
    }
```

mcjwsk:
@vrann are you sure that `$structure = $this->getReaderContext()->getScheduledStructure();` will be always valid if I change the layout plugin from `\Magento\PageCache\Model\Layout\LayoutPlugin::afterGenerateXml` to  `\Magento\PageCache\Model\Layout\LayoutPlugin::afterGenerateElements`?
I did it but the property `$structure->brokenParent` was empty on the first few calls of the method `isCacheable()` still.

magento-engcom-team:
@DariuszMaciejewski thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

mcjwsk:
@magento-engcom-team I'm going to work on this issue at the weekend, so please don't close it. @vrann a comment to my last question would be appreciated.

mcjwsk:
Returning to my comment https://github.com/magento/magento2/pull/10927#issuecomment-332431802 I found other plugins where the `\Magento\Framework\View\Layout::isCacheable()` method is called. It isn't used directly, but inside the `\Magento\PageCache\Model\DepersonalizeChecker::checkIfDepersonalize()` method they use:
```php
public function checkIfDepersonalize(\Magento\Framework\View\LayoutInterface $subject)
{
    return ($this->moduleManager->isEnabled('Magento_PageCache')
        && $this->cacheConfig->isEnabled()
        && !$this->request->isAjax()
        && ($this->request->isGet() || $this->request->isHead())
        && $subject->isCacheable());
}
```

There are plugins including `afterGenerateXml()` method only:
- `\Magento\Catalog\Model\Layout\DepersonalizePlugin`
- `\Magento\Checkout\Model\Layout\DepersonalizePlugin`
- `\Magento\Persistent\Model\Layout\DepersonalizePlugin`

Example:
```php
public function afterGenerateXml(\Magento\Framework\View\LayoutInterface $subject, $result)
{
    if ($this->depersonalizeChecker->checkIfDepersonalize($subject)) {
        $this->catalogSession->clearStorage();
    }
    return $result;
}
```

but we have also plugins including `beforeGenerateXml()` and `afterGenerateXml()` both:
- `\Magento\Tax\Model\Layout\DepersonalizePlugin`
- `\Magento\Customer\Model\Layout\DepersonalizePlugin`

Example:
```php
public function beforeGenerateXml(\Magento\Framework\View\LayoutInterface $subject)
{
    if ($this->depersonalizeChecker->checkIfDepersonalize($subject)) {
        $this->customerGroupId = $this->customerSession->getCustomerGroupId();
        $this->formKey = $this->session->getData(\Magento\Framework\Data\Form\FormKey::FORM_KEY);
    }
    return [];
}

public function afterGenerateXml(\Magento\Framework\View\LayoutInterface $subject, $result)
{
    if ($this->depersonalizeChecker->checkIfDepersonalize($subject)) {
        $this->visitor->setSkipRequestLogging(true);
        $this->visitor->unsetData();
        $this->session->clearStorage();
        $this->customerSession->clearStorage();
        $this->session->setData(\Magento\Framework\Data\Form\FormKey::FORM_KEY, $this->formKey);
        $this->customerSession->setCustomerGroupId($this->customerGroupId);
        $this->customerSession->setCustomer($this->customerFactory->create()->setGroupId($this->customerGroupId));
    }
    return $result;
}
```

Your idea of replacing `afterGenerateXml()` with `afterGenerateElements()` seems to be right (I confirmed that invalid blocks are included in the `$structure->brokenParent` array then), but I'm not sure what to do with plugins implementing before and around plugins that call the `isCacheable()` method. Of course we can't change `beforeGenerateXml()` to `beforeGenerateElements()` because there is no information about broken parents before `generateElements()` method execution.

okorshenko:
Hi @DariuszMaciejewski thank you for your contribution and deep research. Unfortunately, we can't accept this Pull Request as is, since the solution has not been completed. The required solution is more complex and requires more deep refactoring. We have this issues in our backlog (MAGETWO-90967) and we attached your investigation result to the ticket. We are working with the product owner to increase issue priority to have it fixed ASAP. We will inform you once fix is delivered to mainline.

PR #10931, opened 2017-09-17
Return minimal price for configurable products on /rest/V1/products

rogyar:
### Description
Upon getting the products list via API (/rest/V1/products) configurable products always have zero price. Since in product listing on the storefront the minimal price is used as a configurable product price, it's a good idea to use it for product list fetched via API as well.

### Fixed Issues
1. magento/magento2#10644: REST API Configurable Product has price

### Manual testing scenarios
1. Create a configurable product accessible on the storefront
2. Get the products list via API using GET request to the `rest/V1/products?searchCriteria=`
3. You should get a minimal price for the configurable product instead of 0.

### DIFF
--- a/app/code/Magento/Catalog/Model/ProductRepository.php
+++ b/app/code/Magento/Catalog/Model/ProductRepository.php
@@ -20,7 +20,6 @@ use Magento\Framework\Exception\NoSuchEntityException;
 use Magento\Framework\Exception\StateException;
 use Magento\Framework\Exception\ValidatorException;
 use Magento\Framework\Exception\CouldNotSaveException;
-use Magento\ConfigurableProduct\Model\Product\Type\Configurable;
 
 /**
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
@@ -220,10 +219,7 @@ class ProductRepository implements \Magento\Catalog\Api\ProductRepositoryInterfa
             }
             $product->load($productId);
 
-            /* Add price for configurable product */
-            if ($product->getTypeId() == Configurable::TYPE_CODE) {
-                $product->setMinimalPrice($product->getFinalPrice());
-            }
+            $product->setMinimalPrice($product->getFinalPrice());
 
             $this->instances[$sku][$cacheKey] = $product;
             $this->instancesById[$product->getId()][$cacheKey] = $product;
--- a/app/code/Magento/Catalog/Model/ProductRepository.php
+++ b/app/code/Magento/Catalog/Model/ProductRepository.php
@@ -615,6 +615,7 @@ class ProductRepository implements \Magento\Catalog\Api\ProductRepositoryInterfa
         }
         $collection->joinAttribute('status', 'catalog_product/status', 'entity_id', null, 'inner');
         $collection->joinAttribute('visibility', 'catalog_product/visibility', 'entity_id', null, 'inner');
+        $collection->addMinimalPrice();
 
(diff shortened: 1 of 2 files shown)


ishakhsuvarov:
@rogyar Thank you for submission. Please consider covering the updated logic with api-functional or integration test to ensure it's working properly in the future.

orlangur:
Is it purely `2.1.x` issue? I don't see similar logic in `develop`: https://github.com/magento/magento2/blob/develop/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable/Price.php But maybe it's just implemented differently.

vrann:
@rogyar We tested the fix and it works. Couple issues with this PR:
- the same issue exists on the develop branch, it should be fixed there first and backported it here
- the PR fixes getList method of the API, but the same bug is present in get method.

Would you be able to address those?

rogyar:
@vrann sure, I'm going to handle the issue for the get request as well. Thank you.

vrann:
@rogyar another problem with the solution is that it adds minimal price to every product regardless is it simple or configurable. Would it be possible to address that as well?

rogyar:
Hi @vrann. Yes, but the `price` field is only changed for the configurable product. The minimal price is just added as an additional attribute to every product. I guess it might be useful for some users as well. What do you think?

rogyar:
Hi @orlangur, I've just checked the `develop` branch and this issue is present there as well. So, it's not only 2.1 related.

orlangur:
@rogyar I've checked this from code perspective immediately, just that in `develop` the fix could be done slightly differently. So, it makes sense to just check similar steps.

vrann:
@rogyar it sounds meaningless to add this attribute for the simple product because it will always be the same as the price. Let's not add it, unless there is a strong reason to do so.

Can you re-create the PR on develop? This one should be a backport once we accept the one on the develop.

orlangur:
> it sounds meaningless to add this attribute for the simple product because it will always be the same as the price

@vrann, totally agree but don't see a good way to do it from within `ConfigurableProduct` module. Please suggest concrete implementation in corresponding PR if you have an idea.

vrann:
@rogyar @orlangur yes, upon consideration there is no good solution unless we want to filter out the attribute from inside the collection.

Please, forward-port it to develop and we will accept both PRs.

orlangur:
@vrann current changes are incorrect as to me as we cannot refer to `ConfigurableProduct` module from `Catalog` (`ConfigurableProduct` is a removable module).

According to my understanding it's better to preserve modularity by the trade-off of minimal price added to simples as well.

Please clarify exact implementation approach which should be forward-ported.

vrann:
```
According to my understanding it's better to preserve modularity by the trade-off of minimal price added to simples as well.
```
@orlangur I  agree that's what I meant.

@rogyar Let's remove the dependency on the configurable product and have minimal price custom option added to all products.

rogyar:
I've removed the configurable product dependency and also ported the same changes to the develop branch (https://github.com/magento/magento2/pull/11123). Thank you.

vkublytskyi:
@rogyar Please fix failed tests

ishakhsuvarov:
Hi @rogyar 
I am closing this issue for now due to inactivity. Please update code and take a look at the comments if you wish to reopen and continue.
Thank you

erikkubica:
@rogyar Works for me, thank you

PR #10739, opened 2017-09-01
Fix for translated attribute label comparisson.

nowszy:
### Preconditions

1. Clean instance of Magento 2.1.8
2. Store displayed in some foreign languege i.e. pl_PL
3. Simple product with some attribute visible on product page.

### Steps to reproduce

1. Go to product page.
2. See attribute label ('none' translated to your foreign language) in the attributes section/

### Expected result

1. 'none' label translated to other languages i.e. polish lang - 'brak' should not be displayed
2. comparisson in line 37 of app/code/Magento/Catalog/view/frontend/templates/product/view/attribute.phtml is changed to if ($_attributeLabel != __('none')) in order to not display 'none' translated to foreign language

### Actual result

1. In the template app/code/Magento/Catalog/view/frontend/templates/product/view/attribute.phtml line 37 you have followin comparisson: if ($_attributeLabel != 'none') but at this part of code $_attributeLabel is already translated.

PR #10932, opened 2017-09-17
Backport #10739 - fix for translated attribute label comparison.

Januszpl:
### Description

Backport fix from #10739 plus make also `Details` section translatable that uses also `attributes.phtml`

1. magento/magento2#10738: https://github.com/magento/magento2/issues/10738

### Manual testing scenarios

1. set store to use other lang
2. set/create attribute with locale translation label for ex instead of "none" "brak" (polish translation)
3. assing attribute to product
4. on product page in short description and in detail tabs content shouldn't be displayed "none/brak" attribute

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml
+++ b/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml
@@ -109,7 +109,7 @@
                         <argument name="at_call" xsi:type="string">getDescription</argument>
                         <argument name="at_code" xsi:type="string">description</argument>
                         <argument name="css_class" xsi:type="string">description</argument>
-                        <argument name="at_label" xsi:type="string">none</argument>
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=10932) <br/>All committers have signed the CLA.

orlangur:
Whoever will review this PR, please check https://github.com/magento/magento2/issues/10738#issuecomment-328544367 - we should leave only one version of fix (according to my understanding label should be translatable).

orlangur:
Oh, excellent, my suggestion is addressed in https://github.com/magento/magento2/pull/10933. Good job @Januszpl !

orlangur (approved):

vrann:
@Januszpl can you please remind me, why do you want to have `translate=true` on Details section, isn't it affecting only the attribute value `none` which is not displayed anyway?

PR #10936, opened 2017-09-17
#10824 add name for order items grid default renderer block

Zefiryn:
Add name for default renderer block of order items to allow other modules reference it.

1. magento/magento2#10824: Cannot add new columns to item grid in admin sales_order_view layout

### Manual testing scenarios
1. In custom module create file `view/adminhtml/layout/sales_order_view.xml` with the following content
```
    <body>
        <referenceBlock name="order_items">
            <arguments>
                <argument name="columns" xsi:type="array">
                    <item name="test_column" xsi:type="string" translate="true">Test Header</item>
                </argument>
            </arguments>
            <referenceBlock name="default_order_items_renderer">
                <arguments>
                    <argument name="columns" xsi:type="array">
                        <item name="test_column" xsi:type="string" translate="true">col-test</item>
                    </argument>
                </arguments>
            </referenceBlock>

            <block class="Magento\Sales\Block\Adminhtml\Items\Column\DefaultColumn"
                   name="column_test_column"
                   template="Vendor_Module::test.phtml" group="column" />
        </referenceBlock>
    </body>
</page>
```
2. create file view/adminhtml/templates/test.phtml file with content
```
Test column
```
3. Go to the order view page.
4. Column with header "Test Header" is shown with content "Test column"

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/view/adminhtml/layout/sales_order_view.xml
+++ b/app/code/Magento/Sales/view/adminhtml/layout/sales_order_view.xml
@@ -41,21 +41,21 @@
                                 <item name="total" xsi:type="string" translate="true">Row Total</item>
                             </argument>
                         </arguments>
-                        <block class="Magento\Sales\Block\Adminhtml\Order\View\Items\Renderer\DefaultRenderer" as="default" template="Magento_Sales::order/view/items/renderer/default.phtml">
-                        <arguments>
-                            <argument name="columns" xsi:type="array">
-                                <item name="product" xsi:type="string" translate="false">col-product</item>
-                                <item name="status" xsi:type="string" translate="false">col-status</item>
-                                <item name="price-original" xsi:type="string" translate="false">col-price-original</item>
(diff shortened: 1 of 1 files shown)


ishakhsuvarov:
@Zefiryn Thanks for the PR. I think this process may be automated and all blocks which lack name may be updated. Are you interested in that kind of automation?

Zefiryn:
@ishakhsuvarov sure, are you thinking about bin/magento command?

ishakhsuvarov:
@Zefiryn more like one-time fix which updates all the blocks without names. That can be backed up  by static test which would not allow to create new blocks without names.

Zefiryn:
@ishakhsuvarov ok, I will try to work on it in the next 2 weeks and I will push it to this PR.

ihor-sviziev:
@Zefiryn I also prepared similar pull request but for other blocks and got good feedback how it could be fixed in better way. Just take a look at the following comment: https://github.com/magento/magento2/pull/10354#issuecomment-318606882

PR #10937, opened 2017-09-17
Make default sorting fully working when setup on category level

Linek:
Fix an issue #10772. 

### Description
Problem is that category has own configuration to setup Default Sort Attribute, but method to get default sorting attribute didn't take it into consideration and get it from config only.

### Fixed Issues
1. magento/magento2#10772: Unable to sort by store configuration default field when different to category's default sorting field

### Manual testing scenarios

1. In backend, go to Stores => Configuration => Catalog => Catalog
2. Make sure that within the tab "Storefront" the "Product Listing Sort By" is set to "Position".
3. In backend, go to Products => Categories
4. Go ahead and select a category to edit that has at least one product assigned to
5. Within "Display Settings" update the "Default Sort By" to be "Name" and save
6. In frontend, browse to the just changed category
7. Make sure that the default sorting is now by "Name"
8. Change that to sort by "Position"

Before fix, we had 'position' param removed from url and sorting was not applied.
After fix it should work correctly.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Helper/Product/ProductList.php
+++ b/app/code/Magento/Catalog/Helper/Product/ProductList.php
@@ -101,7 +101,7 @@ class ProductList
     public function getDefaultSortField()
     {
         $currentCategory = $this->coreRegistry->registry('current_category');
-        if($currentCategory){
+        if ($currentCategory) {
             $currentCategorySortBy = $currentCategory->getDefaultSortBy();
             return $currentCategorySortBy;
         }
--- a/app/code/Magento/Catalog/Helper/Product/ProductList.php
+++ b/app/code/Magento/Catalog/Helper/Product/ProductList.php
@@ -29,7 +29,11 @@ class ProductList
      * @var \Magento\Framework\App\Config\ScopeConfigInterface
      */
     protected $scopeConfig;
-    protected $coreRegistry;
+
+    /**
+     * @var \Magento\Framework\Registry
+     */
+    private $coreRegistry;
 
     /**
      * Default limits per page
@@ -43,10 +47,10 @@ class ProductList
      */
     public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
-        \Magento\Framework\Registry $coreRegistry
+        \Magento\Framework\Registry $coreRegistry = null
     ) {
         $this->scopeConfig = $scopeConfig;
-        $this->coreRegistry = $coreRegistry;
+        $this->coreRegistry = $coreRegistry ?: \Magento\Framework\App\ObjectManager::getInstance()->get(\Magento\Framework\Registry::class);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/Helper/Product/ProductList.php
```diff
@@ -97,6 +100,12 @@ public function getDefaultViewMode($options = [])
      */
     public function getDefaultSortField()
     {
+        $currentCategory = $this->coreRegistry->registry('current_category');
+        if($currentCategory){
```
vrann: space is needed after if and before {
`if ($currentCategory) {`

vrann (changes requested):

Linek:
@vrann Spaces added and pushed

app/code/Magento/Catalog/Helper/Product/ProductList.php
```diff
@@ -41,9 +42,11 @@ class ProductList
      * @param \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig
      */
     public function __construct(
-        \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig
+        \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
+        \Magento\Framework\Registry $coreRegistry
```
dmanners: I think this dependancy should be injected with a default of null. Then when setting to the `$this->coreRegistry` you can check for null and load from the object manager. See http://devdocs.magento.com/guides/v2.1/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter for a bit more details.

app/code/Magento/Catalog/Helper/Product/ProductList.php
```diff
@@ -29,6 +29,7 @@ class ProductList
      * @var \Magento\Framework\App\Config\ScopeConfigInterface
      */
     protected $scopeConfig;
+    protected $coreRegistry;
```
dmanners: Firstly there is a type hint missing here. Secondly does this need to be protected? My suggestion would be that this can be made private to start with as it is only being used in this class.

app/code/Magento/Catalog/Helper/Product/ProductList.php
```diff
@@ -97,6 +100,12 @@ public function getDefaultViewMode($options = [])
      */
     public function getDefaultSortField()
     {
+        $currentCategory = $this->coreRegistry->registry('current_category');
+        if ($currentCategory) {
+            $currentCategorySortBy = $currentCategory->getDefaultSortBy();
```
dmanners: Is there any reason for setting this as a variable rather than `return $currentCategory->getDefaultSortBy();`. There is no real problem with setting the variable but if it is not needed then I would recommend a quick return.

Linek:
@dmanners Thank you for the review. I've checked Backward Compatible Development doc. That's useful! 

Changes applied.

dmanners:
Thanks @Linek those changes look good to me.

Issue #6478, opened 2016-09-02
Column names row_id and entity_id seem mixed up after upgrading to Magento2 EE 2.1 or 2.1.1 possibly due to catalog staging (module magento-catalog-staging)


malachy-mcconnell:
### MySQL error message
- ERROR 1054: 1054: Unknown column 'entity_id' in 'NEW'
### SOLUTION
1. SOLUTION : set reindex mode to realtime (update on save). This deletes all the old triggers. Then set reindex mode to schedule (Update by Schedule)

### Preconditions

1. 'a normal setup' : linux, nginx, PHP 7.0.8, MySQL 5.6.31 (Ver 14.14)
2. Magento EE. Installed Magento 2.0, maybe updated a few times.
3. **Set the indexer mode to `scheduled`**
4. Was at 2.0.7 but then updated to 2.1.0 using `composer update` && `bin/magento setup:upgrade`.
5. Then 12 hours later updated to 2.1.1 also by `composer update` && `bin/magento setup:upgrade` (due to needing `base_grand_total` column in table `sales_invoice_grid`)
6. This is not a live site. There is no front-end traffic nor any admin use.
### Steps to reproduce

1. get some install of Mageneto 2.1.1
2. disable module Magento_CatalogStaging (and the modules that are dependent on it)
3. drop your entire DB schema
4. run `bin/magento setup:install` and `bin/magento setup:upgrade` to get a new DB created without the CatalogStaging module 
5. set the reindex mode to `schedule` for all indexers
6. observe the table `catalog_product_entity_int`. Note the column called `entiy_id`. **Note the triggers read `... VALUES (NEW.`entity_id`);`**
7. now enable module Magento_CatalogStaging (and enable its dependent set of modules)
8. `bin/magento setup:upgrade` 
9. **observe table `catalog_product_entity_int`. Note the column is now called `row_id` and not `entity_id` however the triggers still read `... VALUES (NEW.`entity_id`);`** which is incorrect and wreakign havoc on a number of process eg product import or if a backup of the DB is taken now, then the SQL dump cannot be re-imported into MySQL.
10. ROOT CAUSE: triggers referring to NEW.entity_id need to be deleted from some tables as part of some upgrade path
### Expected result

1. I think that after updating to Magento EE 2.1.1 when a column name is changed then any triggers referring to the column change need to be deleted and re-instated with the new column name.
### Actual result

1. The database is left in a broken state due to triggers referring to NEW.entity_id instead of NEW.row_id
[...]

PR #10958, opened 2017-09-19
Command to rebuild MySQL Triggers

josephmcdermott:
1. https://github.com/magento/magento2/issues/6478

### Manual testing scenarios
1. Start with a 2.0.x instance with Indexers set to By Schedule.
2. Upgrade to 2.1.8 or 2.1.9.
3. In the DB, attempt to delete a row from catalog_product_entity_varchar.
4. Observe error: `ERROR 1054: 1054: Unknown column 'entity_id' in 'NEW'`.
5. Run CLI command `./bin/magento indexer:rebuild-triggers`.
6. Retry step 3 with success this time.

### Contribution checklist
 - [-] All new or changed code is covered with unit/integration tests (if applicable)
 - [-] All automated tests passed successfully (all builds on Travis CI are green)


vrann:
@josephmcdermott is there any reason why this can't be a part of the UpgradeSchem.php script? I'm a bit hesitant introducing CLI command which will be used just once in a very specific use case (most SI or developers most likely will not use it ever) but it will be supported from version to version of Magento.

josephmcdermott:
No not really, it could be a setup script. I use this kind of approach in local dev from time to time though (manually turning off and on via admin panel index management), as I often find on client DB's that the triggers are missing or are owned by another user, etc. so this can be quite useful.

vrann:
@josephmcdermott would you be willing to convert it to setup script?

josephmcdermott:
Went to have a look, and looks like you guys already have this in hand with the same approach:
https://github.com/magento/magento2/commit/78b796f1cf5df53e2883650cbbec7f36a53d1e8f

The install() method does the 'turn off and turn on' I was proposing, during every install/upgrade in the RecurringData class.

PR #10965, opened 2017-09-19
Add explicit sequence dependency on Magento_Theme to Magento_Customer

CNanninga:
Since Magento_Customer references containers defined by Magento_Theme in layout XML, it should explicitly declare Magento_Theme as a sequence dependency.

It is possible (though highly difficult to predict) for a particular combination of custom extensions to result in Magento_Customer coming before Magento_Theme in the module sequence order.  This can create unintended side effects on the front-end.  If Magento_Customer layout is processed before Magento_Theme, the blocks added to the "content" container in Magento_Customer/view/frontend/layout/default.xml will result in the "content" container being processed and scheduled before the "content.top" container and to precede it in the list of children of "main".  

Ultimately, this will result in the children of "content" being output before the children of "content.top" on the final page.

### DIFF
--- a/app/code/Magento/Customer/etc/module.xml
+++ b/app/code/Magento/Customer/etc/module.xml
@@ -8,6 +8,7 @@
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
     <module name="Magento_Customer" setup_version="2.0.12">
         <sequence>
+            <module name="Magento_Theme"/>
             <module name="Magento_Eav"/>
             <module name="Magento_Directory"/>
         </sequence>


orlangur:
> Since Magento_Customer references containers defined by Magento_Theme in layout XML, it should explicitly declare Magento_Theme as a sequence dependency.

Please provide a basis for such statement. With such approach any module referring to `sidebar.main` container must declare `Magento_Theme` in sequence, for example, the `Magento_Catalog` module.

You must specify `before`/`after` attributes when order or layout elements is essential.

CNanninga:
@orlangur 

The subject isn't related to a custom extension trying to modify layout.

It's the layout order of "content.top" and "content", declared in Magento_Theme/view/frontend/layout/default.xml, that is important.  "content.top" should come before "content".  Because of Magento_Customer's reference to the "content" container, it can disrupt this order simply by being present before Magento_Theme in app/etc/config.php.

Short of providing the code of every custom extension being used in the install where I ran into this, I can't put together an exact reproduction scenario, because this is simply down to the order that modules ultimately end up in due to sequence declarations.  In my case, the module that triggered this scenario was doing only something as benign as referencing Magento_Theme in its own sequence.  Said module does not even have layout updates related to these blocks at all (nor was even enabled at the time).

If the sequence declaration is not the appropriate solution, then before/after attributes should be added to the layout declarations of "content" and "content.top" to make their order explicit.

orlangur:
@CNanninga 

Yeah, let's identify a problem we are trying to solve first.

```
<referenceContainer name="main">
            <container name="content.top" label="Main Content Top"/>
            <container name="content" label="Main Content Area"/>
            <container name="content.aside" label="Main Content Aside"/>
            <container name="content.bottom" label="Main Content Bottom"/>
        </referenceContainer>
```

Are you saying that if `Magento_Customer` is loaded earlier (we can emulate it by specifying an opposite to this PR sequence in `Magento_Theme`) then `content` container will be rendered before `content.top` as it is referenced earlier?

CNanninga:
@orlangur 
Correct

orlangur:
> If the sequence declaration is not the appropriate solution, then before/after attributes should be added to the layout declarations of "content" and "content.top" to make their order explicit.

That seems like a Plan B to me, can we change implementation in such a way that order of declarations is preserved, no matter when they are referenced for the first time?

korostii:
> can we change implementation in such a way that order of declarations is preserved, no matter when they are referenced for the first time?

Not universally. You will still have corner cases where the order is reversed in a different module, for example. Thus it'll only depend on the module load order if you don't add before/after attributes anyway and will remain fragile.

orlangur:
> You will still have corner cases where the order is reversed in a different module, for example

Before/after will be needed if order is essential for layout elements declared in different modules.

What I suggest to fix here is: if we declared ElementTop, ElementContent, ElementBottom in ONE module, there is no way they will change order besides `<move` directive.

vkublytskyi:
@CNanninga Are you still going to work on this PR?

orlangur:
@vkublytskyi https://github.com/magento/magento2/pull/10965#issuecomment-330695383 needs to be checked first and then decided whether we can fix it in Layout Framework: https://github.com/magento/magento2/pull/10965#issuecomment-333462145

CNanninga:
@vkublytskyi - Though I understand am in agreement with the reason the original PR is not an appropriate solution, I unfortunately don't have the capacity to dive into the more complex solution.

okorshenko:
Thank you all for collaboration. Unfortunately, the proposed solution is not acceptable.

PR #10980, opened 2017-09-21
Make returned object of Order::getShippingMethod() more explicit

schmengler:
Replaced the generic `DataObject` with an explicit value object for better developer experience.

### Description
This is a minor refactoring. The new class extends `DataObject` for full backwards compatibility. The magic getters are documented with `@method`, so that static analyzers recognize them.

A `__toString` method is added, so that the `$asObject` parameter in `\Magento\Sales\Model\Order::getShippingMethod()` might be dropped in a later release

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -1208,7 +1208,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
         if (!$asObject) {
             return $shippingMethod;
         } else {
-            return new ShippingMethod($shippingMethod);
+            return ShippingMethod::fromFullShippingMethodCode($shippingMethod);
         }
     }
 
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -13,6 +13,7 @@ use Magento\Framework\Pricing\PriceCurrencyInterface;
 use Magento\Sales\Api\Data\OrderInterface;
 use Magento\Sales\Api\Data\OrderStatusHistoryInterface;
 use Magento\Sales\Model\Order\Payment;
+use Magento\Sales\Model\Order\ShippingMethod;
 use Magento\Sales\Model\ResourceModel\Order\Address\Collection;
 use Magento\Sales\Model\ResourceModel\Order\Creditmemo\Collection as CreditmemoCollection;
 use Magento\Sales\Model\ResourceModel\Order\Invoice\Collection as InvoiceCollection;
@@ -1199,7 +1200,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
      * Retrieve shipping method
      *
      * @param bool $asObject return carrier code and shipping method data as object
-     * @return string|\Magento\Framework\DataObject
+     * @return string|ShippingMethod
      */
     public function getShippingMethod($asObject = false)
     {
@@ -1207,8 +1208,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
         if (!$asObject) {
             return $shippingMethod;
         } else {
-            list($carrierCode, $method) = explode('_', $shippingMethod, 2);
-            return new \Magento\Framework\DataObject(['carrier_code' => $carrierCode, 'method' => $method]);
+            return new ShippingMethod($shippingMethod);
         }
     }
 
--- a/app/code/Magento/Sales/Model/Order/ShippingMethod.php
+++ b/app/code/Magento/Sales/Model/Order/ShippingMethod.php
@@ -66,4 +66,31 @@ class ShippingMethod
     {
         return $this->method;
     }
+
+    /**
+     * Changes carrier code
+     *
+     * @deprecated The value object should be immutable.
+     * @param string $code
+     * @return ShippingMethod
+     */
+    public function setCarrierCode(string $code) : ShippingMethod
+    {
+        $this->carrierCode = $code;
+        return $this;
+    }
+
+    /**
+     * Changes method code
+     *
+     * @deprecated The value object should be immutable.
+     * @param string $medhod
+     * @return ShippingMethod
+     */
+    public function setMethod(string $medhod) : ShippingMethod
+    {
+        $this->method = $medhod;
+        return $this;
+    }
+
 }
--- a/app/code/Magento/Sales/Model/Order/ShippingMethod.php
+++ b/app/code/Magento/Sales/Model/Order/ShippingMethod.php
@@ -5,32 +5,65 @@
  */
 namespace Magento\Sales\Model\Order;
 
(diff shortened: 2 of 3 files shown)


orlangur:
Are you sure creating typified objects as descendants of `\Magento\Framework\DataObject` is encouraged way nowadays? Do we have other similar exemplars in core already?

schmengler:
@orlangur I am not sure but I hope to start a discussion and create a precedence.

I'd prefer to drop the `DataObject` inheritance but that might be one step too far at once

orlangur:
@schmengler ok, thanks, I would like to know the answer as well, maybe those old-school methods in fat ugly model are discouraged for use in general in favor of some `\Magento\Sales\Api\Data\ShippingInterface::getMethod`.

schmengler:
Maybe, because `getShippingMethod` is not part of `OrderInterface`, but then I'd also like to know how I am supposed to access the shipping method in an order before a shipment was created.

dmanners:
@schmengler do you know exactly what parts of the `DataObject` are being used in this case? I for one also would love to see the removal of that object and replace it with a different object that fits each usage.

Once you know what is needed from the object you could build the matching interface(s) and then remove the extends from the dataobject though I feel this could be opening Pandora's box.

vrann:
@schmengler ideal fix for this issue would be to declare methods explicitly in the `ShippingMethod` class. Also, please add annotations to all methods.

schmengler:
@dmanners The only methods that are used in the core, are the getters, I annotaded. And there *should* not be the need for any other methods, but you never know what people are doing...


>  I for one also would love to see the removal of that object and replace it with a different object that fits each usage.

Totally!

@vrann will do. Then the inheritance to `DataObject` can be deprecated and possibly removed in the next major module update.

ishakhsuvarov:
Hi @schmengler 
Are you still interested in proceeding with this PR?

schmengler:
Yes. First I removed the `DataObject` inheritance, then added setters for BC with deprecation notices. So the last commit can be reverted when we want to get rid of them eventually.

I think, the risk that anybody is using the `DataObject` for more than that, is very low, and if they do, keeping the `DataObject` inheritance would still break that code because carrier code and method are now own properties. In that case it's better to fail fast with a "missing method" error.

app/code/Magento/Sales/Model/Order/ShippingMethod.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->method = $method;
+    }
+
+    /**
+     * Shipping method as in shipping_method order attribute
+     *
+     * @param string $fullShippingMethod
+     * @return ShippingMethod
+     */
+    public static function fromFullShippingMethodCode(string $fullShippingMethod) : ShippingMethod
```
vkublytskyi: Introduction of new static methods is forbidden in Magento 2. We may introduce factory class instead:
```php
class ShippingMethodFactory
{
    public function create($carrierCode, $method): ShippingMethod { }
    public function createFromFullCode($code): ShippingMethod { }
}
```

app/code/Magento/Sales/Model/Order/ShippingMethod.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        list($carrierCode, $method) = explode('_', $fullShippingMethod, 2);
```
vkublytskyi: Probably would be good to throw an exception if an invalid code is passed to the method.

app/code/Magento/Sales/Model/Order.php
```diff
@@ -1199,16 +1200,15 @@ public function getTrackingNumbers()
      * Retrieve shipping method
      *
      * @param bool $asObject return carrier code and shipping method data as object
-     * @return string|\Magento\Framework\DataObject
+     * @return string|ShippingMethod
```
vkublytskyi, line 1203: This is backward incompatible change. SippingMethod should be subset of DataObject or [new public method should be introduced](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/) and this declared as `@deprecated`

vkublytskyi (changes requested):
The current implementation is backward incompatible. Introduction of specific value objects definitely is a good improvement but we should allow 3rd party developers to migrate to new approach without breaking them. See my inline comments for more details.

app/code/Magento/Sales/Model/Order.php
schmengler: Extending from DataObject was my first (defensive) approach, but consent seemed to be that it's unnecessary, so I removed it. I'm okay with reverting it though

app/code/Magento/Sales/Model/Order/ShippingMethod.php
schmengler: Please no. Factories for value objects are unnecessary overkill. Can't there be an exception for named constructors? `ShippingMethod` is not meant to be an extensible or replacable object (just as `DataObject` was not)

app/code/Magento/Sales/Model/Order/ShippingMethod.php
schmengler: good catch

app/code/Magento/Sales/Model/Order.php
schmengler: I added back the `DataObject` inheritance, and deprecated the inherited methods as originally intended

schmengler:
@vkublytskyi would you have a second look please?

app/code/Magento/Sales/Model/Order/ShippingMethod.php
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    /**
+     * Shipping method as in shipping_method order attribute
+     *
+     * @param string $fullShippingMethod
+     * @return ShippingMethod
+     * @throws \InvalidArgumentException
+     */
+    public static function fromFullShippingMethodCode(string $fullShippingMethod) : ShippingMethod
```
vkublytskyi, line 35: I'm afraid I must insist. Introduction of new static methods is forbidden and will cause failure of internal tests.

BTW Magento also does not allow to use `new` operator to instantiate objects (except exceptions and embedded classes such as `DateTime`) and this static method is hiding such implementation (in case of a factory it should use object manager to create an instance of a class).

Another one strong reason from my point of view is that knowledge about full shipping method code format now is spread across several classes (this is bad practice and lead to tight coupling). And it is better to avoid infection of this new class by the legacy implementation. Especially in class signature. "Full code" is a foreign concept for Shipping Method. And is actually a bridge from legacy implementation to a new data structure. Thereby, it should be extracted in a separate class.

vkublytskyi (changes requested):

orlangur:
Closing due to inactivity. @schmengler, feel free to reach me out anytime later if you wish to continue work on this pull request and it will be reopened.

PR #10989, opened 2017-09-21
[WIP] Fix import images with same filename in different directories

PieterCappelle:
### Description
When you're trying to import one product with 2 images and both images haves the same filename but different directory only one image will be imported. The solution I proposed in the commit is not the-best-case solution, I need Magento2 architect feedback to complete this PR. Any help would be welcome.

1. https://github.com/magento/magento2/issues/7412

### Manual testing scenarios
See the CSV in attachment as example. [products.csv.zip](https://github.com/magento/magento2/files/1321516/products.csv.zip)

I'll try to explain with an example: if you want to import some product with 2 images. You have one image in directory /images/dir1/food.jpg and you have a second image in /images/dir2/food.jpg only the first image will be imported. The reason for that is the files while importing not are renamed. The check to see if that file already exists was default set to true but must be false.

### Problem
If you run the import twice or more it will create a new image for every image (adding _2, _3, _4, ...). It will keep adding images to the catalog/product directories to keep them unique. This is not best case if you run imports multiple times. I don't see the exact solution right now, any help would be welcome.

Thanks in advance.

### Contribution checklist


dmanners:
Hi @PieterCappelle, thank you for your PR. Some of the integration tests are failing, would you be able to look into this?

PieterCappelle:
@dmanners This is work in progress. I can fix them but this fix is not correct. See my comments above. I would like second opinion from Magento how this should be done.

dmanners:
@PieterCappelle ah sorry I missed that comment. Did you raise this in the slack channel?

PieterCappelle:
I'll be closing this PR because I've created ticket to discuss in https://github.com/magento-engcom/import-export-improvements/issues/22

Issue #10532, opened 2017-08-14
Magento 2.1.7 : Impossible to get UK friendly retail prices


DominicWatts:
Magento 2.1.7. Impossible to get friendly retail prices in UK.  This was an issue for many of our stores in Magento 1 and we had to update core code to work around this issue.  Please I would like to know thoughts on the following issue.

### Steps to reproduce
1. Create Tax Rule - VAT, Retail Customer, Taxable Goods, VAT, 0, 0, 0
2. Create Tax Zone - VAT, United Kingdom, *, *, 20
3. Configure TAX - Stores > Configuration, Sales > Tax:
  -  Default Country : United Kingdom
  -  Display Product Prices in Catalog : Including Tax
4. Edit Random Product - price 3.325 ex VAT (3.99 in VAT)
5. Save
6. Product is rounded to 3.33
7. Product displays on frontend as 4.00 and not 3.99
8. Other prices are affected: try to get 6.99 for example

### Expected result
1. Not round price.  Allow 4 decimal points

### Actual result
1. Price is rounded to 2 decimal points

orlangur:
Looks like `up for grabs` from the past: https://github.com/magento/magento2/issues/1240#issuecomment-99625177

choukalos:
You should be able to manage VAT prices displayed to the customer out of the box if configured appropriately.

Can you clarify your tax configuration:
- Store/Config/Sales/Tax/Tax Calculation - Catalog Prices (should be 'including tax')
- Store/Config/Sales/Shipping Settings/Origin (should be set to UK/so you get 20% tax rate)

If you'd like Magento to have the ability to manage price digits and digit at which it's rounded then @orlangur is spot on that's an up for grabs [feature](https://github.com/magento/magento2/issues/1240#issuecomment-99625177).

DominicWatts:
Hi @choukalos ,

Catalog Prices are configured as excluding tax because that's how the prices are stored in the back office system.  But okay supposing we go through and update the product prices in Magento to be stored inc VAT (and convince the client this is the way forward).  We update the Catalog Prices config to be 'including tax'.  

Origin as UK (It was UK in my configuration)

This problem is resolved.

The reason I didn't entertain that is because previously this has caused us issues with discounts when sending to non-VATable countries.

So I went a few steps further.  If you then want to create a promotional rule with a fixed discount and send to the US (so account with US destination address) for example:

UK
£1.99 Regular Price £3.99 [great]
US
[...]

Issue #1240, opened 2015-05-06
Tax & decimal rounding method


HirokazuNishi:
Magento1.x only uses "round" for tax and decimal number method, but sometimes it causes wrong result.
Also some merchants use "floor" and "ceil" for discarding decimal numbers.

Is it possible to choose decimal discarding method via admin panel for M2?
If we do simply discard decimal numbers, converting currencies causes serious trouble.

For example, 99.99USD is 11,961.15JPY ( 1USD is 119.6235JPY ), but JPY does not use under decimal number. Then correct result is 11,961 JPY. 
If "round" is used for discarding method, sometimes converting result is not correct.

yoavkutner:
+1

choukalos:
Hey @HirokazuNishi in Magento 2 the rounding function should be consolidated for tax.  Should be fairly straight forward to do a plugin and replace that with floor or ceil or even bankers rounding to enable one to create country/taxation specific extension.

As far as choosing the # of decimal places there's a couple constants that really need to be consolidated + the automated test framework used by tax/pricing tests could use some attention to be adapted to changes in # of decimal places.  If you could do that and send in a PR that'd be awesome!  That would enable all sorts of community driven extensions around taxes/pricing.  

Thanks,
Chuck

HirokazuNishi:
@choukalos Of course I'll try to make PR, but I just started to investigation about this issue on M2. so it takes long time.
I already released one extension to solve this issue on M1, but I had to rewrite many classes and had to execute many tests. If M2 aims to be real cross-border ecommerce platform, decimal & tax issues should be solved before stable release.

choukalos:
@HirokazuNishi after you finish investigating please ping me if there are things we can do in core to make it easier to create customer tax extension.  One of our goals in the M2 tax engine was to make it easy to replace or customize.

piotrekkaminski:
Thank you for your submission.

We recently made some changes to the way we process GitHub submissions to more quickly identify and respond to core code issues.

Feature Requests and Improvements should now be submitted to the new [Magento 2 Feature Requests and Improvements](https://community.magento.com/t5/Magento-2-Feature-Requests-and/idb-p/feature-requests) forum (see details [here](https://community.magento.com/t5/News-Announcements/Improvements-to-GitHub-Management/m-p/44572#M96)).

[...]

PR #11006, opened 2017-09-22
Fix rounding error

steros:
### Description
Prices should only rounded to two decimal places in view.

I doubt 2 decimal places is correct here.
If I want a product price of "165" and thus set a base price of "138.6554" with a tax of 19% I get "165.01" because of the rounding to 2 decimal places here.
Setting this to 4 decimal places fixes that issue.

The fact that it is not possible to set a price with 4 decimal places in the backend is another issue that needs to be set separately I suppose.

### Manual testing scenarios
1. set a product price of 138.6554
2. set tax to 19%
3. go to checkout cart
4. final price is 165.01 instead of 165.00

### DIFF
--- a/app/code/Magento/Directory/Model/PriceCurrency.php
+++ b/app/code/Magento/Directory/Model/PriceCurrency.php
@@ -155,6 +155,6 @@ class PriceCurrency implements \Magento\Framework\Pricing\PriceCurrencyInterface
      */
     public function round($price)
     {
-        return round($price, 2);
+        return round($price, 4);
     }
 }


orlangur:
> The fact that it is not possible to set a price with 4 decimal places in the backend is another issue that needs to be set separately I suppose.

> set a product price of 138.6554

Looks a bit contradictory, isn't it?

Some currencies have two decimals, some three, some even zero. Proposed solution may be suitable for your particular needs but in core there should be something more sophisticated covering many scenarios, see https://github.com/magento/magento2/issues/1240#issuecomment-99625177

Problem in general is tracked as https://github.com/magento/magento2/issues/10532 I believe.

steros:
I was going from the point that this is the model not the view. Magento stores prices with 4 decimal places and calculates with them in most places I have seen so far.
The fact that there are currencies with more or less decimals should not be a matter of the model but the view.
There you should be able to set whatever decimals you please.
But the rounding that takes place in the model prevents setting some prices for cases where you need 4 decimals.

orlangur:
> The fact that there are currencies with more or less decimals should not be a matter of the model but the view.

Nope as prices/taxes/etc and corresponding business logic is model responsibility.

> The fact that it is not possible to set a price with 4 decimal places in the backend is another issue that needs to be set separately I suppose.

Such changes, if approved, should be done in scope of the same effort. All necessary scenarios needs to be tested in scope of it. Not just take https://travis-ci.org/magento/magento2/jobs/278635465#L897 and change assetions to the new value :)

Issue #7557, opened 2016-11-24
Backend Security key broken for controllers with frontname not equal to route ID


AlexandreKhayrullin:
### Preconditions
The Magento used is the "develop" branch.
PHP: PHP 7.0.10-1+deb.sury.org~xenial+1

### Steps to reproduce
1. Create a new module with a controller usable in the backend.
2. Define a new route for the controller (new adminhtml/routes.xml file) with a frontname different from the route ID.
3. Add the new controller to the menu and add the appropriate item to the ACL.
4. Log into the backend, locate the new item and click on it.

### Expected result
1. The new controller is executed.

### Actual result
1. The user is redirected back to the dashboard.

After a debugging session, it appears that the security key is incorrectly generated during either the creation of backend URLs or the security key validation.

If you place a breakpoint at the last line of \Magento\Backend\Model\Url::getSecretKey(), you'll notice the following discrepancy:
- While generating the security key used during the creation of the URL, the "$secret" string (the one that is hashed to form the security key) starts with the frontname.
- While checking the security key after having clicked on the link, the "$secret" string starts with the route ID instead.

flancer64:
These are stack traces and secret sources for hashing in my case (Magento ver. 2.2.0-rc30).

Secret key generation for navigation links:
![secret_generate](https://user-images.githubusercontent.com/5052385/30794783-dff69bb2-a1d2-11e7-984e-6312951a9c0c.png)

Secret key validation on route action:
![secret_validate](https://user-images.githubusercontent.com/5052385/30794806-ffe821e8-a1d2-11e7-8db9-f06e4299b889.png)

My module's descriptors for routes & menu:

`./etc/adminhtml/routes.xml`

    <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:noNamespaceSchemaLocation="urn:magento:framework:App/etc/routes.xsd">
        <router id="admin">
    
            <route id="fl32_login_as_route" frontName="loginas">
                <module name="Flancer32_LoginAs"/>
            </route>
    
        </router>
    </config>


`./etc/adminhtml/menu.xml`

    <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Backend:etc/menu.xsd">
        <menu>
            <add id="Flancer32_LoginAs::customer_logged"
                    module="Flancer32_LoginAs"
                    action="loginas/logged"/>
        </menu>
[...]

PR #11020, opened 2017-09-25
Prevent secure key failures if route front name is not equal to route ID

flancer64:
See https://github.com/magento/magento2/issues/7557


dmanners:
Hi @flancer64 thank you for this PR. Before we process this further could you please check the failures in the travis build.

flancer64:
Sorry guys, I checked errors but there are a lot of inappropriate usage of route ID instead of route front name and vice versa. I tried to fix them simply but I failed.

The best solution I found - don't use different route ID & front name. You can leave it "as is" and use "the best solution" too.

dmanners:
@flancer64 thanks for the update. I will check here what the best option is to progress this and get back to you.

dmanners:
@flancer64 unfortunately we cannot accept this in the current state. With all the tests failing and also there will need to be a lot of updates to the code base. Feel free to open a new PR if you have the time to make the updates required.

PR #11024, opened 2017-09-25
Download link does not get rendered in invoice email

jaykobi:
Download link does not get rendered in invoice email


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11024) <br/>All committers have signed the CLA.

miguelbalparda:
This seems to be breaking some tests, can you check them and report back?

miguelbalparda:
Also, could you please edit this PR description to use the correct template?

dmanners:
@skymeissner it looks like the unit test is failing as you have updated the method called in the `getLinks` method and as such you will also need to update the method mocked in the `Magento\Downloadable\Test\Unit\Block\Sales\Order\Email\Items\DownloadableTest` class. Are you able to update this so we can progress with the pull request?

PR #11038, opened 2017-09-26
website scope prices can be randomly deleted by cron

behnamshayani:
### Description
On at least 2 of our projects we face the issue that suddenly all website scope prices are deleted from "catalog_product_entity_decimal" table.
The issue is tracked back to 'catalog_product_outdated_price_values_cleanup' cron job, which under a request where config cache is corrupted can delete website prices due to a loose comparison of null == 0
We have added logs that show us:
$this->scopeConfig->getValue(Store::XML_PATH_PRICE_SCOPE);
can return null while direct SQL query to config table returns 1 for price scope.

I can not prove that config cache can be corrupted but please take my word on that, we had the same issue with M1 for years where locking mechanism of config cache generation goes wrong and we face cache stampede, seems like we have same issue in M2 (I can send you our fix for M1 for checking)

### Manual testing scenarios
1. change config for price scope from website to global and make sure not needed prices are deleted by cron job
2. change config for price scope from global to website, let cron job run and make sure website scope prices are not deleted

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Cron/DeleteOutdatedPriceValues.php
+++ b/app/code/Magento/Catalog/Cron/DeleteOutdatedPriceValues.php
@@ -55,7 +55,8 @@ class DeleteOutdatedPriceValues
     public function execute()
     {
         $priceScope = $this->scopeConfig->getValue(Store::XML_PATH_PRICE_SCOPE);
-        if ($priceScope == Store::PRICE_SCOPE_GLOBAL) {
+        $priceScope = $priceScope === null ? null : (int)$priceScope;
+        if ($priceScope === Store::PRICE_SCOPE_GLOBAL) {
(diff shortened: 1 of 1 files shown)


behnamshayani:
@magento-engcom-team Hi, could please provide some feedback on this PR? do you need something from me? 16 days and no feedback feels like old times again :)

dmanners:
@behnamshayani sorry for the delay. Is this problem specific to develop branch or other version of Magento2? Currently the default branch should be `2.2-develop`. If this problem happens on 2.2 could you recreate the PR going to `2.2-develop`. Then if it is needed in 2.1 we can back port it after this PR has been accepted. Thanks

behnamshayani:
Fixed in 2.1.10 and 2.2.1

PR #11045, opened 2017-09-26
Redirect to checkout after popup login

elzekool:
Redirect user to checkout after login with authentication popup

### Description
The authentication popup is used in the case that guest checkout is
disabled and the customer is not logged in when he clicks on "Proceed to
checkout". So after login it makes sense to redirect to user the
checkout

1. magento/magento2#6015: logging-in in order to checkout when guest checkout is disabled

### Manual testing scenarios
1. Disable guest checkout (Sales/Checkout/Allow Guest Checkout = No)
2. Make sure you have a customer
3. Make sure you are logged out in the frontend
4. Add a product to the cart
5. Click on "Proceed to checkout" in the minicart or cart
6. You are directed to the checkout (before you staid on the same page)

### Contribution checklist


jissereitsma:
Also tested this fix under Magento 2.1.8. Confirmed to be working.

vrann:
@elzekool, unfortunately, cannot accept this change as is, because it adds a dependency from the Customer module to the Checkout module: you require 'checkout' path to be present after customer logged in. If AuthenticationPopup cannot work without the Checkout, it should reside in the Checkout module then.

However, in wider case, AuthenticationPopup should redirect the user to the page, where it came from. Maybe you can redirect him to the last visited page?

vrann:
@elzekool, Please reopen when you want to make needed changes

PR #11052, opened 2017-09-26
Keep maintenance mode on if it was previously enabled

jokeputs:
### Description

When enabling maintenance mode it will get disabled after using one of the following commands:
- `bin\magento module:uninstall`
- `bin\magento setup:backup`
- `bin\magento setup:rollback`
- `bin\magento theme:uninstall`
- `bin\magento deploy:mode:set production`

The changes in this PR will first check or maintenance mode was already enabled. If so, it will skip disabling maintenance mode after executing the command. 

### Fixed Issues

1. magento/magento2#9918: Magento 2 automatically disables maintenance mode after certain actions

### Manual testing scenarios

1. Enable maintenance mode: `bin\magento maintenance:enable`
2. Create a backup: `bin/magento setup:backup --media`
3. Confirm that maintenance mode is still enabled: `bin\magento maintenance:status`

The output will be for example:
```
Maintenance mode already enabled
Media backup is starting...
Media backup filename: 1506422834_filesystem_media.tgz (The archive can be uncompressed with 7-Zip on Windows systems)
Media backup path: /var/www/var/backups/1506422834_filesystem_media.tgz
[SUCCESS]: Media backup completed successfully.
Skipped disabling maintenance mode
```


app/code/Magento/Theme/Console/Command/ThemeUninstallCommand.php
```diff
@@ -375,4 +378,40 @@ private function cleanup(InputInterface $input, OutputInterface $output)
...
         }
     }
+
+    /**
+     * Enable maintenance mode
+     *
+     * @param OutputInterface $output
+     * @return void
+     */
+    private function enableMaintenanceMode(OutputInterface $output)
```
antonkril: Code duplication can be avoided by extracting this behavior to a separate object

jokeputs: How would you do this? Create something like a `MaintenanceModeManagement`? Or would you add this feature to `Magento\Framework\App\MaintenanceMode`? I considered that last one but I don't think `MaintenanceMode` should be responsible for knowing or it should disable maintenance mode.
Do you already have something in mind?

jokeputs:
@antonkril, I've updated the pull request. All logic was moved to `Magento\Framework\App\Console\MaintenanceModeEnabler`.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11052) <br/>All committers have signed the CLA.

app/code/Magento/Deploy/Model/Mode.php
```diff
@@ -83,7 +83,7 @@ class Mode
      * @param OutputInterface $output
      * @param Writer $writer
      * @param Reader $reader
-     * @param MaintenanceMode $maintenanceMode
+     * @param MaintenanceModeEnabler $maintenanceMode
```
vkublytskyi: Such change is backward incompatible and not allowed for a patch releases. You should deprecate old constructor parameter and [add new one](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter).

vkublytskyi (changes requested):
@jokeputs Changing of constructor arguments is a backward incompatible change and should be avoided. See inline comment how it may be resolved.

Also, some little code duplication is still present as for each command we need to enable and disable mode.
In general, Magento does not allow introduce new public methods in patch releases. But in this case, I suppose this is ok so I propose to add new method to 
`Magento\Framework\App\MaintenanceMode`:

```php
public function executeInMaintainanceMode(callable $task)
{
    // check current maintenance mode
    // switch on maintenance mode if not yet enabled
    call_user_func($task);
    // restore initial mode if needed
}
```

Such approach will reduce code duplication to minimum an will not lead to a lot of code changes and constructor arguments (only wrapping some part of logic in anonymous function is required)

lib/internal/Magento/Framework/App/Console/MaintenanceModeEnabler.php
```diff
@@ -0,0 +1,96 @@
...
+        $this->maintenanceMode = $maintenanceMode;
+    }
+
+    /**
+     * Enable maintenance mode
+     *
+     * @param OutputInterface $output
+     * @return void
+     */
+    private function enableMaintenanceMode(OutputInterface $output)
```
vkublytskyi, line 41: I made this function private as it changes internal state. So class interfaces with enable/disable methods creates temporal coupling in a code.

Class `MaintenanceModeEnabler` is preserved in a code as it would be wrong to introduce dependency on `OutputInterface` in `MaintenanceMode` (initial recommendation).

PR #11084, opened 2017-09-27
Fix in stripped min length validation when value has special characters

rubenRP:
Fix in stripped min length validation when value has special characters

### Description
When add a UI Component form with stripped-min-lenght validation and the input has special characters, the console shows a javascript error (Syntax error: Unrecognized expression...) because uses jQuery instead native Javascript function.

1. #9920 

### Manual testing scenarios
1. Create a UI Component based form

2. Add the following validation to the XML form element definition

 ```
<item name="validation" xsi:type="array">
     <item name="required-entry" xsi:type="boolean">true</item>
     <item name="no-whitespace" xsi:type="boolean">true</item>
     <item name="stripped-min-length" xsi:type="string">4</item>
 </item>
```
3. Test in admin using value `a/b` or `ab/` or `a&c`

### Contribution checklist

### DIFF
--- a/app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
+++ b/app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
@@ -244,7 +244,7 @@ define([
         ],
         'stripped-min-length': [
             function (value, param) {
-                return $(value).text().length >= param;
+                return value.length >= param;
             },
             $.mage.__('Please enter at least {0} characters')
         ],
--- a/app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
+++ b/app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
@@ -244,7 +244,7 @@ define([
         ],
         'stripped-min-length': [
             function (value, param) {
-                return value.length >= param;
+                return _.isUndefined(value) || value.length === 0 || utils.stripHtml(value).length >= param;
             },
             $.mage.__('Please enter at least {0} characters')
         ],

### CODE REVIEW

**Comment by magento-cicd2** [comment]
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11084) <br/>All committers have signed the CLA.

File: app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js [hunk 1]
```diff
@@ -244,7 +244,7 @@ define([
         ],
         'stripped-min-length': [
             function (value, param) {
-                return $(value).text().length >= param;
+                return value.length >= param;
```
vkublytskyi: As a rule has `stripped-min-length` name, value should be stripped (use `utils.stripHtml`).

A simple check on `>= param` will make any field with this validation rule required. To avoid this implementation should be aligned with implementation of `min_text_length` rule.

vkublytskyi (changes requested):

app/code/Magento/Ui/view/base/web/js/lib/validation/rules.js
rubenRP: "As a rule has stripped-min-length name, value should be stripped (use utils.stripHtml)." OK

"A simple check on >= param will make any field with this validation rule required. To avoid this implementation should be aligned with implementation of min_text_length rule." When you say "should be aligned" I should validate the values like min_text_length (return _.isUndefined(value) || value.length === 0 || utils.stripHtml(value) >= params;) or I should use min_text_lenght function?

vkublytskyi: Function reuse is preferred as  it reduces code duplication but as there is no much code it is ok just to make same check inside stripped-min-length

rubenRP (commented):
Fix in stripped min lenght validation when value has special characters using alignment with implementation of min_text_length

vkublytskyi (approved):

PR #11133, opened 2017-09-29
Property name should not be prefixed with an underscore to indicate visibility

mayankzalavadia:
I have use https://github.com/magento/marketplace-eqp tool for testing. And as per this tool suggestion "Property name should not be prefixed with an underscore to indicate visibility"

1. Property name should not be prefixed with an underscore to indicate visibility.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Braintree/Test/Unit/Gateway/Http/Client/TransactionSaleTest.php
+++ b/app/code/Magento/Braintree/Test/Unit/Gateway/Http/Client/TransactionSaleTest.php
@@ -90,8 +90,7 @@ class TransactionSaleTest extends \PHPUnit\Framework\TestCase
         $this->adapter->expects($this->once())
             ->method('sale')
             ->with($this->getTransferData())
-            ->willReturn($response)
-        ;
+            ->willReturn($response);
 
         $this->loggerMock->expects($this->once())
             ->method('debug')
--- a/app/code/Magento/Braintree/view/adminhtml/web/js/braintree.js
+++ b/app/code/Magento/Braintree/view/adminhtml/web/js/braintree.js
@@ -211,7 +211,6 @@ define([
             }
 
             if (event.type !== 'fieldStateChange') {
-
                 return false;
             }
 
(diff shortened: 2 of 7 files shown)


orlangur:
Such changes need to be done all over the codebase at once with some automated tool. Are you interested in implementing such tool?

mayankzalavadia:
Yes

orlangur:
Great! Please implement it then, force push to this branch applying result of running all over the code and add the code you used as a comment or gist here.

mayankzalavadia:
Currently, I have use https://github.com/magento/marketplace-eqp tool for testing.  If you have any other tool for that then suggest me.

I think it's such a big task to implementing this changes on all over the codebase manually. If you have any tool for that then suggest me.

orlangur:
I don't know if there any ready-to-use tools for this. Obviously manual changes is not an option as it will not just require a lot of efforts to implement but virtually impossible to review code changes and find possible bugs.

I would suggest starting from https://github.com/squizlabs/PHP_CodeSniffer and change sniff which finds these violation so that it can fix them automatically.

Here is an example of such sniff: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/PSR2/Sniffs/ControlStructures/ElseIfDeclarationSniff.php#L58

orlangur:
> For AdminNotification Module Done

Are you going to perform changes manually? What about automation?

mayankzalavadia:
Automation is possible but in some cases, it's not working and it's causing issues for us. Let me explain you. For example, in any class, if we use any variable like `protected $_syncFlag;` then tool gives an error and for that particular line not for the line which this property is used. So, I think Automation tool also do the same thing it just removes the underscore only form this variable but not able to remove from where it's used. Because Mostly we have used this property like `$this->syncFlag;` That's why. 

If you have any other best why then please suggest me.

orlangur:
Obviously, automation should remove underscores from both property declaration and its usage.

Note that due to backward compatibility policy http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#removing-or-renaming-public-and-protected-properties leading underscore may be removed from `private` members only.

Also, as I mentioned, even if you do all the changes manually, we will be not able to process such PR as it would require insane amount of time for review.

mayankzalavadia:
okay let me try with https://github.com/squizlabs/PHP_CodeSniffer.

mayankzalavadia:
I have tried with https://github.com/squizlabs/PHP_CodeSniffer. But it's not worked. So, now I stop working on this. You just assign it to another one. Because right now I am working on another issue.

orlangur:
Ok, thanks for getting back to me. I have this in my internal TODO list, will report as `'up for grabs'` it will require some teamwork.

PR #11138, opened 2017-09-29
Added template as argument to the store address renderer to allow custom formatting

jokeputs:
### Description
See #10850 for the original pull request.

It's not easy to change the format of the store address used in for example the footer of an e-mail template. By adding the template as an argument in the constructor of `Magento\Store\Model\Address\Renderer` you can easily change it using dependency injection.


orlangur (approved):

orlangur:
@jokeputs thanks for your contribution! It was a pleasure to collaborate with you :)

jokeputs:
@orlangur, and thank you for the constructive feedback and help with the squash commit.

PR #11146, opened 2017-09-29
Add cron expression check when setting cron expression into cron schedule model

adrian-martinez-interactiv4:
Related with closed [PR#9917](https://github.com/magento/magento2/pull/9917)
Reviewed considering @okorshenko [comments](https://github.com/magento/magento2/pull/9917#issuecomment-308881495)
Add cron expression check when setting cron expression into cron schedule model. 

### Description
\Magento\Cron\Model\Schedule::setCronExpr() method only checks if cron expression has between 5 and 6 components. \Magento\Cron\Model\Schedule::trySchedule() allows to check a cron expression against a timestamp value, but there is no way to check if a cron expression is valid by itself, without a timestamp.

\Magento\Cron\Model\ResourceModel\Schedule\ExpressionValidator.php class has been added to add missing functionality, and now it's used by default at \Magento\Cron\Model\Schedule::setCronExpr() method, and when trying to try schedule.

Adapted unit tests to new check when setting cron expression into schedule model.
Added unit tests for new ExpressionValidator.php class.

### Fixed Issues
- Expressions like '2-1 * * * *' were accepted as valid expressions before. New method check does not allow to set a from-to value with the $from value higher than $to value.

### Manual testing scenarios
`php vendor/phpunit/phpunit/phpunit -c dev/tests/unit/phpunit.xml.dist app/code/Magento/Cron/Test/Unit/`

### Contribution checklist
 - [X]  All new or changed code is covered with unit/integration tests (if applicable)

### DIFF
--- a/app/code/Magento/Cron/Model/Schedule.php
+++ b/app/code/Magento/Cron/Model/Schedule.php
@@ -6,8 +6,11 @@
 
 namespace Magento\Cron\Model;
 
-use Magento\Framework\Exception\CronException;
+use Magento\Cron\Model\ResourceModel\Schedule\Expression\PartFactory;
+use Magento\Cron\Model\ResourceModel\Schedule\ExpressionFactory;
+use Magento\Cron\Model\ResourceModel\Schedule\ExpressionInterface;
 use Magento\Framework\App\ObjectManager;
+use Magento\Framework\Exception\CronException;
 use Magento\Framework\Stdlib\DateTime\TimezoneInterface;
 
 /**
@@ -32,6 +35,7 @@ use Magento\Framework\Stdlib\DateTime\TimezoneInterface;
  *
  * @api
  * @since 100.0.2
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 class Schedule extends \Magento\Framework\Model\AbstractModel
 {
@@ -51,12 +55,29 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
     private $timezoneConverter;
 
     /**
-     * @param \Magento\Framework\Model\Context $context
-     * @param \Magento\Framework\Registry $registry
+     * @var ExpressionFactory
+     */
+    private $expressionFactory;
+
+    /**
+     * @var ExpressionInterface
+     */
+    private $expression;
+
+    /**
+     * @var PartFactory
+     */
+    private $partFactory;
+
+    /**
+     * @param \Magento\Framework\Model\Context                        $context
+     * @param \Magento\Framework\Registry                             $registry
      * @param \Magento\Framework\Model\ResourceModel\AbstractResource $resource
-     * @param \Magento\Framework\Data\Collection\AbstractDb $resourceCollection
-     * @param array $data
-     * @param TimezoneInterface $timezoneConverter
+     * @param \Magento\Framework\Data\Collection\AbstractDb           $resourceCollection
+     * @param array                                                   $data
+     * @param TimezoneInterface                                       $timezoneConverter
+     * @param ExpressionFactory                                       $expressionFactory
+     * @param PartFactory                                             $partFactory
      */
     public function __construct(
         \Magento\Framework\Model\Context $context,
@@ -64,10 +85,15 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
         array $data = [],
-        TimezoneInterface $timezoneConverter = null
+        TimezoneInterface $timezoneConverter = null,
+        ExpressionFactory $expressionFactory = null,
+        PartFactory $partFactory = null
     ) {
         parent::__construct($context, $registry, $resource, $resourceCollection, $data);
         $this->timezoneConverter = $timezoneConverter ?: ObjectManager::getInstance()->get(TimezoneInterface::class);
+        $this->expressionFactory = $expressionFactory ?: ObjectManager::getInstance()->get(ExpressionFactory::class);
+        $this->expression = $this->expressionFactory->create();
+        $this->partFactory = $partFactory ?: ObjectManager::getInstance()->get(PartFactory::class);
     }
 
     /**
@@ -79,18 +105,14 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
     }
 
     /**
-     * @param string $expr
+     * @param string $cronExpr
+     *
      * @return $this
      * @throws \Magento\Framework\Exception\CronException
      */
-    public function setCronExpr($expr)
+    public function setCronExpr($cronExpr)
     {
-        $e = preg_split('#\s+#', $expr, null, PREG_SPLIT_NO_EMPTY);
-        if (sizeof($e) < 5 || sizeof($e) > 6) {
-            throw new CronException(__('Invalid cron expression: %1', $expr));
-        }
-
-        $this->setCronExprArr($e);
+        $this->expression->setCronExpr($cronExpr);
         return $this;
     }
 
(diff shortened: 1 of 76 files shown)


app/code/Magento/Cron/Model/Schedule.php
```diff
@@ -51,23 +52,34 @@ class Schedule extends \Magento\Framework\Model\AbstractModel
...
+        TimezoneInterface $timezoneConverter = null,
+        ExpressionValidator $expressionValidator = null
     ) {
         parent::__construct($context, $registry, $resource, $resourceCollection, $data);
-        $this->timezoneConverter = $timezoneConverter ?: ObjectManager::getInstance()->get(TimezoneInterface::class);
+        $this->timezoneConverter = $timezoneConverter
+            ?: ObjectManager::getInstance()->get(TimezoneInterface::class);
+        $this->expressionValidator = $expressionValidator
+            ?: ObjectManager::getInstance()->get(ExpressionValidator::class);
+        $this->expressionValidator->setSchedule($this);
```
vkublytskyi: 1. Only assignment operations allowed in a constructor.
2. `$expressionValidator` is a shared service by default so such design may lead to unpredictable behavior (several instances of Schedule model may inject the same instance of the validator and override values of each other).

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+        $this->fromToModulus = [$fromValue, $toValue, $mod];
+        return true;
+    }
+
+    /**
+     * @param $expr
+     * @return array
+     * @throws CronException
+     */
+    public function getFromToModulusValues($expr)
```
vkublytskyi: This is an additional responsibility of a class. This method is more close to *parser* than to *validator*

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /**
+     * @param string $expr
+     * @return bool
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD.NPathComplexity)
+     */
+    public function checkCronExpressionPart($expr)
```
vkublytskyi: This method used only internally. What is a reason to make it public? Try to keep code private where this possible

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+
+        return $check;
+    }
+
+    /**
+     * @param string $expr
+     * @return bool
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD.NPathComplexity)
```
vkublytskyi: New code should not use warning suppression. Should be refactored at least with breaking 
 logic into private methods. Preferable is an introduction of new classes injected as dependencies.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+            return false;
+        }
+
+        if (sizeof($expr) < self::CRON_EXPRESSION_MIN_PARTS_NUMBER
+            || sizeof($expr) > self::CRON_EXPRESSION_MAX_PARTS_NUMBER
+        ) {
+            return false;
+        }
+
+        $check = $this->checkCronExpressionPart($expr[0])
```
vkublytskyi: `foreach` loop would be more visible solution

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+     */
+    public function checkCronExpr()
+    {
+        $expr = $this->schedule->getCronExprArr();
+
+        if (!$expr) {
+            return false;
+        }
+
+        if (sizeof($expr) < self::CRON_EXPRESSION_MIN_PARTS_NUMBER
```
vkublytskyi: Usage of `count` is more conventional. Simplify task for these who will read code after.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+     * @param string $expr
+     * @return bool
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD.NPathComplexity)
+     */
+    public function checkCronExpressionPart($expr)
+    {
+        $this->lastErrorMsg = '';
+        $this->fromToModulus = [];
```
vkublytskyi: This code introduces temporal coupling which is design antipattern. Please eliminate it.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        return $this->lastErrorMsg;
+    }
+
+    /**
+     * Checks the schedule's cron expression validity
+     *
+     * @return bool
+     */
+    public function checkCronExpr()
```
vkublytskyi: Please avoid temporal coupling: accept schedule module as an argument of the method instead of holding it as internal validator state.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+        // handle ALL match
+        if ($this->isCronExpressionPartAsterisk($expr)) {
+            return true;
+        }
+
+        // handle multiple options
+        if ($this->isCronExpressionPartList($expr)) {
+            foreach (explode(',', $expr) as $e) {
+                if ($this->checkCronExpressionPart($e)) {
+                    return true;
```
vkublytskyi: Error in logic. The loop should be interrupted if some part is invalid. And if all items from the list are valid then list is valid.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+            $toValue = 60;
+        } elseif ($this->isCronExpressionPartRange($expr)) {
+            // handle range
+            $e = explode('-', $expr);
+            if (sizeof($e) !== 2) {
+                $this->lastErrorMsg = __('Invalid cron expression, expecting \'from-to\' structure: %1', $expr);
+                return false;
+            }
+
+            $fromValue = $this->schedule->getNumeric($e[0]);
```
vkublytskyi: `getNumeric` method does not deal with a model state so should not belong to it. To fix this design issue please refactor code:
- extract logic of  `getNumeric` to separate class
- inject it in schedule module and delegate call of `getNumeric` method to created class
- declare `getNumeric` method as `@deprecated`
- inject created class as a dependency into validator
- eliminate schedule model as dependency from validator and pass it as an argument to methods

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        return $this->fromToModulus;
+    }
+
+    /**
+     * @param $expr
+     * @return bool
+     */
+    public function isCronExpressionPartAsterisk($expr)
```
vkublytskyi: Please consider on possibility instead of exposing several public checks provide single public method that would return a *matcher* (e.g. `AsteriskMatcher`, `ListMatcher`, etc.). So code in schedule model instead of sequence of ifs would be in more OO style like:
```
return $this->expressionMatcherFactory->create($expression)->match()
```

Also, this task is out of *validator* responsibility.

vkublytskyi (changes requested):
@adrian-martinez-interactiv4 Please don't be discouraged by the count of comments in the review. Your PR touching a legacy code and intends to introduce new API interfaces so we should design it carefully as we would be limited to modify them in future because of backward compatibility. That's why we should make it aligned with current Magento best practices as possible.

app/code/Magento/Cron/Model/Schedule.php
adrian-martinez-interactiv4: You are right, it was my first approach, I'll change it.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
adrian-martinez-interactiv4: I thought the same when I saw this at the original code in app/code/Magento/Cron/Model/Schedule.php, it uses `sizeof` instead of `count`. I tried to preserve it, but I will change it.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
adrian-martinez-interactiv4: I agree, I'll change this.

app/code/Magento/Cron/Model/ResourceModel/Schedule/ExpressionValidator.php
adrian-martinez-interactiv4: Ok

adrian-martinez-interactiv4:
@vkublytskyi I thought full redesign of this legacy code was out of the scope of this task, but I like the idea of redesigning this part to match the new Magento best practices. I'll be back with an update for this PR.

adrian-martinez-interactiv4:
Hi @vkublytskyi , I leave here some comments about this last PR, and some of the problems I have found accross the redesign.

Cron unit tests often mention http://docs.oracle.com/cd/E12058_01/doc/doc.1014/e12030/cron_expressions.htm as the source of valid cron expressions, and specifies this cron format: 

Name | Required | Allowed Values | Allowed Special Characters
-- | -- | -- | --
Seconds | Y | 0-59 | , - * /
Minutes | Y | 0-59 | , - * /
Hours | Y | 0-23 | , - * /
Day of month | Y | 1-31 | , - * ? / L W C
Month | Y | 0-11 or JAN-DEC | , - * /
Day of week | Y | 1-7 or SUN-SAT | , - * ? / L C #
Year | N | empty or 1970-2099 | , - * /

But https://en.wikipedia.org/wiki/Cron explains this, most often used, cron format:

Field | Required | Allowed values | Allowed special characters | Remarks
-- | -- | -- | -- | --
Minutes | Yes | 0-59 | * , - | -
Hours | Yes | 0-23 | * , - | -
Day of month | Yes | 1-31 | * , - ? L W | -
Month | Yes | 1-12 or JAN-DEC | * , - | -
Day of week | Yes | 0-6 or SUN-SAT | * , - ? L # | -
Year | No | 1970–2099 | * , - | This field is not supported in standard/default implementations.

Test data providers contains examples for the first cron format, meanwhile the \Magento\Cron\Model\Schedule::trySchedule method uses the second format, as shown here:

```
$match = $this->matchCronExpression($cronExprArr[0], strftime('%M', $time))
            && $this->matchCronExpression($cronExprArr[1], strftime('%H', $time))
            && $this->matchCronExpression($cronExprArr[2], strftime('%d', $time))
            && $this->matchCronExpression($cronExprArr[3], strftime('%m', $time))
            && $this->matchCronExpression($cronExprArr[4], strftime('%w', $time));
```

This incoherence was solved at first instance performing a very generic validation for each part, that allows putting 59 as day of month, for example, and bypassing the sixth expression part, breaking the tests because the data provider expressions include allowed special characters, not supported by the implementation. The easy way was took for this part, and that's why I can't (at this point) iterate through the cron parts and perform the same validation for all of them.

I'd like to perform custom validation of each cron part, with allowed values for each one with its allowed special characters, but to achieve that, I need to know two things: cron expression part value, and its index, to know what kind of validation should be done. Unfortunately, because of the generic validation, \Magento\Cron\Model\Schedule::matchCronExpression method (that actually matches a cron expression part, its name is kind of confusing) doesn't accept part index, and as it is part of the api, can't be modified (I also think this method should be private, as \Magento\Cron\Model\ResourceModel\Schedule\Expression\Validator::checkCronExpressionPart).

If I left \Magento\Cron\Model\Schedule::matchCronExpression method out of the tests, I could use another parallel class to proxy the calls maintaining the api, but it still forces the new implementation to have a generic validation for backwards compatibility, even if this method won't be used internally no more.

Would this be the right approach? If using custom validation for each part, should I use the Oracle cron format or the Unix format? How far should I get refactoring with this?

adrian-martinez-interactiv4:
@vkublytskyi ready to review

adrian-martinez-interactiv4:
@vkublytskyi still marked with Needs update label, let me know if you need anything from my side

sidolov:
Decided with @adrian-martinez-interactiv4 close this PR due to high risks with new implementation

PR #11152, opened 2017-09-29
Fix bug reset message must create an instance of "MessageInterface"

jalogut:
On line 245 the object manager should return an instance of `\Magento\Framework\Mail\MessageInterface` same type as original message created on constructor (line 100).
However the objectManager does not use the `MessageInterface`. That causes unexpected results when the message is reseted because `$this->message` does no longer have the type set in the preferences `di.xml` for `\Magento\Framework\Mail\MessageInterface`

### Description
`reset()` method uses different class type for `$message` as the one on the constructor.  

1. If a custom preference is set for `\Magento\Framework\Mail\MessageInterface` and several emails are sent using the same instance of `\Magento\Framework\Mail\Template\TransportBuilder`, only the first email is properly sent. After the first one, following `$messages` no longer match the custom preference type. They are always type of `\Magento\Framework\Mail\Message`.

### Manual testing scenarios

1. Add a preference for `\Magento\Framework\Mail\MessageInterface`

```
    <preference for="Magento\Framework\Mail\MessageInterface" type="Vendor\Package\Mail\CustomMessage"/>
</config>
```

2. Send several messages using same instance of `\Magento\Framework\Mail\Template\TransportBuilder`

```
    $transportBuilder = $objectManager->create(\Magento\Framework\Mail\Template\TransportBuilder::class);
    $transport = $transportBuilder
            ->setTemplateIdentifier('some_email_identifier')
            ->setTemplateOptions(['area' => FrontNameResolver::AREA_CODE, 'store' => Store::DEFAULT_STORE_ID])
            ->setTemplateVars([])
            ->setFrom(['name' => 'integration-tests', 'email' => 'integration-test@custom.mail'])
            ->addTo($recipients);
        
      $useSeveralTimes = [1, 2, 3];
      foreach ($useSeveralTimes as $time) {
           $transport = $transport->getTransport();
           /**
            * Expected: all $message inside $transport to be Vendor\Package\Mail\CustomMessage
            * Actual: only first iteration $message has this type. All others are type Magento\Framework\Mail\Message
            */
      }
```
    
### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
+++ b/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
@@ -72,7 +72,7 @@ class TransportBuilder
     /**
      * Message
      *
-     * @var \Magento\Framework\Mail\Message
+     * @var \Magento\Framework\Mail\MessageInterface
      */
     protected $message;
 
--- a/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
+++ b/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
@@ -242,7 +242,7 @@ class TransportBuilder
      */
     protected function reset()
     {
-        $this->message = $this->objectManager->create(\Magento\Framework\Mail\Message::class);
+        $this->message = $this->objectManager->create(MessageInterface::class);
         $this->templateIdentifier = null;
         $this->templateVars = null;
         $this->templateOptions = null;
--- a/app/code/Magento/Newsletter/Test/Unit/Model/Queue/TransportBuilderTest.php
+++ b/app/code/Magento/Newsletter/Test/Unit/Model/Queue/TransportBuilderTest.php
@@ -26,7 +26,7 @@ class TransportBuilderTest extends \PHPUnit\Framework\TestCase
     protected $templateFactoryMock;
 
     /**
-     * @var \Magento\Framework\Mail\Message | \PHPUnit_Framework_MockObject_MockObject
+     * @var \Magento\Framework\Mail\MessageInterface | \PHPUnit_Framework_MockObject_MockObject
      */
     protected $messageMock;
 
@@ -52,7 +52,7 @@ class TransportBuilderTest extends \PHPUnit\Framework\TestCase
     {
         $objectManagerHelper = new \Magento\Framework\TestFramework\Unit\Helper\ObjectManager($this);
         $this->templateFactoryMock = $this->createMock(\Magento\Framework\Mail\Template\FactoryInterface::class);
-        $this->messageMock = $this->createMock(\Magento\Framework\Mail\Message::class);
+        $this->messageMock = $this->createMock(\Magento\Framework\Mail\MessageInterface::class);
         $this->objectManagerMock = $this->createMock(\Magento\Framework\ObjectManagerInterface::class);
         $this->senderResolverMock = $this->createMock(\Magento\Framework\Mail\Template\SenderResolverInterface::class);
         $this->mailTransportFactoryMock = $this->getMockBuilder(
@@ -159,7 +159,7 @@ class TransportBuilderTest extends \PHPUnit\Framework\TestCase
         )->method(
             'create'
         )->with(
-            $this->equalTo(\Magento\Framework\Mail\Message::class)
+            $this->equalTo(\Magento\Framework\Mail\MessageInterface::class)
         )->will(
             $this->returnValue($transport)
         );
--- a/lib/internal/Magento/Framework/Mail/Test/Unit/Template/TransportBuilderTest.php
+++ b/lib/internal/Magento/Framework/Mail/Test/Unit/Template/TransportBuilderTest.php
@@ -26,7 +26,7 @@ class TransportBuilderTest extends \PHPUnit\Framework\TestCase
     protected $templateFactoryMock;
 
     /**
-     * @var \Magento\Framework\Mail\Message | \PHPUnit_Framework_MockObject_MockObject
+     * @var \Magento\Framework\Mail\MessageInterface | \PHPUnit_Framework_MockObject_MockObject
      */
     protected $messageMock;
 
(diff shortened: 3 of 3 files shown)


lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
```diff
@@ -90,21 +88,18 @@ class TransportBuilder
...
-     * @param MessageInterface $message
+     * @param MessageInterfaceFactory $messageFactory
      * @param SenderResolverInterface $senderResolver
-     * @param ObjectManagerInterface $objectManager
      * @param TransportInterfaceFactory $mailTransportFactory
      */
     public function __construct(
         FactoryInterface $templateFactory,
-        MessageInterface $message,
+        MessageInterfaceFactory $messageFactory,
```
dmanners: I think that this should be updated so that it does not break BC. I would recommend leaving the `MessageInterface $message,` and `ObjectManagerInterface $objectManager,`  but marking them as deprecated and also unused parameters in the __construct doc block. Then adding a new optional injection of the factory. You can find more information on this approach in http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

dmanners: You can use `@SuppressWarnings(PHPMD.UnusedFormalParameter)` on the doc block to mark the fact you are not assigning `$this->message` and `$this->objectManager` anymore but keeping them in the constructor.

dmanners (changes requested):

lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
dmanners: Actually after further reading I think you should continue to assign the older parameters but mark them as deprecated. `Mark properties with the @deprecated tag instead of removing or renaming them. Continue storing the value in the property to preserve the old functionality.`

lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
```diff
@@ -229,20 +224,28 @@ public function setTemplateOptions($templateOptions)
...
-        $mailTransport = $this->mailTransportFactory->create(['message' => clone $this->message]);
+        $mailTransport = $this->mailTransportFactory->create(['message' => clone $this->getMessage()]);
         $this->reset();
 
         return $mailTransport;
     }
 
+    private function getMessage(): MessageInterface
+    {
+        if (!$this->message) {
```
dmanners: Unfortunately this will need to be a new property as to keep backwards compatibility we need to keep the injection of `$this->message` in the `__construct`. In this case creating a new private property and use it here then marke `message` as deprecated.

dmanners (changes requested):
Overall I like the changes here. Just need a couple more changes to fix backwards compatibility issue and then we can process this.

lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
orlangur: If it will be a new property, old customizations will break, isn't it?

orlangur:
Great, this is a new incarnation of https://github.com/magento/magento2/pull/10302#issuecomment-318424929. @okorshenko why `getMessage` needs to be **protected** if we discourage inheritance?

lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
dmanners: With the existing property still being injected via the constructor then this `getMessage` method would never actually call the factory and create the message class and then as such this change would not actually fix the bug. Yes a new property could break old customizations but it should limit the issues caused.

dmanners:
@orlangur I don't think that `getMessage` needs to be protected in this case I think it should be private or it could be set as public if we assume that it will be used during customizations.

orlangur:
@dmanners I tend to **private** as well, just not sure I'm aware of all needed scenarios to be supported.

jalogut:
Ok, let's make it a bit simpler, in order to keep backwards compatibility then.

dmanners:
Thank you for the update @jalogut could you also update the PR to include testing steps? You can see an example at https://github.com/magento/magento2/blob/2.2-develop/PULL_REQUEST_TEMPLATE.md

dmanners:
Also would you be able to look at the failing unit tests https://travis-ci.org/magento/magento2/jobs/282617717 and https://travis-ci.org/magento/magento2/jobs/282617721

dmanners:
Also could you update the type hint for `protected $message;` since you have fixed the fact it could be two different types of objects. In this case `Magento\Framework\Mail\MessageInterface` should be used.

dmanners:
@jalogut thanks for the update. Do you have a testing routine for this other than automated tests?

jalogut:
Hi @dmanners ,

Changes added as requested. Thanks for your feedback. We'll see if tests pass now.

jalogut:
Hi @dmanners ,

No, I do not have another testing routine. How should I check it locally? Is there any doc somewhere to check only specific tests?

dmanners:
@jalogut thanks for the update, with regards to testing. Was there a specific issue that caused you to find this bug? It would be good to have a test cases that we could validate before and after applying this fix.

orlangur:
Some "manual" scenario (involving customization which author tried to implement) is described in issue I referred to: https://github.com/magento/magento2/pull/11152#issuecomment-333337488

Not sure it this PR covered that case though.

jalogut:
Hi @dmanners,

The reason why I found this bug is described on the "manual testing scenario". I wanted to use a `CustomMessage` class for my emails. Then I ran into this error that the `di` preference is only taken into account the first time you call `getTransport()`. That's not a common thing that people would run into and it also has easy workarounds. 
I'll try to add some tests to cover that if it's required. However, IMO this is taking longer than it should. Just changing a line that is clearly wrong should not require so much effort.

dmanners:
@jalogut yeah it should not be too long to process but we are still seeing failures in the unit tests via travis.

jalogut:
@dmanners ok, definitely need to run these tests locally if I want to get that through. I'll check it again next week. Thanks

dmanners:
@jalogut nice, if you have any issues or need help just drop me a message and I would be happy to help you out.

dmanners:
@jalogut do you need any assistance with these unit tests? I would be happy to have a call/screen share and we can work through them together if that helps.

jalogut:
Hi @dmanners 
Sorry, I did not have time to look into it but I did not forget it. I am quite busy atm. It should be better in a week or 2. I will fix it then. Thanks for your offer, I will let you know

dmanners:
Closing this PR due to inactivity. @jalogut please feel free to reopen this PR when ready. It might also be a good idea to look into https://github.com/magento/magento2/pull/10302 to see what changes have been put forward with that PR also.

jalogut:
Hi @dmanners #10302 fixes exactly the same issue. It's a pity that both were closed due to inactivity. I think here is just about fixing the failing tests. I will try to find time some day.

dmanners:
@jalogut yeah. I think they fix the same thing but in different ways. Would be great to get one of the fixes sorted out. Let me know if you get some time to work on this though.

PR #11153, opened 2017-09-30
Fix the correct removal of the images and the removal of all images in the catalog [Backport]

raumatbel:
Related with the PR#[11147](https://github.com/magento/magento2/pull/11147)
Fix the correct removal of the images and the removal of all images in the catalog

### Description

When you want remove image from admin panel, this file is not removed correctly.
If the path of the image saved in the table **catalog_product_entity_media_gallery** is empty, this remove all the catalog.

### Manual testing scenarios

For the first case.
1. Upload image from admin panel
2. Save product
3. Remove this image
4. Save product
The file is not removed in **pub/media/catalog/product/...**

For the second case.
1. Upload image from admin panel
2. Save product
3. Go to the DB to the table **catalog_product_entity_media_gallery** and remove the value of this image
4. Remove this image from admin panel
5. Save product
All the files are removed in **pub/media/catalog/product/...**

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Gallery/UpdateHandler.php
+++ b/app/code/Magento/Catalog/Model/Product/Gallery/UpdateHandler.php
@@ -36,7 +36,7 @@ class UpdateHandler extends \Magento\Catalog\Model\Product\Gallery\CreateHandler
                     $catalogPath = $this->mediaConfig->getBaseMediaPath();
                     $isFile = $this->mediaDirectory->isFile($catalogPath . $image['file']);
                     // only delete physical files if they are not used by any other products and if this file exist
-                    if (!($this->resourceModel->countImageUses($image['file']) > 1) && $isFile) {
+                    if ($isFile && !($this->resourceModel->countImageUses($image['file']) > 1)) {
                         $filesToDelete[] = ltrim($image['file'], '/');
                     }
                 }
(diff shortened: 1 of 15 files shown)


vkublytskyi:
@raumatbel Thank you for your contribution. We will merge your fixes to all branches shortly.

app/code/Magento/Catalog/Model/Product/Gallery/UpdateHandler.php
```diff
@@ -33,8 +33,10 @@ protected function processDeletedImages($product, array &$images)
             if (!empty($image['removed'])) {
                 if (!empty($image['value_id']) && !isset($picturesInOtherStores[$image['file']])) {
                     $recordsToDelete[] = $image['value_id'];
-                    // only delete physical files if they are not used by any other products
-                    if (!$this->resourceModel->countImageUses($image['file']) > 1) {
+                    $catalogPath = $this->mediaConfig->getBaseMediaPath();
+                    $isFile = $this->mediaDirectory->isFile($catalogPath . $image['file']);
+                    // only delete physical files if they are not used by any other products and if this file exist
+                    if (!($this->resourceModel->countImageUses($image['file']) > 1) && $isFile) {
```
ihor-sviziev: Check "$isFile" is more faster, so would be great to swap them.

raumatbel: Can you explain me again? I don't understand well.
Do you want say me that I must not use the variable $isFile and I must specify it in the same line?. 

```// only delete physical files if they are not used by any other products and if this file exist
   if (!($this->resourceModel->countImageUses($image['file']) > 1) && $this->mediaDirectory->isFile($this->mediaConfig->getBaseMediaPath() . $image['file']);) {
```

vkublytskyi: @raumatbel code `if ($isFile && !($this->resourceModel->countImageUses($image['file']) > 1))` would be more optimal from performance perspective as if this is not a file then usage of image will not be checked.
I left up to you to make this fix or not as it may be considered as premature optimization. I think in code like this readability has higher priority.

PR #11155, opened 2017-09-30
Refactor ConfigGenerator to replace/set crypt key instead of append

renttek:
Fix the command "setup:config:set --key=..." to replace existing crypt key instead of append it.

### Description
Originally a call so "bin/magento setup:config:set --key='foobar'" did
not replace the crypt key, but append it with a newline.

This does not only "break" handling of values, encryptwed with the old
key, but also don't allow use of new key, because the new key is
composed of the old one, a newline and the new one.

1. magento/magento2#11089: setup:config:set --key append instead of replace

### Manual testing scenarios
1. Install Magento & get sure, that there is a crypt key defined in env.php
2. Call "bin/magento setup:config:set --key=<new key>"
3. New crypt key in env.php should be your new key, instead of "<old>\n<new>"

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
+++ b/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
@@ -36,7 +36,7 @@ class ConfigDataFactory
      * @param string $fileKey
      * @return ConfigData
      */
-    public function create(string $fileKey) : ConfigData
+    public function create($fileKey)
     {
         return $this->objectManager->create(ConfigData::class, ['fileKey' => $fileKey]);
     }
--- a/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
+++ b/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
@@ -18,7 +18,7 @@ class ConfigDataFactory
      * @param string $fileKey
      * @return ConfigData
      */
-    public function create(string $fileKey)
+    public function create(string $fileKey) : ConfigData
     {
         return new ConfigData($fileKey);
     }
--- a/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
+++ b/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
@@ -6,12 +6,30 @@
 
 namespace Magento\Framework\Config\Data;
 
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\ObjectManagerInterface;
+
 /**
  * Factory for ConfigData
  * @api
  */
 class ConfigDataFactory
 {
+    /**
+     * @var ObjectManager
+     */
+    private $objectManager;
+
+    /**
+     * Factory constructor
+     *
+     * @param ObjectManagerInterface $objectManager
+     */
+    public function __construct(ObjectManagerInterface $objectManager)
+    {
+        $this->objectManager = $objectManager;
+    }
+
     /**
      * Returns a new instance of ConfigData on every call.
      *
@@ -20,6 +38,6 @@ class ConfigDataFactory
      */
     public function create(string $fileKey) : ConfigData
     {
-        return new ConfigData($fileKey);
+        return $this->objectManager->create(ConfigData::class, ['fileKey' => $fileKey]);
     }
 }
--- a/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
+++ b/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Framework\Config\Data;
+
+/**
+ * Factory for ConfigData
+ * @api
+ */
+class ConfigDataFactory
+{
+    /**
+     * Returns a new instance of ConfigData on every call.
+     *
+     * @param string $fileKey
+     * @return ConfigData
+     */
+    public function create(string $fileKey)
+    {
+        return new ConfigData($fileKey);
+    }
+}
--- a/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
+++ b/lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
@@ -11,7 +11,6 @@ use Magento\Framework\ObjectManagerInterface;
 
 /**
  * Factory for ConfigData
- * @api
  */
 class ConfigDataFactory
 {
--- a/setup/src/Magento/Setup/Model/ConfigGenerator.php
+++ b/setup/src/Magento/Setup/Model/ConfigGenerator.php
@@ -82,9 +82,13 @@ class ConfigGenerator
 
         $configData = $this->configDataFactory->create(ConfigFilePool::APP_ENV);
 
-        $key = $data[ConfigOptionsListConstants::INPUT_KEY_ENCRYPTION_KEY]
-            ?? $currentKey
-            ?? $this->cryptKeyGenerator->generate();
+        // Use given key if set, else use current
+        $key = $data[ConfigOptionsListConstants::INPUT_KEY_ENCRYPTION_KEY] ?? $currentKey;
+
+        // If there is no key given or currently set, generate a new one
+        $key = $key ?? $this->cryptKeyGenerator->generate();
+
+        // Chaining of ".. ?? .." is not used to keep it simpler to understand
 
         $configData->set(ConfigOptionsListConstants::CONFIG_PATH_CRYPT_KEY, $key);
 
--- a/setup/src/Magento/Setup/Model/ConfigGenerator.php
+++ b/setup/src/Magento/Setup/Model/ConfigGenerator.php
@@ -232,21 +232,34 @@ class ConfigGenerator
     public function createCacheHostsConfig(array $data)
     {
         $configData = $this->configDataFactory->create(ConfigFilePool::APP_ENV);
+
         if (isset($data[ConfigOptionsListConstants::INPUT_KEY_CACHE_HOSTS])) {
-            $hostData = explode(',', $data[ConfigOptionsListConstants::INPUT_KEY_CACHE_HOSTS]);
-            $hosts = [];
-            foreach ($hostData as $data) {
-                $dataArray = explode(':', trim($data));
-                $host = [];
-                $host['host'] = $dataArray[0];
-                if (isset($dataArray[1])) {
-                    $host['port'] = $dataArray[1];
-                }
-                $hosts[] = $host;
-            }
+            $hosts = explode(',', $data[ConfigOptionsListConstants::INPUT_KEY_CACHE_HOSTS]);
+            $hosts = array_map([$this, 'mapHostData'], $hosts);
             $configData->set(ConfigOptionsListConstants::CONFIG_PATH_CACHE_HOSTS, $hosts);
         }
+
         $configData->setOverrideWhenSave(true);
         return $configData;
+    }
+
+    /**
+     * Splits host data to host and port and returns them as an assoc. array.
+     *
+     * @param string $hostData
+     *
+     * @return array
+     */
+    private function mapHostData($hostData)
+    {
+        list($host, $port) = explode(':', trim($hostData));
+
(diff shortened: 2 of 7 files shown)


setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -50,15 +51,25 @@ class ConfigGenerator
     protected $deploymentConfig;
 
     /**
+     * @var ConfigDataFactory
+     */
+    protected $configDataFactory;
```
dmanners: I think this can be made private rather than protected to start with.

setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -99,7 +109,7 @@ public function createCryptConfig(array $data)
      */
     public function createSessionConfig(array $data)
     {
-        $configData = new ConfigData(ConfigFilePool::APP_ENV);
```
dmanners, line 102: nice work, good to see this being moved out to a factory.

setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -70,15 +81,14 @@ public function createCryptConfig(array $data)
     {
         $currentKey = $this->deploymentConfig->get(ConfigOptionsListConstants::CONFIG_PATH_CRYPT_KEY);
 
-        $configData = new ConfigData(ConfigFilePool::APP_ENV);
+        // TODO: refactor configData to factory in constructor
```
dmanners: I would suggest leaving the create as it is rather than moving it into the factory. You could always make a `getConfigData` method that checks for `$this->configData` being null and if so calls the factory create.

dmanners: Something like https://github.com/magento/magento2/pull/11152/files#diff-dfb3ee5a7f863458afea75f5524cd2e5R233 might work in this case also.

setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -50,15 +51,25 @@ class ConfigGenerator
...
      * @param Random $random
      * @param DeploymentConfig $deploymentConfig
+     * @param ConfigDataFactory $configDataFactory
      */
-    public function __construct(Random $random, DeploymentConfig $deploymentConfig)
-    {
+    public function __construct(
+        Random $random,
+        DeploymentConfig $deploymentConfig,
+        ConfigDataFactory $configDataFactory
```
dmanners: Optional dependency like http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

dmanners (changes requested):
Add the optional dependency and I think we are good to go here.

renttek:
I made the dependency optional.

Also, I refactored the createCryptConfig method a little bit. (Introduced a new Interface on the way)

lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class ConfigDataFactory
+{
+    /**
+     * Returns a new instance of ConfigData on every call.
+     *
+     * @param string $fileKey
+     * @return ConfigData
+     */
+    public function create(string $fileKey)
```
dmanners: You can update this to use PHP 7's return type declaration if you would like. http://php.net/manual/en/migration70.new-features.php#migration70.new-features.return-type-declarations

setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -26,39 +26,49 @@ class ConfigGenerator
...
-     * @param Random $random
      * @param DeploymentConfig $deploymentConfig
+     * @param ConfigDataFactory|null $configDataFactory
+     * @param CryptKeyGeneratorInterface|null $cryptKeyGenerator
      */
-    public function __construct(Random $random, DeploymentConfig $deploymentConfig)
-    {
-        $this->random = $random;
+    public function __construct(
+        DeploymentConfig $deploymentConfig,
```
dmanners: As mentioned we need to keep the `random` parameter here but can mark the property as deprecated.

setup/src/Magento/Setup/Model/CryptKeyGenerator.php
```diff
@@ -0,0 +1,52 @@
...
+ * @api
+ */
+class CryptKeyGenerator implements CryptKeyGeneratorInterface
+{
+    /**
+     * @var Random
+     */
+    private $random;
+
+    public function __construct(Random $random = null)
```
dmanners: Since this is a new class we do not need `random` to be defaulted to null.

renttek:
hey @dmanners I pushed the requested changes

setup/src/Magento/Setup/Model/CryptKeyGenerator.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Generates & returns a string to be used as crypt key.
+     *
+     * The key length is not a parameter, but an implementation detail.
+     *
+     * @return string
+     *
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    public function generate()
```
dmanners: Again these functions can have the php7 return type deceleration such as `: type` http://php.net/manual/en/migration70.new-features.php#migration70.new-features.return-type-declarations

setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -26,39 +27,57 @@ class ConfigGenerator
...
-    public function __construct(Random $random, DeploymentConfig $deploymentConfig)
-    {
-        $this->random = $random;
+    public function __construct(
+        Random $random,
+        DeploymentConfig $deploymentConfig,
+        ConfigDataFactory $configDataFactory = null,
+        CryptKeyGeneratorInterface $cryptKeyGenerator = null
+    ) {
         $this->deploymentConfig = $deploymentConfig;
```
dmanners: Unfortunately we still need to assign `$this->random = $random` here in-case someone is using it in an extension to this class.

renttek: Aah, I thought I added id back.

dmanners:
hey @renttek it looks good to me, there are still a few new functions that could be updated to use the php7 return type declaration but that is not 100% required. If you feel like making that change then please go ahead otherwise I will check the build process and then progress this PR.

renttek:
Hey,

I added some type hints.
But for the many parts, I can't add type hints to keep BC

lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
```diff
@@ -0,0 +1,25 @@
...
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Framework\Config\Data;
+
+/**
+ * Factory for ConfigData
+ * @api
+ */
+class ConfigDataFactory
```
dmanners: Sorry I missed this before but this factory should either be updated to match the standard defined in http://devdocs.magento.com/guides/v2.2/extension-dev-guide/factories.html or actually it might be possible to use the autogenerated factory here.

dmanners:
So with the updated factory or if it works generated factory then I am happy with this code. If you need help with the factory just drop me a message and I will be happy to help out.

renttek:
Hey, I added the OM to the factory and worked on the tests.
If there are still things to improve, let me know

lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
orlangur: Autogenerated factories are not allowed in `Magento\Framework` as far as I know. But, the best option is to commit autogenerated factory here.

setup/src/Magento/Setup/Model/ConfigGenerator.php
```diff
@@ -227,22 +238,35 @@ public function createModeConfig()
...
     }
+
+    /**
+     * Splits host data to host and port and returns them as an assoc. array.
+     *
+     * @param string $hostData
+     *
+     * @return array
+     */
+    private function mapHostData(string $hostData) : array
```
dmanners: I guess we might need to use php mess dectector tags here so we ignore the error with UnusedPrivateMethod. The method is used in the array_map but the mess dectector does not pick this usage up.

setup/src/Magento/Setup/Test/Unit/Module/ConfigGeneratorTest.php
```diff
@@ -18,11 +25,26 @@ class ConfigGeneratorTest extends \PHPUnit\Framework\TestCase
...
+        $objectManagerMock = $this->getMockBuilder(\Magento\Framework\App\ObjectManager::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $objectManagerMock->method('create')->willReturn(new ConfigData('app_env'));
+
+        $configDataFactoryMock = (new ObjectManager($this))
+            ->getObject(ConfigDataFactory::class, ['objectManager' => $objectManagerMock]);
+
+        $this->configGeneratorObject = new ConfigGenerator($randomMock, $deployConfig, $configDataFactoryMock, $cryptKeyGenerator);
```
dmanners: might need to split this over a couple of lines so that the static tests do not fail. https://travis-ci.org/magento/magento2/jobs/281732875#L850

dmanners (approved):

lib/internal/Magento/Framework/Config/Data/ConfigDataFactory.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace Magento\Framework\Config\Data;
+
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\ObjectManagerInterface;
+
+/**
+ * Factory for ConfigData
+ * @api
```
dmanners: So since it is a patch release we cannot add these tag currently. @renttek Can you check all the classes and remove the `@api` tag if you added it in.

setup/src/Magento/Setup/Test/Unit/Module/ConfigGeneratorTest.php
```diff
@@ -5,24 +5,54 @@
...
-class ConfigGeneratorTest extends \PHPUnit\Framework\TestCase
+class ConfigGeneratorTest extends TestCase
 {
     /**
      * @var ConfigGenerator
      */
     private $configGeneratorObject;
 
+    /**
+     * @SuppressWarnings(PHPMD.LongVariable)
```
orlangur: I believe such rule was removed from Magento 2 PHPMD configuration. If it's just Codacy complains, simply ignore this and don't add suppressing annotations. Please remove both `@SuppressWarnings(PHPMD.LongVariable)` introduced.

orlangur (changes requested):

setup/src/Magento/Setup/Test/Unit/Module/ConfigGeneratorTest.php
renttek: Okay, I will remove them. Can such Issues be hidden/ignored on Codacy?

orlangur: Don't know, you can simply ignore it overall. There are some other dumb rules like "never use `else`" :)

renttek:
Hey @orlangur 
I removed the SupressWarnings now

dmanners:
@orlangur are you ok with the changes made? If so I will continue to process this.

orlangur (commented):
Thanks for changes.

orlangur:
@dmanners yeah, please proceed, I didn't do a full review, just small note on phpmd suppression.

Looks like "Changes requested" cannot be changed to neutral resolution (and "Approved" is not suitable if I didn't dig into details).

renttek:
yeah \o/ first code in core

dmanners:
@renttek congrats on the first PR. I have a t-shirt and some stickers for you from the event.

okorshenko:
Hi @renttek could you please accept the invite on the GitHub to join magento team?

renttek:
Hi @okorshenko I just accepted

PR #11161, opened 2017-09-30
Escape backslash in LIKE queries

Zifius:
### Description

Replaces single backslash with a double backslash which is required for the LIKE queries

1. magento/magento2#9817: Unable to find product with slash character in name

### Manual testing scenarios

Steps to reproduce are described in the original issue

### Contribution checklist

### DIFF
--- a/app/code/Magento/Ui/Component/Filters/Type/Input.php
+++ b/app/code/Magento/Ui/Component/Filters/Type/Input.php
@@ -70,7 +70,7 @@ class Input extends AbstractFilter
             if (!empty($value)) {
                 $filter = $this->filterBuilder->setConditionType('like')
                     ->setField($this->getName())
-                    ->setValue(sprintf('%%%s%%', str_replace('\\', '\\\\', $value)))
+                    ->setValue(sprintf('%%%s%%', $value))
                     ->create();
 
                 $this->getContext()->getDataProvider()->addFilter($filter);
--- a/app/code/Magento/Ui/Component/Filters/Type/Input.php
+++ b/app/code/Magento/Ui/Component/Filters/Type/Input.php
@@ -70,7 +70,7 @@ class Input extends AbstractFilter
             if (!empty($value)) {
                 $filter = $this->filterBuilder->setConditionType('like')
                     ->setField($this->getName())
-                    ->setValue(sprintf('%%%s%%', $value))
+                    ->setValue(sprintf('%%%s%%', str_replace('\\', '\\\\', $value)))
                     ->create();
 
                 $this->getContext()->getDataProvider()->addFilter($filter);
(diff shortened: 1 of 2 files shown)


app/code/Magento/Ui/Component/Filters/Type/Input.php
```diff
@@ -70,7 +70,7 @@ protected function applyFilter()
             if (!empty($value)) {
                 $filter = $this->filterBuilder->setConditionType('like')
                     ->setField($this->getName())
-                    ->setValue(sprintf('%%%s%%', $value))
+                    ->setValue(sprintf('%%%s%%', str_replace('\\', '\\\\', $value)))
```
orlangur: I don't think it's responsibility of `Magento_Ui` to be aware of LIKE queries behavior, like necessity of double backslashes. Such replacement should occur as close to database layer as possible.

For example, in  `create` method of `\Magento\Framework\Api\FilterBuilder`, or even `\Magento\Framework\Api\SearchCriteria\CollectionProcessor\FilterProcessor::addFilterGroupToCollection` / `\Magento\Framework\Data\Collection\AbstractDb::addFieldToFilter`.

Also, please implement integration test enforcing this case.

orlangur (changes requested):

app/code/Magento/Ui/Component/Filters/Type/Input.php
Zifius: Agreed, that was easy but not proper place to put the logic. I didn't find a good place to put it in the mentioned classes so it applied just to the LIKE case and moved it to adapter. Let me know if that's ok

What would be the best place to create an integration test? Are there filters tests already somewhere?

lib/internal/Magento/Framework/DB/Adapter/Pdo/Mysql.php
```diff
@@ -2876,6 +2876,9 @@ public function prepareSqlCondition($fieldName, $condition)
                 if (($key == 'in' || $key == 'nin') && is_string($value)) {
                     $value = explode(',', $value);
                 }
+                if ($key == 'like' && strpos($value, '\\') !== false) {
+                    $value = str_replace('\\', '\\\\', $value);
```
orlangur: Great! Let me check if it passes internal CI builds.

orlangur (approved):

orlangur:
A lot of functional tests are failing currently, please check: 

```
1) Magento\Checkout\Test\TestCase\OnePageCheckoutOfflinePaymentMethodsTest::test with data set "OnePageCheckoutTestVariation4_0" (OnePageCheckoutTestVariation4_0)
Exception: Product creation by webapi handler was not successful! Response: {"message":"Consumer is not authorized to access %resources","parameters":{"resources":"Magento_Catalog::products"}}
```

Probably we need more sophisticated logic: first replace all doubled backslashes to single and then apply your current `str_replace`.

orlangur:
@Zifius could you please look into test failures caused by these changes?

orlangur:
Hi @Zifius, are you going to continue work on this PR?

Zifius:
Hi @orlangur, sorry but unlikely I will have time for it in the near future and I don't like the approach taken here for the fix

orlangur:
@Zifius no problem, thanks for your efforts and for the quick feedback!

PR #11164, opened 2017-09-30
Replace module dependency sort algorithm

cmuench:
The new algorithm replaces the old bubble sort with a new topological
sort. This eliminates 53.000 is_array and 35.000 array_unique calls.

### Description
The sort logic is moved into a new Sorter class. In the main class a bunch of code
could be removed. The new logic is much more simple than the old one.

### Manual testing scenarios
Add a new module dependency.
Circular reference check can be tested by adding a circular reference in modules.xml files.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Module/ModuleList/Loader.php
+++ b/lib/internal/Magento/Framework/Module/ModuleList/Loader.php
@@ -48,24 +48,32 @@ class Loader
     private $filesystemDriver;
 
     /**
+     * @var Sorter
+     */
+    private $sorter;
+
+    /**
      * Constructor
      *
      * @param Dom $converter
      * @param Parser $parser
      * @param ComponentRegistrarInterface $moduleRegistry
      * @param DriverInterface $filesystemDriver
+     * @param Sorter $sorter
      */
     public function __construct(
         Dom $converter,
         Parser $parser,
         ComponentRegistrarInterface $moduleRegistry,
-        DriverInterface $filesystemDriver
+        DriverInterface $filesystemDriver,
+        Sorter $sorter = null
     ) {
         $this->converter = $converter;
         $this->parser = $parser;
         $this->parser->initErrorHandler();
         $this->moduleRegistry = $moduleRegistry;
         $this->filesystemDriver = $filesystemDriver;
+        $this->sorter = $sorter ?: \Magento\Framework\App\ObjectManager::getInstance()->get(Sorter::class);
     }
 
     /**
@@ -97,7 +105,7 @@ class Loader
                 $result[$name] = $data[$name];
             }
         }
-        return $this->sortBySequence($result);
+        return $this->sorter->sort($result);
     }
 
     /**
@@ -119,69 +127,5 @@ class Loader
             $filePath = str_replace(['\\', '/'], DIRECTORY_SEPARATOR, "$modulePath/etc/module.xml");
             yield [$filePath, $this->filesystemDriver->fileGetContents($filePath)];
         }
-    }
-
-    /**
-     * Sort the list of modules using "sequence" key in meta-information
-     *
-     * @param array $origList
-     * @return array
-     * @SuppressWarnings(PHPMD.UnusedLocalVariable)
-     */
-    private function sortBySequence($origList)
-    {
-        ksort($origList);
-        $expanded = [];
-        foreach ($origList as $moduleName => $value) {
-            $expanded[] = [
-                'name' => $moduleName,
-                'sequence' => $this->expandSequence($origList, $moduleName),
-            ];
-        }
-
-        // Use "bubble sorting" because usort does not check each pair of elements and in this case it is important
-        $total = count($expanded);
-        for ($i = 0; $i < $total - 1; $i++) {
-            for ($j = $i; $j < $total; $j++) {
-                if (in_array($expanded[$j]['name'], $expanded[$i]['sequence'])) {
-                    $temp = $expanded[$i];
-                    $expanded[$i] = $expanded[$j];
-                    $expanded[$j] = $temp;
-                }
-            }
-        }
-
-        $result = [];
-        foreach ($expanded as $pair) {
-            $result[$pair['name']] = $origList[$pair['name']];
-        }
-
-        return $result;
-    }
-
-    /**
-     * Accumulate information about all transitive "sequence" references
-     *
-     * @param array $list
-     * @param string $name
-     * @param array $accumulated
-     * @return array
-     * @throws \Exception
-     */
-    private function expandSequence($list, $name, $accumulated = [])
-    {
-        $accumulated[] = $name;
-        $result = $list[$name]['sequence'];
-        foreach ($result as $relatedName) {
-            if (in_array($relatedName, $accumulated)) {
-                throw new \Exception("Circular sequence reference from '{$name}' to '{$relatedName}'.");
-            }
-            if (!isset($list[$relatedName])) {
-                continue;
-            }
-            $relatedResult = $this->expandSequence($list, $relatedName, $accumulated);
-            $result = array_unique(array_merge($result, $relatedResult));
-        }
-        return $result;
     }
 }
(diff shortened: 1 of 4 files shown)


orlangur:
For the reference: previous attempt to improve this sorting was performed in https://github.com/magento/magento2/pull/10649

Did you do some benchmarking? Amount of `is_array` calls is not a proper metric for analysis.

cmuench:
@orlangur Yes, you are right. I run "bin/magento" with blackfire. Blackfire shows a ~20% better performance overall.
The module order is called twice during the bootstrap. This results in 4ms per request. The new algorithm should run in ~1ms.

orlangur:
And, before discussing a concrete implementation in details, what do you think about using an existing https://github.com/magento/magento2/pull/10649#issuecomment-324887023 topological sorting implementation?

My main concern regarding current implementation is that it encapsulates pretty abstract algorithm in concrete application. While it would be nice to have some universal `TopologicalGraphSorter` and just use it in one particular place for now.

cmuench:
OK. We can move the algorithm in a new class where we can add elements. Every element should have an "id" attribute which must be unique. Every element can have a "payload" which can have type "mixed". So we can reuse the logic in other parts of the system.

orlangur:
Thanks for quick feedback!

> What do you think about https://github.com/marcj/topsort.php? Not necessary to fully include the library, for example, we can borrow just `ArraySort` implementation.

Could you please evaluate this suggestion? In such case new Magento class will just prepare list of modules in correct format.

cmuench:
I tested this yesterday with Anton during the Magento Hackathon Munich.
The lib works like intended. IMHO Anton likes an own implementation in the Magento  framework. I will discuss this with him.

orlangur:
Ok, thanks. We can have "own" implementation by wrapping a good one in some Magento Framework component by the way :)

lib/internal/Magento/Framework/Module/ModuleList/Loader.php
```diff
@@ -48,24 +48,32 @@ class Loader
...
      * @param DriverInterface $filesystemDriver
+     * @param Sorter $sorter
      */
     public function __construct(
         Dom $converter,
         Parser $parser,
         ComponentRegistrarInterface $moduleRegistry,
-        DriverInterface $filesystemDriver
+        DriverInterface $filesystemDriver,
+        Sorter $sorter
```
dmanners: In this case I think the `Sorter` parameter should be optional http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

lib/internal/Magento/Framework/Module/ModuleList/Sorter.php
```diff
@@ -0,0 +1,83 @@
...
+
+        return $this->sortedModules;
+    }
+
+    /**
+     * @param array $element
+     * @param array $parentElements
+     * @throws \Exception
+     */
+    protected function processElement($element, $parentElements = [])
```
dmanners: Could this be updated to private since I think it is only used inside this class?

lib/internal/Magento/Framework/Module/ModuleList/Sorter.php
```diff
@@ -0,0 +1,83 @@
+<?php
+/**
+ * @copyright Copyright (c) 1999-2017 netz98 GmbH (http://www.netz98.de)
```
dmanners: I assume this is a php storm autocomplete. I guess this should be the Magento docblock instead.

cmuench: :-) Yes, you catched me. I will change the PhpStorm template.

orlangur:
@dmanners @cmuench just to clear the things up - we are sticking to own refactored implementation, not going to intergrate existed third-party toposort implementation?

cmuench:
@orlangur Yes, that is the plan. Anton is not to be a fan of too many 3rd party integrations.

orlangur:
Ok, and I'm not a big fan of reinventing the wheels when we can easily wrap and reuse something :)

As this logic is just a part of CLI and I'm not sure it is covered by any performance test, could you do a little benchmark? No need in performance boost, just to make sure there is no accidental degradation.

> This eliminates 53.000 is_array and 35.000 array_unique calls.

I believe you already did some counting, just do timer start in the beginning and timer stop in the very end. Will do a full review soon.

lib/internal/Magento/Framework/Module/ModuleList/Loader.php
```diff
@@ -97,7 +105,7 @@ public function load(array $exclude = [])
                 $result[$name] = $data[$name];
             }
         }
-        return $this->sortBySequence($result);
+        return $this->sorter->sort($result);
```
orlangur, line 108: Something like `SequenceSorter` seems to be a more appropriate name (both class and injected argument).

lib/internal/Magento/Framework/Module/ModuleList/Sorter.php
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+namespace Magento\Framework\Module\ModuleList;
+
+class Sorter
+{
+    /**
+     * @var array
+     */
+    private $origList;
```
orlangur: orig -> original

lib/internal/Magento/Framework/Module/ModuleList/Sorter.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @var array
+     */
+    private $sortedModules;
+
+    /**
+     * @param array $origList
+     * @return array
+     */
+    public function sort(array $origList)
```
orlangur, line 33: Please specify in PHPDoc of this method exact algorithm used.

lib/internal/Magento/Framework/Module/ModuleList/Sorter.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @param array $element
+     * @param array $parentElements
+     * @throws \Exception
+     */
+    protected function processElement($element, $parentElements = [])
+    {
+        if (isset($parentElements[$element['name']])) {
+            $relatedName = implode('', array_slice(array_keys($parentElements), -1, 1));
```
orlangur: This is quite ineffective operation. Looks like you needed https://stackoverflow.com/a/2348213/8640867

lib/internal/Magento/Framework/Module/Test/Unit/ModuleList/LoaderTest.php
```diff
@@ -54,7 +55,9 @@ protected function setUp()
         $this->parser->expects($this->once())->method('initErrorHandler');
         $this->registry = $this->createMock(\Magento\Framework\Component\ComponentRegistrarInterface::class);
         $this->driver = $this->createMock(\Magento\Framework\Filesystem\DriverInterface::class);
-        $this->loader = new Loader($this->converter, $this->parser, $this->registry, $this->driver);
+
+        $sorter = new Sorter();
```
orlangur: We cannot test **two** classes in **unit** test. Create a new test specifically for `SequenceSorter` and remove anything unrelated from this one.

UPD: great, you already created such test 👍 Just mock `SequenceSorter` here then.

lib/internal/Magento/Framework/Module/Test/Unit/ModuleList/SorterTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Magento\Framework\Module\Test\Unit\ModuleList;
+
+use Magento\Framework\Module\ModuleList\Sorter;
+use PHPUnit\Framework\TestCase;
+
+class SorterTest extends TestCase
```
orlangur, line 12: Please make sure new class is 100% covered. At least there is no test for `Circular sequence reference from` exception currently.

orlangur (changes requested):

orlangur:
Hi @cmuench, any updates on this one?

orlangur:
Closing due to inactivity. Feel free to reach me out anytime later if you wish to continue work on pull request and it will be reopened.

PR #11183, opened 2017-10-02
FIX for #11166 Index Handling Fatal Error

larsroettig:
see #11166
- Add Error Handling
- Implement Test Case

### DIFF
--- a/app/code/Magento/Indexer/Model/Indexer.php
+++ b/app/code/Magento/Indexer/Model/Indexer.php
@@ -418,7 +418,7 @@ class Indexer extends \Magento\Framework\DataObject implements IdxInterface
                 $state->save();
                 $this->getView()->resume();
                 throw $exception;
-            }catch (\Error $error) {
+            } catch (\Error $error) {
                 $state->setStatus(StateInterface::STATUS_INVALID);
                 $state->save();
                 $this->getView()->resume();
--- a/app/code/Magento/Indexer/Model/Indexer.php
+++ b/app/code/Magento/Indexer/Model/Indexer.php
@@ -418,6 +418,11 @@ class Indexer extends \Magento\Framework\DataObject implements IdxInterface
                 $state->save();
                 $this->getView()->resume();
                 throw $exception;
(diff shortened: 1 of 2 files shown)


app/code/Magento/Indexer/Model/Indexer.php
```diff
@@ -418,6 +418,11 @@ public function reindexAll()
                 $state->save();
                 $this->getView()->resume();
                 throw $exception;
+            }catch (\Error $error) {
```
vrann: - please add space before catch
- is there more specific exception you can catch?

vrann (changes requested):

app/code/Magento/Indexer/Model/Indexer.php
larsroettig: It not so easy to a more specific exception:
Currently,  **php7** throw much different but you must catch all to set the correct state for the indexer. After a fatal, the indexer must be on state **invalid**

php__errors_in_php_7_-_manual

vrann (approved):

hostep:
Nice, thanks @larsroettig!

I think this fix should be backported to Magento 2.1.x, would be appreciated if someone is willing to do this.

larsroettig:
Hi, @hostep I cannot so easy backport my Fix because 2.1.X allowed to use **php5.6** and this Error handling requires **php7.0** it is broken change for 2.1.X. To catch Engine Error only for one function is in **php5.6** not so simple. 

Magento  2.1.X:
https://github.com/magento/magento2/blob/13d38880e25347d3aec59494813d643fb6975e0d/composer.json#L11

Magento  2.2.X
https://github.com/magento/magento2/blob/4d2609343a950e9ee6b50cc4444469d50d82b139/composer.json#L11

hostep:
@larsroettig: ah yes, you have a valid point there!

We could add a dependency to https://github.com/symfony/polyfill-php70 to get around this, but that's probably a bit out of scope and might introduce new problems.

PR #11184, opened 2017-10-02
Removed translate=“true” from definition.map.xml fixes crowdin from showing those strings

paales:
Crowdin is showing all sorts of strings that shouldn't be translated:
https://crowdin.com/translate/magento-2/1374/enus-nl#q=settings%2F

If those strings are translated it will potentially cause problems.

### DIFF
--- a/app/code/Magento/Ui/view/base/ui_component/etc/definition.map.xml
+++ b/app/code/Magento/Ui/view/base/ui_component/etc/definition.map.xml
@@ -11,20 +11,17 @@
         <schema name="current">
             <argument name="data" xsi:type="array">
                 <item name="config" xsi:type="array">
-                    <item name="label" type="string" translate="true" xsi:type="xpath">settings/label</item>
+                    <item name="label" type="string" xsi:type="xpath">settings/label</item>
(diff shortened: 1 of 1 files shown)


fooman:
thanks for this contribution @paales - the changes make sense to me but it has tripped a couple of existing tests on travis. Would you mind taking a look at fixing those as part of this pull request.

paales:
Got no idea how to fix the Integration tests the errors are counterintuitive and don't know what to change to solve the issue. If someone else can fix the integration tests..

okorshenko:
Hi @paales could you please look at `\Magento\Ui\Config\Argument\Parser\XpathType::parse`
`translate="true"` attribute only make sense for `type='string'`
`xsi:type="xpath"` means that we will read the value by xpath and translate the value, but not xpath itself. Please, let me know if there is an issue in this logic. Otherwise, we need to reject this PR.

paales:
@okorshenko Alright that makes sense. I was creating the PR because in the past translations caused bugs because some strings were translated that shouldn't have been. In this case all those paths end up in crowdin.. They shoudn't be there..

okorshenko:
@paales I'm closing this PR. Thank you for research

PR #11198, opened 2017-10-03
Reversed order of YES and NO options in Boolean source

giacmir:
Just a tiny change to reverse the appearance order of Yes and No options in Boolean attributes in product edit page

### Description

When you have a boolean attribute without any or null value (I.e. because you've imported it from csv and that was a non required attribute not present in csv) and you go to the product edit page you will have all the select with the YES option selected. This happens because if no value is present the selected option is the firs one. So if you save all those attributes are saved with YES value.

Since the attribute value was null during import it is safe to assume that the user meant NO for the value of that attributes, so this PR changes the order in which those values are showed so that with save those attributes are saved with NO as default.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/Entity/Attribute/Source/Boolean.php
+++ b/app/code/Magento/Eav/Model/Entity/Attribute/Source/Boolean.php
@@ -42,8 +42,8 @@ class Boolean extends \Magento\Eav\Model\Entity\Attribute\Source\AbstractSource
     {
         if ($this->_options === null) {
             $this->_options = [
-                ['label' => __('Yes'), 'value' => self::VALUE_YES],
                 ['label' => __('No'), 'value' => self::VALUE_NO],
+                ['label' => __('Yes'), 'value' => self::VALUE_YES],
             ];
         }
         return $this->_options;


orlangur:
> Since the attribute value was null during import it is safe to assume that the user meant NO for the value of that attributes

No, it is not safe. No matter which order of options is exactly an opposite scenario exists which would cause trouble when you set NO by default.

Any value of any attribute must be explicitly set by content manager. If you need `null` value treated as NO in your particular store, you need to fix it during import and not change core behavior.

hostep:
I agree with @orlangur.
_But_ the current implementation isn't correct either. If the attribute value is `null` in the database, then it should not show up as NO or YES, but as a third option. Just an empty value as the first value of the dropdown for example? I'm not sure how this will behave when the attribute is required btw ...

orlangur:
@hostep according to my understanding when value is `null` during import, default value of corresponding attribute should be used. Probably if there is no default value specified for attribute, import should not be allowed, no matter whether field is required or not?

Current system behavior for Edit Product according to my understanding: set attribute value to default in case it is not specified in database.

hostep:
@orlangur: sorry, you are correct. I was under the impression that a yes/no field was rendered as as dropdown, but it's in a fact a sort of on/off switch which can only have two states.
The default value of such an attribute is required and is one of two options: yes or no. So the current behavior in the PEF (product edit form) makes sense.

I do get the point of @giacmir though, if the value in the database is `null` I don't know how this should be represented on the PEF, both YES and NO are incorrect.
So if there is a real fix here, it's probably to change the importer to not allow you to import `null` values for yes/no attributes, and if you try to import one as `null` it should change to the default option. I haven't tested the importer yet with this scenario, so not sure how it currently behaves.

orlangur:
@giacmir thanks for suggested change, however there is no need to change system behavior like this.

As explained above, the best way would be to change `null` values appropriately before importing. Another option to achieve same behavior as this PR introduced is to specify default values correctly on Edit Product Attribute page (the one with URL similar to `/admin/catalog/product_attribute/edit/attribute_id/175/`).

giacmir:
Thanks @orlangur and @hostep,

I'ma bit late for the discussion but i'd like to add a couple of confiderations:

_"change null values appropriately before importing"_:

this is possibile only if we are speaking of automated procedures, but since the merchant can import a csv file hand-compiled by itself this is not really an option.
Moreover we are speaking about import, but what if we add a boolean attribute after creating products? Same problem and no chance to sanitize the existing products.

_"specify default values correctly on Edit Product Attribute page"_:

this solution is probably the way to go, but at this time it has the same issue of abov: the merchant.
We need to change the attribute validation system and the attribute edit form so that inserting a default value for boolean attributes is mandatory.

orlangur:
It's never too late ;)

> We need to change the attribute validation system and the attribute edit form so that inserting a default value for boolean attributes is mandatory.

It should be not hard to customize this behavior for the needs of particular store. I.e. instead of pasting default boolean attribute value make merchant to explicitly choose Yes or No if there were no value before.

While the issue seem fixable for your particular case as you described, in general there is no difference whether the order is Yes/No or No/Yes. But such change we can probably even break some existed fragile customization which relied on the order.

PR #11199, opened 2017-10-03
Add db-prefix from env conf when command admin:user:create is executed

osrecio:
Read config env.php to get db-prefix when command `admin:user:create` is executed

### Description
If you don't read `db-prefix` in env.php you don't have this information in this function: `Magento\Setup\Model\AdminAccount::getTableName` that is called in first instance in`\Magento\Setup\Model\Installer::installAdminUser`.

1. magento/magento2#11176: IConfigured table prefix is not recognized in CLI admin:user:create

### Manual testing scenarios
1. Install fresh Magento with db-prefix
2. Create admin user: `php bin/magento admin:user:create --admin-user=user --admin-password=pass123Lolailo --admin-email=mail@mail.com --admin-firstname=fname --admin-lastname=lname`

### Contribution checklist

### DIFF
--- a/setup/src/Magento/Setup/Model/Installer.php
+++ b/setup/src/Magento/Setup/Model/Installer.php
@@ -1010,6 +1010,7 @@ class Installer
     public function installAdminUser($data)
     {
         $this->assertDbConfigExists();
+        $data += ['db-prefix' => $this->deploymentConfig->get(ConfigOptionsListConstants::CONFIG_PATH_DB_PREFIX)];
         $setup = $this->setupFactory->create($this->context->getResources());
         $adminAccount = $this->adminAccountFactory->create($setup->getConnection(), (array)$data);
         $adminAccount->save();


orlangur:
Is it an easiest way to make `$this->setup->getTable('admin_user')` return correct table name? I doubt it's a good idea to do it in some particular command.

osrecio:
Where would you put `$this->setup->getTable('admin_user')` in this case ? 
I mean, you don't have access to assign `db-prefix` in other part that inside the command.

PS: maybe i didn't understand your question

orlangur:
I'm referring to the underlying `\Magento\Setup\Model\AdminAccount::saveAdminUser`. Can we inject `db-prefix` on a higher level somehow so that there is no need to do it in each affected command?

osrecio:
Ahh Ok, Maybe would be better approach, but in this case, Only found a reference for `\Magento\Setup\Model\Installer::installAdminUser` in this command.

orlangur:
What about any other interactions with database?

osrecio:
Yeah, I know, but in Internal Ticket: `MAGETWO-62966` was changed `Magento\Setup\Module\Setup` by `Magento\Framework\DB\Adapter\AdapterInterface` in this file [AdminAccountFactory.php](https://github.com/magento/magento2/blob/2.2-develop/setup/src/Magento/Setup/Model/AdminAccountFactory.php).

I can change again but I don't know certainly is good idea.

setup/src/Magento/Setup/Console/Command/AdminUserCreateCommand.php
```diff
@@ -62,7 +75,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return \Magento\Framework\Console\Cli::RETURN_FAILURE;
         }
         $installer = $this->installerFactory->create(new ConsoleLogger($output));
-        $installer->installAdminUser($input->getOptions());
+        $options = $input->getOptions();
+        $options['db-prefix'] = $this->deploymentConfig->get(ConfigOptionsListConstants::CONFIG_PATH_DB_PREFIX);
+        $installer->installAdminUser($options);
```
orlangur: Fix must be moved into model `\Magento\Setup\Model\Installer::installAdminUser`. This class is already aware of `deploymentConfig`, all occurrences of `DB_PREFIX` are already in models.

orlangur (changes requested):

setup/src/Magento/Setup/Console/Command/AdminUserCreateCommand.php
osrecio: Added `Magento\Setup\Module\Setup` to Interface `AdminAccountFactory` inside installAdminUser.

I hope these changes are ok for you.

orlangur: I don't think we should change `AdapterInterface` back to `Setup`.

I suggested just to move DB prefix obtaining from command to `\Magento\Setup\Model\Installer::installAdminUser` model (not to any other model).

osrecio:
Added to Installer the information about `db-prefix`

osrecio:
Squashed commits in one, Thanks @orlangur

orlangur:
Great! 👍

orlangur (approved):

PR #11226, opened 2017-10-04
Remove quantity_and_stock_status attribute from excluded attributes in Catalog Widget

rubenRP:
Remove quantity_and_stock_status from excluded attributes in Catalog Widget

### Description
Remove quantity_and_stock_status from excluded attributes in Catalog Widget in di.xml. It allows load quantity field in filter select when load a Catalog Widget in CMS pages or blocks as in 1.9.x

1. magento/magento2#11187: Quantity Condition not shown in Widget

### Manual testing scenarios
1. Create a CMS page or block
2. Add a Product list widget
3. Add a quantity filter
4. Test in front-end

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogWidget/etc/di.xml
+++ b/app/code/Magento/CatalogWidget/etc/di.xml
@@ -6,11 +6,5 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
-    <type name="Magento\CatalogWidget\Model\Rule\Condition\Combine">
-        <arguments>
-            <argument name="excludedAttributes" xsi:type="array">
-                <item name="0" xsi:type="string">quantity_and_stock_status</item>
-            </argument>
-        </arguments>
-    </type>
+    <type name="Magento\CatalogWidget\Model\Rule\Condition\Combine" />
 </config>


vrann:
@rubenRP this change will introduce another issue, please see here magento/magento2#4108
So it's not enough just to add the attribute, it will require updating the table with the right value when product is out of stock

orlangur:
I don't think filter should be enhanced this way as the attribute is going to be removed: https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Model/Product/Attribute/Backend/Stock.php#L13

vrann:
@naydav please advice what is the future of this attribute and how this issue should be fixed

naydav:
We marked this attribute as a deprecated because we want to remove it from the database...
Values in the product data would remain (and it would be filled from the stock index), but in the DB(in EAV tables) the attribute should be deleted

Global refactoring will be provided in scope of work on [MSI](https://github.com/magento-engcom/msi) project

vrann:
@naydav so in the meantime, it should be fixed, because it is still available, right?

orlangur:
> Values in the product data would remain

@naydav thus workability of this filter will be preserved, right?

> @see `StockStatusInterface` when you want to read the stock data for representation layer (storefront)

Another option could be that widgets should be reworked to use this interface instead of deprecated attribute. Just confirm it is not needed at least in short-term.

vrann:
@orlangur it is not needed because in short term nothing changes, let's proceed with this PR

vrann:
@rubenRP can you please make sure this fix does not introduces regression from this bug magento/magento2#4108

vrann:
@rubenRP cannot accept this fix because it introduces the regression, please re-open with the updated code.

rubenRP:
Ok @vrann

PR #11254, opened 2017-10-05
Fix for #10317 Disable region list when switching to a region optional country.

romainruaud:
This is a fix for issue #10317 

### Description

The PR consist only on disabling the region dropdown element after hiding it when switching to a country not using option based regions.

This prevent the previous value being sent when submitting the form.

1. magento/magento2#10317: Region is being overridden when changing from a required-state country to one that is not required

### Manual testing scenarios

1. Log into your Account on the frontend
2. Edit your billing address and save using EUA for country and pick a state (Alaska) , save it.
3. Edit your billing address again, change from EUA to Uruguay, change the region input for Test
4. Save it and you will see Test state showed. Without this fix, it would still have shown Alaska.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/region-updater.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/region-updater.js
@@ -206,7 +206,7 @@ define([
                     regionInput.removeClass('required-entry');
                 }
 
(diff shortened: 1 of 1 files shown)


romainruaud:
@vrann can you flag this one as partner contribution ? Regards

Issue #10470, opened 2017-08-09
attribute_group_id in eav_attribute_group. Data Type is to small.


SRaromicon:
### Preconditions

1. Magento 2.1.7
2. PHP 7
3. Percona 5.6

### Steps to reproduce

1. Copy Default Attribute Set to New Set. (programmatically)
2. Add New Attribute Group to the new Set.  (programmatically)
3. Create about 6000 Attribute Sets and use the new Set as Skeleton (programmatically)

### Expected result

1. It is possbile to create a lot of Attribute Sets with a lot of Attribute Groups

### Actual result

1. After nearly 5500 Attribute Sets you will recieve this error:

`[PDOException]                                                                                   
  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '65535' for key 'PRIMARY'`

attribute_group_id is a SMALLINT(5) so the max ID is 65535. With 12 Attribute Groups per Set (11 are default), you can only generate 5461 Attribute Sets.

### Solution

Change data type of attribute_group_id in all EAV Tables to INT(10)

vrann:
#mm17es

orlangur:
Hi @SRaromicon, please provide more details. While technically it is possible to reach any limit Magento is supposed to support only more or less realistic scenarios.

5500 attribute sets first of all seems not really manageable with current UX.

Even extra large performance profile contains only 200 attribute sets https://github.com/magento/magento2/blob/2.2-develop/setup/performance-toolkit/profiles/ce/extra_large.xml#L43 thus we have no idea how Magento would perform with x100 bigger value.

orlangur:
Closing due to inactivity, feel free to share additional details anytime later and this issue can be reevaluated.

PR #11281, opened 2017-10-07
allow more attribute sets to be created by increasing id size

patrikpihlstrom:
### Description
This commit increases the size of the attribute set ids in order to allow more attribute sets.

### Fixed Issues
1. magento/magento2#10470: attribute_group_id in eav_attribute_group. Data Type is to small.

### Manual testing scenarios
I've created a gist containing a php-script that will automatically create 7000 attribute sets, which can be found here: https://gist.github.com/patrikpihlstrom/d31921a6d7aa2c5b674e4b49232868b6
1. Download the files in the gist and place them in your magento root directory
2. Run test.php (you'll get some output describing what is happening)
3. Make sure all of your attribute sets were created

### DIFF
--- a/app/code/Magento/Catalog/Setup/InstallSchema.php
+++ b/app/code/Magento/Catalog/Setup/InstallSchema.php
@@ -40,7 +40,7 @@ class InstallSchema implements InstallSchemaInterface
             )
             ->addColumn(
                 'attribute_set_id',
-                \Magento\Framework\DB\Ddl\Table::TYPE_SMALLINT,
+                \Magento\Framework\DB\Ddl\Table::TYPE_INTEGER,
                 null,
                 ['unsigned' => true, 'nullable' => false, 'default' => '0'],
                 'Attribute Set ID'
@@ -671,7 +671,7 @@ class InstallSchema implements InstallSchemaInterface
             )
             ->addColumn(
                 'attribute_set_id',
-                \Magento\Framework\DB\Ddl\Table::TYPE_SMALLINT,
+                \Magento\Framework\DB\Ddl\Table::TYPE_INTEGER,
                 null,
                 ['unsigned' => true, 'nullable' => false, 'default' => '0'],
                 'Attriute Set ID'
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11281) <br/>All committers have signed the CLA.

orlangur:
What was the old limit? Why anyone may need that much attribute sets?

patrikpihlstrom:
The current limit is somewhere around 5500. Apparently some people need more than that.

orlangur:
> Apparently some people need more than that.

I wouldn't say it is obvious, can you share some links maybe?

orlangur:
5500 is strange. I'm too lazy to check the sizing of `SMALLINT` but it should be a power of 2 - 16384, 32768, 65536 etc.

orlangur:
@patrikpihlstrom please provide more details. While technically it is possible to reach any limit Magento is supposed to support only more or less realistic scenarios.

5500 attribute sets first of all seems not really manageable with current UX.

patrikpihlstrom:
This isn't really anything I need or even want. I just figured I'd create a pr since this was a confirmed issue that some people had. Feel free to close this if you want :)

orlangur:
@patrikpihlstrom thanks, I looked into the gist but somehow missed fixed issue mentioned. Checking.

orlangur:
Closing pull request as issue it fixes seems to be invalid. Thanks for collaboration @patrikpihlstrom.

cmuench:
@orlangur We have same problem in a system. The customer is heavily changing the product portfolio via API.
Sets are deleted and created in frequent way. So we reached the limit after several month.
The argumentation that the UX is not possible to handle the amount of data is inconsitent. In our case no one uses the UX. We had the same issues in the checkout with addresses. With the latest releases this problem was fixed. It was possible to add the addresses as well.

Please re-open this issue.

cmuench:
@orlangur Problem btw is not the amount of attribute sets. It's more the limited auto_increment value.

PR #11320, opened 2017-10-09
Fix email not sent when sitemap generation has errors

marinagociu:
### Description

This PR changes the logic for handling errors in the sitemap generation cron. If an exception is thrown when trying to generate any of the sitemaps, the processes is not stopped anymore, but instead the errors are sent by email based on the XML Sitemap configuration. The old `_translateModel` property is not used anymore, and the inline translation is correctly suspended using the `inlineTranslation` property instead.

1. magento/magento2#10502: Fatal error: Call getTranslateInline of null when generating some sitemap with errors

### Manual testing scenarios

1. Create a sitemap.
2. Delete the directory in which the sitemap should be generated, or any other thing that will result in an error when trying to generate the sitemap.
3. In the XML Sitemap configuration enable the cron for generating the sitemap. Setup an email address to receive the errors.
4. Let the cron run and try to generate the sitemap.
5. An email should be received in the configured email address and the cron task should be successful. Before, the cron task went into the status `error` at the first sitemap that could not be generated and the exception message was saved in on the cron task in the message field.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sitemap/Model/Observer.php
+++ b/app/code/Magento/Sitemap/Model/Observer.php
@@ -121,8 +121,7 @@ class Observer
             \Magento\Store\Model\ScopeInterface::SCOPE_STORE
         )
         ) {
-            $translate = $this->_translateModel->getTranslateInline();
-            $this->_translateModel->setTranslateInline(false);
+            $this->inlineTranslation->suspend();
 
             $this->_transportBuilder->setTemplateIdentifier(
                 $this->_scopeConfig->getValue(
--- a/app/code/Magento/Sitemap/Model/Observer.php
+++ b/app/code/Magento/Sitemap/Model/Observer.php
@@ -113,7 +113,6 @@ class Observer
                 $sitemap->generateXml();
             } catch (\Exception $e) {
                 $errors[] = $e->getMessage();
-                throw $e;
             }
         }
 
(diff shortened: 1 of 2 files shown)


vkublytskyi:
@marinagociu Thank you for contribution. Please fix failed **unit** tests (see Travis builds). Actually, unit tests duplicate the error from code so it expects that exception is thrown.

marinagociu:
@vkublytskyi, I think I need you input and help on this as I'm not sure how to proceed. As far as I see it the unit test that fails should be removed since the method should not throw an exception, but instead it should send the exception messages by email. Can you confirm this? Should I remove that test, or replace it with another one? If so, with what kind of test should I replace it? Thanks!

vkublytskyi:
@marinagociu It should be enough just to fix failed unit test:
- remove `@expectedException \Exception` annotation
- add some message to exception (e.g. `Something goes wrong`)
- ensure that for `$this->transportBuilderMock` method `setTemplateVars` is invoked with `['warnings' => 'Something goes wrong']` argument.

You may also face the issue caused by fact that observer use method chaining on transport builder. So you will need to mock all methods to return `transportBuilderMock` or refactor code to call all methods on `$this->_transportBuilder` instead of chaining.

marinagociu:
@vkublytskyi thank you so much for you help! I've updated the unit test with the suggested changes.

PR #11357, opened 2017-10-10
 Allow custom payment gateway commands

luckyraul:
In order to create custom payment gateway command - we need to allow to call this method from any other place

### DIFF
--- a/app/code/Magento/Payment/Model/Method/Adapter.php
+++ b/app/code/Magento/Payment/Model/Method/Adapter.php
@@ -520,7 +520,7 @@ class Adapter implements MethodInterface
     /**
      * @inheritdoc
      */
-    private function executeCommand($commandCode, array $arguments = [])
+    public function executeCommand($commandCode, array $arguments = [])
     {
         if (!$this->canPerformCommand($commandCode)) {
             return null;


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11357) <br/>All committers have signed the CLA.

orlangur:
> we need to allow to call this method from any other place

No, as you are not supposed to call methods not belonging to `MethodInterface`.

If you need to implement custom command, just use `\Magento\Payment\Gateway\Command\CommandManagerInterface::executeByCode`.

Please provide more details on a scenario you need to implement and how you trying to achieve this.

luckyraul:
by this code I access to adapter class 

$order->getPayment()->getMethodInstance()


how can I call command ?

orlangur:
You must call only methods belonging to `\Magento\Payment\Model\MethodInterface`.

You can implement any additional methods in your code, like `\Magento\Paypal\Model\Payflowlink::process`.

GitHub is intended for bug reports, please use [Magento Stack Exchange](https://magento.stackexchange.com) for questions on how particular customization can be implemented.

luckyraul:
I can do whatever I want, so the only one choice now is to extend this class with protected CommandManagerInterface and create custom method, just we same as it is here.

orlangur:
> the only one choice now is to extend this class with protected 

Extending core classes is highly discouraged.

luckyraul:
```
If you need to implement custom command, just use \Magento\Payment\Gateway\Command\CommandManagerInterface::executeByCode.
```
As it sad in documentation I should di.xml  \Magento\Payment\Model\Method\Adapter,  which implements \Magento\Payment\Model\MethodInterface, and have dependency for \Magento\Payment\Gateway\Command\CommandManagerInterface. And now I can't call executeByCode, because it is wrapped inside private function executeCommand. 

Every standard commands(capture, void, etc) have their own public methods inside class. If I create command and add it to pool via di.xml, I am locked in a trap, I can't use it. 

My pull request is about allowing users to resolve this problem without extending classes.

Also I can't modify \Magento\Payment\Model\MethodInterface as it can break other payment methods. 

May be I need to create something like \Magento\Payment\Model\GatewayMethodInterface?

orlangur:
Making some non-intended to be public method public does not solve any problem but just breaks class encapsulation.

You can extend interface with necessary methods, like `\Magento\Vault\Model\VaultPaymentInterface` does. Note that `\Magento\Vault\Model\Method\Vault` does not extend any classes, simply use `$commandExecutor->executeByCode` the same way in your code.

luckyraul:
Yes, Vault case is not suitable to me, you suggest to create whole huge class with very same methods with same code for just one method.

orlangur:
That's basically how it is supposed to work, composition is preferred over inheritance.

```
    /**
     * @inheritdoc
     * @since 100.1.0
     */
    public function validate()
    {
        return $this->getVaultProvider()->validate();
    }
```

Just proxy existing methods with unchanged logic like this.

If you would do it via inheritance or by making some class internals public, technically it would work, but such code would be of poor quality and may lead to problems with next upgrades.

luckyraul:
okay

PR #8685, opened 2017-02-26
[PSR-2 Compliance] Fix #8612: Hundreds of PHPCS-based static tests violations in mainline

orlangur:
## Scope of Work
- Enabled PHPCS-based tests for the whole codebase
- Combined three PHPCS-based tests into one
- Polished up Magento Coding Standard into one `ruleset.xml`
    - no new rules added
    - only those rules were removed which are contained in PSR2
- Made static tests run under PHP 7 which is ~3 times faster than PHP 5.6
- Fixed all existing coding style violations, mostly using automation tools
    - no `@codingStandardsIgnoreFile` annotations removed to reduce size of PR
    - new `@codingStandardsIgnoreFile` only added when really necessary:
        - one wrongly named `interface` marked with `@api` (can be addressed separately, didn't want to mix with cosmetic changes)
        - few `phtml` files which show false-positive with currently used PHPCS 1.5.3 and no violation with latest stable PHPCS 2.8.0 (example: https://github.com/orlangur/magento2/commit/ecf0948ee1c26f40d7ca95003cb9508e80e06a89)

## Code Review Notes
- Changes which require manual review are not so big compared to the size of pull request: https://github.com/magento/magento2/compare/develop...orlangur:phpcs-tests-refactoring-no-auto-fixes?expand=1
- Commits https://github.com/orlangur/magento2/commit/74f3a5aa142986d0ca3a4bbf015001710ffa58f2 and https://github.com/orlangur/magento2/commit/bc79f6d8db09104b72925cd43b96ee42326aabff are obtained using automatic tools, step-by-step instructions are present in commit messages; to review them simply repeat the same steps and assure that no changes shown by `git diff`
- These instructions may be also applied to non-public code if the fix of PHPCS-based tests revealed a lot of violations in it

PR #11362, opened 2017-10-11
PHP Livecodetest testCodeStyle() method does not use whitelist files

adrian-martinez-interactiv4:
### Description
\Magento\Test\Php\LiveCodeTest::testCodeStyle method does not use whitelist files, as it does \Magento\Test\Php\LiveCodeTest::testCodeMess. Because of that, it takes ages to complete when you only want to test code style for a few modified files.

With proposed changes, it will try to find whitelisted files, and if it does not find any, then full whitelist will be checked. By default the behaviour is the same, as long as you have the ability to limit processed files if you create whitelist files from outside.

Improvement

### Manual testing scenarios
1. Create (manually, via git diff, via get_github_changes.php script) Magento/Test/_files/changed_files_local.txt, with references to real modified files to be tested.
2. Run vendor/bin/phpunit --verbose -c dev/tests/static/phpunit-all.xml.dist.

### Contribution checklist

### DIFF
--- a/dev/tests/static/phpunit.xml.dist
+++ b/dev/tests/static/phpunit.xml.dist
@@ -30,6 +30,7 @@
     </testsuites>
     <php>
         <ini name="date.timezone" value="America/Los_Angeles"/>
+        <!-- TESTCODESTYLE_IS_FULL_SCAN - specify if full scan should be performed for test code style test -->
         <const name="TESTCODESTYLE_IS_FULL_SCAN" value="1"/>
         <!-- TESTS_COMPOSER_PATH - specify the path to composer binary, if a relative reference cannot be resolved -->
         <!--<const name="TESTS_COMPOSER_PATH" value="/usr/local/bin/composer"/>-->
--- a/dev/tests/static/phpunit-all.xml.dist
+++ b/dev/tests/static/phpunit-all.xml.dist
@@ -20,5 +20,7 @@
     </testsuites>
     <php>
         <ini name="date.timezone" value="America/Los_Angeles"/>
+        <!-- TESTCODESTYLE_IS_FULL_SCAN - specify if full scan should be performed for test code style test -->
+        <const name="TESTCODESTYLE_IS_FULL_SCAN" value="1"/>
     </php>
 </phpunit>


orlangur:
It was done intentionally in https://github.com/magento/magento2/pull/8685 and should not be ever changed in order to avoid regression of fixed problem. PHPCS run takes only few minutes while PHPMD would take hours.

Behavior of `$changedFiles = self::getChangedFilesList($changedFilesBaseDir);` can be slightly changed in your particular project if current one does not work well.

orlangur:
Current implementation needs to be changed so that:
- 100% whitelisted files check is preserved on any CI
- it is possible to run a reduced check locally

dev/tests/static/phpunit.xml.dist
```diff
@@ -30,6 +30,7 @@
     </testsuites>
     <php>
         <ini name="date.timezone" value="America/Los_Angeles"/>
+        <env name="phpunit.static.testcodestyle.use_whitelist" value="0"/>
```
orlangur: I suggest changing it to `testcodestyle.is_full_scan` and default `1` value.

I'm not sure whether it's better to use env or const, like we do in integration tests `<const name="TESTS_CLEANUP" value="enabled"/>`.

dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php
```diff
@@ -201,11 +201,16 @@ private function getFullWhitelist()
 
     public function testCodeStyle()
     {
+        $useWhiteList = isset($_ENV['phpunit.static.testcodestyle.use_whitelist'])
+            && $_ENV['phpunit.static.testcodestyle.use_whitelist'] === '1';
```
orlangur: Simply `$whitelist = !empty(...is_full_scan...) ? ... `

Please use both `php` and `phtml`.

orlangur (changes requested):

orlangur (approved):

adrian-martinez-interactiv4:
@vrann I updated this PR to include new const also in phpunit-all.xml.dist file, and add comments, could you check the commit associated to this PR? https://github.com/magento/magento2/pull/11362/commits/205c2f902456a1dea6753916afec7b8cd57d62b0

The one added in MAGETWO-81646 - https://github.com/magento/magento2/commit/bf50355dc656bb9b3b192d63fc413964b066865d does not include the last changes, and due to this, conflicts have raised: https://github.com/magento/magento2/pull/11362/conflicts.

orlangur:
@adrian-martinez-interactiv4 it cannot be reverted, please apply these additional changes on top of new `develop` and force push to the same branch.

adrian-martinez-interactiv4:
@orlangur Done

orlangur:
@adrian-martinez-interactiv4 

```
Magento\Test\Php\LiveCodeTest::testCodeStyle
PHPUnit\Framework\Exception: Warning: file_get_contents(.../magento2/dev/tests/static/report/phpcs_report.txt): failed to open stream: No such file or directory in .../magento2/dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php:212.

...1/magento2/dev/tests/static/framework/bootstrap.php:56
.../magento2/dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php:212
```

Somehow it is not passing on Bamboo currently, I need some time to find out why. No action required from your side, just some patience :)

ihor-sviziev:
@orlangur it looks like after https://github.com/magento/magento2/commit/f59cfe252d2f04fcc7a7e04dea95202d1e8ffc15 issue should gone away. Could you re-check?

orlangur:
@ihor-sviziev it is just a workaround, I need to find out how such changes broke the Bamboo CICD loop.

PR #11369, opened 2017-10-11
Database backup doesn't include triggers #9036

denisristic:
ADDED getTableTriggersSql() and now it returns valid SQL with drop & create queries for triggers

1. magento/magento2#9036

### DIFF
--- a/app/code/Magento/Backup/Model/Db.php
+++ b/app/code/Magento/Backup/Model/Db.php
@@ -173,6 +173,7 @@ class Db implements \Magento\Framework\Backup\Db\BackupDbInterface
             }
         }
         $backup->write($this->getResource()->getTableForeignKeysSql());
+        $backup->write($this->getResource()->getTableTriggersSql());
         $backup->write($this->getResource()->getFooter());
 
         $this->getResource()->commitTransaction();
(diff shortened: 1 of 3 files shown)


orlangur:
Shouldn't they be recreated with `bin/magento setup:upgrade`? Please check https://github.com/magento/magento2/pull/10958#issuecomment-330806589

denisristic:
@orlangur I think that this issue is not about indexer triggers but SQL database triggers. If you create inside SQL database custom trigger than, before my fix, those wouldn't be dumped to a backup file (you could't recreate them from backup)

orlangur:
I don't think making any changes to database, like adding triggers, manually is a Magento way.

Why not simply implement it in `RecurringData`, similar to the example with `->unsubscribe()->subscribe()`?

denisristic:
@orlangur I know that this isn't Magento way to do something, but I didn't report this issue nor I've put it on to-do list on SQUASHTOBERFEST :)
You are able to do a backup (dump) of database and it makes sense to dump triggers (for any reason) as well

hostep:
@orlangur: please read the issue report https://github.com/magento/magento2/issues/9036, it will make everything clear ;)

orlangur:
@hostep I read it before my very first comment here :) Just that I don't think such case is valid.

Does Magento backup database tables created manually?

It is fine to dump triggers when you do it with some tooling. From within Magento I don't think it's a good idea to dump anything not application-specific.

Let me raise this question in maintainers group, maybe I'm missing something.

PR #11393, opened 2017-10-12
[Backport 2.2] Fix getImage method when sending product alert email

Jeroenll:
Backport of #10208.

### Description
When sending product alert emails from cron jobs it will do so from the backend area. An exception is thrown `Unable to resolve the source file for 'frontend/_view/nl_NL/Magento_Catalog/images/product/placeholder/.jpg'` because the product image can't be loaded.

Although Magento will mark the subscription send, in fact it's not. To solve this I created an emulation to the product image is loaded in the fronted area.

1. None that I could find.

### Manual testing scenarios
1. Add yourself to the mailing list for product alerts (e.g. out of stock products).
2. Run the cron job which is responsible for sending the email if a product is back in stock.
3. Magento will mark the subscription as send but in fact it's not.

### Contribution checklist

### DIFF
--- a/app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
+++ b/app/code/Magento/ProductAlert/Block/Email/AbstractEmail.php
@@ -3,24 +3,21 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
-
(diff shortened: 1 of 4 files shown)


Jeroenll:
Not reproducible on 2.2-develop.

PR #11397, opened 2017-10-12
Fix for #9566: Show the correct label in the admin

michielgerritsen:
### Description
When creating an order status it is possible to set different labels for the different store views. This label was used in the backend as well, this fix makes sure the general status label is used in the backend.

### Fixed Issues
1. magento/magento2#9566: Status label is wrong in admin

### Manual testing scenarios
1. Create order
2. Open edit order page in admin and note "Order Status"
3. Go to Admin: Stores>Settings>Order Status and edit row with status that has order
4. Change label for Default Store View = "Processing on store view"
5. Refresh order edit form

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Config.php
+++ b/app/code/Magento/Sales/Model/Order/Config.php
@@ -206,7 +206,6 @@ class Config
             return $this->stateStatuses[$key];
         }
         $statuses = [];
-        $area = $this->state->getAreaCode();
 
         if (!is_array($state)) {
             $state = [$state];
@@ -218,7 +217,7 @@ class Config
                 foreach ($collection as $item) {
                     $status = $item->getData('status');
                     if ($addLabels) {
-                        $statuses[$status] = $area == 'adminhtml' ? $item->getLabel() : $item->getStoreLabel();
+                        $statuses[$status] = $this->getStatusLabel($status);
                     } else {
                         $statuses[] = $status;
                     }
--- a/app/code/Magento/Sales/Model/Order/Config.php
+++ b/app/code/Magento/Sales/Model/Order/Config.php
@@ -122,8 +122,14 @@ class Config
      */
     public function getStatusLabel($code)
     {
-        $code = $this->maskStatusForArea($this->state->getAreaCode(), $code);
+        $area = $this->state->getAreaCode();
+        $code = $this->maskStatusForArea($area, $code);
         $status = $this->orderStatusFactory->create()->load($code);
+
+        if ($area == 'adminhtml') {
+            return $status->getLabel();
+        }
+
         return $status->getStoreLabel();
     }
 
--- a/app/code/Magento/Sales/Model/Order/Config.php
+++ b/app/code/Magento/Sales/Model/Order/Config.php
@@ -206,6 +206,7 @@ class Config
             return $this->stateStatuses[$key];
         }
         $statuses = [];
+        $area = $this->state->getAreaCode();
 
         if (!is_array($state)) {
             $state = [$state];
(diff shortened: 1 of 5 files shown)


dev/tests/integration/testsuite/Magento/Sales/Model/Order/StatusTest.php
```diff
@@ -0,0 +1,48 @@
...
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Sales\Model\Order;
+
+/**
+ * Class ShipmentTest
+ * @package Magento\Sales\Model\Order
+ */
+class StatusTest extends \PHPUnit\Framework\TestCase
```
okorshenko, line 13: you can have I test + dataPorviderer instead of 2 similar tests

okorshenko (changes requested):

michielgerritsen:
Hi @okorshenko,

I've updated the test to use a data provider.

dmanners:
Hi @michielgerritsen thanks for the PR so far it looks good and the status is changed on the main order detail display. What would your thoughts be on also updating the "Notes for this order" section's drop-down also?

order-status

michielgerritsen:
Hi David,

Thanks, makes sense. I just updated the PR with changes to use the correct label in that dropdown.

dev/tests/integration/testsuite/Magento/Sales/Model/Order/StatusTest.php
dmanners: @okorshenko are you happy with this updated test?

app/code/Magento/Sales/Model/Order/Config.php
```diff
@@ -211,7 +218,7 @@ public function getStateStatuses($state, $addLabels = true)
                 foreach ($collection as $item) {
                     $status = $item->getData('status');
                     if ($addLabels) {
-                        $statuses[$status] = $item->getStoreLabel();
+                        $statuses[$status] = $area == 'adminhtml' ? $item->getLabel() : $item->getStoreLabel();
```
dmanners: Great looks like this fixes the problem. It does appear that this is the same check as in `getStatusLabel`, assuming that `$item` here and `$status` in the `getStatusLabel` are the same type. Is it possible to use the same method for getting the label based on the area?

dmanners (changes requested):
Thanks for this change. It does look like there is a bit of code duplication here. Could we consider using the same method for both cases?

michielgerritsen:
Makes sense. I just changed the code by calling the `getStatusLabel`, which holds the area check.

dmanners:
Thanks @michielgerritsen code looks much cleaner now. Could you take a look at the unit tests that are failing. You should be able to run the failing test cases by running `./vendor/bin/phpunit --config dev/tests/unit/phpunit.xml.dist app/code/Magento/Sales/Test/Unit/Model/Order/ConfigTest.php`

```
1) Magento\Sales\Test\Unit\Model\Order\ConfigTest::testGetStatuses with data set "pending state" ('pending', true, array(Magento\Framework\DataObject Object (...)), array('Pending label'))
Error: Call to a member function load() on null
/home/travis/build/magento/magento2/app/code/Magento/Sales/Model/Order/Config.php:127
/home/travis/build/magento/magento2/app/code/Magento/Sales/Model/Order/Config.php:220
/home/travis/build/magento/magento2/app/code/Magento/Sales/Test/Unit/Model/Order/ConfigTest.php:150
```

michielgerritsen:
Hi @dmanners,

It took some time to find time, but i just updated the code to fix the test. Fingers crossed that Travis will mark it as green :-)

dmanners (approved):

michielgerritsen:
Does @okorshenko maybe have a little bit time to look at this on the last day of #squashtoberfest?

PR #11405, opened 2017-10-12
Allow setting of http response status code in a Redirection

gabrielqs-redstage:
Extensions currently cannot define http response status code when redirecting. It is always overwritten to 302

1. magento/magento2#9028: You cannot set a 303 redirect response using a result factory

### Manual testing scenarios

1. Perform a 303 redirect from any controller (see code below)
2. Send a request to that controller endpoint
3. Examine the response code

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Controller/Result/Redirect.php
+++ b/lib/internal/Magento/Framework/Controller/Result/Redirect.php
@@ -9,6 +9,8 @@ namespace Magento\Framework\Controller\Result;
 use Magento\Framework\App;
 use Magento\Framework\App\Response\HttpInterface as HttpResponseInterface;
 use Magento\Framework\Controller\AbstractResult;
+use Magento\Framework\App\Response\RedirectInterface;
+use Magento\Framework\UrlInterface;
 
 /**
  * In many cases controller actions may result in a redirect
@@ -18,13 +20,14 @@ use Magento\Framework\Controller\AbstractResult;
  */
 class Redirect extends AbstractResult
 {
+
     /**
-     * @var \Magento\Framework\App\Response\RedirectInterface
+     * @var RedirectInterface
      */
     protected $redirect;
 
     /**
-     * @var \Magento\Framework\UrlInterface
+     * @var UrlInterface
      */
     protected $urlBuilder;
 
@@ -37,11 +40,11 @@ class Redirect extends AbstractResult
      * Constructor
      *
      * @param App\Response\RedirectInterface $redirect
-     * @param \Magento\Framework\UrlInterface $urlBuilder
+     * @param UrlInterface $urlBuilder
      */
     public function __construct(
         App\Response\RedirectInterface $redirect,
-        \Magento\Framework\UrlInterface $urlBuilder
+        UrlInterface $urlBuilder
     ) {
         $this->redirect = $redirect;
         $this->urlBuilder = $urlBuilder;
@@ -70,6 +73,7 @@ class Redirect extends AbstractResult
     }
 
     /**
+     * URL Setter
      * @param string $url
      * @return $this
      */
@@ -97,7 +101,11 @@ class Redirect extends AbstractResult
      */
     protected function render(HttpResponseInterface $response)
     {
-        $response->setRedirect($this->url);
+        if (empty($this->httpResponseCode)) {
+            $response->setRedirect($this->url);
+        } else {
+            $response->setRedirect($this->url, $this->httpResponseCode);
+        }
         return $this;
     }
 }
--- a/lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
+++ b/lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
@@ -6,10 +6,13 @@
 
 namespace Magento\Framework\Controller\Test\Unit\Result;
 
-use Magento\Framework\App\Response\HttpInterface as HttpResponseInterface;
+use \PHPUnit\Framework\TestCase;
+use \Magento\Framework\App\Response\HttpInterface as HttpResponseInterface;
+use \Magento\Framework\App\Response\RedirectInterface;
 use \Magento\Framework\Controller\Result\Redirect;
+use \Magento\Framework\UrlInterface;
 
-class RedirectTest extends \PHPUnit\Framework\TestCase
+class RedirectTest extends TestCase
 {
     /** @var \Magento\Framework\Controller\Result\Redirect */
     protected $redirect;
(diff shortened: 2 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11405) <br/>All committers have signed the CLA.

okorshenko:
Hi @gabrielqs-redstage could you please sign CLA?

gabrielqs-redstage:
Signed. Please let me know if you'd like me to refactor it following the codacy suggestion.

Best,
-G

orlangur:
@gabrielqs-redstage please sign CLA using the *gabrielqs* account (GitHub recognized this account as commit author).

lib/internal/Magento/Framework/Controller/Result/Redirect.php
```diff
@@ -93,11 +100,27 @@ public function setPath($path, array $params = [])
...
+            in_array($statusCode, self::REDIRECT_RESPONSE_STATUS_CODES);
+    }
+
+    /**
      * {@inheritdoc}
      */
     protected function render(HttpResponseInterface $response)
     {
-        $response->setRedirect($this->url);
+        if ($this->isValidHttpResponseCode($this->httpResponseCode)) {
```
orlangur: Please replace to `if (!empty($this->httpResponseCode)) {`

as there is already similar check in ZF

```
if ((300 <= $code) && (307 >= $code)) {
            $this->_isRedirect = true;
        } else {
            $this->_isRedirect = false;
        }
```

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
```diff
@@ -67,16 +70,44 @@ public function testSetPath()
...
+     */
+    public function testRender($url, $statusCode, $expectedStatusCode)
+    {
+        $this->redirect->setUrl($url);
+        $this->redirect->setHttpResponseCode($statusCode);
+
+        $this->response
+            ->expects($this->once())
+            ->method('setRedirect')
+            ->with($this->equalTo($url), $this->equalTo($expectedStatusCode));
```
orlangur: `$this->equalTo` calls are not necessary here, please remove.

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
```diff
@@ -67,16 +70,44 @@ public function testSetPath()
...
         );
         $this->assertInstanceOf(
-            \Magento\Framework\Controller\Result\Redirect::class,
+            Redirect::class,
             $this->redirect->setPath($path, $params)
         );
     }
 
-    public function testRender()
+    public function testRenderDataProvider()
```
orlangur: Starting data provider with `test` is not a good idea. Method name should describe data provided and not method under test,

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
```diff
@@ -67,16 +70,44 @@ public function testSetPath()
...
+        return [
+            ['http://test.com', null, 302],
+            ['http://test.com', 0, 302],
+            ['http://test.com', '', 302],
+            ['http://test.com', 301, 301],
+            ['http://test.com', 302, 302],
+            ['http://test.com', 303, 303],
+            ['http://test.com', 200, 302],
+            ['http://test.com', 404, 302],
+            ['http://test.com', 503, 302],
```
orlangur: Too much cases as to me for such a tiny method.

First of all move URL out of data provider. Secondly, it's enough to check
```
null, 302
302, 302
303, 303 
```

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
```diff
@@ -67,16 +70,44 @@ public function testSetPath()
...
+        ];
+    }
+
+    /**
+     * @param string $url
+     * @param mixed $statusCode
```
orlangur: Avoid `mixed` pseudo type.

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
```diff
@@ -67,16 +70,44 @@ public function testSetPath()
...
+     * @param int $expectedStatusCode
+     * @dataProvider testRenderDataProvider
+     */
+    public function testRender($url, $statusCode, $expectedStatusCode)
```
orlangur: Please change order of argument similar to assert: expected, actual

gabrielqs:
@orlangur ah, ok. Done.

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
```diff
@@ -67,16 +70,44 @@ public function testSetPath()
...
+
+        $this->response
+            ->expects($this->once())
+            ->method('setRedirect')
+            ->with($this->equalTo($url), $this->equalTo($expectedStatusCode));
+
         $this->assertInstanceOf(
-            \Magento\Framework\Controller\Result\Redirect::class,
+            Redirect::class,
             $this->redirect->renderResult($this->response)
```
orlangur: Please remove this pretty useless assert and simply call the method instead.

orlangur (changes requested):

orlangur:
@gabrielqs-redstage thanks for making the code much cleaner using aliasing! After doing necessary changes, please commit them with amend and force push to the same branch.

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
gabrielqs-redstage: I did remove it, but are you sure this assert is useless? removing the return from renderResult could possibly lead to breaking things if something outside depends on the return.

gabrielqs-redstage:
@orlangur, all set. please see my comment about the assert removal.

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
orlangur: Mainly because I wouldn't treat `return $this` as class behavior but rather a particular implementation details :) Just assume that for every `return $this` in code we must have an assertion in some test. I believe it would be better to control it with dedicated static test and it matters in unit test only in scope of some bugfix, when return was previously missing.

I do understand your opinion however and removing "useless" (as to me) assert obviously does not deserve a separate commit.

orlangur (approved):

lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php
gabrielqs-redstage: gotcha, thanks for the explanation!

PR #11422, opened 2017-10-13
[Backport 2.2] Translate order getCreatedAtFormatted() to store locale

Jeroenll:
### Description
Backport of version of https://github.com/magento/magento2/pull/10491.

Add `getDefaultStoreLocale()` to allow fetching scoped values. Use this in `getCreatedAtFormatted()` so `created_at` date of order will be translated in emails to locale being used in that store view.

1. None that I could find.

### Manual testing scenarios
1. Set default locale of your store to any language but English.
2. Create order and send confirmation email.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -7,6 +7,8 @@ namespace Magento\Sales\Model;
 
 use Magento\Directory\Model\Currency;
 use Magento\Framework\Api\AttributeValueFactory;
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\Locale\ResolverInterface;
 use Magento\Framework\Pricing\PriceCurrencyInterface;
 use Magento\Sales\Api\Data\OrderInterface;
 use Magento\Sales\Api\Data\OrderStatusHistoryInterface;
@@ -213,6 +215,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $_currencyFactory;
 
     /**
+     * @var \Magento\Eav\Model\Config
+     */
+    private $_eavConfig;
+
+    /**
      * @var \Magento\Sales\Model\Order\Status\HistoryFactory
      */
     protected $_orderHistoryFactory;
@@ -268,6 +275,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $timezone;
 
     /**
+     * @var ResolverInterface
+     */
+    private $localeResolver;
+
+    /**
      * @param \Magento\Framework\Model\Context $context
      * @param \Magento\Framework\Registry $registry
      * @param \Magento\Framework\Api\ExtensionAttributesFactory $extensionFactory
@@ -295,6 +307,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
      * @param \Magento\Framework\Model\ResourceModel\AbstractResource $resource
      * @param \Magento\Framework\Data\Collection\AbstractDb $resourceCollection
      * @param array $data
+     * @param ResolverInterface $localeResolver
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
     public function __construct(
@@ -324,7 +337,8 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
         \Magento\Catalog\Model\ResourceModel\Product\CollectionFactory $productListFactory,
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
-        array $data = []
+        array $data = [],
+        ResolverInterface $localeResolver = null
     ) {
         $this->_storeManager = $storeManager;
         $this->_orderConfig = $orderConfig;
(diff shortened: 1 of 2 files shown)


app/code/Magento/Sales/Model/Order.php
```diff
@@ -213,6 +215,11 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     protected $_currencyFactory;
 
     /**
+     * @var \Magento\Eav\Model\Config
+     */
+    private $_eavConfig;
```
dmanners: If this is a new variable you do not need the `_` at the beginning.

Jeroenll: It isn't a new variable, just a missing phpdoc.

dmanners: In version `2.2-develop` of this file there is no such variable. https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Sales/Model/Order.php It seems be have been added to `develop` in you original PR. I wonder if it is a left over from when the file was created. I will see what I can find out.

Jeroenll: @dmanners I remember, variable was already there but never declared as a property.

https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Sales/Model/Order.php#L338

dmanners: Yup seems to be. My guess is we can remove the property and the mapping but leave the injection for BiC reasons. But I will dig a bit deeper to make sure.

Jeroenll: Sure, ping me if you need any changes done in this PR.

dmanners: Yup so this looks like a left over from the M1 approach of doing things. I cannot find a usage in this class and in the history it was last changed 4 years ago. I would suggest you can remove the new property, remove the assignment in the constructor `$this->_eav` and add the tag `@SuppressWarnings(PHPMD.UnusedFormalParameter)` to the constructor phpdocs.

dmanners (changes requested):

Jeroenll:
@dmanners I removed the variable and added the `@SuppressWarnings(PHPMD.UnusedFormalParameter)`

Jeroenll:
@dmanners Can you proceed merging? Would love to see this released in 2.2.2

app/code/Magento/Sales/Model/Order.php
```diff
@@ -310,7 +318,6 @@ public function __construct(
         \Magento\Catalog\Model\Product\Visibility $productVisibility,
         \Magento\Sales\Api\InvoiceManagementInterface $invoiceManagement,
         \Magento\Directory\Model\CurrencyFactory $currencyFactory,
-        \Magento\Eav\Model\Config $eavConfig,
```
dmanners: you cannot remove the injection via the `__constuct` without this breaking BC but you can remove the setting of `$this->_eavConfig = $eavConfig;` which you have done. If you add back this to the construct and docblock the rest seems fine to me

dmanners (changes requested):

dmanners (approved):

dmanners:
@JeroenVanLeusden while testing this our team found the following.

```
I couldn't see what was fixed here.
Here are the steps I used:
1. Install German thought console:
composer require splendidinternet/mage2-locale-de-de
php bin/magento setup:upgrade
php bin/magento cron:run && php bin/magento indexer:reindex && php bin/magento cache:clean
2. Go to Stores->Configurations->General->Locale Otions->Locale-> German (Germany)
3. Create a product, open storefront and order it as registered customer
4. Check Email of the customer
Expected result: 
Date of order will be translated in emails to locale being used in that store view
Actual result:
There is no difference in Emails
```

It seems that in the testing process both the branch for this PR and the `2.2-develop` branch had the same date information in the resulting emails. Are there any special test cases that should be taken into account?

Jeroenll:
@dmanners Will find some time this week to check this. Will ping you with the results!

dmanners:
Thanks @JeroenVanLeusden if you get stuck and need some help feel free to drop me a message here on in the engcom slack.

Jeroenll:
@dmanners I've reproduced using the following steps:

1. Install NL language pack, `composer require honl/magento2-nl-nl:^1@dev`.
2. Set store language to Dutch (NL_nl).
3. Create order and send confirmation email.
4. Without this patch the date is printed as `Geplaatst op 2017 M10 11 19:56:05 GMT+2` (http://cloud.h-o.nl/nmts).
4. With this patch the date is printed as `Geplaatst op 11 oktober 2017 19:56:05 CEST` (http://cloud.h-o.nl/nn0T).

Using the DE language pack will give me the same results. http://cloud.h-o.nl/nmZE and with patch http://cloud.h-o.nl/nnBu

This is tested from commit d2cbddd1b7ab15ced43de73446efab3468de3b4a

dmanners:
Thank you for the update @JeroenVanLeusden will keep you updated on the progress.

Jeroenll:
@dmanners I've been able to reproduce it using the following steps:
1. Clean install from 2.2-develop branch, commit 2b789948bc2de3cc8267a150003f3c078496e470.
1. Install NL language pack, composer require honl/magento2-nl-nl:^1@dev.
2. Set store language to Dutch (NL_nl).
3. Create order and send confirmation email (date is translated correctly).
4. Login the backend and open new order and press `Send Email` to send the email again.
5. Date is now formatted using US locale.

Issue #6244, opened 2016-08-19
Promo code label not used


valibus:
When setting a promo code label, it is not displayed in cart but the promocode is used instead
### Preconditions
1. version : magento 2.1 public release
2. Setting a promo code with a name : promocodename
3. seeting a code for this promo with a code : truc
4. Setting a default label : labeltest
   ![image](https://cloud.githubusercontent.com/assets/6961470/17805905/0b966f30-6602-11e6-92a1-513c56aeca86.png)
   ![image](https://cloud.githubusercontent.com/assets/6961470/17805913/15b52b96-6602-11e6-8fc8-fc182157c5ed.png)
   ![image](https://cloud.githubusercontent.com/assets/6961470/17805999/6d7246de-6602-11e6-9835-27ffeb44732d.png)
### Steps to reproduce
1. add the promo code to cart, and watch the result
### Expected result

1. in the session message confirmation in green the right label should be displayed
2. in the cart resumee the right label should be displayed
### Actual result

1. in the session message confirmation in green the promo.code appear
2. in the cart resumee the promo.code appear

![image](https://cloud.githubusercontent.com/assets/6961470/17806051/a5122794-6602-11e6-8719-583066825f4d.png)

sevos1984:
Thanks for reporting, this is a known issue, internal tiket is MAGETWO-48435

magento-team:
Internal ticket to track issue progress: MAGETWO-69533

PR #11438, opened 2017-10-13
Issue #11428: Cart Price Rule Label is not working

mrodespin:
### Description

Cart Price Rule Label is not working. The Total label for discounts was not the correct one, after this fix the label was "Discount" and with this change is the Cart Rule Label.

### Steps to reproduce

1. While creating the cart price rule, I have given the Label value as "Buy 2 and Get 1 Free".
2. In front end, Checkout page, It will show only the label as "Discount" not the given name has been reflected.

The following modification fixes this magento 2 issue #11428 

### Contribution checklist

### DIFF
--- a/app/code/Magento/SalesRule/view/frontend/web/template/summary/discount.html
+++ b/app/code/Magento/SalesRule/view/frontend/web/template/summary/discount.html
@@ -7,7 +7,7 @@
 <!-- ko if: isDisplayed() -->
 <tr class="totals discount">
     <th class="mark" scope="row">
-        <span class="title" data-bind="text: title"></span>
+        <span class="title" data-bind="text: getCouponLabel()"></span>
         <span class="discount coupon" data-bind="text: getCouponCode()"></span>
     </th>
     <td class="amount">


adrian-martinez-interactiv4:
I leave here a reference to the related issue: #11428

ghost (changes requested):
We have tested with this solution it's working but it's displayed "Discount" Before label. I think It's should not be display If a label is set. @mrodespin  Please consider covering the updated flow with the test.

mrodespin:
@emiprotech I didn't remove the Discount label because it how it works in previous magento versions.

mrodespin:
@magento-team i seen than you added a needs update label, but i don't know why. What i have to modify??

okorshenko:
Hi @mrodespin we are trying to align your solution with the documentation http://docs.magento.com/m2/ce/user_guide/marketing/price-rules-cart.html

mrodespin:
@okorshenko ok  i  will wait then. Can you add the partner contribution label to my PRs?

okorshenko:
Hi @mrodespin, unfortunately, we can not accept this changes. Please see the comment here: https://github.com/magento/magento2/issues/11428#issuecomment-337990525

pmoreno1980:
@okobchenko here we have 2 situations for this issue (Cart Price Rule Label is not working):

Shopping Cart Rules with Coupon Code and this is the solution for that https://github.com/magento/magento2/issues/6244

Shopping Cart Rules without Coupon Code, i mean automatic discount, we don't have solution for this yet, this #11438 doesn't work.

this is a very important problem that is not solved in Magento 2. we didn't have this issue on Magento 1

thanks for you help

pmoreno1980:
any update here? its a big bug, i know that is fixed on enterprise version.

please let me know

versdivers:
I have the same issue.  Filled in all the labels. Frontend just says 'Discount'

soniclouds:
Any updates on this???  Using 2.4.2 and same exact issue.  Labels not working when applying cart rule without coupon code.

PR #11440, opened 2017-10-13
Add missing translations in Magento_UI

Jeroenll:
### Description
Add missing translations in UI module taken from `mage/validation.js` file.

1. magento/magento2#5956: Untranslatable string "Please enter the same value again."

### Manual testing scenarios
1. Install dutch language pack.
2. Set locale of frontend to dutch.
3. Create new user on the storefront, Fill both password fields with a different value.
4. String should be translated.

### Contribution checklist


fooman:
@JeroenVanLeusden thanks for this pull request. I am currently unable to confirm that this patch does indeed fix the issue.

Couple of questions:

1. Are there any other steps needed to verify the translations works?
I have used `honl/magento2-nl-nl` (which didn't seem to include this term so I added a dummy translation)  and run `bin/magento clear:cache` and have also wiped the pub/static content in developer mode, but am still seeing:

screen shot 2017-10-17 at 11 24 09


2. Is there a reason that this added to the UI module? It seems that at least the string
"Please enter the same value again." is already present [here](https://github.com/magento/magento2/blob/2.2-develop/lib/web/i18n/en_US.csv#L105).

fooman:
I have now tested in production mode which successfully translated
"Please enter the same value again."
for me. However this was without the changes of this PR applied.

Jeroenll:
@fooman Even in production the mode the string isn't translated. Could it be your dummy translations you've added earlier in the process?

I did the following steps to reproduce the issue:
- `composer require honl/magento2-nl-nl`
- Create file `src/lib/web/i18n/nl_NL.csv` and add `"Please enter the same value again.","Vul dezelfde waarde nog een keer in."`
- `php bin/magento setup:static-content:deploy nl_NL -j 1 -f`
- In `src/pub/static/frontend/Magento/luma/nl_NL/js-translation.json` the string isn’t translated.
- In `src/app/code/Magento/Ui/i18n/nl_NL.csv` add `"Please enter the same value again.","Vul dezelfde waarde nog een keer in, Magento_UI"`
- `php bin/magento setup:static-content:deploy nl_NL -j 1 -f`
- In `src/pub/static/frontend/Magento/luma/nl_NL/js-translation.json` the string is translated with the Magento_UI prefix.

I added them to `Magento_UI` since most translations from `mage/validation.js` are located there. The strings located [here](https://github.com/magento/magento2/blob/2.2-develop/lib/web/i18n/en_US.csv#L105) are not used at all.

fooman:
@JeroenVanLeusden can you try the following on your end:

- composer require honl/magento2-nl-nl
- add   
 `"Please enter the same value again.","Vul dezelfde waarde nog een keer in.",module,Magento_Ui`   to `vendor/honl/magento2-nl-nl/crowdin.csv`
- then run `bin/magento setup:static-content:deploy nl_NL en_US -j 1 -f`

After the above I get the below

screen shot 2017-10-25 at 20 42 14

I do believe there is still some oddity somewhere (possibly in the quick deploy strategy) as just running `bin/magento setup:static-content:deploy nl_NL -j 1 -f` did not work.

Jeroenll:
@fooman Following your steps resulted in a translated string. Running `bin/magento setup:static-content:deploy nl_NL -j 1 -f` wil also translate the string in my case. So far so good.

I decided to add them to `Magento_UI` because `php bin/magento i18n:collect-phrases -o "strings.csv" -m` collects the string as `"Please enter the same value again.","Please enter the same value again.",lib,lib/web/mage/validation.js` and when adding that to `vendor/honl/magento2-nl-nl/crowdin.csv` it results in a non translated string.

fooman:
Thanks @JeroenVanLeusden for confirming. So this narrows it down to there being an issue with either the output of collect-phrases not being correct or `lib,lib/web/mage/validation.js` not working correctly when applied.

For now your suggested approach seems a reasonable workaround considering that this seems to have already been done in the past. I'll proceed with merging this one - however can you please take a look at line 203.

app/code/Magento/Ui/i18n/en_US.csv
```diff
@@ -179,3 +179,25 @@ settings/addAfter,settings/addAfter
...
+"Please enter a valid number.","Please enter a valid number."
+"Admin is a required field in the each row.","Admin is a required field in the each row."
+"Please fix this field.","Please fix this field."
+"Please enter a valid date (ISO).","Please enter a valid date (ISO)."
+"Please enter only digits.","Please enter only digits."
+"Please enter the same value again.","Please enter the same value again."
+"Please enter no more than {0} characters.","Please enter no more than {0} characters."
+"Please enter at least {0} characters.","Please enter at least {0} characters."
+"Please enter a value between {0} and {1} characters long.","Please enter a value between {0} and {1} characters long."
+Please enter a value between {0} and {1}.
```
fooman: "Please enter a value between {0} and {1}." Misses second part and quotes

ghost (changes requested):

Jeroenll:
@fooman Thanks for the update. I also corrected line 203.

ghost (approved):

PR #4532, opened 2016-05-16
Virtual products not showing totals in cart

degaray:

Issue #2991, opened 2016-01-13
Products added to cart with REST API give total prices equal to zero


degaray:
### Preconditions (*)

1. Magento 2.3.0(release), 2.3-develop, 2.2-develop

### Steps to reproduce (*)

1.Execute:
```
endpoint="http://m2ce.conf/rest"
shippingEmail="earthworm.jim@gmail.com"

admin_token=$(curl -X POST "$endpoint/V1/integration/admin/token" \
--header "Content-Type: application/json" \
-d '{"username":"admin","password":"123123q"}') && echo $admin_token && admin_token=$(echo $admin_token | tr -d '"')

curl -X POST "$endpoint/V1/categories" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $admin_token" \
-d '{"category":{"parent_id":2,"name":"Category 1","is_active":true,"level":2,"path":"1/2/3","include_in_menu":true,"custom_attributes":[{"attribute_code":"is_anchor","value":"1"},{"attribute_code":"url_key","value":"category-1"},{"attribute_code":"url_path","value":"category-1"}]}}' | json_pp

curl -X POST "$endpoint/V1/categories" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $admin_token" \
-d '{"category":{"parent_id":2,"name":"Category 2","is_active":true,"level":2,"path":"1/2/4","include_in_menu":true,"custom_attributes":[{"attribute_code":"is_anchor","value":"1"},{"attribute_code":"url_key","value":"category-2"},{"attribute_code":"url_path","value":"category-1"}]}}' | json_pp

category_id="3"; productSKU="Simple Product SSM 1" && curl -X POST "$endpoint/V1/products" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer $admin_token" \
-d '{"product":{"sku":"'"$productSKU"'","name":"'"$productSKU"'","attribute_set_id":4,"price":10,"status":1,"visibility":4,"type_id":"simple","created_at":"2018-01-01 00:00:01","updated_at":"2018-01-01 00:01:00","weight":1,"extension_attributes":{"category_links":[{"position":0,"category_id":"'"$category_id"'"}],"stock_item":{"qty":1000,"is_in_stock":true,"is_qty_decimal":false}}}}' | json_pp

category_id="3"; productSKU="Simple Product SSM 2" && curl -X POST "$endpoint/V1/products" \
-H "Content-Type: application/json" \
[...]

PR #11458, opened 2017-10-14
Products added to cart with REST API give total prices equal to zero

peterjaap:
Products added to cart with REST API give total prices equal to zero

### Description
When adding a product trough the (web) API to an empty cart, the product gets a price of zero.

To quote @degaray in https://github.com/magento/magento2/issues/2991#issuecomment-174325730, "The problem is here: `Magento\Quote\Model\Quote\TotalsCollector::collect` because all totals are set to zero and then all totals are calculated per address set. Since no address is set, then no totals are calculated, and as result, it does not calculate any total."

Adding these collect shipping method calls will set a blank address, and cause the price to show up correctly.

1. magento/magento2#2991: Products added to cart with REST API give total prices equal to zero

### Related Issues
1. magento/magento2#4532 Virtual products not showing totals in cart (PR)

### Manual testing scenarios

1. See https://github.com/magento/magento2/issues/2991#issuecomment-174598611

### Contribution checklist

(sorry for the commits mess, apparently I suck at rebasing)

### DIFF
--- a/app/code/Magento/Quote/Model/QuoteManagement.php
+++ b/app/code/Magento/Quote/Model/QuoteManagement.php
@@ -232,7 +232,6 @@ class QuoteManagement implements \Magento\Quote\Api\CartManagementInterface
         $quote->setShippingAddress($this->quoteAddressFactory->create());
 
         try {
-            $quote->getBillingAddress()->setCollectShippingRates(true);
             $quote->getShippingAddress()->setCollectShippingRates(true);
 
             $this->quoteRepository->save($quote);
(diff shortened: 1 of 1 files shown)


okorshenko:
Hi @peterjaap 
Could you please look at broken unit tests?

peterjaap:
@okorshenko I have no idea why the unit test fails; the address is created 2 lines right above where the setCollectShippingRates() is called...

okorshenko:
I will fix the tests

ghost:
@peterjaap I experienced the same bug with a logged in customer. In that case it uses the "createEmptyCartForCustomer" method. Your fix also works for that method.

PR #11462, opened 2017-10-14
#7241: Always add empty option for prefix and/or suffix if optional

avstudnitz:
### Description
This solves part of #7241. If the prefix and/or suffix field is set to "optional" in the configuration, we automatically add a blank first option if it isn't entered in the "Dropdown Options" configuration field. 

1. magento/magento2#7241: No option to start with blank option for prefix and suffix in checkout. (partly fixed)

### Manual testing scenarios

1. Go to Admin panel > Stores > Configuration > Customers > Customer Configuration
2. Change "Show Prefix" and/or "Show Suffix" to "Optional"
3. Populate "Prefix Dropdown Options" with : "Ms;Miss;Mrs;Mr;Dr;Atty;Hon;Prof;Pres;VP;MA"
4. Populate "Suffix Dropdown Options" with : "Jr;Sr;Ph.D.;M.D.;R.N.;Ret."
5. Click Save.
6. Switch to the front-end, and add something to your cart.
7. Go to the checkout page or customer registration page
8. You'll notice that there is an empty first option for the Prefix and Suffix field.

### Contribution checklist


app/code/Magento/Customer/Model/Options.php
```diff
@@ -53,23 +57,31 @@ public function getNamePrefixOptions($store = null)
...
     /**
      * Unserialize and clear name prefix or suffix options
+     * If field is optional, add an empty first option.
      *
      * @param string $options
+     * @param bool $isOptional
      * @return array|bool
      */
-    protected function _prepareNamePrefixSuffixOptions($options)
+    private function prepareNamePrefixSuffixOptions($options, $isOptional = false)
```
dmanners, line 87: As part of the BC  guidelines you cannot remove a protected method http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#public-and-protected-method-removal I would suggest that you keep the private method and the internal calls to this private method and then simply add back the protected method that simply calls the private method and mark it as deprecated.

dmanners (changes requested):

dmanners:
@avstudnitz do you think you will have some time to continue this PR?

avstudnitz:
Yes, I will, probably next week.

avstudnitz:
I re-added the protected method as you said @dmanners.

dmanners:
Thanks @avstudnitz I will take another look over this PR for you.

app/code/Magento/Customer/Model/Options.php
```diff
@@ -53,23 +57,43 @@ public function getNamePrefixOptions($store = null)
...
+            $this->addressHelper->getConfig('suffix_show', $store) == NooptreqSource::VALUE_OPTIONAL
+        );
+    }
+
+    /**
+     * @param $options
+     * @param bool $isOptional
+     * @return array|bool
+     *
+     * @deprecated See prepareNamePrefixSuffixOptions
```
dmanners: You should also be able to use http://docs.phpdoc.org/references/phpdoc/tags/see.html here to link to the new method. Something like the following should work.

```
@see prepareNamePrefixSuffixOptions()
```

dmanners (changes requested):

app/code/Magento/Config/Model/Config/Source/Nooptreq.php
```diff
@@ -11,15 +11,19 @@
  */
 class Nooptreq implements \Magento\Framework\Option\ArrayInterface
 {
+    const VALUE_NO = '';
```
dmanners, line 14: Sadly while putting this to the `2.2-develop` branch and with this class marked as `@api` we cannot add any new constants. For a PR to `2.3-develop` we could add these new constants. For this PR we will have to change this back.

app/code/Magento/Customer/Model/Options.php
```diff
@@ -42,7 +43,10 @@ public function __construct(
      */
     public function getNamePrefixOptions($store = null)
     {
-        return $this->_prepareNamePrefixSuffixOptions($this->addressHelper->getConfig('prefix_options', $store));
+        return $this->prepareNamePrefixSuffixOptions(
+            $this->addressHelper->getConfig('prefix_options', $store),
+            $this->addressHelper->getConfig('prefix_show', $store) == NooptreqSource::VALUE_OPTIONAL
```
dmanners, line 48: Unfortunately we will need to change this since we cannot add the new constants.

app/code/Magento/Customer/Model/Options.php
```diff
@@ -53,23 +57,43 @@ public function getNamePrefixOptions($store = null)
      */
     public function getNameSuffixOptions($store = null)
     {
-        return $this->_prepareNamePrefixSuffixOptions($this->addressHelper->getConfig('suffix_options', $store));
+        return $this->prepareNamePrefixSuffixOptions(
+            $this->addressHelper->getConfig('suffix_options', $store),
+            $this->addressHelper->getConfig('suffix_show', $store) == NooptreqSource::VALUE_OPTIONAL
```
dmanners, line 62: Unfortunately we will need to change this since we cannot add the new constants.

dmanners (changes requested):

dmanners:
@avstudnitz have you had anytime to look into this PR further? There is still the part about showing the empty value in the check and also the backwards compatibility changes also. Thanks

app/code/Magento/Config/Model/Config/Source/Nooptreq.php
avstudnitz: Can you explain why adding new constants is an issue? It doesn't affect backwards compatibility and only improves things, I don't see any disadavantage of that... sounds like a quite arbitrary rule to me, but perhaps it gets clearer if you explain.

app/code/Magento/Customer/Model/Options.php
avstudnitz: Changed, makes sense.

avstudnitz:
@dmanners I made the requested changes. Would be glad about an explanation of the forbidden constants.
This PR doesn't cover preselecting the empty value in the checkout as I wasn't able to do that (see https://github.com/magento/magento2/issues/7241#issuecomment-336636441). It still improves UX and makes the core better even if it doesn't solve all of #7241.

dmanners:
@avstudnitz thank you for the update. I will see if I can get someone to help out with the front-end part, it would be nice to have both together.

With regards to the constant addition. I also agree that it is worth having in this case. Generally speaking because constants are public and the class is marked as `@api` the public face of the class is changing and people need to be able to rely on this. In this case I will talk with the team and see if we can get it approved.

Could you revert the revert commit and then I can ask from some approval here. Sorry for the pain.

avstudnitz:
Thanks @dmanners, I reverted the commit. I agree that removing or changing public elements of the class would be bad, but I don't see the point for newly added functions or constants because they can be ignored. Would be great if you could get approval for that.

dmanners:
@avstudnitz still trying to sort out that front end part for you. I will aim to get around to this PR in the next week. Sorry for the delay.

app/code/Magento/Config/Model/Config/Source/Nooptreq.php
okorshenko: Hi @dmanners in general, you are right. But in this particular, we can accept new constants. There are some cases when new constants can break client code, but the chances a very low.
@dmanners @avstudnitz thank you for collaboration

dmanners (approved):

Issue #11350, opened 2017-10-10
Infinite redirect issue when switching to https 


scholtz:
### Preconditions

1. Magento 2.2, Magento 2.3-dev
2. Any version of PHP
3. Webserver that supports Strict-Transport-Security

### Steps to reproduce
1. Install magento using only http
2. Set up HTTPS on webserver
3. Test webserver address
4. Go to magento administration

### Expected result
1. Magento login screen is displayed

### Actual result
1. Infinite redirect loop

It is because webserver redirects user to https version, and magento redirects user to http version.

I am aware of the fact that user can open it in other browser that does not have the https version cached yet, but I believe that Magento should not allow inifinite redirects. It is quite easy to check on server side if there is infinite redirect loop, by storing the address to session and on next redirect checking this address.. Even more you can make a counter, and show warning on third same redirect..

Also note that it might be wise not to redirect user to http if they already use https even if http is set as magento default by config.

orlangur:
> It is because webserver redirects user to https version, and magento redirects user to http version.

Seems to be an incorrect server setup to me. Magento is not supposed to check if you configured server properly with some hacks.

scholtz:
Let me opose.. The server is not set up incorrectly.. Use of http after Use of https is severe security issue!

scholtz:
Also infinite redirect is huge magento flaw.. What if developer by mistake redirects user in a loop? Eshop owners will loose clients because they will not get any reasonable response from eshop?

orlangur:
Just checked what https://www.google.com/search?q=Strict-Transport-Security&oq=Strict-Transport-Security&sourceid=chrome&ie=UTF-8 is.

> Install magento using only http

You shouldn't do this if you enabled such feature of your web server.

> Also infinite redirect is huge magento flaw.. 

In your scenario it is not caused by Magento. It is caused by incompatible web server setup.

> What if developer by mistake redirects user in a loop? 

There is internal counter disallowing more than 100 internal redirects. External redirects are out of control by Magento, just make an appropriate web server setup.

scholtz:
i have made PR request.. hopefully there is someone else then  orlangur  to check it up

just to summarize what i found during fixing this issue..

the behavior of not transferring to http from https stream is already coded in the main url class, but magento administration only verifies 2 parameters.. if secure_is_forced parameter is checked, then the connection will be secured. When web/secure/use_in_adminhtml parameter is not set it will use http connection, and will lead to infinite redirect loop.

[...]

Issue #11264, opened 2017-10-06
Magento 2.2.0 does not support nullable return types on PHP 7.0.24


esimov:
### Preconditions

1. Magento 2.2.0
2. PHP 7.0.24

### Steps to reproduce

Checking the system requirements for Magento 2.2.0 on http://devdocs.magento.com/guides/v2.2/install-gde/system-requirements-tech.html#php it shows that PHP 7.0 is supported, however the VarTag.php file under vendor/zendframework/zend-code/src/Reflection/DocBlock/Tag/ contains a few methods which returns nullable return types. Like the methods below:
```php
public function getVariableName() : ?string
{
    return $this->variableName;
}

public function getDescription() : ?string
{
    return $this->description;
}
```

The nullable return types are not yet supported on PHP7.0. They were introduced from PHP7.1 on. This means that the actions using the methods above (like the login action) thrown syntax error. Please correct the system requirements.

orlangur:
Strange that it wasn't revealed by integration tests. Is this code actually used in Magento 2? Could you provide some scenario?

Looks like `3.1.0` release of `Zend_Code` didn't support PHP 7.1: https://github.com/zendframework/zend-code/blob/release-3.1.0/composer.json#L16
While `3.2.0` release does not support PHP 7.0: https://github.com/zendframework/zend-code/blob/release-3.2.0/composer.json#L16

Thus there are only two options for Magento 2 to solve the problem directly:
- drop PHP 7.0 support
- maintain separate composer files for PHP 7.0 and PHP 7.1

Indirect solution is to fully cover all code which relies on `Zend_Code` with integration tests so that PHP 7.0 workability is preserved.

orlangur:
According to @ytorbyk we can observe problem with following scenario:

1. Use PHP 7.1
2. Create new Magento project via Composer
3. Try performing DI compilation
4. Exception observed due to Magento compiler not supporting nullable `?string` notation

If we compile using PHP 7.0, it will work just fine under PHP 7.1 then.

orlangur:
Just found out Magento is using `3.1.0` version of `zend-code`: https://github.com/magento/magento2/blob/2.2/composer.lock#L2022

So, there is no problem with PHP 7.0 compatibility, just that this component does not officially support PHP 7.1 but maybe it does not affect Magento.

Those who somehow installed `3.2.0` version of `zend-code` seems to be in trouble though.

esimov:
[...]

PR #11467, opened 2017-10-15
FIX Strict-Transport-Security bug in magento backend

scholtz:
this commit fixes the infinite redirect loop in magento administration when user tries to update from non http to https.

default magento behavior is that if user opens https connection, it uses the https stream. If magento administration does not contain any configuration wheater to use https, only http is used, and redirection by browser does inifinite redirection. If magento config does not setup strict https (because with upgrade the https is not enabled at first) magento redirects administrator from https stream to http which is Strict-Transport-Security bug.

this PR fixes issue #11350 and  #11356 

please do not let @orlangur to process this one, because he believes that inifinite redircts in magento are not an issue, and he believes that magento behavior against Strict-Transport-Security is not an issue. Please note that it is severe issue, and should be fixed ASAP.

### Description
Magento should not redirect https stream to http

1. magento/magento2#11350
2. magento/magento2#11356 

### Manual testing scenarios
1. Set up magento using only http
2. Allow https in web browser
3. This issue fixes the issue that user then cannot get to admin login screen because he gets inifinite redirect loop

### Contribution checklist
yes
yes
not possible because it requires changes in web browser
todo

### DIFF
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -168,7 +168,7 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
             return true;
         }
 
-        // backend specific checks 
+        // backend specific checks
         // if ($this->hasData('secure_is_forced') already processed in parent class
 
         if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -167,9 +167,9 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
         if ($this->hasData('secure_is_forced')) {
             return $this->getData('secure');
         }
-		if($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')){
-			return $this->getData('secure');
-		}
+        if($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')){
+            return $this->getData('secure');
+        }
         return parent::_isSecure();
     }
 
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -164,13 +164,24 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
      */
     protected function _isSecure()
     {
-        if ($this->hasData('secure_is_forced')) {
-            return $this->getData('secure');
-        }
-        if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
+        if (true === parent::_isSecure()) {
             return true;
         }
-        return parent::_isSecure();
+
+        // backend specific checks 
+        // if ($this->hasData('secure_is_forced') already processed in parent class
+        
+        if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
+            if ($this->getRouteParamsResolver()->hasData('secure')) {
+                // we still respect options provided programatically as for secure request in parent class
+                return (bool) $this->getRouteParamsResolver()->getData('secure');
+            }    
+            return true;
+        } else {
+            // if web/secure/use_in_adminhtml == 0 that means we do not care but not that HTTP should be used
+        }
+
+        return false;
     }
 
     /**
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -158,6 +158,7 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
 
     /**
      * Retrieve is secure mode for ULR logic
+     * If secure mode is not set up, use default magento behavior
      *
      * @return bool
      */
@@ -166,7 +167,10 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
         if ($this->hasData('secure_is_forced')) {
             return $this->getData('secure');
         }
-        return $this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml');
+		if($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')){
+			return $this->getData('secure');
+		}
+        return parent::_isSecure();
     }
 
     /**
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -168,7 +168,7 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
             return $this->getData('secure');
         }
         if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
-            return $this->getData('secure');
+            return true;
         }
         return parent::_isSecure();
     }
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -170,12 +170,12 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
 
         // backend specific checks 
         // if ($this->hasData('secure_is_forced') already processed in parent class
-        
+
         if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
             if ($this->getRouteParamsResolver()->hasData('secure')) {
                 // we still respect options provided programatically as for secure request in parent class
                 return (bool) $this->getRouteParamsResolver()->getData('secure');
-            }    
+            }
             return true;
         } else {
             // if web/secure/use_in_adminhtml == 0 that means we do not care but not that HTTP should be used
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -167,7 +167,7 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
         if ($this->hasData('secure_is_forced')) {
             return $this->getData('secure');
         }
-        if($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')){
+        if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
             return $this->getData('secure');
         }
         return parent::_isSecure();
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -164,12 +164,16 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
      */
     protected function _isSecure()
     {
-        if (true === parent::_isSecure()) {
+        if ($this->_request->isSecure()) {
+            if ($this->getRouteParamsResolver()->hasData('secure')) {
+                return (bool) $this->getRouteParamsResolver()->getData('secure');
+            }
             return true;
         }
 
-        // backend specific checks
-        // if ($this->hasData('secure_is_forced') already processed in parent class
+        if ($this->getRouteParamsResolver()->hasData('secure_is_forced')) {
+            return (bool) $this->getRouteParamsResolver()->getData('secure');
+        }
 
         if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
             if ($this->getRouteParamsResolver()->hasData('secure')) {


app/code/Magento/Backend/Model/Url.php
```diff
@@ -166,7 +167,10 @@ protected function _isSecure()
         if ($this->hasData('secure_is_forced')) {
             return $this->getData('secure');
         }
-        return $this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml');
+        if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
+            return $this->getData('secure');
```
orlangur: Here it must return `true`, note the previous line. It looks like for your particular case you can use the same `secure_is_forced` flag.

app/code/Magento/Backend/Model/Url.php
```diff
@@ -166,7 +167,10 @@ protected function _isSecure()
...
+        }
+        return parent::_isSecure();
```
orlangur: This is a _really_ risky fallback. Please provide exact scenario you trying to fix with it, as now it is pretty easy to get a non-secure admin open for MITM attacks and such.

orlangur:
>  inifinite redircts in magento are not an issue, and he believes that magento behavior against Strict-Transport-Security is not an issue. Please note that it is severe issue, and should be fixed ASAP.

That's true, infinite redirects in case of webserver misconfiguration is not a Magento issue.

Previously you suggested some counter of redirects on Magento side just to overcome incorrect webserver configuration which does not seem to be a good idea to me. Proposed PR does not implement such idea and does not look "hacky" in general. I have no objections against such changes if they do not break some of the existing scenarios related to backend security.

Not assigning to myself to make it explicit that somebody other than me should process this PR as author requested.

app/code/Magento/Backend/Model/Url.php
scholtz: I iam not going to provide any more scenarios.. now there is a vulnerability in magento, and this fixes it.. Please someone other make look at it or think again.. if you are not wise ehough to understand why it should be merged to magento, i am very disappointed on how magento picks human resources :(

app/code/Magento/Backend/Model/Url.php
```diff
@@ -166,7 +167,10 @@ protected function _isSecure()
         if ($this->hasData('secure_is_forced')) {
             return $this->getData('secure');
         }
-        return $this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml');
+        if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
```
fooman: this change would currently make the setting `web/secure/use_in_adminhtml` = 0 ineffective

scholtz: yes, exactly.. if he sets not to use secure connection, but accesss it with secure connection, secure connection should be better then non secure connection.. 
when he sets non secure connection at the first magento setup, and then he wants to secure it, how should he proceed? 1) if he sets up this option prior setting web browser, magento admin will not work any more 2) if he sets up this option after setting web browser to https, then magento will try to redirect him to http, and he gets infinite redirects, so magento will not work.. the only option is to set it up at once, but really?? what if something goes not well?

app/code/Magento/Backend/Model/Url.php
```diff
@@ -166,7 +167,10 @@ protected function _isSecure()
...
+            return true;
```
scholtz: @orlangur are you sure? becase the previous if returns $this->getData('secure'); .. it may be some kind of object what is expected from this function..

app/code/Magento/Backend/Model/Url.php
scholtz: or do you want to create new option "web/secure/force_unsecure"? :)

app/code/Magento/Backend/Model/Url.php
scholtz: i checked the parent class and it also returns mostly true/false, so to return true is probably best solution

orlangur: > it also returns mostly true/false

omg. maybe just return true or false randomly then?

implementation must be changed so that `web/secure/use_in_adminhtml = 0` is still effective.

scholtz: @orlangur please stop commenting my issues please.. i am really angry on you because you are not wise enough to understand that there is a security bug in magento and web/secure/use_in_adminhtml = 0 is causing it.. Also your opinion that infinite redirects is not magento issue is crazy. You have told that you will leave this PR to the other developers, and still you are blocking the release..

Question to someone else: What needs to be done for web/secure/use_in_adminhtml = 0 to get obsolete in favor of fixing the security bug in magento?

orlangur: There is no "security bug" if Magento behaves exactly how you configured it. You cannot simply remove a part of functionality making `web/secure/use_in_adminhtml = 0` ineffective.

app/code/Magento/Backend/Model/Url.php
```diff
@@ -166,7 +167,10 @@ protected function _isSecure()
         if ($this->hasData('secure_is_forced')) {
             return $this->getData('secure');
```
scholtz: Shoud not this return just true?

app/code/Magento/Backend/Model/Url.php
scholtz: You do not understand even written text that i do NOT want reply from YOU? Is Vlad Veselov only magento developer? If so, i am leaving magento forever..

app/code/Magento/Backend/Model/Url.php
orlangur: > when he sets non secure connection at the first magento setup, and then he wants to secure it, how should he proceed?

I already answered on this here: https://github.com/magento/magento2/issues/11356#issuecomment-335609411

> what if something goes not well?

You can apply recipe from my comment as many times as you want.

app/code/Magento/Backend/Model/Url.php
orlangur: It does not work like that. During code review valuable feedback from *any* Community member should be incorporated.

scholtz: so stop spamming, and let other community members tell their review.. i believe that you have expressed enough your intelligence quotient at the thread..

maksek:
Hi @scholtz , interesting point of view regarding this issue. Would you mind to join community engineering channel, so we can have chat about possible fixes?

scholtz:
@maksek How can I join community engineering channel?

maksek:
@scholtz - send email request to engcom[@]magento.com and I will add you.

vkublytskyi:
@scholtz thank you for this contribution and your insistence in strive to make Magento better.

I do agree with you that described behavior is an issue or at least legacy implementation that requires improvement.

Magento as a modern software should follow best practices and one of them is encouraging of HTTPS usage over HTTP. That means that Magento should not redirect a user to HTTP if HTTPS is used. To provide more control `web/secure/use_in_adminhtml` may be used by Magento admin to force HTTPS usage regardless web server configuration. But if this option is not switched on it does not mean HTTP should be forced (Magento should not change protocol). This rule should be applied for storefront and for admin.

Regarding provided fix I think it will be better to check parent implementation before custom logic:
```php
if (true === parent::_isSecure()) {
    return true;
}

// backend specific checks 
// if ($this->hasData('secure_is_forced') already processed in parent class

if ($this->_scopeConfig->isSetFlag('web/secure/use_in_adminhtml')) {
    if ($this->getRouteParamsResolver()->hasData('secure')) {
          // we still respect options provided programatically as for secure request in parent class
          return (bool) $this->getRouteParamsResolver()->getData('secure');
    }
    return true;
} else {
    // if web/secure/use_in_adminhtml == 0 that means we do not care but not that HTTP should be used
}

return false;
```

Unfortunately, the root of the issue is in the legacy implementation of `Magento\Framework\Url` which depends on HTTP request and PHP session. Ideally, we should have clear URL model that will depend only on a Magento configuration (domain model layer). A result of this model should be passed to another service (application layer) that will modify it corresponding to current user session and request. Providing such implementation would require deep refactoring and backward incompatible changes so hardly we can make them now. Because of this, any fix would not be perfect but some compromise.

orlangur:
@vkublytskyi,

> That means that Magento should not redirect a user to HTTP if HTTPS is used

Not really as with http-only and secure cookie flags no sensitive information can be sent over HTTP. Generally there were some plans to get rid of mixed (http+https) mode at all, until then some odd scenarios may still exist thus making changes in current implementation may be not worth efforts.

Thanks for a clever suggestion, in such form the fix may really do the trick and does not break possibility to fully configure Magento.

orlangur:
@vkublytskyi looking at the code of `parent::_isSecure()` I don't think we can use it like this.

```
if (!$this->_getScope()->isUrlSecure()) {
            return false;
        }

        if (!$this->getRouteParamsResolver()->hasData('secure')) {
            if ($this->_getType() == UrlInterface::URL_TYPE_LINK) {
                $pathSecure = $this->_urlSecurityInfo->isSecure('/' . $this->_getActionPath());
                $this->getRouteParamsResolver()->setData('secure', $pathSecure);
            } elseif ($this->_getType() == UrlInterface::URL_TYPE_STATIC) {
                $isRequestSecure = $this->_getRequest()->isSecure();
                $this->getRouteParamsResolver()->setData('secure', $isRequestSecure);
            } else {
                $this->getRouteParamsResolver()->setData('secure', true);
            }
        }
```

these lines will ruin admin configurability making frontend-related options taking precedence over admin-related.

Do we need any logic from parent class except 
```
if ($this->_request->isSecure()) {
            return true;
        }
```
?

scholtz:
the CI passes.. @vkublytskyi please merge it in. i have looked also to parent class, and any setting that could force https for frontend should force https to admin area as the admin area should be more secure than the store.. Vlad is narrowminded and first he was against using https at all, later he thought that the solution to call parent function is briliant, and later he wants only to use the if($this->_request->isSecure()) {return true;}  .. in either case thank god tha vlad does not see that this option makes the option web/secure/use_in_adminhtml == 0 unusable.. I think that someone should modify the documentation accordingly..

orlangur:
> that the solution to call parent function is briliant

Until I looked into the code :) Making http/https admin/storefront unconfigurable is generally not a good idea, those options shouldn't affect each other in unpredictable manner, I hope all combinations and corner cases will be tested properly during PR processing. Otherwise bugs may be reported and this change will be reverted.

@scholtz please squash all changes into a single commit and force push to the same branch.

vkublytskyi:
@scholtz, I think in this case @orlangur is right. I made my proposal based on interface declaration but in fact `Magento\Framework\Url\ScopeInterface::isUrlSecure` will fallback to `Magento\Store\Model\Store::isFrontUrlSecure`. I agree with you that in real life it is not a reasonable configuration to make HTTPS enabled for storefront but disabled for admin. But while Magento provides to separate configuration options they should not have an impact on each other. As an alternative, we may disable `Use Secure URLs in Admin` switcher if `Use Secure URLs on Storefront` is set to `Yes` but this change requires product owners approval as is product behavior change.

Usually not to call parent method and copying code is a terrible practice but in this case looks like we do not have another way without huge refactoring and backward incompatible change.

So I think to avoid undesirable side effects instead of 
```php
if (true === parent::_isSecure()) {
    return true;
}
```
we should copy first part from parent class
```php
if ($this->_request->isSecure()) {
    if ($this->getRouteParamsResolver()->hasData('secure')) {
        return (bool) $this->getRouteParamsResolver()->getData('secure');
    }
    return true;
}

if ($this->getRouteParamsResolver()->hasData('secure_is_forced')) {
    return (bool) $this->getRouteParamsResolver()->getData('secure');
}
```

@scholtz please also remove unnecessary comments from implementation and combine all changes into a single commit (you can safely use `git push -f`).

scholtz:
@vkublytskyi "combine all changes into a single commit " how can i do it though web interface?

vkublytskyi:
@scholtz I'm not sure that this is possible with web interface. I do it with git CLI and force push of head branch of the PR. As an alternative, you may close this PR and create a new one with final changes.

scholtz:
@vkublytskyi to which branch do you want to make new PR?

orlangur:
@scholtz better force push to the same `patch-3`. Otherwise new branch name is insignificant.

vkublytskyi:
@scholtz `2.2-develop` is the default and is preferable.

orlangur:
@scholtz are you going to continue work on this pull request?

scholtz:
@orlangur Go on and merge it.. (i give u persmission to create new PR with this content) i am very busy, and dont have time to work on it.. Because someone put there label duplicate and non issue, i thought that someone already took care of it.. 

Also it seems to me, that you are blocking the release by "Changes requested" even these changes has been filled in..

i was also highly demotivated by your centsorship at issue https://github.com/magento/magento2/issues/11264  .. you have closed commenting the issue by your intention even though you were not right.. You do not accept feedback, and are trying to hide your mistakes no matter what other people will think about magento.. ("Whoever face with this issue will notice parse error immediately" .. this is lie and you should remember it from previous communication)

Also i asked you not to communicate with me, and repeatingly you are behaving as you do not understand..

You may be good developer, but you should not work on communication with clients (developers)..

orlangur:
Hello @scholtz, thanks for the quick feedback.

> Also i asked you not to communicate with me, and repeatingly you are behaving as you do not understand..

I really doubt pull request / issue processing is supposed to work this way. In particular such attitude does not look like "Showing empathy towards other community members" in my eyes.

https://help.github.com/articles/blocking-a-user-from-your-personal-account/ could be helpful in your case probably.

Thanks for all your endeavors and good luck.

orlangur:
Closing due to inactivity. Feel free to reach any of us out anytime later if you wish to continue work on this pull request and it will be reopened.

PR #11473, opened 2017-10-15
Fix for remove 'product_list_toolbar' block from layout in XML #9413

mariuscris:
Fix for remove 'product_list_toolbar' block from layout in XML #9413

### Description
This solution solves the issue when you want to not have toolbar block set on product listing.  The reason why toolbar persisted even after that was explicit removed from layout configuration, it's because the method Magento\Catalog\Block\Product\ListProduct:getToolbarBlock always returns a toolbar block.

I think the main reason was build to return a toolbar block is because the product collection needs to use pagination.  But there are cases when you have listing with less then 20 products and you want to not have toolbar in place. In that case you will be forced to create a custom product listing block class because you cannot remove it from the layout.

The product list use modes for listing and this depends on toolbar block to get that information. Because the toolbar block has a default block, this fix will use this block to get the default mode. You can change the listing mode on layout like this:

```
<referenceBlock name="category.products.list">
            <action method="setMode">
                <argument name="name" xsi:type="string">list</argument>
            </action>
</referenceBlock>
``` 

1. magento/magento2#9413: Cannot remove product_list_toolbar in XML

### Manual testing scenarios
1. Create Magento_Catalog/layout/catalog_category_view.xml in a custom theme
2. Add the following code:
```
<?xml version="1.0"?>
    <body>
            <referenceBlock name="product_list_toolbar" remove="true"/>
        </referenceContainer>
    </body>
</page>
```
3. Activate the custom theme
4. Open a category page with products


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11473) <br/>All committers have signed the CLA.

app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -33,7 +37,7 @@ class ListProduct extends AbstractProduct implements IdentityInterface
     /**
      * Product Collection
      *
-     * @var AbstractCollection
+     * @var \Magento\Eav\Model\Entity\Collection\AbstractCollection
```
orlangur: FQCN must not be added.

dev/tests/integration/testsuite/Magento/Catalog/Block/Product/ListTest.php
```diff
@@ -55,6 +55,10 @@ public function testToolbarCoverage()
         $parent = $this->_getLayout()->createBlock(\Magento\Catalog\Block\Product\ListProduct::class, 'parent');
 
         /* Prepare toolbar block */
+        $this->_getLayout()
+            ->createBlock(\Magento\Catalog\Block\Product\ProductList\Toolbar::class, 'product_list_toolbar');
+        $parent->setToolbarBlockName('product_list_toolbar');
```
orlangur, line 60: Why need this logic? According to my understanding existing integration tests should work without any changes, otherwise existing Magento instances may be broken by this PR.

app/code/Magento/Catalog/Block/Product/ListProduct.php
mariuscris: I will remove the FQCN.

dev/tests/integration/testsuite/Magento/Catalog/Block/Product/ListTest.php
mariuscris: My main reason why I added the above code for toolbar coverage, it was to update with the correct setup for product listing block. The toolbar block needs to be set to a product listing block only from the layout and the test will check if content for the toolbar is not empty. 

The previous implementation used a default toolbar block class, in case this was not provided in the layout, which created the  problem  when you didn't  want to have a toolbar on the product listing. 

The normal way, I think, is to have toolbar declared on layout which in this case you need to have a test that sets the toolbar block in the product listing block and check if it's ok.

I totally agree with you about the tests should work without with any changes, but in my opinion, for this case, the test was used in a situation where the product listing block provided a toolbar block. The product listing will use a toolbar only if it's provided from layout. If you use a default toolbar, we are back to the problem reported, meaning you can not have a product listing without a toolbar block. 

Please let me know your opinion on this.

orlangur: Such change looks backward-incompatible to me for those who previously relied on toolbar block created automatically. Maybe more suitable for `2.3` but let's give somebody else to look into it.

mariuscris: Yes, it will create a backward-incompatibility for those who relay on toolbar block created automatically, but I do not see how the toolbar can be removed without creating your custom product listing block where you will be forced to remove it.

app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -7,13 +7,17 @@
 namespace Magento\Catalog\Block\Product;
 
 use Magento\Catalog\Api\CategoryRepositoryInterface;
+use Magento\Catalog\Block\Product\Context;
```
omiroshnichenko: Import of Magento\Catalog\Block\Product\Context is not necessary.

app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -7,13 +7,17 @@
...
 use Magento\Catalog\Block\Product\ProductList\Toolbar;
 use Magento\Catalog\Model\Category;
```
omiroshnichenko: Import of Magento\Catalog\Model\Category is never used.

app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -40,17 +44,17 @@ class ListProduct extends AbstractProduct implements IdentityInterface
     /**
      * Catalog layer
      *
-     * @var \Magento\Catalog\Model\Layer
+     * @var Layer
```
omiroshnichenko, line 52: Undefined class Layer, use FQCN.

orlangur: Disagree, why not add a proper import instead?

omiroshnichenko: Or add proper import:)

app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -233,10 +302,10 @@ protected function _getConfig()
     /**
      * Prepare Sort By fields from Category Data
      *
-     * @param \Magento\Catalog\Model\Category $category
-     * @return \Magento\Catalog\Block\Product\ListProduct
+     * @param Category $category
+     * @return $this
      */
-    public function prepareSortableFieldsByCategory($category)
+    public function prepareSortableFieldsByCategory(Category $category)
```
omiroshnichenko: Method signature is changed. This change breaks backward compatibility. Please, revert this.

app/code/Magento/Catalog/Block/Product/ListProduct.php
mariuscris: Solved. Thank you for your review.

app/code/Magento/Catalog/Block/Product/ListProduct.php
mariuscris: This is used on  prepareSortableFieldsByCategory method for $category parameter.

app/code/Magento/Catalog/Block/Product/ListProduct.php
mariuscris: Solved. Thank you for your review.

app/code/Magento/Catalog/Block/Product/ListProduct.php
mariuscris: I will revert this change. I added this change because  if $category is not an Category object, it will crash at calling this code  $category->getAvailableSortByOptions() if there is no available orders set on.

PR #11475, opened 2017-10-15
FR#ProductAlertsCron_22 Magento Product Alerts cron is not generated …

mrodespin:
Magento 2 product alerts cron configuration was not working. When a administrator wants to enable the functionality of product alerts the frequency was not saving in cron_schedule table.

### Description
The commit solves the issue on saving magento 2 automatic product alerts configuration. With these modifications the cron job is working properly.

### Manual testing scenarios
1.  go to admin>store>configurtion>catalog>catalog
2. turn on generation and set up time.
3. Execute php bin/magento cron:run
3. Go to cron_schedule table and the cron job catalog_product_alert is not in the table

With this modification if we do the same process the cron job appears in the table (If the time 
configured is nearby)

### Related PRs
#11567

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cron/Model/Config/Backend/Product/Alert.php
+++ b/app/code/Magento/Cron/Model/Config/Backend/Product/Alert.php
@@ -82,22 +82,14 @@ class Alert extends \Magento\Framework\App\Config\Value
         $cronExprString = join(' ', $cronExprArray);
 
         try {
-            $this->_configValueFactory->create()->load(
-                self::CRON_STRING_PATH,
-                'path'
-            )->setValue(
-                $cronExprString
-            )->setPath(
-                self::CRON_STRING_PATH
-            )->save();
-            $this->_configValueFactory->create()->load(
-                self::CRON_MODEL_PATH,
-                'path'
-            )->setValue(
-                $this->_runModelPath
-            )->setPath(
-                self::CRON_MODEL_PATH
-            )->save();
+            $cronStringPath = $this->_configValueFactory->create()->load(self::CRON_STRING_PATH, 'path');
+            $cronStringPath->setValue($cronExprString);
+            $cronStringPath->setPath(self::CRON_STRING_PATH);
+            $cronStringPath->save();
+            $cronModelPath = $this->_configValueFactory->create()->load(self::CRON_MODEL_PATH, 'path');
+            $cronModelPath->setValue($this->_runModelPath);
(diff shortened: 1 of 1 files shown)


dmanners:
@mrodespin thanks for this PR. I will have a look at this and start to process it. If I need any feedback then I will contact you.

app/code/Magento/Cron/Model/Config/Backend/Product/Alert.php
```diff
@@ -82,22 +82,14 @@ public function afterSave()
         $cronExprString = join(' ', $cronExprArray);
```
dmanners: Not specifically a problem as join and implode are the same thing but if during this pr you have changed the code I would suggest we keep them the same between the two versions so I would request either revert that change in the other pr or make the same change here. The choice is yours. https://github.com/magento/magento2/pull/11567/files#diff-482e1a34ed1fa0e0e666785335496ef8R82

mrodespin: @dmanners I didn't modify this function it was the phpcsfixer utility. I need to change it?

dmanners: @mrodespin ah ok, would be good to keep these 2 files a close as possible between 2.2 and 2.1 If you dont mind making that change that would be cool.

mrodespin:
@dmanners done!

dmanners:
@mrodespin I have tested both your branch and the 2.2-develop branch and with both branches I was able with the right configuration to get the `catalog_product_alert` cron to appear in the cron table.

Was there a specific test case that was not working on the 2.2 system?

mrodespin:
@dmanners No, I did this PR because I made this other PR #11476 for fix xml sitemap cron generation that wasn't working and the when I tried to configure cron products alert I couldn't do it, so I used the same code for this.

dmanners:
Hmm ok cause on the `2.2-develop` branch (without your changes) I can run the following.

1. Install Magento and setup correct timezone for admin,
2. Setup Product alerts cron to run in 5 mins time,
3. Run cron:run cli command,
4. Check DB for job_code = "catalog_product_alert",

With this test case everything works as expected. The code you have changed is not actually about the cron generation but the config saving. On both branches the core config table is updated in the same way. Though I do think your code changes make the code more readable I do not think that there is a bug to fix here.

mrodespin:
Okey, we can close the PR then

PR #11499, opened 2017-10-16
Ensure database is cleared/Magento reinstalled when TESTS_CLEANUP is enabled

joshuaswarren:
### Description
When TESTS_CLEANUP is set to enabled, the database should be cleared and Magento reinstalled each time the integration tests are run. This is the expected behavior and the documented behavior according to http://devdocs.magento.com/guides/v2.2/test/integration/integration_test_execution.html#the-testscleanup-constant

In this pull request, I've added a check for the TESTS_CLEANUP variable which we use and only restore the database from a previous dump (instead of installing from scratch) if TESTS_CLEANUP is not enabled. 

1. https://github.com/magento/magento2/issues/10025 Integration tests don't reset the database

### Manual testing scenarios
1. Follow the steps in http://devdocs.magento.com/guides/v2.2/test/integration/integration_test_execution.html
2. Ensure that TESTS_CLEANUP is enabled
3. Make some manual change to the integration test database
4. Run the integration tests again
5. Ensure that your change you made in step 3 is no longer present in the database.

### Contribution checklist

### DIFF
--- a/dev/tests/integration/framework/Magento/TestFramework/Application.php
+++ b/dev/tests/integration/framework/Magento/TestFramework/Application.php
@@ -438,7 +438,7 @@ class Application
      * @return void
      * @throws \Magento\Framework\Exception\LocalizedException
      */
-    public function install()
+    public function install($cleanup)
     {
         $dirs = \Magento\Framework\App\Bootstrap::INIT_PARAM_FILESYSTEM_DIR_PATHS;
         $this->_ensureDirExists($this->installDir);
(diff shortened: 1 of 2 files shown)


dev/tests/integration/framework/Magento/TestFramework/Application.php
```diff
@@ -438,7 +438,7 @@ public function cleanup()
      * @return void
      * @throws \Magento\Framework\Exception\LocalizedException
      */
-    public function install()
+    public function install($cleanup)
```
okorshenko, line 441: Could you please take a look at `Magento\TestFramework\WebApiApplication::install()` this method signature must be compatible with `Magento\TestFramework\Application::install($cleanup)`

okorshenko (changes requested):

PR #11502, opened 2017-10-16
MAGETWO-10704: refactoring free shipping for live rates

JosephMaxwell:
*In progress:*
https://github.com/magento/magento2/issues/10704

The original issue stated that cart price rules for free shipping were broken on UPS. In replicating this issue, I found that it was broken for UPS, USPS and Fede. Since most of the code was committed back in 2011, this likely is a leftover issue from back then.

### Description
At face value, fixing the live rates free shipping should be easy. Instead, I ended up refactoring the code surrounding the free shipping calculation.

One thing that is an assumption is that Row Weight and Free Method Weight are mutually exclusive. If there are 10 items in the cart and 7 of them qualify for free shipping while the other three don't and they all weigh 1lb, the Free Method Weight would be 7lbs and the Row Weight would be 3lbs.

I also took the opportunity to refactor the `\Magento\Shipping\Model\Carrier\AbstractCarrier::_updateFreeMethodQuote()` method. This method was reduced from 51 lines to 10 lines (four other functions were created in the process).

1. magento/magento2#10704: Free shipping code is not working for UPS for magento 2

### Manual testing scenarios
1. Follow the testing scenario in the #10704.
2. Ideally, configure your Magento installation to use test account credentials from USPS, UPS and Fedex.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Fedex/Model/Carrier.php
+++ b/app/code/Magento/Fedex/Model/Carrier.php
@@ -366,9 +366,7 @@ class Carrier extends AbstractCarrierOnline implements \Magento\Shipping\Model\C
 
         $weight = $this->getTotalNumOfBoxes($request->getPackageWeight());
         $r->setWeight($weight);
-        if ($request->getFreeMethodWeight() != $request->getPackageWeight()) {
-            $r->setFreeMethodWeight($request->getFreeMethodWeight());
-        }
+        $r->setFreeMethodWeight($request->getFreeMethodWeight());
 
         $r->setValue($request->getPackagePhysicalValue());
         $r->setValueWithDiscount($request->getPackageValueWithDiscount());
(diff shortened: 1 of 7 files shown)


joeshelton-wagento:
Love to continue seeing work on this. Client is not happy after an upgrade from 2.0 caused his main promotion to stop working.

dazz397:
This fix causes problems with a module that was working correctly in Magento 2.1.10.  After applying the changes, another module I had installed called matrix rates started only showing 0 as the package weight.  I am not sure if it affects other shipping modules since I only use UPS and matrix rates from web shop apps.

okorshenko:
Hi All,
QA department also reported the issue is still present on this PR with applying the discount on shipping amount.
We are closing this PR due to inactivity. Feel free to reopen it once ready

markalston:
@JosephMaxwell  This pull almost works but there are a few other issues that need to be fixed with cart promotions.  

That said, your change to how free shipping is calculated (by subtracting the shipping cost for the free shipping items from the total shipping cost), while intuitive from an overall standpoint, leads to some confusing things for the customer.

With those changes the following situation happens: 

 A customer has a cart with non-free shipping items in it (and possibly free items as well).   They check the shipping amount and it is X dollars.  If the customer then adds some more free shipping items and goes back to see the shipping they will see that the new shipping amount is different.  

This is not what we would intuitively think should happen but does so because the shipping rates are not linear based on weight.  By subtracting out the price of shipping the free items from the total shipping price the final ship prices moves around as we add and remove free shipping items (or non-free for that matter).  This results in the case where adding free shipping items can reduce or increase the cost of shipping non-free items.

PR #11503, opened 2017-10-16
MAGETWO-81245: All arguments for mapping methods  must be compatible with call_user_func_array

kirmorozov:
Making standard data structure for use in map().

### Description
Making all elements of criteria arrays for use in call_user_func_array. 
Throw exception if it not. 

1. magento/magento2#7678: StockItemCriteria::setProductsFilter doesn't work with array of ids

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/DB/AbstractMapper.php
+++ b/lib/internal/Magento/Framework/DB/AbstractMapper.php
@@ -124,7 +124,7 @@ abstract class AbstractMapper implements MapperInterface
             $mapperMethod = 'map' . $camelCaseKey;
             if (method_exists($this, $mapperMethod)) {
                 if (!is_array($value)) {
-                    $value = [$value];
+                    throw new \InvalidArgumentException('Wrong type of argument, expecting array for '. $mapperMethod);
                 }
                 call_user_func_array([$this, $mapperMethod], $value);
             }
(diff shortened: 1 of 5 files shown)


lib/internal/Magento/Framework/DB/AbstractMapper.php
```diff
@@ -124,7 +124,7 @@ public function map(CriteriaInterface $criteria)
             $mapperMethod = 'map' . $camelCaseKey;
             if (method_exists($this, $mapperMethod)) {
                 if (!is_array($value)) {
-                    $value = [$value];
+                    throw \Exception('Wrong type of argument, expecting array for '. $mapperMethod);
```
orlangur: `new` operator is missing here. Looks like `InvalidArgumentException` is more suitable.

Besides fixing an existing unit test
```
There was 1 error:
1) Magento\Framework\DB\Test\Unit\AbstractMapperTest::testMap with data set #0 (array('mapMyMapperMethodOne', 'mapMyMapperMethodTwo'), array('my-test-value1', 'my-test-value2'))
Error: Call to undefined function Exception()
/home/travis/build/magento/magento2/lib/internal/Magento/Framework/DB/AbstractMapper.php:127
/home/travis/build/magento/magento2/lib/internal/Magento/Framework/DB/Test/Unit/AbstractMapperTest.php:141
```
please cover this new logic with additional test method.

orlangur (changes requested):

lib/internal/Magento/Framework/DB/AbstractMapper.php
kirmorozov: Thanks. Will update test also in next commit.

kirmorozov: Thanks. Will update test also in next commit.

kirmorozov:
@orlangur Seems to be good now.

orlangur:
@kirmorozov thanks. Since this is a significant system behavior change could you please add a test case ensuring exception is thrown in case of passing not array as a value?

Please squash changes into single commit and force push to the same branch.

PR #11513, opened 2017-10-17
Modify Report processor to return 500

### LINKED ISSUE #11512: Incorrect use of 503 status code
More generally, Magento appears to use the "503 (service unavailable)" status code when an "unexpected error" happens. This ranges, but a good example of such an event is anything that generates the "report" format pages.

The problem is, this status code is generally reserved for cases in which the server is suffering a temporary problem[1][2]. When the server is in an error case, I would have expected something such as a 500 (internal server error) to be expressed, clearly indicating an "abnormal" event, or one that is not an expected failure mode.

This has implications for monitoring. When setting up alerting rules, it's unclear how to express that a problem is temporary in nature. For example, if during a period of high traffic there are several 503s, this does not warrant immediate resolution. However, if there is a spike in 503s as a result of an SQL constraint issue, this requires immediate attention, and will not self-resolve.

I would suggest modifying the report page to return a 500, rather than a 503. Additionally, some guidelines as to what status code was used when would be useful.

[1] - https://stackoverflow.com/questions/2786595/what-is-the-correct-http-status-code-to-send-when-a-site-is-down-for-maintenance
[2] - https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

andrewhowdencom:
The report processor is currently returning a HTTP 503 status
code; generally used for temporarily failures to connect to a service
such as when that service is in maintenance mode, when an upstream
proxy is unavailable etc.

This commit modifies the report HTTP code to be a 500. The author
believes this to be a better reflection that the error is miscellaneous
in nature, and that action is required in order to change it (i.e.
it is not a temporary condition)

Fixes #11512

### DIFF
--- a/pub/errors/processor.php
+++ b/pub/errors/processor.php
@@ -215,7 +215,7 @@ class Processor
     public function processReport()
     {
         $this->pageTitle = 'There has been an error processing your request';
-        $this->_response->setHttpResponseCode(503);
+        $this->_response->setHttpResponseCode(500);
 
         $this->showErrorMsg = false;
         $this->showSentMsg  = false;


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11513) <br/>All committers have signed the CLA.

orlangur (approved):

andrewhowdencom:
CI failed due to timeout. Please rerun

orlangur:
Yeah, simply ignore it, main source of truth is Bamboo running internally.

andrewhowdencom:
While I understand that, I do feel I need to point out that if CI is ignored, it is pointless. I am somewhat grossly passionate about this topic as I find people regularly ignoring builds, but I feel it's reasonable to point this out here.

Or, in other words, *old man grumbling*

orlangur:
Sorry, I didn't give enough context (as usual), I'm a big fan of CICD done the right way as well :)

I meant "ignore functional tests in Travis, I hope somebody is working on permanent solution for these failures". Just found out if was fixed for `develop` in https://github.com/magento/magento2/commit/f291ca3abe796971d65ef45a9c9defa3a2c39086#diff-354f30a63fb0907d4ad57269548329e3 (by testsuite reduction but it's totally fine considering limited Travis resources and that those tests are still running on Bamboo). Raised this problem in Slack so that it will be backported to `2.1-develop` and `2.2-develop` (or maybe fixed another way).

andrewhowdencom:
HAHA sweet. I sometimes feel guilty for bitching, but unfortunately someone has to and I am okay being the jerk. 

<3 for patience, and particularly <3 for super fast feedback. This is my first PR, and it's cool!

andrewhowdencom:
Uhh @magento-team what needs an update? And @okorshenko I'm hesitant to call this a feature request as:

- it's changing behaviour, not adding new functionality. That's at most a refactoring, but not a feature request
- feature requests are a disallowed here, and are often closed.

Please advise

orlangur:
Hello @andrewhowdencom, the only problem here was an actual issue reproduction to see 500/503 HTTP code. I will check it today and I believe we will proceed with merging.

Don't worry about labels, all PRs are processed manually and won't get closed accidentally. Probably related to some massive issue processing.

Issue #8145, opened 2017-01-15
Catalog image resizing performance problems


hostep:
### Preconditions
1. Magento develop branch (1856c28 - January 15 2017)
2. PHP 7.0.14

### Steps to reproduce
1. Have a shop with a bunch of products with images assigned to them
2. Run `bin/magento catalog:images:resize`

### Expected result
1. The command should only produced images for the themes which are currently active in the frontend
2. The command shouldn't produce multiple files which are binary exactly the same

### Actual result
1. The command produces files for all the installed themes, even if they aren't being used in the frontend
2. The command produced files multiple times which are binary exactly the same

### Discussion
While trying to figure out why it takes over 12 hours to run the command `bin/magento catalog:image:resize` on a very beefy server with a Magento CE 2.1.2 shop with about 6500 products and 8500 images, I found a couple of performance problems in the code:

1. The command generates resized images for all themes, so if you are **not** using `Magento/blank` or `Magento/luma` on the frontend, you still get images resized for those themes, and they take a lot of time to generate, so this is not good
2. After fixing nr 1, I looked at the generated files and compared them using a hashing function to see if there were duplicated files in the result. And this turned out to be the case.
It was always caused by a difference in the `image type` (thumbnail, small_image, image, ...). All the other parameters (width, height, keep frame, transparency, quality, ...)  to generate a unique file are correct I think. But I don't think the distinction on `image type` is important, but maybe I'm missing something?

### Possible solution
1. I created a PR over here: https://github.com/magento/magento2/pull/8142
2. In [Magento\Catalog\Model\Product\Image::getMiscParams](https://github.com/magento/magento2/blob/eb30343/app/code/Magento/Catalog/Model/Product/Image.php#L934), remove the line:
`'image_type' => $this->getDestinationSubdir(),`

### Result
I ran a very small benchmark using a test shop with 3 products and 5 images in total. Only `Magento/blank` and `Magento/luma` themes are installed and only `Magento/luma` is active in frontend:

1. Originally: 11 seconds, 165 files are generated.
After applying PR: 9 seconds, 135 files are generated
2. Originally: 11 seconds, 165 files are generated.
After removing line mentioned above: 9 seconds, 130 files are generated

Combined result of both optimizations:
* Originally: 11 seconds, 165 files are generated.
After both optimizations: 7 seconds, 105 files are generated

I think this is a significant enough improvement for you guys to at least consider optimizing this :)

Thanks!

choukalos:
[...]

PR #8142, opened 2017-01-15
Only resize images for themes which are being used instead of all installed themes.

hostep:
Currently Magento resizes catalog images in the frontend for all the installed themes.
So if you have the `Magento/blank`, `Magento/luma` and your own custom theme, it would resize the images for those three themes. Even if you only use your own custom theme in the frontend.

This PR should fix this, it will only resize catalog images for themes which are actually active.

This code is being executed when running the command: `bin/magento catalog:images:resize` and also in the `afterSave` method of the `Magento\Catalog\Model\Product` class.

PR #11514, opened 2017-10-17
Only resize images (cache) for themes which are being used instead of all installed themes

denisristic:
Only resize images (cache) for themes which are being used instead of all installed themes

### Description
Pull request is using [@vrann fix](https://github.com/magento/magento2/pull/8142/commits/35e3225f0478a7b952119dca2a800a76b9ae5f3e#diff-dbf04232abcedbe2e67996b043041b48) and it is upgraded to check (as suggested in comments) for used (custom) themes in products/categories/cms pages.

1. magento/magento2#10363
2. magento/magento2#8145
3. magento/magento2#8142

### Manual testing scenarios
1. Add custom theme for product/category/cms page
2. Disable some theme
3. Upload product image
4. Check if cache image is generated for disabled theme

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Image/Cache.php
+++ b/app/code/Magento/Catalog/Model/Product/Image/Cache.php
@@ -29,7 +29,7 @@ class Cache
     /**
      * @var ThemeCustomizationConfig
      */
-    protected $themeCustomizationConfig;
+    private $themeCustomizationConfig;
 
     /**
      * @var ImageHelper
@@ -160,7 +160,7 @@ class Cache
      *
      * @return array
      */
-    public function getThemesInUse()
+    private function getThemesInUse(): array
     {
         $themesInUse = [];
 
@@ -175,7 +175,9 @@ class Cache
 
         $connection = $this->attribute->getResource()->getConnection();
 
-        $productCustomDesignAttributeId = $this->attribute->loadByCode(4, 'custom_design')->getId();
+        $productEntityTypeId = \Magento\Catalog\Setup\CategorySetup::CATALOG_PRODUCT_ENTITY_TYPE_ID;
+
+        $productCustomDesignAttributeId = $this->attribute->loadByCode($productEntityTypeId, 'custom_design')->getId();
 
         $productSql = $connection
             ->select()
@@ -198,7 +200,9 @@ class Cache
             }
         }
 
-        $categoryCustomDesignAttributeId = $this->attribute->loadByCode(3, 'custom_design')->getId();
+        $categoryEntityTypeId = \Magento\Catalog\Setup\CategorySetup::CATEGORY_ENTITY_TYPE_ID;
+
+        $categoryCustomDesignAttributeId = $this->attribute->loadByCode($categoryEntityTypeId, 'custom_design')->getId();
 
         $categorySql = $connection
             ->select()
--- a/app/code/Magento/Catalog/Model/Product/Image/Cache.php
+++ b/app/code/Magento/Catalog/Model/Product/Image/Cache.php
@@ -8,8 +8,12 @@ namespace Magento\Catalog\Model\Product\Image;
 use Magento\Catalog\Helper\Image as ImageHelper;
 use Magento\Catalog\Model\Product;
 use Magento\Theme\Model\ResourceModel\Theme\Collection as ThemeCollection;
+use Magento\Theme\Model\Config\Customization as ThemeCustomizationConfig;
 use Magento\Framework\App\Area;
 use Magento\Framework\View\ConfigInterface;
+use Magento\Eav\Model\Config;
+use Magento\Eav\Model\Entity\Attribute;
+use Magento\Framework\ObjectManagerInterface;
 
 class Cache
 {
@@ -24,9 +28,29 @@ class Cache
     protected $themeCollection;
 
     /**
+     * @var ThemeCustomizationConfig
+     */
+    protected $themeCustomizationConfig;
+
+    /**
      * @var ImageHelper
      */
     protected $imageHelper;
+
+    /**
+     * @var Config
+     */
+    protected $eavConfig;
+
+    /**
+     * @var Attribute
+     */
+    protected $attribute;
+
+    /**
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
 
     /**
      * @var array
@@ -41,11 +65,19 @@ class Cache
     public function __construct(
         ConfigInterface $viewConfig,
         ThemeCollection $themeCollection,
-        ImageHelper $imageHelper
+        ImageHelper $imageHelper,
+        ThemeCustomizationConfig $themeCustomizationConfig,
+        Config $eavConfig,
+        Attribute $attribute,
+        ObjectManagerInterface $objectManager
     ) {
         $this->viewConfig = $viewConfig;
         $this->themeCollection = $themeCollection;
         $this->imageHelper = $imageHelper;
+        $this->themeCustomizationConfig = $themeCustomizationConfig;
+        $this->eavConfig = $eavConfig;
+        $this->attribute = $attribute;
+        $this->objectManager = $objectManager;
     }
 
     /**
@@ -58,8 +90,10 @@ class Cache
     protected function getData()
     {
         if (!$this->data) {
+            $themesInUse = $this->getThemesInUse();
+
             /** @var \Magento\Theme\Model\Theme $theme */
-            foreach ($this->themeCollection->loadRegisteredThemes() as $theme) {
+            foreach ($themesInUse as $theme) {
                 $config = $this->viewConfig->getViewConfig([
                     'area' => Area::AREA_FRONTEND,
                     'themeModel' => $theme,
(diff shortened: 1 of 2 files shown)


app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -24,11 +27,27 @@ class Cache
     protected $themeCollection;
 
     /**
+     * @var ThemeCustomizationConfig
+     */
+    protected $themeCustomizationConfig;
```
ishakhsuvarov: Adding new `protected` properties is expanding the contract of the class, which is not really desired in this case. Please change to `private`

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,93 @@ protected function processImageData(Product $product, array $imageData, $file)
 
         return $this;
     }
+
+    /**
+     * Get themes in use
+     *
+     * @return array
+     */
+    public function getThemesInUse()
```
ishakhsuvarov: Does this method need to be public?
Please consider extracting it into appropriate service, as `getThemesInUse` does not seem to have too much in common with Product Image Cache Model.

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,93 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+        $registeredThemes = $this->themeCollection->loadRegisteredThemes();
+        $storesByThemes   = $this->themeCustomizationConfig->getStoresByThemes();
+
+        foreach ($registeredThemes as $registeredTheme) {
+            if (array_key_exists($registeredTheme->getThemeId(), $storesByThemes)) {
+                $themesInUse[] = $registeredTheme;
+            }
+        }
+
+        $connection = $this->attribute->getResource()->getConnection();
```
ishakhsuvarov, line 176: Getting dependencies in such way (one dependency via another one) violates the principle of least knowledge. 
Please consider refactoring this.

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,93 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+
+        $productCustomDesignAttributeId = $this->attribute->loadByCode(4, 'custom_design')->getId();
+
+        $productSql = $connection
```
ishakhsuvarov: Direct access to the data layer is out of responsibilities scope for Magento Models. I suggest to move it into appropriate resource model. This would also help to resolve the commend above (on the least knowledge principle).

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,93 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+
+        foreach ($registeredThemes as $registeredTheme) {
+            if (array_key_exists($registeredTheme->getThemeId(), $storesByThemes)) {
+                $themesInUse[] = $registeredTheme;
+            }
+        }
+
+        $connection = $this->attribute->getResource()->getConnection();
+
+        $productCustomDesignAttributeId = $this->attribute->loadByCode(4, 'custom_design')->getId();
```
ishakhsuvarov: Is there a way to avoid hardcoding the `4`?

app/code/Magento/Catalog/Model/Product/Image/Cache.php
ishakhsuvarov: Please add PHP 7.0 specific type hinting  and return types to any newly introduced methods.

ishakhsuvarov (changes requested):

app/code/Magento/Catalog/Model/Product/Image/Cache.php
denisristic: Where would you suggest I should put it? Inside themes?

app/code/Magento/Catalog/Model/Product/Image/Cache.php
denisristic: changed to private

app/code/Magento/Catalog/Model/Product/Image/Cache.php
denisristic: fixed

app/code/Magento/Catalog/Model/Product/Image/Cache.php
denisristic: added

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,97 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+            if (array_key_exists($registeredTheme->getThemeId(), $storesByThemes)) {
+                $themesInUse[] = $registeredTheme;
+            }
+        }
+
+        $connection = $this->attribute->getResource()->getConnection();
+
+        $productEntityTypeId = \Magento\Catalog\Setup\CategorySetup::CATALOG_PRODUCT_ENTITY_TYPE_ID;
+
+        $productCustomDesignAttributeId = $this->attribute->loadByCode($productEntityTypeId, 'custom_design')->getId();
```
sidolov, line 180: Please use \Magento\Eav\Model\Config::getAttribute instead of load entity by model

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,97 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+
+        foreach ($registeredThemes as $registeredTheme) {
+            if (array_key_exists($registeredTheme->getThemeId(), $storesByThemes)) {
+                $themesInUse[] = $registeredTheme;
+            }
+        }
+
+        $connection = $this->attribute->getResource()->getConnection();
+
+        $productEntityTypeId = \Magento\Catalog\Setup\CategorySetup::CATALOG_PRODUCT_ENTITY_TYPE_ID;
```
sidolov, line 178: Please use constant directly instead of assign to variable

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -41,11 +60,17 @@ class Cache
     public function __construct(
         ConfigInterface $viewConfig,
         ThemeCollection $themeCollection,
-        ImageHelper $imageHelper
+        ImageHelper $imageHelper,
+        ThemeCustomizationConfig $themeCustomizationConfig,
```
sidolov, line 64: Dependencies added with backward compatibility standards violation, right way described in http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/ section: Adding a constructor parameter

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,97 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+     */
+    private function getThemesInUse(): array
+    {
+        $themesInUse = [];
+
+        $registeredThemes = $this->themeCollection->loadRegisteredThemes();
+        $storesByThemes   = $this->themeCustomizationConfig->getStoresByThemes();
+
+        foreach ($registeredThemes as $registeredTheme) {
+            if (array_key_exists($registeredTheme->getThemeId(), $storesByThemes)) {
```
sidolov, line 171: Use $registeredTheme->getCode() instead of $registeredTheme->getThemeId() otherwise this expression will return false permanently and $themesInUse array will be always empty. Without this fix cache directory is empty.

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,97 @@ protected function processImageData(Product $product, array $imageData, $file)
 
         return $this;
     }
+
+    /**
+     * Get themes in use
+     *
+     * @return array
+     */
+    private function getThemesInUse(): array
```
sidolov, line 163: Consider refactoring this logic with introducing composite object (with configuration via DI) for themes in use and move logic to corresponded modules (pages to CMS module; product, category -> Catalog module etc)

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -23,11 +26,27 @@ class Cache
...
+
     /**
      * @var ImageHelper
      */
     protected $imageHelper;
 
+    /**
+     * @var Config
+     */
+    protected $eavConfig;
```
sidolov, line 42: Please use private modifier for object dependencies

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -23,11 +26,27 @@ class Cache
...
+
+    /**
+     * @var Attribute
+     */
+    protected $attribute;
```
sidolov, line 47: Please use private modifier for object dependencies

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,97 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+        if (count($productThemeIds)) {
+            foreach ($productThemeIds as $productThemeId) {
+                if (array_key_exists($productThemeId, $storesByThemes)
+                    && !array_key_exists($productThemeId, $themesInUse) ) {
+                    $themesInUse[] = $this->themeCollection->load($productThemeId);
+                }
+            }
+        }
+
+        $categoryEntityTypeId = \Magento\Catalog\Setup\CategorySetup::CATEGORY_ENTITY_TYPE_ID;
```
sidolov, line 203: Please use constant directly instead of assign to variable

app/code/Magento/Catalog/Model/Product/Image/Cache.php
```diff
@@ -127,4 +154,97 @@ protected function processImageData(Product $product, array $imageData, $file)
...
+
+        $categoryCustomDesignAttributeId = $this->attribute->loadByCode($categoryEntityTypeId, 'custom_design')->getId();
```
sidolov, line 205: Please use \Magento\Eav\Model\Config::getAttribute instead of load entity by model

sidolov (changes requested):

sidolov:
@denisristic,  I am closing this PR now due to inactivity.
Thank you for collaboration

eharvest:
Firstly I really want to thank @sidolov for making such effort to offer this solution and I want to give a bit feedback of how it worked for me.

I have a Magento 2.2.3 store (now in production). It has 20k products. Before I run catalog:images:resize the overall size is less than 20G, after run the command a few times (I always delete product image cache before run the command), the size is now growing to nearly 40G. And this what I get:

- without applying the fix, the command took up to 22 hours to finish and my product image cache grows to 22G!!!
- after applying the fix, the command took about half the time and my product image cache still as big as 18G. I saved 4G but it's still a lot bigger than I expected. I'm using a child theme, would that be the issue?

I can see there're quite a few developers have been complaining about this, and I really think this is a big performance issue, now we have been forced to use a much bigger server. Hopefully this can be fixed in 2.3 where catalog:images:resize can take args and only apply to the theme in use.

PR #11556, opened 2017-10-19
Fix #10583: Checkout place order exception when using a new address

### LINKED ISSUE #10583: Checkout place order exception when using a new address
### Preconditions
1. Magento 2.1.8 CE
2. PHP 7.0
3. Have a plugin on `\Magento\Customer\Api\Data\CustomerInterface` (any function, the plugin doesn't have to do anything)

### Steps to reproduce
1. Register as new customer
2. Add a product to your cart
3. Go to the checkout page
4. Fill in your address details and go through the checkout steps
5. Place an order

### Expected result
1. Order is placed and the address is added to your address book

### Actual result
1. The checkout displays the error message "An error occurred on the server. Please try to place the order again."

The actual error message is this: InvalidArgumentException: Getter return type must be specified using @return annotation. See Magento\Customer\Model\Data\Customer\Interceptor::getDefaultBilling() in vendor/magento/framework/Reflection/TypeProcessor.php on line 305

This error is caused in: `\Magento\Quote\Model\Quote::updateCustomerData()` by 
`$this->dataObjectHelper->mergeDataObjects(get_class($quoteCustomer), $quoteCustomer, $customer);`

The `$quoteCustomer` is the interceptor, and the `TypeProcessor` can't retrieve the required docblock information from it.

YevSent:
### Fixed Issues
1. magento/magento2#10583: Checkout place order exception when using a new address

### Contribution checklist

### DIFF
--- a/dev/tests/integration/testsuite/Magento/Quote/Model/QuoteTest.php
+++ b/dev/tests/integration/testsuite/Magento/Quote/Model/QuoteTest.php
@@ -393,7 +393,7 @@ class QuoteTest extends \PHPUnit\Framework\TestCase
             \Magento\Customer\Model\Data\Customer::DOB => '2014-02-03 00:00:00',
             \Magento\Customer\Model\Data\Customer::EMAIL => 'qa@example.com',
             \Magento\Customer\Model\Data\Customer::FIRSTNAME => 'Joe',
-            \Magento\Customer\Model\Data\Customer::GENDER => 'Male',
+            \Magento\Customer\Model\Data\Customer::GENDER => 0,
             \Magento\Customer\Model\Data\Customer::GROUP_ID =>
                 \Magento\Customer\Model\GroupManagement::NOT_LOGGED_IN_ID,
             \Magento\Customer\Model\Data\Customer::ID => 1,
--- a/lib/internal/Magento/Framework/Reflection/Test/Unit/Fixture/TSample.php
+++ b/lib/internal/Magento/Framework/Reflection/Test/Unit/Fixture/TSample.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\Reflection\Test\Unit\Fixture;
+
+class TSample implements TSampleInterface
+{
+    /**
+     * @inheritdoc
+     */
+    public function getPropertyName()
+    {
+        return '';
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getName()
+    {
+        return '';
+    }
+}
--- a/lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest.php
+++ b/lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest.php
@@ -6,8 +6,9 @@
 // @codingStandardsIgnoreStart
 namespace Magento\Framework\Reflection\Test\Unit;
 
-use Zend\Code\Reflection\ClassReflection;
 use Magento\Framework\Exception\SerializationException;
+use Magento\Framework\Reflection\Test\Unit\Fixture\TSample;
+use Zend\Code\Reflection\ClassReflection;
 
 /**
  * Type processor Test
@@ -259,5 +260,36 @@ class TypeProcessorTest extends \PHPUnit\Framework\TestCase
     public function testGetOperationName()
     {
         $this->assertEquals("resNameMethodName", $this->_typeProcessor->getOperationName("resName", "methodName"));
+    }
+
+    /**
+     * Checks a case when method has only `@inheritdoc` annotation.
+     */
+    public function testGetReturnTypeWithInheritDocBlock()
+    {
+        $expected = [
+            'type' => 'string',
+            'isRequired' => true,
+            'description' => null,
+            'parameterCount' => 0
+        ];
+
+        $classReflection = new ClassReflection(TSample::class);
+        $methodReflection = $classReflection->getMethod('getPropertyName');
+
+        self::assertEquals($expected, $this->_typeProcessor->getGetterReturnType($methodReflection));
+    }
+
+    /**
+     * Checks a case when method and parent interface don't have `@return` annotation.
+     *
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage Getter return type must be specified using @return annotation. See Magento\Framework\Reflection\Test\Unit\Fixture\TSample::getName()
+     */
+    public function testGetReturnTypeWithoutReturnTag()
+    {
+        $classReflection = new ClassReflection(TSample::class);
+        $methodReflection = $classReflection->getMethod('getName');
+        $this->_typeProcessor->getGetterReturnType($methodReflection);
     }
 }
--- a/lib/internal/Magento/Framework/Reflection/TypeProcessor.php
+++ b/lib/internal/Magento/Framework/Reflection/TypeProcessor.php
@@ -8,6 +8,7 @@ namespace Magento\Framework\Reflection;
 use Magento\Framework\Exception\SerializationException;
 use Magento\Framework\Phrase;
 use Zend\Code\Reflection\ClassReflection;
+use Zend\Code\Reflection\DocBlock\Tag\ReturnTag;
 use Zend\Code\Reflection\DocBlockReflection;
 use Zend\Code\Reflection\MethodReflection;
 use Zend\Code\Reflection\ParameterReflection;
@@ -16,6 +17,7 @@ use Zend\Code\Reflection\ParameterReflection;
  * Type processor of config reader properties
  *
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ * @SuppressWarnings(PHPMD.ExcessiveClassComplexity) this suppress MUST be removed after removing deprecated methods.
  */
 class TypeProcessor
 {
@@ -275,22 +277,7 @@ class TypeProcessor
      */
     public function getGetterReturnType($methodReflection)
     {
-        $methodDocBlock = $methodReflection->getDocBlock();
-        if (!$methodDocBlock) {
-            throw new \InvalidArgumentException(
-                "Each getter must have description with @return annotation. "
-                . "See {$methodReflection->getDeclaringClass()->getName()}::{$methodReflection->getName()}()"
-            );
-        }
-        $returnAnnotations = $methodDocBlock->getTags('return');
-        if (empty($returnAnnotations)) {
-            throw new \InvalidArgumentException(
-                "Getter return type must be specified using @return annotation. "
-                . "See {$methodReflection->getDeclaringClass()->getName()}::{$methodReflection->getName()}()"
-            );
-        }
-        /** @var \Zend\Code\Reflection\DocBlock\Tag\ReturnTag $returnAnnotation */
-        $returnAnnotation = current($returnAnnotations);
+        $returnAnnotation = $this->getMethodReturnAnnotation($methodReflection);
         $types = $returnAnnotation->getTypes();
         $returnType = current($types);
         $nullable = in_array('null', $types);
@@ -362,7 +349,7 @@ class TypeProcessor
                 self::NORMALIZED_INT_TYPE,
                 self::NORMALIZED_FLOAT_TYPE,
                 self::NORMALIZED_DOUBLE_TYPE,
-                self::NORMALIZED_BOOLEAN_TYPE
+                self::NORMALIZED_BOOLEAN_TYPE,
             ]
         );
     }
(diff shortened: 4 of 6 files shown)


lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest.php
```diff
@@ -25,6 +27,12 @@ class TypeProcessorTest extends \PHPUnit\Framework\TestCase
     protected function setUp()
     {
         $this->_typeProcessor = new \Magento\Framework\Reflection\TypeProcessor();
+        $loader = new ClassLoader();
+        $loader->addPsr4(
```
orlangur: Why do we need to hack autoloader instead of placing fixtures accordingly?

lib/internal/Magento/Framework/Reflection/Test/Unit/_files/TSample.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\Reflection\Test\Unit;
+
+class TSample implements TSampleInterface
+{
+    /**
+     * {@inheritdoc}
```
orlangur: Such syntax only inherits method description, not type hints for parameters and return value.

See https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#61-making-inheritance-explicit-using-the-inheritdoc-tag

Please change implementation in such a way that we do not allow misuse of syntax with curly braces and only inherit return/param in case of `@inheritdoc`.

dev/tests/integration/testsuite/Magento/Quote/Model/QuoteTest.php
```diff
@@ -393,7 +393,7 @@ protected function _getCustomerDataArray()
             \Magento\Customer\Model\Data\Customer::DOB => '2014-02-03 00:00:00',
             \Magento\Customer\Model\Data\Customer::EMAIL => 'qa@example.com',
             \Magento\Customer\Model\Data\Customer::FIRSTNAME => 'Joe',
-            \Magento\Customer\Model\Data\Customer::GENDER => 'Male',
+            \Magento\Customer\Model\Data\Customer::GENDER => 0,
```
orlangur, line 396: Why need this change?

orlangur (changes requested):

lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest.php
YevSent: To do not use `_files` word in the class namespace.

dev/tests/integration/testsuite/Magento/Quote/Model/QuoteTest.php
YevSent: Fixed fixture implementation according to interface doc block annotation https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Customer/Api/Data/CustomerInterface.php#L269. After replacing `get_class` by namespace this fixture breaks test.

lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest.php
orlangur: You don't need to use `_files` and do manual include. Just place fixtures in something like `lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest/Fixture/TSample.php`.

lib/internal/Magento/Framework/Reflection/Test/Unit/_files/TSample.php
YevSent: According to Magento internal guidelines, this annotation should look like `{@inheritdoc}`.

orlangur: Where do you see it? http://devdocs.magento.com/guides/v2.2/coding-standards/docblock-standard-general.html#inheritdoc is compliant with what I'm saying.

lib/internal/Magento/Framework/Reflection/Test/Unit/TypeProcessorTest.php
YevSent: If you open integration or unit tests you will see all fixtures located in `_files` directory. I don't want to change such "standard".

orlangur: There is no such standard. I remember such cases in integration tests just because there is no class usually.

When we can make our code `PSR-4` compliant we should do so. Please remove unnecessary `ClassLoader` complication, other `_files` occurrences containing PHP classes should be considered as technical debt.

lib/internal/Magento/Framework/Reflection/Test/Unit/_files/TSample.php
YevSent: From internal Magento wiki but I agree with `@inheritdoc` usage, even I think it's better to use it.
screenshot from 2017-10-19 15-38-51

orlangur: Great 👍 Yeah, seems to be some outdated documentation, please change/remove it or delegate this to whoever is in charge of it.

YevSent:
The class complexity is 103 but Codacy configured complexity threshold is 50.

orlangur:
@joni-jones I don't think we should any suppressions when `testCodeMess` is passed but Codacy complains.

YevSent:
@orlangur, this class has 7 deprecated methods, which increase complexity, I can't them remove because it will be BIC. Magento class complexity configured as 100 (https://github.com/magento/magento2/blob/2.2-develop/dev/tests/static/testsuite/Magento/Test/Php/_files/phpmd/ruleset.xml#L24) but codacy complexity is 50. So without full refactoring, the codacy test for this class will never be green.

orlangur:
@joni-jones do you mean green in Travis or green in Codacy? The latter is pretty odd, like forbidding  `else` usage, contains TooLongVariable rule which was intentionally removed from Magento PHPMD ruleset.

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
```diff
@@ -708,4 +695,58 @@ private function getNormalizedType($type)
...
+            /** @var ClassReflection $parentClassReflection */
+            foreach ($implemented as $parentClassReflection) {
+                if (!$parentClassReflection->hasMethod($methodName)) {
+                    continue;
+                }
+                $returnAnnotations = $this->getReturnFromDocBlock(
+                    $parentClassReflection->getMethod($methodName)
+                );
+
+                break;
```
orlangur: Currently logic of this cycle is not so easy to understand. I suggest something like

```
foreach (...) {
    if ($parent...->hasMethod()) {
        $returnAnnotations = ...;
        break;
    }
}
```

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
```diff
@@ -708,4 +695,58 @@ private function getNormalizedType($type)
...
+     *
+     * @param MethodReflection $methodReflection
+     * @return ReturnTag
+     */
+    private function getReturnFromDocBlock(MethodReflection $methodReflection)
+    {
+        $methodDocBlock = $methodReflection->getDocBlock();
+        if (!$methodDocBlock) {
+            throw new \InvalidArgumentException(
+                "Each getter must have dock block. "
```
orlangur: Typo dock -> doc. "have **a** doc block"?

YevSent:
I mean green Codacy. Removing deprecated methods will decrease complexity, but the class also requires refactoring like the composition usage, unit tests. I think it should be done in some release but not in a patch release and not during this bug fix.

lib/internal/Magento/Framework/Reflection/Test/Unit/Fixture/TSample.php
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function getPropertyName()
+    {
+        return '';
+    }
+
+    /**
+     * @inheritdoc
```
orlangur, line 19: Would be really nice to add test case with `{@inheritdoc}` to make it explicit such syntax is not supposed to work. I do not insist though, this PR is already excellent.

orlangur (changes requested):

orlangur:
@joni-jones ok, thanks. According to my understanding we should ignore invalid Codacy complaints but I'm not sure if we have a requirement to make it green (especially when "making green" is just adding suppression).

Please remove this suppression if it is not mandatory to make Codacy green.

lib/internal/Magento/Framework/Reflection/Test/Unit/Fixture/TSample.php
YevSent: I thought about it, but a lot of Magento code has such annotation. The current unit test covers such case for parsing return tag because the behavior just checks if a method has `@return` annotation.

YevSent:
This suppress needed for Travis and Bamboo builds because the class has complexity 103, after small refactoring, without changing functionality I've got - 101. To reduce it need to remove deprecated methods and do deep refactoring.

orlangur:
Aaah, ok. So green Bamboo actually, thanks. I'll raise a topic regarding Codacy in Slack, we should do something to it.

lib/internal/Magento/Framework/Reflection/Test/Unit/Fixture/TSample.php
orlangur: > a lot of Magento code has such annotation

We should fix it anyway. Not saying it must be a scope of this PR, but it could be.

YevSent: Absolutely agree, even more, I think there should be a static test which won't allow adding this annotation.

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
YevSent: Done

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
YevSent: Done

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
orlangur: Don't see it :(

YevSent: Please, see a list of changed files, I amended commit.

orlangur: Hm, I was answering to next thread actually O.o

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
orlangur: > +                "Each getter must have a dock block. "

I meant don't see this changed.

YevSent: https://github.com/magento/magento2/pull/11556/files#diff-292261de777138aa6bb4342b9757a3a5R744

orlangur: "dock" is correct spelling? I was not certain regarding "a" but "dock" seems to be not correct.

YevSent: Yes, sure. Fixed.

orlangur (approved):

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
```diff
@@ -708,4 +695,56 @@ private function getNormalizedType($type)
...
+     * @return ReturnTag
+     * @throws \InvalidArgumentException if doc block is empty or `@return` annotation doesn't exist
+     */
+    private function getMethodReturnAnnotation(MethodReflection $methodReflection)
+    {
+        $methodName = $methodReflection->getName();
+        $returnAnnotations = $this->getReturnFromDocBlock($methodReflection);
+        if (empty($returnAnnotations)) {
+            // method can inherit doc block from implemented interface, like for interceptors
+            $implemented = $methodReflection->getDeclaringClass()->getInterfaces();
```
VoidWalker, line 712: Sir, you're missing docs from extended class here. $methodReflection->getDeclaringClass()->getParentClass() should be included in this array

VoidWalker (commented):
Missing inheridocs from extended classes

lib/internal/Magento/Framework/Reflection/TypeProcessor.php
orlangur: Hi @VoidWalker, please report an issue using a template consisting of full steps to reproduce.

PR #11563, opened 2017-10-19
Add price calculation improvement for product option value price

marinagociu:
### Description

The price calculation for the product option value relied on the getFinalPrice value, which if called without the qty parameter will not consider tier prices.
This method was updated to use the base price as calculated by the priceInfo object, just like in Magento\Catalog\Model\Product\Option.

1. magento/magento2#5774: Tier price and custom options give bad results

### Manual testing scenarios

1. Create a product with a regular price.
2. Add a tier price with quantity 1, a customer group and a smaller price than the regular one.
3. Add a custom option using a type that will allow adding values (for example, Checkbox).
4. Add a value with Price Type: Percent, and a price value (for example, 100).
5. Login with a customer from the customer group used when adding the tier price.
6. Go to the product page, the product price should be the tier price added for the customer group and the price of the custom option value should be calculated based on the corresponding product tier price, and not the product regular price.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Option/Value.php
+++ b/app/code/Magento/Catalog/Model/Product/Option/Value.php
@@ -11,6 +11,7 @@ namespace Magento\Catalog\Model\Product\Option;
 use Magento\Catalog\Model\Product;
 use Magento\Catalog\Model\Product\Option;
 use Magento\Framework\Model\AbstractModel;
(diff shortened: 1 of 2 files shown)


dmanners (approved):

dmanners:
Thank you for this PR @marinagociu I will get this processed through our QA team for you now.

PR #11580, opened 2017-10-20
#5770 FIX Catalog Rule Product Index is not reindexing after save

thiagolima-bm:
**Description**
After editing a catalog rule, the rule is never applied.

**Fixed Issues (if relevant)**
#5770

### DIFF
--- a/app/code/Magento/CatalogRule/Model/Rule/Job.php
+++ b/app/code/Magento/CatalogRule/Model/Rule/Job.php
@@ -55,6 +55,7 @@ class Job extends \Magento\Framework\DataObject
     {
         try {
             $this->ruleProcessor->markIndexerAsInvalid();
+            $this->ruleProcessor->reindexAll();
             $this->setSuccess(__('Updated rules applied.'));
         } catch (\Magento\Framework\Exception\LocalizedException $e) {
             $this->setError($e->getMessage());


vrann:
@thiagolima-bm indexer should be invoked just in case if index is in "Update on Save" mode. Otherwise, it should be scheduled for later. Have you tested the use-case when the indexer is in "Update By Schedule" mode? Will indexer be invoked rather way?

orlangur:
Need to check internal JIRA tickets of the time of this feature implementation but offhand Catalog Rule Product cannot work in Update on Save mode due to it's nature.

Main problem is rule chaining feature which in worst case may lead to full reindex in case of only one rule changed. Triggering `reindexAll` on every save will obviously kill your store.

There was a UX improvement reported to make it explicit in UI that "Update on Save" is not available for this index type.

To make "Update on Save" working properly a new mechanism should be implemented which would determine all affected data which needs to be changed. Not sure whether it's hard, super-hard or impossible task.

thiagolima-bm:
It is my understanding that when you click "save and apply", the rule should be applied right away.
I may be wrong... 
Maybe we have here two different problems.

1) Save and apply is not working.
2) Reindex rules are not working.

Another approach to solving this issue would use the rule_id parameter to instantiate the rule and reindex just that rule.

Magento\CatalogRule\Controller\Adminhtml\Promo\Catalog\Save.php, line 95

`if ($this->getRequest()->getParam('auto_apply')) {
                    $this->getRequest()->setParam('rule_id', $model->getId());
                    $this->_forward('applyRules');
} else ....`

at Magento\CatalogRule\Controller\Adminhtml\Promo\Catalog\ApplyRules.php

instead of 
`$ruleJob->applyAll()`
use
`$rule->reindex()`

kokoc:
@thiagolima-bm existing "save and apply" button actually only schedules the update, that will be performed asynchronously. Please setup the cron jobs, if you want to fix the issue on your Magento instance. 

@orlangur is right, the button was disabled due to performance issues in this functionality. 

We have an internal epic, that should improve UX and performance of the rules. Also, we are going to eliminate "update on save" mode for this indexer and remove "apply" button from UI (MAGETWO-39129)

Regarding partial updates - in general case, existing data structure and requirements don't allow partial updates, mostly because of "rule chaining" functionality.

Thank you for the contribution! However, we can't accept the fix, because it may lock the UI and lead to "gateway timeout" errors on some installations. Closing...

PR #11345, opened 2017-10-10
Use US English spelling for "Optimization".

davidangel:
### Contribution checklist
 - [√] Pull request has a meaningful description of its purpose
 - [√] All commits are accompanied by meaningful commit messages
 - [√] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

PR #11584, opened 2017-10-20
Use U.S. English spelling for 'optimization'.

davidangel:
### Contribution checklist

### DIFF
--- a/app/code/Magento/Cms/view/adminhtml/ui_component/cms_page_form.xml
+++ b/app/code/Magento/Cms/view/adminhtml/ui_component/cms_page_form.xml
@@ -138,7 +138,7 @@
             </formElements>
         </field>
     </fieldset>
-    <fieldset name="search_engine_optimisation" sortOrder="20">
+    <fieldset name="search_engine_optimization" sortOrder="20">
         <settings>
             <collapsible>true</collapsible>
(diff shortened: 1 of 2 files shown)


orlangur (changes requested):
1. We must not change code of third-party libraries (like KnockoutJS)
2. Changing fieldset name is backward-incompatible as to me: http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#magento-apis

orlangur:
@davidangel, thanks for your contribution, it looks like all places which could be changed are already fixed in https://github.com/magento/magento2/pull/11345.

davidangel:
1. Reverted. Is there a reason knockout.js isn't included via npm?
2. I was not able to find any other instances of `search_engine_optimisation` in the codebase. I was able to find 7 other results for `search_engine_optimization`. Are we sure this change is not desired?

orlangur:
>  Is there a reason knockout.js isn't included via npm?

Yes. We don't want NodeJS on prod. It could be included via some Composer plugins but for now it works like that.

> Are we sure this change is not desired?

Yeah, third-party customizations are relying on names. Keeping them unbroken is lesser evil than a typo in caption not shown to end user.

PR #11588, opened 2017-10-20
Fix Issue #7225 - Remove hardcoding of apply_to when saving attributes

MartinPeverelli:
Remove hardcoding of apply_to when saving attributes

### Description
If an empty dropdown is created via InstallData script, and its constrained to a product type, when later adding options manually via the Admin, the constrain is removed.
This PR removes the hardcoding of an empty `apply_to` value when saving attributes.

1. [[BUG] [Magento 2.1.2] Programmatically creating an empty dropdown attribute, "apply_to" is set to NULL (from "simple") after adding options through store admin](https://github.com/magento/magento2/issues/7225)

### Manual testing scenarios
1. Create an empty dropdown product attribute named "test_dropdown" programmatically through InstallData.php, set "apply_to" to "simple".
1. After setup:upgrade, things are going fine. In store admin, the "test_dropdown" attribute is shown on simple product create or edit page (an empty dropdown), and is hidden on other products create or edit page. The content of "apply_to" field of the current attribute in database table of "catalog_eav_attribute" is also "simple".
3. After adding attribute options through store admin, the "test_dropdown" remains constrained to the "simple" product type. The content of "apply_to" field of the current attribute in database table of "catalog_eav_attribute" is still "simple".

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
@@ -193,7 +193,7 @@ class Save extends \Magento\Catalog\Controller\Adminhtml\Product\Attribute
                 );
             }
 
-            $data += ['is_filterable' => 0, 'is_filterable_in_search' => 0, 'apply_to' => []];
+            $data += ['is_filterable' => 0, 'is_filterable_in_search' => 0];
 
             if (is_null($model->getIsUserDefined()) || $model->getIsUserDefined() != 0) {
                 $data['backend_type'] = $model->getBackendTypeByInput($data['frontend_input']);


orlangur:
> After adding attribute options through store admin, the "test_dropdown" remains constrained to the "simple" product type. 

Such change would break product type transition. Product type is changed among simple/virtual/configurable/downloadable automatically thus you must not have attributes with `apply_to` limited to only one of these types.

Note that unlike M1 in M2 `apply_to` was intentionally removed from Admin UI.

okorshenko:
Hi @orlangur 
Your comment is valid. But we should consider 2 use cases:
1. Developer creates attribute which is not `user defined` (so it is system attribute). In this case admin user does not have an ability to change it. Developer can control transition of this attribute

2. Developer creates attribute which is `user defined`. Admin user can edit the attribute. But the developer must implement the logic to support attribute transition and catch all possible modifications by the admin user. So it is up to the developer's decision how to handle that.

okorshenko (approved):

PR #11601, opened 2017-10-20
Changed Validation to get store for block

osrecio:
Changed validation to `getIsUniqueBlockToStores`
### Description

Changed validation to get `DEFAULT_STORE_ID` when the app is setted with `SingleStoreMode` in the other cases get store from object

1. magento/magento2#4831: Inconsistent cms block validation and save methods
2. ...

### Manual testing scenarios
1. Make sure that you have only one store view and SingleStoreMode option to `no`
2. Create cms block example for All Store views
3. Create another cms block example for first store view only

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cms/Model/ResourceModel/Block.php
+++ b/app/code/Magento/Cms/Model/ResourceModel/Block.php
@@ -7,10 +7,10 @@ namespace Magento\Cms\Model\ResourceModel;
 
 use Magento\Cms\Api\Data\BlockInterface;
 use Magento\Framework\DB\Select;
+use Magento\Framework\EntityManager\EntityManager;
+use Magento\Framework\EntityManager\MetadataPool;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Model\AbstractModel;
-use Magento\Framework\EntityManager\MetadataPool;
-use Magento\Framework\EntityManager\EntityManager;
 use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
 use Magento\Framework\Model\ResourceModel\Db\Context;
 use Magento\Store\Model\Store;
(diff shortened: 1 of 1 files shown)


dmanners:
So as far I a see this fixes the issue raised but there is another issue with the block validation. In my testing I setup a multi store system so there are two store views for a website. I then added 2 blocks with the id example for store 1 and store 2. I then went to add a 3rd block to store 2 with the id example and the system lets me do this. My thinking is that the validation should check to make sure you are not adding the same block twice to the same store.

blocks

app/code/Magento/Cms/Model/ResourceModel/Block.php
```diff
@@ -183,7 +183,7 @@ public function getIsUniqueBlockToStores(AbstractModel $object)
         $entityMetadata = $this->metadataPool->getMetadata(BlockInterface::class);
         $linkField = $entityMetadata->getLinkField();
 
-        if ($this->_storeManager->hasSingleStore()) {
+        if ($this->_storeManager->isSingleStoreMode()) {
             $stores = [Store::DEFAULT_STORE_ID];
         } else {
             $stores = (array)$object->getData('stores');
```
dmanners, line 189: The store validation seems to be broken here as the data set on the object is under `store_id` and not `stores`.

dmanners (changes requested):
Store validation seems to not work anyway.

app/code/Magento/Cms/Model/ResourceModel/Block.php
dmanners: Or maybe better still would be to use the `getStores` method that is against the Block model.

dmanners (approved):
The validation issue I raised has been solved in https://github.com/magento/magento2/pull/11802

osrecio:
Yes is my teamate @thiagolima-bm , What que do?

dmanners:
@osrecio nothing to do now with this pr. We will process both of them to improve/fix the cms block admin experience. Thank you for your code.

osrecio:
Solved this PR in magento/magento2/pull/11802

Issue #11347, opened 2017-10-10
Payment Failed Email not working as intended


ebaschiera:
According to [documentation](http://docs.magento.com/m2/ce/user_guide/sales/checkout-payment-failed-emails.html), an email should be sent "if the payment method that is chosen during checkout fails to complete the transaction."

Looking at the code, the `sendPaymentFailedEmail()` method is called only in `Magento\Multishipping\Controller\Checkout\OverviewPost` class and in `Magento\Checkout\Controller\Onepage\SaveOrder` class.

I made a couple of tests and in my environment, trying a couple of different payment methods, Magento never run those controllers. So, even if the payment fails, the email notification is not sent.

This behaviour is very different from Magento 1, where the merchant receives effective notifications and can monitor checkout/payment problems.

### Preconditions

1. Magento CE 2.1.8
2. PHP 7.0.22-2+ubuntu16.04.1+deb.sury.org+4
3. MySQL 5.7.19-0ubuntu0.16.04.1

### Steps to reproduce

1. Enable Cybersource payment method (or any other that require configuration) without configuring it
2. Add a product to cart and proceed to checkout
2. select Cybersource and input fake credit card information
3. get the "An error occurred on the server. Please try to place the order again." error message at checkout

### Expected result

1. the merchant should receive a failed payment email notification

### Actual result

1. no emails are sent

magento-engcom-team:
@ebaschiera, thank you for your report.
We've created internal ticket(s) MAGETWO-81571 to track progress on the issue.

adrian-martinez-interactiv4:
I'm working on this

gx2g:
please fix this, my company is having the same issue and we are live taking orders.

okorshenko:
hi @gx2g feel free to join code review and testing of proposed solutions (see linked Pull Requests)
Please let us know if the proposed solution works for you
Thank you

ebaschiera:
I confirm that https://github.com/magento/magento2/pull/11602 works fine.

gx2g:
what do you mean it says on that page that the issue can be replicated in all version of magento 2 it seems. plus the solution is on hold until 2.2 pr?

gx2g:
@ebaschiera @okobchenko  what files to I update I don't understand when look at the linked pull requests I am not seeing a solution to the problem can you guide me on what files I need to update or where I can copy and paste them into my site files. I don't see any solution I may not be understanding pull requests.

ebaschiera:
[...]

PR #11603, opened 2017-10-20
[Backport 2.2-develop] #11347 Fix payment failed email not being sent

adrian-martinez-interactiv4:
According to documentation, an email should be sent "if the payment method that is chosen during checkout fails to complete the transaction."

### Description
Payment failed email is not being sent.

1. magento/magento2#11347: Payment Failed Email not working as intended

### Manual testing scenarios
As described in  magento/magento2#11347

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
+++ b/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
@@ -6,8 +6,10 @@
 
 namespace Magento\Checkout\Model;
 
-use Magento\Quote\Api\CartRepositoryInterface;
+use Magento\Framework\App\ObjectManager;
 use Magento\Framework\Exception\CouldNotSaveException;
+use Magento\Quote\Api\CartRepositoryInterface;
+use Magento\Quote\Model\QuoteIdMask;
 
 /**
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
@@ -51,12 +53,19 @@ class GuestPaymentInformationManagement implements \Magento\Checkout\Api\GuestPa
     private $logger;
 
     /**
+     * @var \Magento\Checkout\Helper\Data
+     */
+    private $checkoutHelper;
+
+    /**
      * @param \Magento\Quote\Api\GuestBillingAddressManagementInterface $billingAddressManagement
      * @param \Magento\Quote\Api\GuestPaymentMethodManagementInterface $paymentMethodManagement
      * @param \Magento\Quote\Api\GuestCartManagementInterface $cartManagement
      * @param \Magento\Checkout\Api\PaymentInformationManagementInterface $paymentInformationManagement
      * @param \Magento\Quote\Model\QuoteIdMaskFactory $quoteIdMaskFactory
      * @param CartRepositoryInterface $cartRepository
+     * @param \Magento\Checkout\Helper\Data $checkoutHelper
+     * @param \Psr\Log\LoggerInterface $logger
      * @codeCoverageIgnore
      */
     public function __construct(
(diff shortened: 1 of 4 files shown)


app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -141,7 +157,7 @@ public function getPaymentInformation($cartId)
     private function getLogger()
     {
         if (!$this->logger) {
-            $this->logger = \Magento\Framework\App\ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
+            $this->logger = ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
```
dmanners: How about we take this a step further? Since we are changing the `__construct` anyway why not optionally inject the logger also and then we can remove this `getLogger` method completely.

app/code/Magento/Checkout/Model/PaymentInformationManagement.php
```diff
@@ -144,23 +159,8 @@ public function getPaymentInformation($cartId)
     private function getLogger()
     {
         if (!$this->logger) {
-            $this->logger = \Magento\Framework\App\ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
+            $this->logger = ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
```
dmanners: How about we take this a step further? Since we are changing the __construct anyway why not optionally inject the logger also and then we can remove this getLogger method completely.

dmanners (changes requested):
Good work with cleaning up some of the old deprecated methods. Could also take it one step further a clean up the old `getLogger` method.

adrian-martinez-interactiv4:
@dmanners Done

dmanners:
@adrian-martinez-interactiv4 thanks for those changes. I will progress this PR through our system.

dmanners (approved):

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -65,14 +74,19 @@ public function __construct(
         \Magento\Quote\Api\GuestCartManagementInterface $cartManagement,
         \Magento\Checkout\Api\PaymentInformationManagementInterface $paymentInformationManagement,
         \Magento\Quote\Model\QuoteIdMaskFactory $quoteIdMaskFactory,
-        CartRepositoryInterface $cartRepository
+        CartRepositoryInterface $cartRepository,
+        \Magento\Checkout\Helper\Data $checkoutHelper = null,
```
YevSent, line 78: You should not use helpers.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -88,12 +102,17 @@ public function savePaymentInformationAndPlaceOrder(
         try {
             $orderId = $this->cartManagement->placeOrder($cartId);
         } catch (\Magento\Framework\Exception\LocalizedException $e) {
+            /** @var $quoteIdMask QuoteIdMask */
+            $quoteIdMask = $this->quoteIdMaskFactory->create()->load($cartId, 'masked_id');
+            /** @var \Magento\Quote\Model\Quote $quote */
+            $quote = $this->cartRepository->getActive($quoteIdMask->getQuoteId());
+            $this->checkoutHelper->sendPaymentFailedEmail($quote, __($e->getMessage()));
```
YevSent, line 109: Not only payment methods can throw `LocalizedException`, your implementation will send email even if something wrong not only with a payment transaction.

YevSent (changes requested):

dmanners:
@adrian-martinez-interactiv4 I will close this PR for 2 reasons.

1. Not all payment providers use the code in this manor and as such this will only fix some cases.
2. As @joni-jones has pointed out the exception is not payment specific and as such we could be sending payment failure emails that are unrelated to payment failure.

Feel free to reach out if you have any questions about this.

PR #11606, opened 2017-10-21
[Backport 2.1-develop] Fix AcountManagementTest unit test fail randomly

adrian-martinez-interactiv4:
Surprisingly, after testing succesfully locally unit tests, I've found this error in Travis, in unit tests performed by \Magento\Customer\Test\Unit\Model\AccountManagementTest class:
captura de pantalla 2017-10-21 a las 1 25 35
Simply rebuilding the job worked and all the tests passed, but this error may happen again at any time, due to how are mocks built in that class:

- A datetime variable is set from current time:
captura de pantalla 2017-10-21 a las 3 36 13

- Below, customerSecure mock expects setRpTokenCreatedAt method to be called with previous datetime value will return self:
captura de pantalla 2017-10-21 a las 3 37 44

- But testing the method, the customerSecure::setRpTokenCreatedAt method is called with the current datetime, that may be NOT the previous one, so tests fails because method was not expected to be called with new date value:
captura de pantalla 2017-10-21 a las 3 42 41

- The test assumes the mock creation and the mocked method call will happen in the same second, thing that may not be true and it's irrelevant for tests using \Magento\Customer\Test\Unit\Model\AccountManagementTest::prepareInitiatePasswordReset method:
captura de pantalla 2017-10-21 a las 3 44 35

Removing expected date fix this rare, weird, but actually happening, issue.

### Manual testing scenarios
1. To simulate this, change 
`$dateTime = date(\Magento\Framework\Stdlib\DateTime::DATETIME_PHP_FORMAT);`
 to
`$dateTime = '2017-10-20 00:00:00';` 
in \Magento\Customer\Test\Unit\Model\AccountManagementTest::prepareInitiatePasswordReset method.
2. Run unit test over this class.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Model/AccountManagement.php
+++ b/app/code/Magento/Customer/Model/AccountManagement.php
@@ -12,7 +12,6 @@ use Magento\Customer\Api\CustomerRepositoryInterface;
 use Magento\Customer\Api\Data\AddressInterface;
 use Magento\Customer\Api\Data\CustomerInterface;
 use Magento\Customer\Api\Data\ValidationResultsInterfaceFactory;
-use Magento\Customer\Model\EmailNotificationInterface;
 use Magento\Customer\Helper\View as CustomerViewHelper;
 use Magento\Customer\Model\Config\Share as ConfigShare;
 use Magento\Customer\Model\Customer as CustomerModel;
@@ -22,6 +21,7 @@ use Magento\Framework\Api\ExtensibleDataObjectConverter;
 use Magento\Framework\App\Area;
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\App\ObjectManager;
+use Magento\Framework\DataObjectFactory as ObjectFactory;
 use Magento\Framework\Encryption\EncryptorInterface as Encryptor;
 use Magento\Framework\Encryption\Helper\Security;
 use Magento\Framework\Event\ManagerInterface;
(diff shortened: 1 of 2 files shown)


app/code/Magento/Customer/Model/AccountManagement.php
```diff
@@ -1159,8 +1165,8 @@ public function isResetPasswordLinkTokenExpired($rpToken, $rpTokenCreatedAt)
 
         $expirationPeriod = $this->customerModel->getResetPasswordLinkExpirationPeriod();
 
-        $currentTimestamp = (new \DateTime())->getTimestamp();
-        $tokenTimestamp = (new \DateTime($rpTokenCreatedAt))->getTimestamp();
+        $currentTimestamp = ($this->dateTimeFactory->create())->getTimestamp();
+        $tokenTimestamp = ($this->dateTimeFactory->create($rpTokenCreatedAt))->getTimestamp();
```
orlangur: Too fast :) 2.1 still supports PHP5.6:
```
PHP Parse error:  syntax error, unexpected '->' (T_OBJECT_OPERATOR) in /home/travis/build/magento/magento2/app/code/Magento/Customer/Model/AccountManagement.php on line 1168
```

orlangur (changes requested):

app/code/Magento/Customer/Model/AccountManagement.php
adrian-martinez-interactiv4: Ouch, I forgot remove wrapping parenthesis in this one, I've reviewed also 2.2 and 2.3 PRs and they are ok. Updated.

adrian-martinez-interactiv4: I've been also playing with PhpStorm to configure code reformatting, static tests are still complaining about that, I'm fixing it.

PR #11613, opened 2017-10-21
FR#7689_22 quote_address_id not copied when converting quote address to order address

mrodespin:
When an order is created the quote address id are not saving in the order address table.

### Description
I modified the place order function for save the quote_addres_id in order_address table properly. It was necessary to modify the Order Address Api interface for adding new getters and setters methods and Quote Management Model Logic.

1. magento/magento2#7689: quote_address_id not copied when converting quote address to order address

### Manual testing scenarios
1. Add products to the cart
2. Place a guest order 
3. Check table order_address and detect that quote_addres_id is empty

With these modifications this field is saving properly.

### Related PRs
#11614

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Api/Data/OrderAddressInterface.php
+++ b/app/code/Magento/Sales/Api/Data/OrderAddressInterface.php
@@ -32,11 +32,6 @@ interface OrderAddressInterface extends \Magento\Framework\Api\ExtensibleDataInt
     const CUSTOMER_ADDRESS_ID = 'customer_address_id';
 
     /**
-     * Quote address ID.
-     */
-    const QUOTE_ADDRESS_ID = 'quote_address_id';
-
-    /**
      * Region ID.
      */
     const REGION_ID = 'region_id';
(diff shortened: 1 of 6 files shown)


app/code/Magento/Sales/Api/Data/OrderAddressInterface.php
```diff
@@ -182,6 +187,13 @@ public function getCountryId();
     public function getCustomerAddressId();
 
     /**
+     * Gets the quote address ID for the order address.
+     *
+     * @return int|null Quote address ID.
+     */
+    public function getQuoteAddressId();
```
dmanners: Unfortunately you cannot add these new methods to an interface without breaking BC http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#introduction-of-a-method-to-a-class-or-interface in this case I would recommend adding a new interface to handle these new methods.

dmanners (changes requested):
We will need a new interface for BC reasons. Something like `OrderQuoteAddressInterface` should work here.

app/code/Magento/Sales/Api/Data/OrderAddressInterface.php
mrodespin: Okey, i didnt know that policy, i will implement your solution and update the PR as soon as i a can. Thanks!! @dmanners

dmanners:
@mrodespin unfortunately the new methods to the class marked as `@api` cannot be done in a patch release so we cannot accept these changes to `2.2-develop` and `2.1-develop`. For these two releases you should be able to make the same changes with the extension attributes feature. For `2.3-develop` the new methods and interface should be ok. Please feel free to make a new pull request using the extension attributes.

PR #11617, opened 2017-10-21
Re saving product attribute

raumatbel:
### Description

To re save product attribute, the attribute code is not retrieve correctly.

1. magento/magento2#[6770](https://github.com/magento/magento2/issues/6770): M2.1.1 : Re-saving a product attribute with a different name than it's code results in an error

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
@@ -5,79 +5,95 @@
  * See COPYING.txt for license details.
  */
 
-// @codingStandardsIgnoreFile
-
 namespace Magento\Catalog\Controller\Adminhtml\Product\Attribute;
 
+use Magento\Backend\App\Action\Context;
+use Magento\Backend\Model\View\Result\Redirect;
+use Magento\Catalog\Controller\Adminhtml\Product\Attribute;
+use Magento\Catalog\Model\Product\AttributeSet\BuildFactory;
+use Magento\Catalog\Helper\Product;
+use Magento\Catalog\Api\Data\ProductAttributeInterface;
+use Magento\Catalog\Model\ResourceModel\Eav\AttributeFactory;
+use Magento\Eav\Model\Entity\Attribute\Set;
+use Magento\Eav\Model\Adminhtml\System\Config\Source\Inputtype\Validator;
+use Magento\Eav\Model\Adminhtml\System\Config\Source\Inputtype\ValidatorFactory;
+use Magento\Eav\Model\ResourceModel\Entity\Attribute\Group\CollectionFactory;
+use Magento\Framework\Cache\FrontendInterface;
 use Magento\Framework\Controller\ResultFactory;
+use Magento\Framework\Controller\Result\Json;
 use Magento\Framework\Exception\AlreadyExistsException;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Filter\FilterManager;
+use Magento\Framework\Registry;
+use Magento\Framework\View\LayoutFactory;
+use Magento\Framework\View\Result\PageFactory;
 
 /**
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
-class Save extends \Magento\Catalog\Controller\Adminhtml\Product\Attribute
+class Save extends Attribute
 {
     /**
-     * @var \Magento\Catalog\Model\Product\AttributeSet\BuildFactory
+     * @var BuildFactory
      */
     protected $buildFactory;
 
     /**
-     * @var \Magento\Framework\Filter\FilterManager
+     * @var FilterManager
      */
     protected $filterManager;
 
     /**
-     * @var \Magento\Catalog\Helper\Product
+     * @var Product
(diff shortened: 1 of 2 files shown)


app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
```diff
@@ -124,9 +141,18 @@ public function execute()
                 }
             }
 
+            /* @var $model ModelAttribute */
+            $model = $this->attributeFactory->create();
+
             $attributeId = $this->getRequest()->getParam('attribute_id');
-            $attributeCode = $this->getRequest()->getParam('attribute_code')
-                ?: $this->generateCode($this->getRequest()->getParam('frontend_label')[0]);
+            if($attributeId){
```
orlangur: Please remove ` @codingStandardsIgnoreFile` in the top of file, such formatting violates PSR-2 but it is not checked until there is an annotation.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
```diff
@@ -124,9 +141,18 @@ public function execute()
...
+                $model->load($attributeId);
+                $attributeCode = $model->getAttributeCode();
+            }else{
+                $attributeCode = $this->getRequest()->getParam('attribute_code')
```
orlangur: Are we actually passing `attribute_code` in some scenarios in current implementation?

app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
```diff
@@ -9,75 +9,92 @@
...
+        FrontendInterface $attributeLabelCache,
+        Registry $coreRegistry,
+        PageFactory $resultPageFactory,
+        BuildFactory $buildFactory,
+        AttributeFactory $attributeFactory,
+        ValidatorFactory $validatorFactory,
+        CollectionFactory $groupCollectionFactory,
+        FilterManager $filterManager,
+        Product $productHelper,
+        LayoutFactory $layoutFactory
```
orlangur, line 96: Thanks for cleanup 👍

orlangur (changes requested):

app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
raumatbel: After of this, in this function checks if the attributeCode is valid or not using Zend_Validate_Regex. I think that this I can improve it in the next commit.

orlangur: I mean, is `attribute_code` present in UI or it is only auto-generated and then displayed as disabled field?

orlangur: Ah, ok, just checked admin may enter attribute code manually when creating new attribute.

raumatbel:
@orlangur can you check this PR with the new changes please?

orlangur:
@raumatbel sure, just need some time to understand the whole logic of the method, will have time for review in 5-7 hours.

app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php
```diff
@@ -127,8 +143,10 @@ public function execute()
             $attributeId = $this->getRequest()->getParam('attribute_id');
             $attributeCode = $this->getRequest()->getParam('attribute_code')
                 ?: $this->generateCode($this->getRequest()->getParam('frontend_label')[0]);
-            if (strlen($attributeCode) > 0) {
-                $validatorAttrCode = new \Zend_Validate_Regex(['pattern' => '/^[a-z\x{600}-\x{6FF}][a-z\x{600}-\x{6FF}_0-9]{0,30}$/u']);
+            if (!$attributeId && strlen($attributeCode) > 0) {
```
orlangur: I believe it would be safer to validate attribute code even for existing attributes like it was before.

orlangur: Basically here is the appropriate logic change as to me: https://github.com/magento/magento2/issues/6770#issuecomment-255696090

We just should not obtain attribute code from request or generate it for already existing attributes.

orlangur (changes requested):

raumatbel:
@orlangur Codacy has failed by an error in the name of parameter of Magento. That name has not been modified by me.

orlangur:
@raumatbel surely, Codacy can be ignored safely. Changes looks good to me, I first was concerned about the `$model->load` but then noticed you simply moved it from the very bottom like it was before and not introduced by yourself.

Please squash them into single commit.

raumatbel:
Yes @orlangur. I have moved the model to up. I have not used repository as in the example to don't add new param in the constructor.

Done in 1 commit.

Also, I have changed ```$this->messageManager->addError()``` by ```$this->messageManager->addErrorMessage()``` and the same with **succesMessage()** and **exceptionMessage()**

orlangur (approved):

PR #11628, opened 2017-10-22
2.2 develop 11032 Unable to add new options to swatch attribute

springerin:
ProductAttributeOptionManagementInterface#add() doesn't do anything.

### Description

On the beforeBeforeSave(Attribute $attribute) in the case of a swatch attribute, expects an array with key swatchtext / swatchvisual ( first in case of textswatch, second in case of visualswatch) that overwrites the option array originally given, so if this is empty, it will be emptied as well. So I added a plugin that right before doing this, creates the expected array that will be used later for the creation of the actual option. 

1. magento/magento2#11032: Unable to add new options to swatch attribute
2. magento/magento2#9410: Impossible to add swatch options via Service Contracts if there is no existing swatch option for attribute

### Manual testing scenarios

1.  Call the rest api http://<magento-url>/index.php/rest/V1/products/attributes/color/options to add a new color option with the following input 
`{
  "option": {
    "label": "new color",
    "value": "",
    "sort_order": 0,
    "is_default": true,
    "store_labels": [
      {
        "store_id": 0,
        "label": "new color"
      },
      {
        "store_id": 1,
        "label": "new color"
      }
    ]
  }
}`
2. Check from the admin area that your new option is there or call the rest api http://<magento-url>/index.php/rest/V1/products/attributes/color/options

### Contribution checklist


springerin:
gonna fix what failed this weekend.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+{
+    const SWATCH_VISUAL = 'visual';
+    const SWATCH_TEXT = 'text';
+    const PREFIX_OPTION = 'option';
+    const PREFIX_SWATCH = 'swatch';
+
+    /**
+     * @var Config
+     */
+    protected $eavConfig;
```
dmanners: since we have a new class here the default is to stay away from protected in properties and functions. Recommended is private unless the method is needed outside the class.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+     * @param Config           $eavConfig
+     * @param SwatchHelper     $swatchHelper
+     * @param OptionCollection $attrOptionCollectionFactory
+     */
+    public function __construct(
+        Config $eavConfig,
+        SwatchHelper $swatchHelper,
+        OptionCollection $attrOptionCollectionFactory
+    ) {
+        $this->eavConfig                   = $eavConfig;
```
dmanners: spacing looks a bit off here

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public function beforeAdd(OptionManagement $subject, $entityType, $attributeCode, $option)
+    {
+        $attribute = $this->eavConfig->getAttribute(Product::ENTITY, $attributeCode);
+
+        $isSwatch = $this->isSwatch($attribute);
+        $optionKey = self::PREFIX_OPTION . $isSwatch;
+        $swatchKey = self::PREFIX_SWATCH . $isSwatch;
+
+        if (!empty($isSwatch) && $attribute->getData($optionKey) === null
```
dmanners: Could you maybe update this to look as follows for readability:

```
if (
    !empty($isSwatch)
    && $attribute->getData($optionKey) === null
    && $attribute->getData($swatchKey) === null
) {
```

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+     * @param $attributeId
+     * @param $optionKey
+     *
+     * @return array
+     */
+    protected function prepareAttributeDataForNewOption($attributeId, $optionKey)
+    {
+        $options       = $this->getOptionsByAttributeIdWithSortOrder($attributeId);
+        $attributeData = $this->getOptionsForSwatch($options, $optionKey);
+        return $attributeData;
```
dmanners: You could just `return $this->getOptionsForSwatch($options, $optionKey);` here rather than set `$attributeData`.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * @param $optionId
+     *
+     * @return array
+     */
+    protected function getStoreLabels($optionId)
+    {
+        $optionCollection = $this->optionCollection->create();
+        $connection                  = $optionCollection->getConnection();
```
dmanners: Again spacing seems a bit off

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+     * @param OptionCollection $attrOptionCollectionFactory
+     */
+    public function __construct(
+        Config $eavConfig,
+        SwatchHelper $swatchHelper,
+        OptionCollection $attrOptionCollectionFactory
+    ) {
+        $this->eavConfig                   = $eavConfig;
+        $this->swatchHelper                = $swatchHelper;
+        $this->optionCollection = $attrOptionCollectionFactory;
```
dmanners: Technically at this point this is the factory still right? Might be good to keep that obvious in the naming of the property.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+
+        return [$entityType, $attributeCode, $option];
+    }
+
+    /**
+     * @param $attribute
+     *
+     * @return null|string
+     */
+    protected function isSwatch($attribute)
```
dmanners: The naming of this method is a bit confusing to me. `is` would suggest to me that it will return a boolean but it actually returns a mix of string and null. I think it might be a bit more readable if there was two methods, 1 for returning a boolean and one for getting the swatch type or something like that.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+
+        return $optionsArray;
+    }
+
+    /**
+     * @param $optionId
+     *
+     * @return array
+     */
+    protected function getStoreLabels($optionId)
```
dmanners: Part of me thinks that both this method and `getOptionsByAttributeIdWithSortOrder` should not be in the plugin but in another class.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
```diff
@@ -0,0 +1,195 @@
...
+        return $attributeData;
+    }
+
+    /**
+     * @param []     $options
+     * @param string $optionKey
+     *
+     * @return array
+     */
+    protected function getOptionsForSwatch($options, $optionKey)
```
dmanners: we can make sure an array is passed in here with `array $options` if it is always going to be an array.

dmanners (changes requested):
There are a couple of comments that should be addressed, also I am still not 100% sure about this PR if it is the "correct" way to fix this bug.

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed for getOptionsByAttributeWithSortOrder but still to be fixed for storeLabels

app/code/Magento/Swatches/Plugin/Catalog/AddSwatchDataToAddOption.php
springerin: fixed

dmanners:
@springerin the following API test is failing. `dev/tests/api-functional/testsuite/Magento/Catalog/Api/ProductAttributeOptionManagementInterfaceTest.php` could you have a quick look into this failure for me?

springerin:
sure I will check it out today and let you know.

EliasZ:
:+1:  It is just embarrassing what kind of code you currently need to add a swatch.

springerin:
locally tested with the last updates and still green for me. Waiting for more info to reproduce the error.

dmanners:
@springerin it seems that this has now be fixed by https://github.com/magento/magento2/pull/12036. Using the `2.2-develop` branch I followed your test case and now the swatch options have been added.

PR #11632, opened 2017-10-22
Fix #8820 #8171 check typeId on product save

### LINKED ISSUE #8820: WEBAPI: Unable to save Product when typeId and status is missing(Magento 2.1.5.)
### Preconditions
Magento 2.1.5 SOAP API

### Steps to reproduce
1. Create a product without typeId and status (see attached XML request)
2. Send the XML to Magento
3. You will receive error;: cannot save product
4. Product is created in backend!

image

### Expected result
1. The product should not be created in the backend
image

### Actual result
Product is created in backend!

image

gomencal:
… #8820 #8171 
When saving a product, check typeId control. If it's value isn't one of available, set the default one before saving.

### Description
Set the default product type before saving it on repository, for prevent error on create stockitem and the undesired attributes deleting on \Magento\Catalog\Model\Product\Type\AbstractType::_removeNotApplicableAttributes()

### Fixed Issues
1. magento/magento2#8820: WEBAPI: Unable to save Product when typeId and status is missing
2. magento/magento2#8171 Price not saved when creating product via REST API

### Manual testing scenarios
Same testing described at issues #8820 and #8171. Save product via API.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ImageExtractor.php
+++ b/app/code/Magento/Catalog/Model/ImageExtractor.php
@@ -5,10 +5,11 @@
  */
 namespace Magento\Catalog\Model;
 
-use Magento\Catalog\Model\Product\Attribute\Backend\Media\ImageEntryConverter;
 use Magento\Catalog\Helper\Image;
+use Magento\Catalog\Model\Product\Attribute\Backend\Media\ImageEntryConverter;
+use Magento\Framework\View\Xsd\Media\TypeDataExtractorInterface;
 
-class ImageExtractor implements \Magento\Framework\View\Xsd\Media\TypeDataExtractorInterface
+class ImageExtractor implements TypeDataExtractorInterface
 {
     /**
      * Extract configuration data of images from the DOM structure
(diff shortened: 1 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11632) <br/>All committers have signed the CLA.

orlangur:
> There is a problem when trying to save a product with typeId "" (empty string).

Where this problematic  code is located? What is current behavior in case of using empty typeId while saving product via Admin UI?

gomencal:
When saving a product with typeId "" a exception is launched, in version 2.2 it can be seen easily:

> Fatal error: Uncaught Error: Call to undefined method Magento\Catalog\Model\Product\Type\Simple::getTypeId() in /Users/misterge/Sites/magento2_2.2-develop/app/code/Magento/CatalogInventory/Model/Stock/StockItemRepository.php on line 161

If typeId isn't passed the default type (simple) is set during save and the exception is not launched. So I think the easiest way is to ensure that the typeId passed is one of the available.

orlangur:
@gomencal thanks for checking. Why not simply disable empty `typeId` in WebAPI scheme if it never worked as expected?

gomencal:
RIght, I thought that, but I considered useful to check also that `typeId` is valid, not only the empty value.

orlangur:
>  check also that typeId is valid, not only the empty value.

That is the part which scared me the most in this PR actually :) I don't think there is any need in such validation, `ProductRepository` class is quite bloated already. Let's just fix one simple thing - empty `typeId` leading to incorrect entities created via API.

gomencal:
I understand, I tried to get the typeId list without injecting it in constructor but I wasn't able to make it so -in a gentle way-.
I changed my PR to check only empty `typeId`. I don't change webapi because I consider it a very simple complexity check and we don't force the client to make a validation on its side. 
Many thanks @orlangur ! :)

gomencal:
Hi @orlangur , I changed my PR to set the empty product type to the default one (_simple_) because I've seen working on issue #8171 that when running `\Magento\Catalog\Model\Product\Type\AbstractType::_removeNotApplicableAttributes`
on BeforeSave it unsets the attributes when typeId is not set at the product.

gomencal:
I'm sorry, I didn't mean to close this pr. I created new one unified:
https://github.com/magento/magento2/pull/12039

PR #11634, opened 2017-10-22
Change Precisions to 4 to solve #10790

jonashrem:
Change Precisions to 4 to solve #10790

### Description
I changed the rounding precision in app/code/Magento/Directory/Model/PriceCurrency.php to 4, to avoid issues with rounding in some situations. See   #10790

solves  #10790

### Manual testing scenarios
Download the Dump provided here https://github.com/magento/magento2/issues/10790#issuecomment-331135502
Install Magento and use this dump as Database
Add 2 of each Product A and B to you shopping card
Apply Code "100percent"
Total Order should be 0

### Contribution checklist

### DIFF
--- a/app/code/Magento/Directory/Model/PriceCurrency.php
+++ b/app/code/Magento/Directory/Model/PriceCurrency.php
@@ -155,6 +155,6 @@ class PriceCurrency implements \Magento\Framework\Pricing\PriceCurrencyInterface
      */
     public function round($price)
     {
-         return round($price, 4);
+        return round($price, 4);
     }
 }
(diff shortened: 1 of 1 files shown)


jonashrem:
Can please someone more experienced have a look at this?

After my investigation, I think this is the only way to avoid the issue and as far as I can tell, this works, but it could break functionality somewhere else.

orlangur:
Hi @jonashrem, although this change may fix some particular scenarios, it may break some other which work correctly now. Something more sophisticated and configurable is needed instead.

Please check https://github.com/magento/magento2/pull/11006#issuecomment-331516790 for more details.

jonashrem:
Hi @orlangur 

I see, thank you for the feedback.

PR #11645, opened 2017-10-23
Fix #8820 #8171 [backport 2.2] check typeId on product save

### LINKED ISSUE #8820: WEBAPI: Unable to save Product when typeId and status is missing(Magento 2.1.5.)

gomencal:
… #8820 #8171 
When saving a product, check typeId control. If it's value isn't one of available, set the default one before saving.

### Description
Set the default product type before saving it on repository, for prevent error on create stockitem and the undesired attributes deleting on \Magento\Catalog\Model\Product\Type\AbstractType::_removeNotApplicableAttributes()

### Fixed Issues
1. magento/magento2#8820: WEBAPI: Unable to save Product when typeId and status is missing
2. magento/magento2#8171 Price not saved when creating product via REST API

### Manual testing scenarios
Same testing described at issues #8820 and #8171. Save product via API.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ProductRepository.php
+++ b/app/code/Magento/Catalog/Model/ProductRepository.php
@@ -7,6 +7,7 @@
 namespace Magento\Catalog\Model;
 
 use Magento\Catalog\Api\Data\ProductInterface;
+use Magento\Catalog\Api\ProductTypeListInterface;
 use Magento\Catalog\Model\Product\Gallery\MimeTypeExtensionMap;
 use Magento\Catalog\Model\ResourceModel\Product\Collection;
 use Magento\Framework\Api\Data\ImageContentInterface;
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/Model/ProductRepository.php
```diff
@@ -326,6 +326,10 @@ private function cacheProduct($cacheKey, \Magento\Catalog\Api\Data\ProductInterf
      */
     protected function initializeProductData(array $productData, $createNew)
     {
+        if (empty($productData['type_id'])) {
+            $productData['type_id'] = \Magento\Catalog\Model\Product\Type::DEFAULT_TYPE;
```
orlangur, line 330: Hi @gomencal,

So, we don't have any kind of schema validation for WebAPI, right? Like a way to forbid empty type_id.

Where exactly error occurs in case of empty type_id? Maybe it's better to put such assignment there instead of repository?

orlangur (changes requested):

orlangur:
Closing due to inactivity. @gomencal, feel free to reach me out anytime later if you wish to continue work on this pull request and it will be reopened.

PR #11651, opened 2017-10-23
[BUGFIX] Solved error while upgrading from 2.1 to 2.2

lewisvoncken:
Given error while running setup:upgrade:

	`postcode is a required field`

Problem is solved by setting tax_postcode to * if it was NULL

### Description

When you try to upgrade from 2.1 to 2.2 and you have the following two tax rates in your 2.1 installation:

**Rate 1:**
- tax_calculation_rate_id = 1
- tax_country_id = US
- tax_region_id = 43
- tax_postcode = NULL
- code = US-CA-*-Rate 1
- rate = 8.25
- zip_is_range = NULL
- zip_from = NULL
- zip_to = NULL

**Rate 2:**
- tax_calculation_rate_id = 5
- tax_country_id = US
- tax_region_id = 12
- tax_postcode = NULL
- code = US-NY-*-Rate 1
- rate = 8.3750
- zip_is_range = NULL
- zip_from = NULL
- zip_to = NULL

Insert query for these rates is:

```INSERT INTO `tax_calculation_rate` (`tax_calculation_rate_id`, `tax_country_id`, `tax_region_id`, `tax_postcode`, `code`, `rate`, `zip_is_range`, `zip_from`, `zip_to`) VALUES
(1, 'US', 12, NULL, 'US-CA-*-Rate 1', 8.2500, NULL, NULL, NULL),
(5, 'US', 43, NULL, 'US-NY-*-Rate 1', 8.3750, NULL, NULL, NULL)```

1. magento/magento2#11095: Magento_Tax "postcode is a required field" when upgrading from 2.1.9 to 2.2

### Manual testing scenarios

1. Insert or change rules so they are the same as the above rates
2. Set setup_module version Magento_Tax from 2.0.2 to 2.0.1 (this will simulate an upgrade from M2.1 to M2.2 for the Magento_Tax module)
3. Run setup:upgrade and the error is given

Apply changes and the data will be changed automatically and this manual query is no longer needed:

`UPDATE tax_calculation_rate SET tax_postcode = '*' WHERE tax_postcode IS NULL`

tax_calculation_rate table before:
image

tax_calculation_rate table after:
image

### Contribution checklist

### DIFF
--- a/app/code/Magento/Tax/Setup/UpgradeData.php
+++ b/app/code/Magento/Tax/Setup/UpgradeData.php
@@ -81,7 +81,7 @@ class UpgradeData implements UpgradeDataInterface
                 false
             );
         }
-        if (version_compare($context->getVersion(), '2.0.2', '<')) {
+        if (version_compare($context->getVersion(), '2.0.3', '<')) {
             //Update the tax_region_id
             $taxRateList = $this->taxRateRepository->getList($this->searchCriteriaFactory->create());
             /** @var \Magento\Tax\Api\Data\TaxRateInterface $taxRateData */
--- a/app/code/Magento/Tax/Setup/UpgradeData.php
+++ b/app/code/Magento/Tax/Setup/UpgradeData.php
@@ -91,6 +91,9 @@ class UpgradeData implements UpgradeDataInterface
                     /** @var \Magento\Directory\Model\Region $region */
                     $region = $this->directoryRegionFactory->create();
                     $region->loadByCode($regionCode, $taxRateData->getTaxCountryId());
+                    if ($taxRateData->getTaxPostcode() === null) {
+                        $taxRateData->setTaxPostcode('*');
+                    }
                     $taxRateData->setTaxRegionId($region->getRegionId());
                     $this->taxRateRepository->save($taxRateData);
                 }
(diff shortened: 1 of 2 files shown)


orlangur:
Any idea when it was broken? Seems like a https://github.com/magento/magento2/issues/10138 issue but for a different field.

So, maybe `null` value needs to be treated same as asterisk instead of changing data in database.

lewisvoncken:
@orlangur I think it already does and if you try to edit it through the admin panel it is a required field. I think it only happens when you migrate from Magento 1 to Magento 2. Besides that the error in #10138 is because of bad rate amount which has impact on Tax Calculation and this data change has no impact because if it is NULL it used as *

orlangur:
Yeah, `Zip/Post Code` is required even in `2.1.7` instance I have in my hand.

`2.0.2` data upgrade is idempotent, right?

We cannot change `UpgradeData` the way you proposed as it won't fix already upgraded installations. However, as `2.0.2` logic seems fine when running multiple times, to avoid unnecessary two cycles you can bump module data version to `2.0.3` and change corresponding condition in `UpgradeData`.

lewisvoncken:
@orlangur thank you for all the advice and I just updated according the feedback

orlangur (approved):

PR #11658, opened 2017-10-23
[TASK] Added dispatch event so it easy to add a custom attribute to t…

lewisvoncken:
…he SalesRule Address Conditions in a Third Party Module

### Description

Currently when you want to add a custom address attribute to the SalesRule Conditions select you have to write a plugin which sets different values. But it would be easier to only add new values instate of updating the existing error so I have added a simple dispatch event so Third Party Modules can easily add extra Address Attribute Conditions

The added dispatch event is:

`salesrule_rule_condition_address_custom_attribute_options`

### Contribution checklist

### DIFF
--- a/app/code/Magento/SalesRule/Model/Rule/Condition/Address.php
+++ b/app/code/Magento/SalesRule/Model/Rule/Condition/Address.php
@@ -28,11 +28,17 @@ class Address extends \Magento\Rule\Model\Condition\AbstractCondition
     protected $_paymentAllmethods;
 
     /**
+     * @var \Magento\Framework\Event\ManagerInterface
+     */
+    protected $_eventManager;
+
+    /**
      * @param \Magento\Rule\Model\Condition\Context $context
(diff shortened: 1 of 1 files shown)


orlangur:
New events MUST NOT be added when some goal is achievable using plugins-interceptors.

I don't see any problem for extensibility here, an after plugin for `loadAttributeOptions` method calling `setAttributeOption` could do the trick.

lewisvoncken:
@orlangur the only problem I see with that it is possible that a Third Party module will reset the default attribute options

orlangur:
> a Third Party module will reset the default attribute options

Sometimes it may be the whole point of customization. Event introduction does not prevent this reset in any way as plugins for the existing public method can still be written in third party modules.

If some options are crucial for your store workability, it's better to write an integration test to enforce their presence.

Thanks for the use case provided! It's always interesting to hear how people are customizing a particular piece of code, here it seems like core implementation is customizable enough.

PR #11660, opened 2017-10-23
[BUGFIX] Solved problem with incorrect store urls in footer and heade…

lewisvoncken:
…r of emails

### Description

Always use Use Absolute Links in email templates.

1. magento/magento2#9369: Store URLs incorrect in email templates

### Manual testing scenarios

1. Create a custom footer or header email template
2. Add {{store url="test"}} to this template
3. Configure so it is used as header or footer in the email
Design > Design > Configuration > Edit your Theme > Transactional Emails > Header or Footer
4. Place a Guest Order
5. The Url will be displayed in the email as `http://www.example.com/test/index/index`
6. Apply changes
7. Re-sent email through the Admin Interface
8. The Url will now be displayed in the email as `http://www.example.com/test`

### Contribution checklist

### DIFF
--- a/app/code/Magento/Email/Model/AbstractTemplate.php
+++ b/app/code/Magento/Email/Model/AbstractTemplate.php
@@ -654,7 +654,7 @@ abstract class AbstractTemplate extends AbstractModel implements TemplateTypesIn
     {
         if (empty($this->templateFilter)) {
             $this->templateFilter = $this->getFilterFactory()->create();
-            $this->templateFilter->setUseAbsoluteLinks($this->getUseAbsoluteLinks())
+            $this->templateFilter->setUseAbsoluteLinks(true)
                 ->setStoreId($this->getDesignConfig()->getStore())
                 ->setUrlModel($this->urlModel);
         }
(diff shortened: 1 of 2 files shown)


lewisvoncken:
@okorshenko if this pr is approved I will backport it for 2.2 and 2.1

okorshenko:
Hi @lewisvoncken thank you for pull request. Ubnformatuntelly we can not accept it. This change of the behavior is backward incompatible. We may break someone who relays on this logic.

The root problem is with the inclusion of child template.
There is a call of `\Magento\Email\Model\AbstractTemplate::getTemplateInstance` where we are losing this flag. The flag itself is a very bad approach to implement the logic. The flag must be replaced with the parameter.
To achieve the same behavior you can use `{{store direct_url='privacy'}}` I think it should work

Also, I believe you saw this comment, but I want to highlight it again.

```
the "_absolute" key is not supported by underlying services
probably this happened because of multitude of refactorings in past
The original intent of _absolute parameter was to simply append specified path to a base URL
bypassing any kind of processing.
For example, normally you would use {{view url="css/styles.css"}} directive which would automatically resolve
into something like http://example.com/pub/static/area/theme/en_US/css/styles.css
But with _absolute, the expected behavior is this: {{view url="favicon.ico" _absolute=true}} should resolve
into something like http://example.com/favicon.ico

To fix the issue, it is better not to maintain the _absolute parameter anymore in undrelying services,
but instead just create a different type of directive, for example {{baseUrl path="favicon.ico"}}
```

PR #11668, opened 2017-10-24
#10457 System config backend_model not used when reading from ScopeConfigInterface::getValue()

adrian-martinez-interactiv4:
System config backend_model not used when reading from ScopeConfigInterface::getValue(), as explained in #10457.

This PR is a proposal of how to solve this problem. Any other idea will be welcome. Testing is missing, I'll add afterwards if this proposal is accepted as the way to solve the problem.

### Description
This PR uses a **config postprocessor** to transform values through their backend models, if they have one defined in their system.xml definition.
For this purpose, a new indexer has been added, to feed the config postprocessor. This new indexer `core_config_field_data` retrieves field related data defined in indexer.xml file via `\Magento\Config\Model\Config\Structure\Reader::read($scope = null)` method, and stores it in a flat table. It is done this way, and not directly in the postprocessor for several reasons:
- This index is not tied to any model or event, and must be run manually or by schedule, but always asynchronously, not at the same time config is generated.
- Calls to `\Magento\Config\Model\Config\Structure\Reader::read($scope = null)` are expensive, and since changes in system.xml to redefine backend models aren't made frequently, there is **no need to perform this operation every time config cache gets invalidated** and config gets collected before being cached.
- Also, because of the previous, if cache is being generated during frontend navigation, it would result in an additional delay for the customer every time config cache gets erased.
- Fields defined in indexer.xml allows to extend and redefine indexer fieldset to save other config field data defined in system.xml files to be used elsewhere.

1. https://github.com/magento/magento2/issues/10457: System config backend_model not used when reading from ScopeConfigInterface::getValue()

### Manual testing scenarios

1. Get this Gist: https://gist.github.com/adrian-martinez-interactiv4/7686616361c9d4c4b66b3f80f616eeae, put it in the project base folder. It simulates frontend area to retrieve `newrelicreporting/general/insights_insert_key` config info via scopeConfig object. We are using this config field as an encrypted field example, using backend model to encrypt / decrypt data.

#### Initial status (before applying changes, in 2.3-develop branch)

2. In Admin -> Stores -> Configuration, Under General tab, go to New Relic Reporting section.

3. In General group, Insights API Key field, type `demo api key` or whatever you want, and save.

4. As expected, it is encrypted in bbdd:
captura de pantalla 2017-10-24 a las 2 50 47

5. Clean cache and run provided Gist; in frontend area, it still appears encrypted, because config value has not been processed by backend model:
captura de pantalla 2017-10-24 a las 2 55 23

#### Final status (after applying proposed changes)

6. Apply changes proposed in this PR, run `bin/magento setup:upgrade`and clean cache. A new table should appear, core_config_field_data_flat. You may need to reindex it:
captura de pantalla 2017-10-24 a las 3 01 28

7. You should see something like this (filtered search):
captura de pantalla 2017-10-24 a las 2 59 25

8. Clean cache and run provided Gist; in frontend area, it should appear unencrypted, as backend model has processed it:
captura de pantalla 2017-10-24 a las 3 05 43

For any question, suggestion, mention me in comments pls.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Config/App/Config/Type/System.php
+++ b/app/code/Magento/Config/App/Config/Type/System.php
@@ -5,9 +5,9 @@
  */
 namespace Magento\Config\App\Config\Type;
 
+use Magento\Config\App\Config\Type\System\Reader;
 use Magento\Framework\App\Config\ConfigTypeInterface;
 use Magento\Framework\App\ObjectManager;
-use Magento\Config\App\Config\Type\System\Reader;
 
 /**
  * System configuration type
@@ -146,17 +146,16 @@ class System implements ConfigTypeInterface
             }
             return $this->data[$pathParts[0]];
         }
+
         $scopeType = array_shift($pathParts);
         if ($scopeType === 'default') {
-            if (!isset($this->data[$scopeType])) {
-                $this->data = array_replace_recursive($this->loadDefaultScopeData($scopeType), $this->data);
-            }
+            $this->data = array_replace_recursive($this->data, $this->loadDefaultScopeData($scopeType));
             return $this->getDataByPathParts($this->data[$scopeType], $pathParts);
         }
+
         $scopeId = array_shift($pathParts);
-        if (!isset($this->data[$scopeType][$scopeId])) {
-            $this->data = array_replace_recursive($this->loadScopeData($scopeType, $scopeId), $this->data);
-        }
+        $this->data = array_replace_recursive($this->data, $this->loadScopeData($scopeType, $scopeId));
+
         return isset($this->data[$scopeType][$scopeId])
             ? $this->getDataByPathParts($this->data[$scopeType][$scopeId], $pathParts)
             : null;
--- a/app/code/Magento/Config/Model/Processor/BackendModel.php
+++ b/app/code/Magento/Config/Model/Processor/BackendModel.php
@@ -1,0 +1,176 @@
+<?php
+/**
+ * Backend model configuration values processor. Process configuration values with backend models if needed.
+ *
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Config\Model\Processor;
+
+use Magento\Config\Model\Config\Structure\Reader as ConfigReader;
+use Magento\Framework\App\Area;
+use Magento\Framework\App\Config\Spi\PostProcessorInterface;
+
+/**
+ * Backend model configuration values processor. Process configuration values with backend models if needed.
+ */
+class BackendModel implements PostProcessorInterface
+{
+    /**
+     * @var \Magento\Framework\ObjectManagerInterface
+     */
+    private $objectManager;
+
+    /**
+     * @var ConfigReader
+     */
+    private $configReader;
+
+    /**
+     * Avoid circular dependencies while creating backend models,
+     * some of them rely on configuration, running this processor again and trying to create again same
+     * backend model
+     *
+     * @var bool
+     */
+    private $processed = false;
+
+    /**
+     * BackendModel constructor.
+     * @param \Magento\Framework\ObjectManagerInterface $objectManager
+     * @param ConfigReader $configReader
+     */
+    public function __construct(
+        \Magento\Framework\ObjectManagerInterface $objectManager,
+        ConfigReader $configReader
+    ) {
+        $this->objectManager = $objectManager;
+        $this->configReader = $configReader;
+    }
+
+    /**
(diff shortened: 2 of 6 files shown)


sidolov:
Hi @adrian-martinez-interactiv4, looks like described behavior caused by configuration mismatch. In case with New Relic Reports you should declare backend_model in config.xml file for "insights_insert_key" field: 
`<insights_insert_key backend_model="Magento\Config\Model\Config\Backend\Encrypted"/>`

after this fix expression `$this->scopeConfig->getValue('newrelicreporting/general/insights_insert_key')` will return decrypted key.

Could you please take a look at provided solution and leave feedback?
Thanks!

sidolov:
Issue https://github.com/magento/magento2/issues/10457 was closed by reporter as not relevant

adrian-martinez-interactiv4:
Hi @sidolov , thanks for pointing out this solution. I already knew this posibility when I started to develop this PR. As you also point out, maybe it makes sense to show raw value for some cases, but since for admin area backend models are always used (because of loading of system.xml files in that area), not using them in frontend seems more like a bug than a feature.

The main point of changes proposed in this PR are the following:
- I've heard people complaining about this behaviour since Magento 1, where you could also apply this solution, so I've tried to find a solution that does not involves modifying config.xml file, and works globally.
-  **Avoid having defined twice the backend model**, once in system.xml and another in config.xml. I see this solution as a workaround for a inconsistency between different config files loaded in different areas.
- **Even this solution may be inconsistent too by itself**: if you change the backend model in system.xml and leave it unchanged it in config.xml, results are unpredictable, and you may not get the expected value. You'll have to maintain two backend model definitions (or more) for each field, and it is not easy to find the references in config.xml files given a config field, when changing a system.xml backend model definition.

Having this in mind, we could discuss what is a more place to put that backend model definition. From my point of view, config.xml file is only used to preset default config values, so **config.xml should not be responsible of defining the config field backend model**, as the whole config field definition is put in system.xml.

If we agree in the previous, we find that when config values are loaded to be saved in cache from the frontend area, system.xml files aren't loaded, so there was no way to know the backend model for a config field in the frontend area. That is what this PR tries to solve, aside of the concrete problem described in the issue https://github.com/magento/magento2/issues/10457. This PR is related to that issue because it would also solve that issue without modifying the config.xml file, and that's why I hadn't proposed that solution at first instance.

sidolov:
Hi @adrian-martinez-interactiv4, I understand that two different places for field configuration more difficult and may cause unpredictable results but at the same time this solution more flexible (developer can define two different backend models for configuration page and frontend usage).

I agree that we may have some kind of fallback to default backend_model defined in system.xml but after implementation fallback mechanism we will need to verify all fields usages because, I think, we have a lot of usages raw values that will be decoded, unserialized, etc in client code. 

About provided solution: it's a bad idea to use indexer for static data, for this purpose we use file cache mechanism. Indexers using only for dynamic data, like prices, stocks etc. 

So, if we want have described fallback mechanism, we need additionally read system.xml configuration, retrieve backend_model property and use it if this property not defined in config.xml and as I say above - verify all usages. 

In case of huge performance degradation, we need to choose preferably option: leave fast compromise variant with two places for configuration, or slow solution with fallback and more consistent behavior. This decision depends on performance measurements results.

adrian-martinez-interactiv4:
@sidolov Yep, I agree with all you said, including using of indexer for this purpose. The first approach used only config postprocessor to read system.xml files, and when I had it working, I noticed the performance degradation when trying to read system.xml files, so that was why I added the indexer, as a way to aggregate system.xml data asynchronously, not over demand, and only when really needed. If there was another place or way to aggregate this data without using an indexer, it would be a better solution.

In case indexer is discarded, and without an alternative to it, and keeping in mind that loading system.xml files on the go everytime config is reloaded has performance problems, specially for production environments, I'd say to leave this as it already is, with the fast compromise variant.

sidolov:
Hi @adrian-martinez-interactiv4, any updates on this issue?

adrian-martinez-interactiv4:
Hi @sidolov , still not, I've been busy, but I have some free time now to check this, I'll try to update you asap.

adrian-martinez-interactiv4:
Hi @sidolov , I'm facing some issues with circular dependencies, still working on it. I have also found it works properly reading system.xml files when cache is cleaned instead of flushed.

sidolov:
Hi @adrian-martinez-interactiv4 , sounds good, let me know when you finish!

sidolov:
Decided with @adrian-martinez-interactiv4 leave current behavior as compromised and find another reliable solution

PR #11670, opened 2017-10-24
save invoice ID on credit memo when using API method salesRefundInvoiceV1

AntonEvers:
### Description

- Sets the invoice ID when an invoice is set on the credit memo
- Loads the invoice by id on getInvoice if an invoice ID is set on the credit memo

1. magento/magento2#11669: API salesRefundInvoiceV1 does no save invoice ID on credit memo 

### Manual testing scenarios

1. Run testing scenario from magento/magento2#11669

### Contribution checklist


app/code/Magento/Sales/Model/Order/Creditmemo.php
```diff
@@ -1525,5 +1539,12 @@ public function setExtensionAttributes(\Magento\Sales\Api\Data\CreditmemoExtensi
         return $this->_setExtensionAttributes($extensionAttributes);
     }
 
+    /**
+     * @return InvoiceRepositoryInterface|mixed
+     * @deprecated
+     */
+    private function getInvoiceRepository() {
+        return $this->invoiceRepository ?: ObjectManager::getInstance()->get(InvoiceRepositoryInterface::class);
```
orlangur: Please avoid such methods creation and move this logic to constructor.

AntonEvers: Okay

AntonEvers:
@orlangur tests passing

ishakhsuvarov:
Fixes same issue as magento/magento2#11687

ishakhsuvarov:
@ajpevers I would suggest aligning solutions as #11687 seems to be more lightweight (however from code review only I can not see if it really fixes all cases), but this PR includes integration test, which is preferred.

AntonEvers:
Hello @ishakhsuvarov,

Although the solution there is more lightweight, it is not in line with the rest of the code. The problem in #11669 brings to light that a credit memo - invoice relation is not saved automatically. In this case in the salesRefundInvoiceV1 API call. Other relations are always saved, such als the credit memo - order relation. No matter how or from which part of the code you save the credit memo, because it is done in the same _beforeSave as my addition. My solution is in line with the credit memo - order relation persistence code. It works in all instances. Not only in the API. Especially with these essential entities in the system I would not go for end of pipe solutions such as #11687.

As far as the unit testing goes. I would not know how to test the _beforeSave of a resource model. For now that goes beyond my knowledge of unit testing. As far as I can see the unit test for checking if the order ID is persisted on the credit memo is also lacking. So it would be a shame to keep this PR open for that reason in my opinion.

app/code/Magento/Sales/Model/Order/Creditmemo.php
```diff
@@ -379,6 +389,9 @@ public function canRefund()
      */
     public function getInvoice()
     {
+        if (!$this->getData('invoice') instanceof \Magento\Sales\Model\Order\Invoice && $this->getInvoiceId()) {
```
ishakhsuvarov: Is there a reason not to use `\Magento\Sales\Api\Data\InvoiceInterface` here?

AntonEvers: Good point. No there is no reason.

PR #11685, opened 2017-10-24
module downloadable overwrites result in checkout_allow_guest event

jaykobi:
Module Downloadable overwrites other listener's changes to $result->isAllowed in checkout_allow_guest event.

When you listen to that event and set is_allowed to false it won't make a difference because the downloadable moduke sets it to true again afterwards.

Instead, the module should not set is_allowed to true, but only set is_allowed to false if a downloadable product is found in cart.


jaykobi:
Already fixed in develop...

PR #11686, opened 2017-10-24
 Fix error when generating urn catalog for empty misc.xml

tdgroot:
Fix error when generating urn catalog for empty misc.xml

### Description
Fix error when generating urn catalog for existing empty `misc.xml`. I fixed this by creating another DomDocumentFactory, for PhpStorm specific `misc.xml`. This factory checks if the passed data is empty, and if so, create the `misc.xml` xml structure that `\Magento\Developer\Model\XmlCatalog\Format\PhpStorm` expects. 

I also changed `\Magento\Framework\DomDocument\DomDocumentFactory` to follow the Magento 2 guidelines and added a Test class for this factory to define the behavior. 

1. magento/magento2#5188

### Manual testing scenarios

1. `rm .idea/misc..xml`
2. `touch .idea/misc.xml`
3. `bin/magento dev:urn-catalog:generate .idea/misc.xml`

### Contribution checklist


tdgroot:
Codacy is not successful, because I used an `} else {` statement. Any suggestions?

dmanners:
@tdgroot thanks for the PR. I will start to process this now. Do not worry about the codacy check for the else that result seems a bit odd to me.

app/code/Magento/Developer/Model/XmlCatalog/Format/PhpStorm.php
```diff
@@ -23,20 +23,28 @@ class PhpStorm implements FormatInterface
...
      * @param ReadFactory $readFactory
-     * @param \Magento\Framework\Filesystem\File\WriteFactory $fileWriteFactory
+     * @param FileWriteFactory $fileWriteFactory
+     * @param DomDocumentFactory $domDocumentFactory
      */
     public function __construct(
         ReadFactory $readFactory,
-        \Magento\Framework\Filesystem\File\WriteFactory $fileWriteFactory
+        FileWriteFactory $fileWriteFactory,
+        DomDocumentFactory $domDocumentFactory
```
dmanners: This change will break BC try adding an optional dependency here as described in http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#adding-a-constructor-parameter

lib/internal/Magento/Framework/DomDocument/DomDocumentFactory.php
```diff
@@ -3,20 +3,45 @@
...
+    /**
      * Create empty DOM document instance.
      *
-     * @return \DOMDocument
+     * @param string $data the data to be loaded into the object
+     *
+     * @return DOMDocument
      */
-    public function create()
+    public function create($data = null)
```
dmanners: I will need to double check if this will break BC also as according to http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#adding-parameters-in-public-methods you cannot add new parameters to public methods. Though all BC breaks can be approved let me check internally here and get back to you on this one.

dmanners (changes requested):
At least 1 change to be made for BC reasons. I will get back to you about the changes to the factory class.

app/code/Magento/Developer/Model/XmlCatalog/Format/PhpStorm.php
tdgroot: I was already wondering how BC was being handled! I'll make the constructor parameter optional for BC.

dmanners: It might also be a good idea to revert the change made to `\Magento\Framework\Filesystem\File\WriteFactory $fileWriteFactory` as this is not specific to your change and I am not 100% sure how BC would work in this case.

dmanners:
Ok @tdgroot here is what I am thinking so that we do not break BC and then this can be merged into 2.2.

1. It is not recommended to use inheritance inside of Magento2 but rather composition. So what I would suggest we do here is update `app/code/Magento/Developer/Model/XmlCatalog/Format/PhpStorm/DomDocumentFactory.php` so that it does not extends the `lib/internal/Magento/Framework/DomDocument/DomDocumentFactory.php`.
2. Revert the changes to `lib/internal/Magento/Framework/DomDocument/DomDocumentFactory.php` as I am not sure how we can make these without breaking BC.
3. Add all the required methods into the new factory so that it will work as expected.

After this has been done and merged we could consider making the more detailed change into the 2.3 release in which case we can make a few more changes with regards to BC.

How does that sound to you?

tdgroot:
@dmanners thank you for your feedback. 

They seem fair points, I'll need to find some time coming week to propose the changes based on your feedback. I have one question, though. 

How did the changes in `lib/internal/Magento/Framework/DomDocument/DomDocumentFactory.php` break BC? The `$data` parameter is optional, so it should work fine for piece of code that uses the method. Or is it just against Magento's guidelines?

dmanners:
@tdgroot yes sorry that was my mistake I missed the fact it was optional. That will be fine.

tdgroot:
Hi @dmanners. I added BC to the code and tweaked some parts as well!

lib/internal/Magento/Framework/DomDocument/DomDocumentFactory.php
```diff
@@ -3,20 +3,46 @@
...
     /**
+     * @var \Magento\Framework\ObjectManagerInterface
+     */
+    private $objectManager;
+
+    /**
+     * DomDocumentFactory constructor.
+     * @param \Magento\Framework\ObjectManagerInterface $objectManager
+     */
+    public function __construct(\Magento\Framework\ObjectManagerInterface $objectManager)
```
dmanners: Do we need to make the changes to this file or are we always using the new Dom factory class? My worry here is that a change like this will break backwards compatibility.

tdgroot: I see, this breaks BC. I can make the `$objectManager` dependency optional?

dmanners: Technically this is a private class so we could choose to break BC in this case. Otherwise you could also make the object manager a static call. Another thing to note is that the factory should only be concerned with creating the object and not filling it. In this case `initializeDocument` in the other factory and calling `$dom->loadXML($data)` load in the other factory is not really the way to go in my mind.

tdgroot: You're right. Factories should only instantiate objects. Wouldn't it be better to migrate all the domain specific code back to `\Magento\Developer\Model\XmlCatalog\Format\PhpStorm`?

Sorry for all the corrections, refactorings, etc. I'm learning a lot from creating Magento 2 pull requests :).

dmanners: No worries. I see two options here.

1. Migrate all the domain specific code back to the `\Magento\Developer\Model\XmlCatalog\Format\PhpStorm`.
2. Create our own classes for the two dom documents that have the public methods that we need. Then change the new factories to instantiate these new objects.

Not sure what you would prefer to do. If you do go with the idea of new classes then please have a look through http://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines/technical-guidelines.html#2-class-design and of course let me know if you need some assistance.

dmanners: @tdgroot do you think you will have time to look into this?

tdgroot: @dmanners I'm currently too busy to get working on this PR. Next week I hope to find some time.

dmanners:
Hi @tdgroot if you would like help with this pull request I would be happy to take over and see if we can get this sorted before the end of 2018. How does that sound?

tdgroot:
@dmanners Sure! Rounding up 2017 is getting really busy unfortunately..

dmanners (approved):

PR #11690, opened 2017-10-24
Add a health check to the NGINX configuration sample

### LINKED ISSUE #11157: nginx.sample.conf missing heath_check.php?
I've updated to 2.2, and regularly get the error below in my nginx error logs. I believe it's because health_check which is used in varnish config is not catered for in the nginx sample vhost configuration

### Preconditions
1. Magento 2.2.0
2. nginx 1.12.1
3. OS X 10.12.6

### Steps to reproduce
1. Update to Magento 2.2
2. View home page of site

### Expected result
1. No errors

### Actual result
nginx error log
1. 2017/09/30 10:42:59 [error] 7848#0: *115 open() "/usr/local/Cellar/nginx/1.12.1/html/pub/health_check.php" failed (2: No such file or directory), client: 127.0.0.1, server: localhost, request: "GET /pub/health_check.php HTTP/1.1", host: "127.0.0.1"

andrewhowdencom:
With the release of the 2.2 version of Magento Varnish has been configured 
to determine the state of the upstream HTTP proxy with a HTTP probe. This
probe polls a script called "health_check.php", located in the pub folder.

Unfortunately in the sample configuration, this script is not whitelisted as a PHP
executable script, and thus fails to get executed as expected. Varnish then marks
the host as down, refusing to serve any traffic.

This commit adds the health_check script to the executable PHP block, allowing
the varnish proxy to correctly determine the health of the application.

Fixes #11157

### DIFF
--- a/nginx.conf.sample
+++ b/nginx.conf.sample
@@ -162,7 +162,7 @@ location /media/import/ {
 }
 
 # PHP entry point for main application
-location ~ (index|get|static|report|404|503)\.php$ {
+location ~ (index|get|static|report|404|503|health_check)\.php$ {
     try_files $uri =404;
     fastcgi_pass   fastcgi_backend;
     fastcgi_buffers 1024 4k;


andrewhowdencom:
Untested in this particular build; just reprod the change in our environment while debugging varnish fuckery

andrewhowdencom:
Also note this doesn't work for things that request `pub/health_check.php` (i.e. the way varnish is exported by default); it is the authors opinion that this should be handled outside the scope of this change as index.php et. al are

davidalger (approved):

PR #11709, opened 2017-10-25
#7065 Allow page.main.title block to decide if title should be translated

adrian-martinez-interactiv4:
Related to https://github.com/magento/magento2/issues/7065.

Layout block with name 'page.main.title', related with block `\Magento\Theme\Block\Html\Title`, always translates title before being returned. Unfortunately, category and product pages get their titles translated when they shouldn't (if wanted so, a different value should be defined for category / product at store view level in admin).

### Description
This may be the desired behaviour for most parts of the system, but 'page.main.title' always translates the title. Because of that, that block is now capable to handle the boolean argument `translate_title`, injected in constructor and defined via layout xml, to decide whether it has to translate the title or not.

As it used to translate all titles before, impact has been minified making `\Magento\Theme\Block\Html\Title` translate all titles by default, unless it has been told to do the opposite. This is the case of category and product pages by now, but there could be other pages interested in title not being translated.

I'm aware that arguments translation and be done via `translate` attribute, like in `Magento/Catalog/view/frontend/layout/catalog_product_compare_index.xml`:
```
<page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" layout="1column" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
    <body>
        <referenceBlock name="page.main.title">
            <action method="setPageTitle">
                <argument translate="true" name="page_title" xsi:type="string">Compare Products</argument>
            </action>
        </referenceBlock>
            <block class="Magento\Catalog\Block\Product\Compare\ListCompare" name="catalog.compare.list" template="Magento_Catalog::product/compare/list.phtml" cacheable="false"/>
        </referenceContainer>
    </body>
</page>
```

But this only works for static values, that get translated when layout files are parsed and layout is generated. For products and categories, this value is set afterwards in _prepareLayout method.

1. magento/magento2#7065: page.main.title is translating title

### Manual testing scenarios
1. As proposed in https://github.com/magento/magento2/issues/7065, I created a product with name 'Save'
2. I have installed translation packages for es_ES, and configured store to use that locale.
3. The result was the product name translated in product page view (Save => Guardar):
captura de pantalla 2017-10-25 a las 2 31 28
4. After applying this changes, product name does not get translated anymore:
captura de pantalla 2017-10-25 a las 3 08 54

Also checked home page, and it still gets it title translated.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/layout/catalog_category_view.xml
+++ b/app/code/Magento/Catalog/view/frontend/layout/catalog_category_view.xml
@@ -42,12 +42,14 @@
                 </arguments>
             </block>
         </referenceContainer>
-        <referenceBlock name="page.main.title">
-            <arguments>
-                <argument name="id" xsi:type="string">page-title-heading</argument>
-                <argument name="add_base_attribute_aria" xsi:type="string">page-title-heading toolbar-amount</argument>
-            </arguments>
-            <block class="Magento\Catalog\Block\Category\Rss\Link" name="rss.link" template="Magento_Catalog::category/rss.phtml"/>
-        </referenceBlock>
+        <referenceContainer name="columns.top">
+            <block class="Magento\Theme\Block\Html\RawTitle" name="page.main.title" template="Magento_Theme::html/title.phtml" before="-">
+                <arguments>
+                    <argument name="id" xsi:type="string">page-title-heading</argument>
+                    <argument name="add_base_attribute_aria" xsi:type="string">page-title-heading toolbar-amount</argument>
+                </arguments>
+                <block class="Magento\Catalog\Block\Category\Rss\Link" name="rss.link" template="Magento_Catalog::category/rss.phtml"/>
+            </block>
+        </referenceContainer>
     </body>
 </page>
--- a/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml
+++ b/app/code/Magento/Catalog/view/frontend/layout/catalog_product_view.xml
@@ -17,12 +17,14 @@
         <referenceBlock name="head.components">
             <block class="Magento\Framework\View\Element\Js\Components" name="checkout_page_head_components" template="Magento_Catalog::js/components.phtml"/>
         </referenceBlock>
-        <referenceBlock name="page.main.title">
-            <arguments>
-                <argument name="css_class" xsi:type="string">product</argument>
-                <argument name="add_base_attribute" xsi:type="string">itemprop="name"</argument>
-            </arguments>
-        </referenceBlock>
+        <referenceContainer name="columns.top">
+            <block class="Magento\Theme\Block\Html\RawTitle" name="page.main.title" template="Magento_Theme::html/title.phtml" before="-">
+                <arguments>
+                    <argument name="css_class" xsi:type="string">product</argument>
+                    <argument name="add_base_attribute" xsi:type="string">itemprop="name"</argument>
+                </arguments>
+            </block>
+        </referenceContainer>
         <referenceBlock name="root">
             <arguments>
                 <argument name="add_attribute" xsi:type="string">itemscope itemtype="http://schema.org/Product"</argument>
--- a/app/code/Magento/Theme/Block/Html/RawTitle.php
+++ b/app/code/Magento/Theme/Block/Html/RawTitle.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Theme\Block\Html;
+
+use Magento\Framework\View\Element\Template;
+
+/**
+ * Html page raw title block
+ *
+ * @method $this setTitleId($titleId)
+ * @method $this setTitleClass($titleClass)
+ * @method string getTitleId()
+ * @method string getTitleClass()
+ */
+class RawTitle extends Template
+{
+    /**
+     * Own page title to display on the page
+     *
+     * @var string
+     */
+    private $pageTitle;
+
+    /**
+     * Provide own page title or pick it from Head Block
+     *
+     * @return string
+     */
+    public function getPageTitle()
+    {
+        if (!empty($this->pageTitle)) {
+            return $this->pageTitle;
+        }
+        return $this->pageConfig->getTitle()->getShort();
+    }
+
+    /**
+     * Provide own page content heading
+     *
+     * @return string
+     */
+    public function getPageHeading()
+    {
+        if (!empty($this->pageTitle)) {
+            return $this->pageTitle;
+        }
+        return $this->pageConfig->getTitle()->getShortHeading();
+    }
+
+    /**
+     * Set own page title
+     *
+     * @param string $pageTitle
+     * @return void
+     */
+    public function setPageTitle($pageTitle)
+    {
+        $this->pageTitle = $pageTitle;
+    }
+}
--- a/app/code/Magento/Theme/Test/Unit/Block/Html/RawTitleTest.php
+++ b/app/code/Magento/Theme/Test/Unit/Block/Html/RawTitleTest.php
@@ -1,0 +1,113 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Theme\Test\Unit\Block\Html;
+
+use Magento\Framework\TestFramework\Unit\Helper\ObjectManager as ObjectManagerHelper;
+
+class RawTitleTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @var ObjectManagerHelper
+     */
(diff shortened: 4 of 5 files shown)


app/code/Magento/Theme/Block/Html/Title.php
```diff
@@ -27,6 +27,25 @@ class Title extends Template
...
+
+    /**
+     * Constructor
+     *
+     * @param Template\Context $context
+     * @param array $data
+     */
+    public function __construct(Template\Context $context, array $data = [])
+    {
+        $this->translateTitle = isset($data['translate_title']) ? (bool) $data['translate_title'] : true;
```
orlangur: Having flags in code to switch its behavior is almost never a good idea.
 
> category and product pages get their titles translated when they shouldn't

Why not simply introduce another title block implementation, `RawTitle` or something, and use it on mentioned pages?

General note: please always start contribution from `2.2-develop` branch, a default one in this repo.

orlangur (changes requested):

app/code/Magento/Theme/Block/Html/Title.php
adrian-martinez-interactiv4: Noted

adrian-martinez-interactiv4:
@orlangur Updated with suggested changes

app/code/Magento/Catalog/view/frontend/layout/catalog_category_view.xml
```diff
@@ -42,12 +42,15 @@
...
             </block>
         </referenceContainer>
-        <referenceBlock name="page.main.title">
-            <arguments>
-                <argument name="id" xsi:type="string">page-title-heading</argument>
-                <argument name="add_base_attribute_aria" xsi:type="string">page-title-heading toolbar-amount</argument>
-            </arguments>
-            <block class="Magento\Catalog\Block\Category\Rss\Link" name="rss.link" template="Magento_Catalog::category/rss.phtml"/>
-        </referenceBlock>
+        <referenceBlock name="page.main.title" remove="true"/>
```
orlangur: I believe I read somewhere that changing block name is backward-incompatible change but http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#magento-apis is quite vague about it.

Can we preserve `page.main.title` name but just change block type on needed places?

orlangur (changes requested):

app/code/Magento/Catalog/view/frontend/layout/catalog_category_view.xml
adrian-martinez-interactiv4: I have thought the same thing, I had problems in the past removing blocks and trying to re-add them with the same name. I'm trying now to override block class definition by setting the same block name, so full block node gets overriden during merge, I keep you updated.

orlangur: Probably it can be done via reference: http://devdocs.magento.com/guides/v2.0/frontend-dev-guide/layouts/xml-manage.html#set_template

adrian-martinez-interactiv4: I think this is for defining / changing arguments for a given block, like the template, but for a given (already instantiated) block. I've achieved changing block without changing name redefining the block, see https://github.com/magento/magento2/pull/11709/commits/584cfd33f3ab7a0bab86e2f303a28c1c95c477da

orlangur: Good! But, will replacing `class` using `referenceBlock` work? It would allow to not add additional `referenceContainer` and generally referencing is safer that redeclaring.

adrian-martinez-interactiv4: I don't think this is possible without rewriting xml node block redefining it with the same name, I've been investigating this possibility, and this is what I found.

Here is the xsd definition for `referenceBlock`, from `Magento/Framework/View/Layout/etc/elements.xsd`:
```
<xs:complexType name="blockReferenceType" mixed="true">
        <xs:choice minOccurs="0" maxOccurs="unbounded">
            <xs:element name="action" type="actionType" minOccurs="0" maxOccurs="unbounded">
                <xs:annotation>
                    <xs:documentation>
                        Argument name must be unique in scope of action.
                    </xs:documentation>
                </xs:annotation>
                <xs:key name="blockReferenceActionArgumentName">
                    <xs:selector xpath="argument"></xs:selector>
                    <xs:field xpath="@name"></xs:field>
                </xs:key>
            </xs:element>
            <xs:element ref="arguments" minOccurs="0" maxOccurs="1"/>
            <xs:element ref="block" minOccurs="0"/>
            <xs:element name="container" type="containerType" minOccurs="0"/>
            <xs:element ref="referenceBlock" minOccurs="0" />
            <xs:element ref="referenceContainer" minOccurs="0"/>
            <xs:element ref="uiComponent" minOccurs="0" />
        </xs:choice>
        <xs:attribute type="elementNameType" name="name" use="required"/>
        <xs:attribute type="xs:string" name="template" use="optional"/>
        <xs:attribute type="xs:boolean" name="display" default="true" use="optional"/>
        <xs:attribute type="xs:boolean" name="remove" use="optional"/>
    </xs:complexType>
```

To change the class, I think it should be done redefining class as a `referenceBlock` attribute: ```<referenceBlock name="page.main.title" class="Magento\Theme\Block\Html\RawTitle">```, thing that is not allowed by xsd definition (it only allows name, template, display and remove attributes for `referenceBlock`).

The rest of the xsd declaration are arguments for the already defined block class, or actions over it, or declaring children blocks, or referencing other blocks, containers or uiComponents.

I have even tried to find any example of any layout xml file trying to perform an action like this with no success. No clue of this kind of usage even in layout directives tests under Magento/Framework/View/_files/layout_directives_test.

If you still don't like the idea of doing this task the way it is done, I could still use `$this->eventManager->dispatch('layout_render_before_' . $this->request->getFullActionName());` event placed in `\Magento\Framework\View\Result\Layout::renderResult`, or tweak it in some way in `\Magento\Catalog\Block\Category\View::_prepareLayout` and `\Magento\Catalog\Block\Product\View::_prepareLayout`, but this way I think you may lose some control about this, control that you may have per theme editing layout xml files.

orlangur: > thing that is not allowed by xsd definition

Thanks, that's pretty answers my question. Current implementation with declarative approach instead of some PHP code seems best to me. Will check if there was some intentional decision to disallow class overriding.

Did class overwriting work properly by the way if you expand XSD?

adrian-martinez-interactiv4: Surprisingly, yes, it does, this way:
- At Magento/Framework/View/Layout/etc/elements.xsd:
captura de pantalla 2017-10-26 a las 17 03 36

- At Magento/Catalog/view/frontend/layout/catalog_product_view.xml:
captura de pantalla 2017-10-26 a las 17 07 11

- Result:
captura de pantalla 2017-10-26 a las 17 08 25

- Only removing class attribute from referenceBlock:
captura de pantalla 2017-10-26 a las 17 09 22

- Result:
captura de pantalla 2017-10-26 a las 17 09 58

So yes, it works. Have you found any intentional decision for this not being allowed by xsd?

orlangur: Nah, didn't ask yet. Asking now. Big thanks for checking!

adrian-martinez-interactiv4:
@orlangur I'm facing this issue with static tests, new block is expected to be marked as @api if referenced in default layouts:
captura de pantalla 2017-10-26 a las 17 53 28

I thought I wasn't supposed to add @api annotation for any added class, what should I do?

orlangur:
Yeah-yeah, I noticed it on Bamboo as well, forgot we already run all static tests on Travis :)

No need to do anything from your side, I'll ask if architect approval is required and add `@api` on our side if we did it right.

ihor-sviziev:
Hi @orlangur ,
Are you interested in finalizing this PR? Any results after discussing with architecs?

sivaschenko:
According to @orlangur this pull request can be substantially improved after the merge of https://github.com/magento/magento2/pull/18933

orlangur (changes requested):
Hi there @adrian-martinez-interactiv4, the time has come to rework implementation considering https://github.com/magento/magento2/pull/18933 😎

sidolov:
@adrian-martinez-interactiv4 , @orlangur hey guys, any progress on this?

orlangur:
@sidolov I've reached Adrian in Slack yesterday, he is now aware of this PR and there are high chances to update it :)

sivaschenko:
@adrian-martinez-interactiv4 , @orlangur any updates?

sidolov:
@adrian-martinez-interactiv4 , I am closing this PR now due to inactivity.

PR #11715, opened 2017-10-25
magento/magento2#80862 Custom Product Attribute changes backend_type property

keyurshah070:

1. magento/magento2#9219: Custom Product Attribute changes 'backend_type' when 'is_user_defined = 1' and get updated/saved in Admin Backend
2. User define product attribute backend type change from backend save.
3. So if we have backend type as varchar(on Installer script) with its value in a string and after creating attribute when we update its property from backend then it changes backend type to integer. so we cannot save its attribute value on a product level

### Manual testing scenarios

1. Created Installer script with backendtype as varchar and change its property from backend and after saving attribute still, I have backendtype as varchar

### Contribution checklist


omiroshnichenko:
Hi, @keyurshah070. Could you please verify failed tests. Also you have opened concurrent PR #11837 that will have conflict with current. Can you, please, let us know if you would continue working on it or we should close it. Thank you.

PR #11728, opened 2017-10-25
Check variable existence in prepareOptionIds(array) in EavAttribute.php

angelo983:
### Description
The existence of $optionsArray['delete'] does not guarantee the existence of $optionsArray['delete'][$optionId]

1. Php 500 Error when $optionsArray['delete'][$optionId] is not set

### DIFF
--- a/app/code/Magento/Swatches/Model/Plugin/EavAttribute.php
+++ b/app/code/Magento/Swatches/Model/Plugin/EavAttribute.php
@@ -174,7 +174,7 @@ class EavAttribute
     {
         if (isset($optionsArray['value']) && is_array($optionsArray['value'])) {
             foreach (array_keys($optionsArray['value']) as $optionId) {
-                if (isset($optionsArray['delete']) && $optionsArray['delete'][$optionId] == 1) {
+                if (isset($optionsArray['delete'][$optionId]) && $optionsArray['delete'][$optionId] == 1) {
                     unset($optionsArray['value'][$optionId]);
                 }
             }


app/code/Magento/Swatches/Model/Plugin/EavAttribute.php
```diff
@@ -174,7 +174,7 @@ protected function prepareOptionIds(array $optionsArray)
     {
         if (isset($optionsArray['value']) && is_array($optionsArray['value'])) {
             foreach (array_keys($optionsArray['value']) as $optionId) {
-                if (isset($optionsArray['delete']) && $optionsArray['delete'][$optionId] == 1) {
+                if (isset($optionsArray['delete'][$optionId]) && $optionsArray['delete'][$optionId] == 1) {
```
orlangur, line 177: > Php 500 Error when $optionsArray['delete'][$optionId] is not set

This is just a notice, 500 will be displayed only in developer mode.

Can `$optionsArray['delete'][$optionId]` contain anything but `1`? Why not do unconditional `unset($optionsArray['value'][$optionId])`?

angelo983: Simply misunderstood! Read the function definition comment: "Get options array without deleted items"
so... not all array_keys($optionsArray['value']) are present in $optionsArray['delete'][$optionId] as indexes of the $optionsArray['delete']

orlangur: Yeah, I've got the point of bugfix, just trying to simplify the code.

Is check for `== 1` essential or we can simply do `unset` without condition?

angelo983: At first glance I would say yes, but checking getData() method of Attribute I can't be totally sure.
Anyway a condition must be verified, so the choice should be between:
(isset($optionsArray['delete'][$optionId]))
and
(isset($optionsArray['delete'][$optionId]) && $optionsArray['delete'][$optionId] == 1)

orlangur: Ok, it is already merged, I was trying to understand if it can be just `if (!empty($optionsArray['delete'][$optionId]))`.

PR #11737, opened 2017-10-25
[Backport 2.1-develop] Add option to keep Files in Rollback

osrecio:
Add option to keep files in rollback and improve Readability and PSR of Backup Library
### Description
Ask to user when launch command: `bin/magento setup:rollback` if he wants keep files to future rollbacks. Also I changed variable name protected with underscore to acomplish PSR and I imported all classes to use in class.

1. magento/magento2#6460: [2.1.1 CE] Rollback/Restore deletes database (--db) backup file in ${webroot}/var/backups.

### Manual testing scenarios
1. create backup  e.g: `bin/magento setup:backup --media`
2. rollback backup e.g: `bin/magento setup:rollback -m 1508950479_filesystem_media.tgz`
3. Answer `y` to question: `you want to keep the backups?[y/N]`
4. Check that your backup was kept in `var/backups`

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Backup/AbstractBackup.php
+++ b/lib/internal/Magento/Framework/Backup/AbstractBackup.php
@@ -319,6 +319,11 @@ abstract class AbstractBackup implements BackupInterface
         return $this;
     }
 
+    /**
+     * Get if keep files of backup
+     *
+     * @return boolean
+     */
     public function keepSourceFile()
     {
         return $this->keepSourceFile;
--- a/lib/internal/Magento/Framework/Backup/AbstractBackup.php
+++ b/lib/internal/Magento/Framework/Backup/AbstractBackup.php
@@ -5,6 +5,9 @@
  */
 namespace Magento\Framework\Backup;
 
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Phrase;
+
 /**
  * Class to work with archives
  *
@@ -67,6 +70,12 @@ abstract class AbstractBackup implements BackupInterface
      * @var string
      */
     protected $_lastErrorMessage;
+
+    /**
+     *
+     * @var boolean
+     */
+    protected $keepSourceFile = false;
 
     /**
      * Set Backup Extension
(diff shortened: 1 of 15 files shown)


lib/internal/Magento/Framework/Backup/AbstractBackup.php
```diff
@@ -17,56 +17,62 @@
      *
      * @var string
      */
-    protected $_name;
+    protected $name;
```
orlangur: Unfortunately, such kind of change is only allowed for `private` properties, see http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#removing-or-renaming-public-and-protected-properties

Please always create PR targeting `2.2-develop` first, ideally PRs targeting other branches should be created after it is merged.

orlangur (changes requested):

lib/internal/Magento/Framework/Backup/AbstractBackup.php
osrecio: I will change all protected variables and functions to old state.

orlangur: 👍 

Please create a new PR targeting `2.2-develop` with these changes, don't close this PR - just force push into it after `2.2` PR is processed.

osrecio: I have `2.2-develop` in local ready to push. I will take care in next PR

osrecio:
@orlangur Squashed and revert name of variables and functions `protected`

dmanners:
@osrecio unfortunately these new methods will break BC for this class. I think we should focus on getting these changes into `2.3-develop` and then work on the backport.

For the backport I think we might be able to make the file removals inside the `doRollback` method but I am not sure how that will work for all types of rollback, such as ftp and others. Would you have sometime to look into this?

First let's work on getting these new methods and interfaces into https://github.com/magento/magento2/pull/11750 and then we can go back and work on the backport to 2.2 and 2.1. How does that sound?

PR #11742, opened 2017-10-25
[Backport 2.2-develop] Add option to keep Files in Rollback

osrecio:
Add option to keep files in rollback and improve Readability and PSR of Backup Library
### Description
Ask to user when launch command: `bin/magento setup:rollback` if he wants keep files to future rollbacks. Also I changed variable name protected with underscore to acomplish PSR and I imported all classes to use in class.

1. magento/magento2#6460: [2.1.1 CE] Rollback/Restore deletes database (--db) backup file in ${webroot}/var/backups.

### Manual testing scenarios
1. create backup  e.g: `bin/magento setup:backup --media`
2. rollback backup e.g: `bin/magento setup:rollback -m 1508950479_filesystem_media.tgz`
3. Answer `y` to question: `you want to keep the backups?[y/N]`
4. Check that your backup was kept in `var/backups`

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Backup/AbstractBackup.php
+++ b/lib/internal/Magento/Framework/Backup/AbstractBackup.php
@@ -5,12 +5,15 @@
  */
 namespace Magento\Framework\Backup;
 
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Phrase;
+
 /**
  * Class to work with archives
  *
  * @api
  */
-abstract class AbstractBackup implements BackupInterface
+abstract class AbstractBackup implements BackupInterface, SourceFileInterface
 {
     /**
      * Backup name
@@ -67,6 +70,11 @@ abstract class AbstractBackup implements BackupInterface
      * @var string
      */
     protected $_lastErrorMessage;
+
+    /**
+     * @var boolean
+     */
+    private $keepSourceFile = false;
 
     /**
      * Set Backup Extension
@@ -138,14 +146,14 @@ abstract class AbstractBackup implements BackupInterface
      * Set root directory of Magento installation
      *
      * @param string $rootDir
-     * @throws \Magento\Framework\Exception\LocalizedException
+     * @throws LocalizedException
      * @return $this
      */
     public function setRootDir($rootDir)
     {
         if (!is_dir($rootDir)) {
-            throw new \Magento\Framework\Exception\LocalizedException(
-                new \Magento\Framework\Phrase('Bad root directory')
+            throw new LocalizedException(
+                new Phrase('Bad root directory')
             );
         }
 
@@ -295,5 +303,25 @@ abstract class AbstractBackup implements BackupInterface
         $name = str_replace(' ', '_', $name);
 
         return $name;
+    }
+
+    /**
+     * Set if keep files of backup
+     *
+     * @param bool $keepSourceFile
+     * @return $this
+     */
+    public function setKeepSourceFile(bool $keepSourceFile)
+    {
+        $this->keepSourceFile = $keepSourceFile;
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    public function keepSourceFile()
+    {
+        return $this->keepSourceFile;
     }
 }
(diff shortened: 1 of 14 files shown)


lib/internal/Magento/Framework/Backup/AbstractBackup.php
```diff
@@ -69,6 +72,13 @@
     protected $_lastErrorMessage;
 
     /**
+     * Keep Source files in Backup
+     *
+     * @var boolean
+     */
+    protected $keepSourceFile;
```
dmanners: We recommend that by default new properties are private.

lib/internal/Magento/Framework/Backup/BackupInterface.php
```diff
@@ -68,4 +68,19 @@ public function getType();
      * @return $this
      */
     public function setBackupsDir($backupsDir);
+
+    /**
+     * Check if keep files of backup
+     *
+     * @return boolean
+     */
+    public function keepSourceFile();
```
dmanners: unfortunately you cannot add new classes to interfaces (http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#introduction-of-a-method-to-a-class-or-interface) in this case I would recommend adding a new interface specific for the source file flags.

lib/internal/Magento/Framework/Backup/AbstractBackup.php
```diff
@@ -296,4 +306,21 @@ protected function _filterName($name)
...
         return $name;
     }
+
+    /**
+     * Set if keep files of backup
+     *
+     * @param boolean $keepSourceFile
+     * @return $this
+     */
+    public function setKeepSourceFile($keepSourceFile)
```
dmanners: you can update this to be `setKeepSourceFile(bool $keepSourceFile)`

lib/internal/Magento/Framework/Backup/AbstractBackup.php
dmanners: Also wondering if we should have a default value here to make it a bit more explicit that it is a boolean. I guess the php will behave the same way but might be nice to add in the default.

dmanners (changes requested):
A couple of changes with regards to the new property and methods, also the interface will need to be changed.

osrecio:
Changes Made at  #mm17es

lib/internal/Magento/Framework/Backup/Db.php
```diff
@@ -45,7 +45,9 @@ public function rollback()
         foreach ($file as $statement) {
             $this->getResourceModel()->runCommand($statement);
         }
-        @unlink($source);
+        if ($this->keepSourceFile() === false) {
```
dmanners, line 48: @osrecio so this works when doing a DB backup but does not with the other types of backup. Can you also update the other backup types to include the check for this flag.

dmanners (changes requested):
Works but for only 1 type of backup.

lib/internal/Magento/Framework/Backup/Db.php
osrecio: Made changes for `Fs.php`  and `Ftp.php`

dmanners (approved):

lib/internal/Magento/Framework/Backup/AbstractBackup.php
```diff
@@ -296,4 +304,24 @@ protected function _filterName($name)
...
         return $name;
     }
+
+    /**
+     * Set if keep files of backup
+     *
+     * @param bool $keepSourceFile
+     * @return $this
+     */
+    public function setKeepSourceFile(bool $keepSourceFile)
```
dmanners, line 314: So as this class has been marked with the `@api` notation we cannot accept new methods here.

dmanners:
@osrecio unfortunately these new methods will break BC for this class. I think we should focus on getting these changes into `2.3-develop` and then work on the backport.

For the backport I think we might be able to make the file removals inside the `doRollback` method but I am not sure how that will work for all types of rollback, such as ftp and others. Would you have sometime to look into this?

First let's work on getting these new methods and interfaces into https://github.com/magento/magento2/pull/11750 and then we can go back and work on the backport to 2.2 and 2.1. How does that sound?

PR #11744, opened 2017-10-25
Minification erroneously removes // from JavaScript strings

navarr:
This pull request adds a failing unit test to identify and test for the resolution of a bug in the phtml template minification process. 

### Description
When the minifier encounters a // in the middle of JavaScript, even when in a string, the minification will ignore it and everything afterwards.

### Contribution checklist


okorshenko:
Hi @navarr 
Thank you for the PR, but we cannot accept the broken test. If would be great if you can fix the logic on the minification or you can convert this to the issue and attach your test

PR #11762, opened 2017-10-26
[BUGFIX] Removed Empty PasswordHash from Confirmation Key Logic

ghost:
Confirmation key is emptied on customer save.

### Description
If you do a save on a customer object before the activated themselves confirmation values gets truncated.

### Manual testing scenarios
1. STEPS TO REPRODUCE
- Change setting customer/create_account/confirm to 1 / true
- Register a new customer account via the frontend /customer/account/create/
- Don't confirm your account yet!
- A confirmation code is saved in the database. customer_entity table confirmation column
- Go to the backend of Magento and edit the newly registered customer and save changes
- Check the customer_entity table. The confirmation value has been cleared.
- BUG. The confirmation link in the email does not work any more
- BUG. The account is activated by accident.

### DIFF
--- a/app/code/Magento/Customer/Model/ResourceModel/Customer.php
+++ b/app/code/Magento/Customer/Model/ResourceModel/Customer.php
@@ -140,7 +140,7 @@ class Customer extends \Magento\Eav\Model\Entity\VersionControl\AbstractEntity
         }
 
         // set confirmation key logic
-        if ($customer->getForceConfirmed() || $customer->getPasswordHash() == '') {
+        if ($customer->getForceConfirmed()) {
             $customer->setConfirmation(null);
         } elseif (!$customer->getId() && $customer->isConfirmationRequired()) {
             $customer->setConfirmation($customer->getRandomConfirmationKey());


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11762) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=11762) before we can accept your contribution.<br/>

avoelkl:
Hi @dheesbeen!
Thanks for the PR!
I can reproduce the problem, but I don't think removing `getPasswordHash()` is the right way to solve it.
Without further debugging I think it's a problem of the save action from the admin or there needs to be another condition checking if the confirmation code is set.

Any thoughts on this?

ghost:
Hi @avoelkl 

Of course i also was curious why this code was added. The code "getPasswordHash()" seems to be added in a [commit](https://github.com/magento/magento2/commit/bde2759e16acc8b619afb0beb70df5d8ebef04c8) that fixes or added a feature for the create customer from the checkout success page.  But looking at the code i could not understand why i was there. I also tested the register functionality on the success page after i remove the getPasswordHash() check and it still seems to work properly.

- There are al lot of situations where you do a save on the customer object without having the password hash. Because the hash isn't load by default on the object. 
- There is all ready a method which makes it possible to skip confirmation ($customer->getForceConfirmed()) in a unique situation. From my point of view this should be used in the [commit](https://github.com/magento/magento2/commit/bde2759e16acc8b619afb0beb70df5d8ebef04c8) of the customer successpage registration 

Seems to me that !$customer->getPasswordHash() is to generic and causes a lot of problems. In Enterprise for example a observer of a Enterprise module does a save directly after registration. This also empties the confirmation code. 

Main question is why is this piece of code added in the first place. I could not yet think of a valid reason. Do you think you can find the answer? If needed i will change the PR

Thanks

avoelkl:
Hi @dheesbeen,
Thanks for your reply. Actually I don't know either but will try to find out something more about it.
I already got a hint from @vrann about a reference to the internal ticket MAGETWO-40511.
I'll keep you updated.

orlangur:
@avoelkl @vrann was there any actual followup here?

orlangur (changes requested):

orlangur:
Thanks for collaboration!

PR #11650, opened 2017-10-23
Fix ModelInstance RegEx pattern for class names.

MartinPeverelli:
The updated XSD files forbid the usage of digits in the class or namespaces for the modelInstance value.

### Description
Changes `'[A-Za-z_\\]+'` for `'[A-Za-z0-9_\\]+'` to support namespaces with digits.

### Fixed Issues (if relevant)
1. [import.xsd doesn't allow numbers in vendor name and class name #4470](https://github.com/magento/magento2/issues/4470)
2. [Exception error on Module name when it has number(s) on product_options.xml - Regex issue #9985](https://github.com/magento/magento2/issues/9985)

### Manual testing scenarios
Create a module with a vendor name containing a digit. The following XML should be valid.
```
<!-- app/code/Interactiv4/TestProduct/etc/product_types.xml -->
<?xml version="1.0"?>
<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Catalog:etc/product_types.xsd">
    <type name="test" label="Test Product" modelInstance="Interactiv4\TestProduct\Model\Product\Type\Course" />
</config>
```

### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #11765, opened 2017-10-26
Allows modules with underscores in name to add blocks to layout via XML

bentideswell:
Allow modules with underscores in the name to add blocks to the layout using XML.

### Description
Magento 2 allows module names to include the underscore character (Vendor_Module_Name, which is app/code/Vendor/Module_Name) but doesn't then allow the module to add a block to the layout via XML.

This only occurs in developer mode and produces the following error:

```
1 exception(s):
Exception #0 (Magento\Framework\Config\Dom\ValidationException): Element 'block', attribute 'class': [facet 'pattern'] The value 'Vendor\Module_Name\Block\OpenGraph' is not accepted by the pattern '[A-Z][a-zA-Z\d]*(\\[A-Z][a-zA-Z\d]*)*'.
Line: 632

Element 'block', attribute 'class': 'Vendor\Module_Name\Block\OpenGraph' is not a valid value of the atomic type 'blockClassType'.
Line: 632


Exception #0 (Magento\Framework\Config\Dom\ValidationException): Element 'block', attribute 'class': [facet 'pattern'] The value 'Vendor\Module_Name\Block\OpenGraph' is not accepted by the pattern '[A-Z][a-zA-Z\d]*(\\[A-Z][a-zA-Z\d]*)*'.
Line: 632

Element 'block', attribute 'class': 'Vendor\Module_Name\Block\OpenGraph' is not a valid value of the atomic type 'blockClassType'.
Line: 632
```
The facet pattern is defined in lib/internal/Magento/Framework/View/Layout/etc/elements.xsd:

```
<xs:simpleType name="blockClassType">
    <xs:restriction base="xs:string">
        <xs:pattern value="[A-Z][a-zA-Z\d]*(\\[A-Z][a-zA-Z\d]*)*"/>
    </xs:restriction>
</xs:simpleType>
```
This pattern does not allow the underscore character to appear in the block class type attribute. I have fixed this by changing the pattern to:

`<xs:pattern value="[A-Z][a-zA-Z\d]*(\\[A-Z][_a-zA-Z\d]*)*"/>`

This allows an underscore to be used in the module name. This could be further to changed to allow an underscore to be used in the vendor name by changing it to:

`<xs:pattern value="[A-Z][_a-zA-Z\d]*(\\[A-Z][_a-zA-Z\d]*)*"/>`

### Manual testing scenarios
1. Enable developer mode using php bin/magento deploy:mode:set developer
2. Create a module at app/code/Vendor/Some_Module
3. Add a frontend layout XML file to the module and include a block from the module
4. Flush the cache and view the frontend

### DIFF
--- a/lib/internal/Magento/Framework/View/Layout/etc/elements.xsd
+++ b/lib/internal/Magento/Framework/View/Layout/etc/elements.xsd
@@ -138,7 +138,7 @@
 
     <xs:simpleType name="blockClassType">
         <xs:restriction base="xs:string">
-            <xs:pattern value="[A-Z][a-zA-Z\d]*(\\[A-Z][a-zA-Z\d]*)*"/>
+            <xs:pattern value="[A-Z][_a-zA-Z\d]*(\\[A-Z][_a-zA-Z\d]*)*"/>
         </xs:restriction>
     </xs:simpleType>
 


orlangur:
Underscores are not allowed according to [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md). Why do you need them?

lib/internal/Magento/Framework/View/Layout/etc/elements.xsd
```diff
@@ -138,7 +138,7 @@
 
     <xs:simpleType name="blockClassType">
         <xs:restriction base="xs:string">
-            <xs:pattern value="[A-Z][a-zA-Z\d]*(\\[A-Z][a-zA-Z\d]*)*"/>
+            <xs:pattern value="[A-Z][a-zA-Z\d]*(\\[A-Z][_a-zA-Z\d]*)*"/>
```
orlangur: For consistency with other places, like https://github.com/magento/magento2/pull/11650/files, were underscore was allowed previously, please allow underscore in vendor name also, perform an amend commit and force push into the same branch.

orlangur (changes requested):

lib/internal/Magento/Framework/View/Layout/etc/elements.xsd
bentideswell: I have made this change. Now both Vendor and Module allow underscores.

orlangur: Changes look good now. Just squash all changes into single commit and do a force push to the same branch please.

bentideswell: I think I have done this but let me know if I've made a mistake or you need anything else.

orlangur (approved):

bentideswell:
Just a quick update but I have found the same issue in the XSD file that handles the system.xml configuration options. I have created a [new pull request](https://github.com/magento/magento2/pull/14397) that supplies the same fix as this (adding the underscore to the regular expression).

The new pull request can be seen at https://github.com/magento/magento2/pull/14397

PR #11766, opened 2017-10-26
#8421: uiComponent sortOrder is not applied properly if some layout items are missing sortOrder value

RomaKis:
### Description
All ui components, which don't has SortOrder undefined will has SordOrder depends on as items are rendered as they appear in layout file

1. magento/magento2#8421: Issue title

### Manual testing scenarios
1. Create a new uiComponent (or use existing one just for testing)
2. Try to add this to checkout and order after summary item in summary displayArea

```xml 
<referenceBlock name="checkout.root">
            <arguments>
                <argument name="jsLayout" xsi:type="array">
                    <item name="components" xsi:type="array">
                        <item name="checkout" xsi:type="array">
                            <item name="children" xsi:type="array">
                                <item name="sidebar" xsi:type="array">
                                    <item name="children" xsi:type="array">
                                        <item name="new_component" xsi:type="array">
                                            <item name="component" xsi:type="string">uiComponent</item>
                                            <item name="displayArea" xsi:type="string">summary</item>
                                            <item name="sortOrder" xsi:type="string">900</item>
                                            <item name="config" xsi:type="array">
                                                <item name="template" xsi:type="string">new_component/template</item>
                                            </item>
                                        </item>
                                    </item>
                                </item>
                            </item>
                        </item>
                    </item>
                </argument>
            </arguments>
        </referenceBlock>
```
3. This component must be placed as a last element as it has sortOrder 900.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
+++ b/app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
@@ -435,6 +435,8 @@ define([
             if (parent && parent.component) {
                 name = child.name || child;
 
+                this.setChildSortOrderIfUndefined(parent, child);
+
                 this.insert(name, parent.name, child.sortOrder);
             }
 
@@ -600,6 +602,25 @@ define([
             });
 
             return result;
+        },
+
+        /**
+         * Set sort order for child if it is undefined.
+         *
+         * @param {Object} parent
+         * @param {Object} child
+         */
+        setChildSortOrderIfUndefined: function (parent, child) {
+            var key, sortOrderIndex = 0;
+
+            for (key in parent.children) {
+                if (child.index === key
+                    && _.isUndefined(parent.children[key].sortOrder)) {
+                    child.sortOrder = sortOrderIndex;
+                    break;
+                }
+                sortOrderIndex++;
+            }
         }
     });
 
--- a/app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
+++ b/app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
@@ -614,8 +614,7 @@ define([
             var key, sortOrderIndex = 0;
 
             for (key in parent.children) {
-                if (child.index === key
-                    && _.isUndefined(parent.children[key].sortOrder)) {
+                if (child.index === key && _.isUndefined(parent.children[key].sortOrder)) {
                     child.sortOrder = sortOrderIndex;
                     break;
                 }


app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
```diff
@@ -600,6 +602,25 @@ define([
...
             return result;
+        },
+
+        /**
+         * Get sort order for child if it is undefined.
+         *
+         * @param {Object} parent
+         * @param {Object} child
+         */
+        getChildSortOrderIfUndefined: function (parent, child) {
```
ishakhsuvarov: This method does not return anything, but naming suggests so (`get`).

app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
```diff
@@ -600,6 +602,25 @@ define([
...
+            var key, i = 0;
```
ishakhsuvarov: While `i` is a common name for index in the loops, it is still suggested to use conscious names.

app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
```diff
@@ -600,6 +602,25 @@ define([
...
+
+            for (key in parent.children) {
+                if (child.index === key
+                    && typeof parent.children[key].sortOrder === "undefined") {
```
ishakhsuvarov: Looks like this component depends on `underscore` already, so it is possible to use it's method to detect undefined.

ishakhsuvarov (changes requested):
Please attempt to update JS or integration tests as well to support this case.

app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
RomaKis: Fixed.

app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
RomaKis: Fixed.

app/code/Magento/Ui/view/base/web/js/core/renderer/layout.js
RomaKis: Fixed.

ishakhsuvarov (approved):

omiroshnichenko:
Hi, @RomaKis, this pull request make no sense. I've described correct use of sortOrder in [issue comment](https://github.com/magento/magento2/issues/8421#issuecomment-342845954).

PR #11787, opened 2017-10-26
Fix #10438: Potential error on order edit page when address has extension attributes

### LINKED ISSUE #10438: Potential error on order edit page when address has extension attributes
Potential error on order edit page when address has extension attributes

### Preconditions
1. Magento 2.1.7
2. PHP 7.0.18

### Steps to reproduce
1. Suppose you init extension attributes of the order address on the `sales_order_address_load_after` event (any other way is also ok): 
```php
        $address = $observer->getEvent()->getAddress();
        $extensionAttributes = $address->getExtensionAttributes();
        if (!$extensionAttributes) {
            $extensionAttributes = $this->extensionAttributesFactory->create();
            $address->setExtensionAttributes($extensionAttributes);
        }
```
2. Go to order's admin panel and click "Edit"

### Expected result
1. Order edit page opens

### Actual result
1. Order edit page fails with the following error: `Recoverable Error: Object of class Magento\Sales\Api\Data\OrderAddressExtension could not be converted to string in .../module-sales/Model/AdminOrder/Create.php on line 503`

The code on this line does the following:
```php
        $shippingAddress = $order->getShippingAddress();
        if ($shippingAddress) {
            $addressDiff = array_diff_assoc($shippingAddress->getData(), $order->getBillingAddress()->getData());
            unset($addressDiff['address_type'], $addressDiff['entity_id']);
            $shippingAddress->setSameAsBilling(empty($addressDiff));
        }
```

The problem is that `$shippingAddress` object could potentially have `extension_attributes` field in the data and that field is an object (which is correct).
When you diff arrays with object in a field you get the error:

```cli
php > var_dump(array_diff_assoc(['foo' => new stdClass], ['foo' => new stdClass]));
PHP Catchable fatal error:  Object of class stdClass could not be converted to string in php shell code on line 1
``` 

`array_diff_assoc` converts values to strings according to the [documentation](http://php.net/manual/en/function.array-diff-assoc.php):

> Two values from key => value pairs are considered equal only if (string) $elem1 === (string) $elem2 . In other words a strict check takes place so the string representations must be the same.

YevSent:
### Fixed Issues
1. magento/magento2#10438: Potential error on order edit page when address has extension attributes

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/AdminOrder/Create.php
+++ b/app/code/Magento/Sales/Model/AdminOrder/Create.php
@@ -10,9 +10,12 @@ namespace Magento\Sales\Model\AdminOrder;
 
 use Magento\Customer\Api\AddressMetadataInterface;
 use Magento\Customer\Model\Metadata\Form as CustomerForm;
+use Magento\Framework\Api\ExtensibleDataObjectConverter;
 use Magento\Framework\App\ObjectManager;
 use Magento\Quote\Model\Quote\Address;
 use Magento\Quote\Model\Quote\Item;
+use Magento\Sales\Api\Data\OrderAddressInterface;
+use Magento\Sales\Model\Order;
 
 /**
  * Order create model
@@ -236,6 +239,11 @@ class Create extends \Magento\Framework\DataObject implements \Magento\Checkout\
     private $serializer;
 
     /**
+     * @var ExtensibleDataObjectConverter
+     */
+    private $dataObjectConverter;
+
+    /**
      * @param \Magento\Framework\ObjectManagerInterface $objectManager
      * @param \Magento\Framework\Event\ManagerInterface $eventManager
      * @param \Magento\Framework\Registry $coreRegistry
@@ -265,6 +273,7 @@ class Create extends \Magento\Framework\DataObject implements \Magento\Checkout\
      * @param \Magento\Quote\Model\QuoteFactory $quoteFactory
      * @param array $data
      * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @param ExtensibleDataObjectConverter|null $dataObjectConverter
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
     public function __construct(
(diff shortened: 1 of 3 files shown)


app/code/Magento/Sales/Test/Unit/Model/AdminOrder/CreateTest.php
```diff
@@ -295,43 +220,50 @@ public function testUpdateQuoteItemsWithConfiguredOption()
...
+        $address->method('setCollectShippingRates')
+            ->with(true)
+            ->willReturnSelf();
+        $address->method('setFreeShipping')
+            ->with(0)
+            ->willReturnSelf();
 
         $object = $this->adminOrderCreate->applyCoupon($couponCode);
-        $this->assertEquals($this->adminOrderCreate, $object);
+        self::assertEquals($this->adminOrderCreate, $object);
```
orlangur, line 267: `$this->assert*` must be used, change all occurrences.

YevSent: @orlangur, `assert*` it's a `static` function. Also, I didn't find any requirements about `$this->...*` usage instead `self::...` for tests.

orlangur: Easy, it's a recommendation of PHPUnit author, we should not rely on 15 years old things from Java: https://github.com/sebastianbergmann/phpunit/issues/1914#issuecomment-148407321

Maybe I'll implement static test enforcing this someday but till then please keep things consistent, most of calls in Magento codebase are `$this->assert` currently.

YevSent: As I see, it's just a recommendation. We are using `self::assert` more than 1 year in the tests.

Another point, it's calling `static` methods as `non-static` could be changed in future versions, the same we have got for PHP 5.6 for calling `non-static` as `static` - http://php.net/manual/en/migration56.deprecated.php

orlangur: Of course you can use `self::` technically but it's lame.

31352 occurrences of `$this->assert` vs 296 of `self::assert`, hm, not too much tests written during more than a year :)

There is no point in continuing discussion but now I see that a test enforcing best practices is really needed.

app/code/Magento/Sales/Model/AdminOrder/Create.php
```diff
@@ -296,7 +304,8 @@ public function __construct(
         \Magento\Sales\Api\OrderManagementInterface $orderManagement,
         \Magento\Quote\Model\QuoteFactory $quoteFactory,
         array $data = [],
-        \Magento\Framework\Serialize\Serializer\Json $serializer = null
+        \Magento\Framework\Serialize\Serializer\Json $serializer = null,
+        ExtensibleDataObjectConverter $dataObjectConverter = null
```
omiroshnichenko, line 309: Please, update docblock accordingly with the parameters that you accept.

YevSent: Done.

PR #11790, opened 2017-10-26
Fixed minimum order amount multi currency notice message wrong conver…

vijay-wagento:
During currency switch notice message shows wrong calculation of minimum order amount.

### Description
This is about notice message of min order amount on cart page.When we have set min order amount and on cart page someone change currency, user see same amount and only currency symbol changed. It should recalculate amount also.
Instead of currency convert, we use price helper then it gives us re calculated price amount in notice message.

### Manual testing scenarios

Preconditions
1. - Any Product must be created and enabled in frontend with the price greater then 0 or less then 1000  

Steps to reproduce 

1. - Enable Minimum Order Amount 
Login to [Magento Admin Panel]/ Stores -> Configuration -> Sales -> Sales -> Minimum Order Amount and do the following

  - Set Enable "Yes"
  - Enter Amount for example 1000
  - Leave Description Message text blank 
  - Leave Error to Show in Shopping Cart blank 
  - Save the Configurations by clicking on "Save Config" button on top right corner

2. - Now setup Multiple currencies for that follow below steps 

Go to section Stores -> Configuration->General->Currency Setup

Now select multiple currencies from Allowed Currencies by pressing ctrl key 
select Brazilian Real, Indian Rupee, Japanese Yen, Russian Ruble, US Dollar

3. - Import Currency Rates 
Go to Stores -> Under Currency Title click on Currency Rates. 
Select any of service and click on import rates it will import live currency rates. 

4. - Go to Frontend Add any product in shopping cart add any product 1 qty which has price less then 1000. 

5. - Go to cart page "View and Edit Cart" link which you can see on Minicart Popup or by browsing url [Magento base url]/heckout/cart/

6. - You can see the default message on "Minimum order amount is $1,000.00" 

7. - Now change the currencies from header upper right corner. You can see Minimum order amount is $1,000.00 in any currency you selected 

Expected result

1000 must be changed according to the live currency rates in case of multi currency.

Current Result:
shopping-cart-actual-result-indiancurrency
shopping-cart-actual-result-japanesecurrency

Expected Result
shopping-cart-expected-result-indiancurrency
shopping-cart-expected-result-japanesecurrency

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11790) <br/>All committers have signed the CLA.

app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
```diff
@@ -30,11 +34,13 @@ class ValidationMessage
     public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Magento\Store\Model\StoreManagerInterface $storeManager,
-        \Magento\Framework\Locale\CurrencyInterface $currency
+        \Magento\Framework\Locale\CurrencyInterface $currency,
```
vrann: please add these dependencies in a backward compatible manner according to http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
```diff
@@ -18,23 +18,22 @@ class ValidationMessage
...
      * @param \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig
      * @param \Magento\Store\Model\StoreManagerInterface $storeManager
-     * @param \Magento\Framework\Locale\CurrencyInterface $currency
+     * @param \Magento\Framework\Pricing\Helper\Data $priceHelper
      */
     public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Magento\Store\Model\StoreManagerInterface $storeManager,
-        \Magento\Framework\Locale\CurrencyInterface $currency
+        \Magento\Framework\Pricing\Helper\Data $priceHelper
```
vrann, line 32: please don't remove parameters from constructor and
add dependency in a backward compatible manner according to http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

vrann (changes requested):

okorshenko:
Hi @vijay-wagento 
We are closing this PR due to inactivity. Feel free to reopen it once ready

PR #11802, opened 2017-10-27
#8236 FIX CMS blocks

thiagolima-bm:
FIX >> CMS blocks are not validated against having same store and identifier #8236


thiagolima-bm:
@okorshenko @veloraven  Related issue #8236

dmanners (approved):

PR #11809, opened 2017-10-27
8003: Using System Value for Base Currency Results in Config Error.

nmalevanec:
### Description
Fix for "Using System Value for Base Currency Results in Config Error".

1. magento/magento2#8003: Using System Value for Base Currency Results in Config Error

### Manual testing scenarios
1. Log into Admin.
2. Navigate to Stores > Settings > Configuration > General > Currency Setup.
3. Select Use system value for Base Currency.
4. Click Save Config if necessary and flush the cache.
5. Navigate to Stores > Currency > Currency Rates.
6. Try Importing Currency Rates via any of the three services.
7. Check currency rates are successfully imported for Allowed Currencies.

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11809) <br/>All committers have signed the CLA.

dmanners (changes requested):
I think that this plugin is not needed. From what I can see the methods `getConfigDefaultCurrencies` and `getConfigBaseCurrencies` should behave like `getConfigAllowCurrencies`. This method reads the data from the database but then also checks for the value set in the config. In this case it means that you will get all currencies set in the DB but also the default. Does that make sense?

nmalevanec:
@dmanners correct me, if I'm wrong, but in https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Directory/Model/Currency.php#L348 I see that allowed currencies reads only from DB, and later adds base currency(which indeed reads from db and system config) to allowed currencies array. So, if in admin I'll check 'Use system value' for Allowed Currencies, I'll get only base currency code.
 I've created plugin, because both Model/Currency and ResourceModel/Currency marked with @api. And as far as I know, removing or marking as deprecated methods from such classes should be avoided(I mean ResourceModel/Currency::getConfigCurrencies()). And add system config, which reads config files, as dependency to ResourceModel, which interacts with DB looks like not the best idea for me.

dmanners:
@nmalevanec with regards to `getConfigAllowCurrencies` yes that does seem odd that it will add the base currency to the allowed currencies but I think that is another issue.

My concern with using the plugin is that this is a bug with the code code but fixing it with a plugin is not that obvious to all developers. If someone is interested in finding out how this code works it seems to be hidden.

I would actually love to see the following either been added to `app/code/Magento/Directory/Model/Currency.php::getConfigDefaultCurrencies` and `app/code/Magento/Directory/Model/Currency.php::getConfigBaseCurrencies`

1. Call resource model that checks database values,
2. Call new class that checks the config values,
3. Combine the results of both for the system,

In this way we would cover the database and config values and since both getConfigDefaultCurrencies and getConfigBaseCurrencies are due to return an array it should work with 1 or more currencies.

How does that sound?

nmalevanec:
@dmanners Well, it's pretty doable. The only thing I'm concerning about(And this is actually one of the initial reasons to move logic into plugin) is Model/Currency already has coupling between objects more than allowed and suppressed warning about this. So adding to it another dependency like Magento\Config\App\Config\Type\System... Well, I'm not sure it's valid scenario. At least previously it wasn't.

dmanners:
@nmalevanec a very valid point.

We could always consider changing the `app/code/Magento/Directory/Model/Currency/Import/AbstractImport.php::_getDefaultCurrencyCodes` to use a new class that is specific for our purpose. Then we could deprecate the other methods and use the `@see` notation to highlight the new class and methods.

nmalevanec:
@dmanners Done.

app/code/Magento/Directory/Model/CurrencySystemConfig.php
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Currency constructor.
+     *
+     * @param System $systemConfig
+     * @param StoreManagerInterface $storeManager
+     */
+    public function __construct(
+        System $systemConfig,
+        StoreManagerInterface $storeManager,
+        ResourceConnection $resources
```
dmanners: I guess we do not need this if it is never assigned to anything.

app/code/Magento/Directory/Model/Currency.php
```diff
@@ -88,6 +95,7 @@ public function __construct(
         \Magento\Framework\Locale\CurrencyInterface $localeCurrency,
         \Magento\Framework\Model\ResourceModel\AbstractResource $resource = null,
         \Magento\Framework\Data\Collection\AbstractDb $resourceCollection = null,
+        CurrencySystemConfig $currencyConfig = null,
```
dmanners: Should this not be after the `$data` otherwise it will break backwards compatibility?

app/code/Magento/Directory/Model/Currency.php
```diff
@@ -378,7 +390,8 @@ public function getConfigDefaultCurrencies()
      */
     public function getConfigBaseCurrencies()
     {
-        $defaultCurrencies = $this->_getResource()->getConfigCurrencies($this, self::XML_PATH_CURRENCY_BASE);
+        $defaultCurrencies = $this->_getResource()->getConfigCurrencies($this, self::XML_PATH_CURRENCY_BASE) ?:
```
dmanners: What happens in the case that you have multiple stores, 1 with a database value and one without. In this case I guess we need to merge the values here.

dmanners (changes requested):

app/code/Magento/Directory/Model/Currency.php
nmalevanec: Fixed.

app/code/Magento/Directory/Model/Currency.php
nmalevanec: Fixed.

app/code/Magento/Directory/Model/CurrencySystemConfig.php
nmalevanec: Removed.

PR #11817, opened 2017-10-27
GITHUB-8970: Cannot assign products to categories not under tree root.

p-bystritsky:
### Description
Added ability to assign products to categories if those products are already assigned to category tree root (id=1).

1.  magento/magento2#8970: Cannot assign products to categories not under tree root.

### Manual testing scenarios
1. Create a Product, assign it to any category.
2. Open your Magento database table "catalog_category_product", locate your Product table entry and change it's "category_id" value to 1.
3. Assign your Product to some category through admin interface and save it.
4. "You saved the product." message should appear.

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogUrlRewrite/Model/ProductScopeRewriteGenerator.php
+++ b/app/code/Magento/CatalogUrlRewrite/Model/ProductScopeRewriteGenerator.php
@@ -207,9 +207,9 @@ class ProductScopeRewriteGenerator
      */
     public function isCategoryProperForGenerating(Category $category, $storeId)
     {
-        $parentId = $category->getParentId();
-        if ($parentId != Category::ROOT_CATEGORY_ID && $parentId != Category::TREE_ROOT_ID) {
-            list(, $rootCategoryId) = $category->getParentIds();
+        $parentIds = $category->getParentIds();
+        if (count($parentIds) >= 2) {
+            $rootCategoryId = $parentIds[1];
             return $rootCategoryId == $this->storeManager->getStore($storeId)->getRootCategoryId();
         }
         return false;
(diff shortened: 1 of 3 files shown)


dmanners:
@p-bystritsky thanks for this PR. Could you please look at the test failures.

```
There was 1 failure:
1) Magento\CatalogUrlRewrite\Test\Unit\Model\ProductScopeRewriteGeneratorTest::testGenerationForGlobalScope
Expectation failed for method name is equal to <string:getParentId> when invoked 2 time(s).
Method was expected to be called 2 times, actually called 1 times.
```

And

```
1) Magento\Test\Php\LiveCodeTest::testCodeStyle
PHP Code Sniffer detected 2 violation(s): 
FILE: .../app/code/Magento/CatalogUrlRewrite/Model/ProductScopeRewriteGenerator.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 211 | ERROR | [x] Expected 0 spaces after opening bracket; newline found
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------
```

dmanners:
@p-bystritsky I do not think this fixes the issue fully. In the case of your fix you will no longer be able to have malfunctioning categories and the save via the admin should still work, which is good, but the core of the issue raised was the usage of `list(, $rootCategoryId) = $category->getParentIds();` which is still in place.

As mentioned in the issue #8970 `Generally speaking, it is advisable to avoid relying on a specific order of operation, as this may change again in the future.` I would suggest that a "nicer" option for now would be to check that `$category->getParentIds()` returns an array of at least 2 entries and that the 2nd one is the store's root category. This way we cover both the broken category tables and also the miss use of the list option.

Do you think you will be able to make this change?

dmanners (changes requested):
Would be good to remove the usage of `list` as this has changed between php5 and php7 and as noted `Generally speaking, it is advisable to avoid relying on a specific order of operation, as this may change again in the future.`

p-bystritsky:
Hello, @dmanners. I'll try to remove list usage with the hint you provided.

dmanners (approved):

PR #11822, opened 2017-10-27
Fix #1779: CarrierInterface additional requirement

### LINKED ISSUE #1779: CarrierInterface additional requirement

dverkade:
Fixed for issue #1779 as mentioned in closed PR #11293 :
- Moved methods from AbstractCarrierInterface to CarrierInterface
- Deprecated AbstractCarrierInterface
- Removed references to AbstractCarrierInterface in fthe CarrierFactoryInterface, changed them to CarrierInterface.
- Removed AbstractCarrier model implementing the AbstractCarrierInterface.
- Changed references from AbstractCarrierInterface to CarrierInterface in de Config model.


app/code/Magento/Shipping/Model/Carrier/AbstractCarrier.php
```diff
@@ -17,7 +17,7 @@
  * @api
  * @since 100.0.2
  */
-abstract class AbstractCarrier extends \Magento\Framework\DataObject implements AbstractCarrierInterface
+abstract class AbstractCarrier extends \Magento\Framework\DataObject
```
omiroshnichenko: Please, revert this change because it is backward incompatible.

app/code/Magento/Shipping/Model/Carrier/AbstractCarrierInterface.php
```diff
@@ -5,199 +5,12 @@
...
 
-use Magento\Quote\Model\Quote\Address\RateRequest;
-
 /**
  * Interface AbstractCarrierInterface
+ *
+ * @deprecated
+ * @see CarrierInterface
  */
 interface AbstractCarrierInterface
```
omiroshnichenko: You must extend CarrierInterface here.

app/code/Magento/Shipping/Model/CarrierFactoryInterface.php
```diff
@@ -5,7 +5,7 @@
  */
```
omiroshnichenko: Please, revert all these changes because they cause backward incompatible issue.

app/code/Magento/Shipping/Model/Config.php
```diff
@@ -8,7 +8,7 @@
```
omiroshnichenko: Please, revert all these changes because they cause backward incompatible issue.

omiroshnichenko (changes requested):
Sync your branch with 2.3-develop.

omiroshnichenko:
@dverkade Will you continue to work on this PR?

app/code/Magento/Shipping/Model/Carrier/AbstractCarrierInterface.php
```diff
@@ -5,199 +5,12 @@
  */
```
omiroshnichenko, line 5: Leave methods isStateProvinceRequired, isCityRequired, debugData in this interface. These methods must be also deprecated.

dverkade: As described in this comment by @okorshenko the AbstractCarrierInterface should be empty
https://github.com/magento/magento2/pull/11293#issuecomment-339484781

omiroshnichenko: As described in comment that you mentioned "We need to move all the “correct” methods to CarrierInterface", so I please you to leave unused methods in deprecated interface. When deprecated interface will be deleted, unused methods also will be removed.

omiroshnichenko:
@dverkade Will you continue to work on this PR?

omiroshnichenko (approved):

dverkade:
Thanks for chaning this, I was not able to make time last week. I see this is now approved and will be processed further.

omiroshnichenko:
Hi @dverkade, after internal discussion, we came to that fact that we need to introduce new interface that will have correct methods, for this we must make some HLD.So I will close this PR as current solution will be backward incompatible and will not bring any benefit. Thank you for your contribution.

PR #11823, opened 2017-10-27
signing in after selecting checkout button, will not end up to checko…

bohemiorulo:
…ut page! #10834

signing in after selecting checkout button, will not end up to checkout page! #10834

### Description
Apparently I was not taking into account the "context" parameter that comes in the form with the value of "checkout", and not controlling this did not have a url to redirect.

1.- After login redirect the checkout page, previously clicking on the checkout button

### Manual testing scenarios
1.- disable guest checkout
2.- open magento store
3.- add product to your cart and select checkout button
4.- fill the login window details
5.- and you should select checkout button once again after signing in

result: after signing in, user should redirected to checkout page, as he want this before signing in

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Controller/Ajax/Login.php
+++ b/app/code/Magento/Customer/Controller/Ajax/Login.php
@@ -175,14 +175,15 @@ class Login extends \Magento\Framework\App\Action\Action
             $this->customerSession->setCustomerDataAsLoggedIn($customer);
             $this->customerSession->regenerateId();
             $redirectRoute = $this->getAccountRedirect()->getRedirectCookie();
+            
+            if(isset($credentials['context']) && $credentials['context'] == 'checkout'){
+                $response['redirectUrl'] = $this->_redirect->success($this->_url->getUrl('checkout', ['_secure' => true]));
+                $this->getAccountRedirect()->clearRedirectCookie();
+            }
+            
             if (!$this->getScopeConfig()->getValue('customer/startup/redirect_dashboard') && $redirectRoute) {
                 $response['redirectUrl'] = $this->_redirect->success($redirectRoute);
                 $this->getAccountRedirect()->clearRedirectCookie();
-            }else{
-                if(isset($credentials['context']) && $credentials['context'] == 'checkout'){
-                    $response['redirectUrl'] = $this->_redirect->success($this->_url->getUrl('checkout', ['_secure' => true]));
-                    $this->getAccountRedirect()->clearRedirectCookie();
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11823) <br/>All committers have signed the CLA.

okorshenko:
Similar PR https://github.com/magento/magento2/pull/11045

bohemiorulo:
I'm not sure why it did not pass the TEST_SUITE = static (https://travis-ci.org/magento/magento2/jobs/295922237), they could help me detect what was my mistake so I could fix it, thanks.

app/code/Magento/Customer/Controller/Ajax/Login.php
```diff
@@ -57,6 +57,11 @@ class Login extends \Magento\Framework\App\Action\Action
      * @var ScopeConfigInterface
      */
     protected $scopeConfig;
+    
+    /**
+     * @var url
+     */
+    protected $_url;
```
ihor-sviziev: 1. New properties should be private
2. properties should not be prefixed with underscore
3. Looks like phpdoc block is not correct for this variable

app/code/Magento/Customer/Controller/Ajax/Login.php
```diff
@@ -169,6 +175,13 @@ public function execute()
             $this->customerSession->setCustomerDataAsLoggedIn($customer);
             $this->customerSession->regenerateId();
             $redirectRoute = $this->getAccountRedirect()->getRedirectCookie();
+            
+            if (isset($credentials['context']) && $credentials['context'] == 'checkout') {
+                $url_checkout = $this->_url->getUrl('checkout', ['_secure' => true]);
```
ihor-sviziev: Variable names should be in camel case ($urlCheckout or $checkoutUrl is better)

ihor-sviziev (changes requested):

ihor-sviziev:
@bohemiorulo look at https://github.com/magento/magento2/pull/11045#issuecomment-333969727. Your change is doing the same. Probably you need to update this with proposed changes to get accepted

okorshenko:
Hi @bohemiorulo 
Please reiew @ihor-sviziev comment. This is valid consern. Currently Customer module has a dependency on Checkout, but it make sense to redirect the user to the page, where it came from. This partial fix of the problem.
Please reopen PR when you want to make needed changes

PR #11836, opened 2017-10-28
CMS Block/Page admin improvements

adrian-martinez-interactiv4:
CMS Admin Improvements

### Description
This PR will:
- Use CMS Block repository to retrieve/save/delete data, instead of model load/save/delete
- Use CMS Page repository to retrieve/save/delete data, instead of model load/save/delete
- Remove deprecated calls to $messageManager
- Update CMS Page and Block unit tests, also covering throwing NoSuchEntityException and more cases than before

None, code improvement

### Contribution checklist


dmanners:
@adrian-martinez-interactiv4 I love the effort here. My gut feeling is that this will not be able to make it into 2.2 for BC reasons, though I think 2.3 will be perfectly manageable here. I will keep you updated on it.

ihor-sviziev:
@adrian-martinez-interactiv4 @dmanners I think most of these changes can be updated to not brake BC. 
Do you think replacing $model->load() to load via repository is  BC change?

adrian-martinez-interactiv4:
@ihor-sviziev @dmanners I've been reviewing this, and even if at first look it seems that changing $model->load() to load via repository is a backward compatible change, I find that may be some reasons to consider it is not a backward compatible change:

- In CMS page case, loading page via repository, is actually triggering $model->load():
captura de pantalla 2017-11-06 a las 0 49 16
At the end it uses `\Magento\Framework\Model\AbstractModel::load`, that uses directly resource model loading.
 
- In CMS block case, loading block via repository is actually triggering block resource model load:
captura de pantalla 2017-11-06 a las 0 51 20
I understand that if block model had implemented load method (like page model does), by loading block via repository I would be bypassing model load method, because of loading block directly from resource model, but $block->load() uses `\Magento\Framework\Model\AbstractModel::load`, that also uses directly resource model loading:
captura de pantalla 2017-11-06 a las 0 54 10
So both $block->load and repository load use directly resource model loading. In the current scope, I understand that currently there is no difference to load a block using load method or repository.

But, regarding BC, I'd say this change is an incompatible change if someone is attached to deprecated model method load, extending cms/page model and redefining load method, or via plugin (beforeLoad, aroundLoad, afterLoad)...

Anyway, it could exist a way to make changes backward compatible; for the blocks case, I think that changing `$this->resource->load($block, $blockId)` to `$block->load($blockId)` in `\Magento\Cms\Model\BlockRepository::getById` method would solve the inconvenients exposed in previous paragraph.

Let me know what you think about this.

ihor-sviziev:
@adrian-martinez-interactiv4 Looks reasonable. I think it's good idea to add all of these changes to 2.3-develop branch, and add backward compatible changes to 2.2-develop.
Also for 2.3-develop would be great to replace $model->load() in cms pages repository to resource model.

dmanners:
@adrian-martinez-interactiv4 so I will put this on-hold just now to focus on the `2.3-develop` pr first. Then once that is sorted out we can work on a back port of these changes. How does that sound?

dmanners:
@adrian-martinez-interactiv4 we have had a few PRs come in for smaller refactoring of this part of the admin section. With that in mind I will close this PR. If you feel that you would like to still provide some of this code, feel free to rebase it onto the latest 2.2-develop branch and reopen a pull request.

Thank you

PR #11846, opened 2017-10-29
Fixed a js bug where ui_component labels have the wrong sort order.

deiserh:
### Description
If you extend an dynamic-row element in a ui_component and add a sort order attribute with a amount between the others, the Fields are in the correct order but the label is always the last one.

### Manual testing scenarios
1. Extend an dynamic-row in a ui_component with a sort order so your element isn't the last one.
Result: The label are in the correct sort order (your Field Label isn't the last one).


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11846) <br/>All committers have signed the CLA.

app/code/Magento/Ui/view/base/web/js/dynamic-rows/dynamic-rows.js
```diff
@@ -548,6 +548,13 @@ define([
                     });
 
                     this.labels.push(data);
+
+                    /**
+                     * Sort the array after an element was added to fix an bug where
+                     * additional added field labels in ui_components haven't the right
+                     * sort order.
+                     */
+                    this.labels.sort(this._compare);
```
omiroshnichenko: Could you, please, move push and sort calls outside of loop for performance improvement. Also, I think it's better to use `_.sortBy()` function and remove custom compare function. An example:
```js
initHeader: function () {
    var labels = [],
        data;

    if (!this.labels().length) {
        _.each(this.childTemplate.children, function (cell) {
            data = this.createHeaderTemplate(cell.config);
            cell.config.labelVisible = false;
            _.extend(data, {
                label: cell.config.label,
                name: cell.name,
                required: !!cell.config.validation,
                columnsHeaderClasses: cell.config.columnsHeaderClasses,
                sortOrder: cell.config.sortOrder
            });

            labels.push(data);
        }, this);
        this.labels(_.sortBy(labels, 'sortOrder'));
    }
}
```

app/code/Magento/Ui/view/base/web/js/dynamic-rows/dynamic-rows.js
```diff
@@ -548,6 +548,13 @@ define([
                     });
```
omiroshnichenko, line 549: It's would be nice if you will mark [dynamic-rows-tier-price.js](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/view/adminhtml/web/js/components/dynamic-rows-tier-price.js#L6-L29) as deprecated, because it was created by mistake for fix same issue, but just for tier price functionality.

omiroshnichenko (changes requested):

omiroshnichenko (approved):

omiroshnichenko:
Hi, @deiserh. I have made update for your code. Thank you for contribution.

PR #11857, opened 2017-10-30
CMS Page - CMS Page - Force validate layout update xml in production mode when saving CMS Page - Handle layout update xml validation exceptions

adrian-martinez-interactiv4:
When saving custom layout update xml in CMS page, **in production mode**, xml is not validated against schema file, due to `\Magento\Framework\App\Arguments\ValidationState::isValidationRequired`method:
```
    /**
     * Retrieve current validation state
     *
     * @return boolean
     */
    public function isValidationRequired()
    {
        return $this->_appMode == \Magento\Framework\App\State::MODE_DEVELOPER;
    }
```

Actual result:
captura de pantalla 2017-10-30 a las 3 46 56
captura de pantalla 2017-10-30 a las 3 47 15

Desired behaviour would be check the layout update xml against the schema, and show error message to the user if it is invalid:
captura de pantalla 2017-10-30 a las 3 58 10

Also, when layout update is validated, validation exceptions are unhandled, showing a report error in production mode, and showing this error in developer mode:
captura de pantalla 2017-10-30 a las 1 03 01

For example, uncomment demo layout handle in home page:
captura de pantalla 2017-10-30 a las 1 02 34

Desired behaviour would be handle the exception properly instead of crash, and show error message to the user:
captura de pantalla 2017-10-30 a las 1 26 21

### Description
- According to doc block, `\Magento\Cms\Controller\Adminhtml\Page\PostDataProcessor::validate` should always return a boolean value. It uses `\Magento\Framework\View\Model\Layout\Update\Validator::isValid` method, that may throw an exception if xml is not valid, so `\Magento\Cms\Controller\Adminhtml\Page\PostDataProcessor::validate` must handle possible exceptions thrown by xml validator, in order to always return a boolean value.
- `\Magento\Cms\Model\Page\DomValidationState` has been added, and it is injected from post data processor to force layout update xml validation, only upon cms page save.

This PR include changes of PR https://github.com/magento/magento2/pull/11865

### Manual testing scenarios
1. Enable developer mode
2. Edit or create a CMS page, enter some invalid xml in Layout Update XML textarea
3. Try to save the page, application crashes

1. Enable production mode
2. Edit or create a CMS page, enter some invalid xml in Layout Update XML textarea
3. Try to save the page, xml does not get validated

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
+++ b/app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
@@ -84,15 +84,15 @@ class PostDataProcessor
     public function validate($data)
     {
         if (!empty($data['layout_update_xml']) || !empty($data['custom_layout_update_xml'])) {
-            /** @var $validatorCustomLayout \Magento\Framework\View\Model\Layout\Update\Validator */
-            $validatorCustomLayout = $this->validatorFactory->create(
+            /** @var $layoutXmlValidator \Magento\Framework\View\Model\Layout\Update\Validator */
+            $layoutXmlValidator = $this->validatorFactory->create(
                 [
                     'validationState' => $this->validationState,
                 ]
             );
 
-            if (!$this->validateData($data, $validatorCustomLayout)) {
-                $validatorMessages = $validatorCustomLayout->getMessages();
+            if (!$this->validateData($data, $layoutXmlValidator)) {
+                $validatorMessages = $layoutXmlValidator->getMessages();
                 foreach ($validatorMessages as $message) {
                     $this->messageManager->addErrorMessage($message);
                 }
@@ -131,17 +131,17 @@ class PostDataProcessor
      * Validate data, avoid cyclomatic complexity
      *
      * @param array $data
-     * @param \Magento\Framework\View\Model\Layout\Update\Validator $validatorCustomLayout
+     * @param \Magento\Framework\View\Model\Layout\Update\Validator $layoutXmlValidator
      * @return bool
      */
-    private function validateData($data, $validatorCustomLayout)
+    private function validateData($data, $layoutXmlValidator)
     {
         try {
-            if (!empty($data['layout_update_xml']) && !$validatorCustomLayout->isValid($data['layout_update_xml'])) {
+            if (!empty($data['layout_update_xml']) && !$layoutXmlValidator->isValid($data['layout_update_xml'])) {
                 return false;
             }
             if (!empty($data['custom_layout_update_xml']) &&
-                !$validatorCustomLayout->isValid($data['custom_layout_update_xml'])
+                !$layoutXmlValidator->isValid($data['custom_layout_update_xml'])
             ) {
                 return false;
             }
--- a/app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
+++ b/app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
@@ -6,6 +6,15 @@
  */
 namespace Magento\Cms\Controller\Adminhtml\Page;
 
+use Magento\Cms\Model\Page\DomValidationState;
+use Magento\Framework\App\ObjectManager;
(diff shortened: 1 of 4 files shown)


app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
```diff
@@ -108,4 +113,26 @@ public function validateRequireEntry(array $data)
...
+     * @param array $data
+     * @param \Magento\Framework\View\Model\Layout\Update\Validator $validatorCustomLayout
+     * @return bool
+     * @throws \Exception
+     */
+    private function validateData($data, $validatorCustomLayout)
+    {
+        $hasError = false;
+        if (!empty($data['layout_update_xml']) && !$validatorCustomLayout->isValid($data['layout_update_xml'])) {
+            $hasError = true;
```
dmanners: I would consider just a return true here

app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
```diff
@@ -108,4 +113,26 @@ public function validateRequireEntry(array $data)
...
+        }
+        if (!empty($data['custom_layout_update_xml']) &&
+            !$validatorCustomLayout->isValid($data['custom_layout_update_xml'])
+        ) {
+            $hasError = true;
```
dmanners: I would consider just a return true here

app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
```diff
@@ -108,4 +113,26 @@ public function validateRequireEntry(array $data)
...
+        }
+        return $hasError;
```
dmanners: I would consider a `return false;` and then we could remove the `$hasError` variable here.

app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
```diff
@@ -65,23 +68,25 @@ public function filter($data)
...
-            }
-            foreach ($validatorCustomLayout->getMessages() as $message) {
-                $this->messageManager->addError($message);
+            try {
+                $hasError = $this->validateData($data, $validatorCustomLayout);
+            } catch (ValidationException $e) {
+            } catch (ValidationSchemaException $e) {
+            } catch (\Exception $e) {
+                $this->messageManager->addExceptionMessage($e);
+            } finally {
```
dmanners: I understand that you want to catch all other exceptions here but from a validation point I think there is only the `ValidationException` and `ValidationSchemaException` unless I am mistaken. In that case I would prefer the following.

```
try {
    $hasError = $this->validateData($data, $validatorCustomLayout);
} catch (ValidationException $e || ValidationSchemaException $e) {
    $validatorMessages = $validatorCustomLayout->getMessages();
    $hasError = $hasError || (bool) count($validatorMessages);
    foreach ($validatorMessages as $message) {
        $this->messageManager->addErrorMessage($message);
    }
} catch (\Exception $e) {
    $this->messageManager->addExceptionMessage($e);
}
```

What do you think?

dmanners (changes requested):
Just a few suggestions on the exceptions and validation methods.

app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php
adrian-martinez-interactiv4: I hadn't seen that before, I mean, I didn't know I could catch several exceptions in a single catch block, I used finally to avoid logic duplication, but also that code must be executed if no exception is caught, so it can't be just there in the catch block. We have to find an intermediate solution.

adrian-martinez-interactiv4: Is it really allowed to do that?
captura de pantalla 2017-10-30 a las 12 38 37

If not, may I use instanceof a generic exception catch?

dmanners (approved):

PR #11878, opened 2017-10-30
[BUGFIX] Made method public so a plugin is possible.

ghost:
Customer isConfirmationRequired method exists in two places. In the [customer Object](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Customer/Model/Customer.php) and in the [accountManagement object ](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Customer/Model/AccountManagement.php)

The method in the customer object is public. Which makes it possible to plugin it. 
The method in the accountManagement object is protected. Which makes it impossible to plugin it. 

This PR only makes the method in isConfirmationRequired in AccountManagement class public. 

- This makes it possible to develop custom business logic to decide if confirmation is required yes/no for certain customers.
- This makes it possible to implement a soft confirmation

- Can plugin the isConfirmationRequired method in AccountManagement


omiroshnichenko:
Hi @dheesbeen. Change of method from protected to public leads to backward compatibility issue. To solve you problem you can create new class that will contain `isConfirmationRequired` method, that will be public, inject this class into AccountManagement and Customer classes, replace all places where old methods used and make deprecation of old methods.

orlangur:
> inject this class into AccountManagement and Customer classes, replace all places where old methods used and make deprecation of old methods.

... or just use this new class inside of `isConfirmationRequired` methods

omiroshnichenko (approved):

orlangur:
@dheesbeen could you please squash changes into a single commit?

Cc @omiroshnichenko

omiroshnichenko:
Hi, @dheesbeen. What is a use case to make plugin on isConfirmationRequired method? What goal you want to achieve?

orlangur:
@omiroshnichenko 
> This makes it possible to develop custom business logic to decide if confirmation is required yes/no for **certain** customers.

It seems pretty clear to me. Logic like

```
if (customer corresponds to some conditions) {
    skip confirmation;
}
```

Probably it's possible with existing code by some `setSkipConfirmation`?

PR #11891, opened 2017-10-31
[Backport 2.3-develop] Update wrong layout update xml handle installed in CMS Home Page by default

adrian-martinez-interactiv4:
Preinstalled xml layout update handle in Home Page is invalid, element action is not expected anymore under referenceContainer node:
captura de pantalla 2017-10-30 a las 1 02 34

captura de pantalla 2017-10-30 a las 1 03 01

Exception handling for this validation will be done when changes from PR https://github.com/magento/magento2/pull/11857 are applied.

### Description
This legacy code shows a demo of how can be removed `right.reports.product.viewed` and `right.reports.product.compared` from right column, but it is outdated; the instructions are not valid and referenced blocks have changed.

This PR suggest changing that preinstalled layout update handle, adapting reference to compared block and removing the viewed block reference, since this last one is now handled via widget and is not at the right column by default, so there is no need to try to remove it: 
```
<referenceContainer name="right">
    <referenceBlock name="catalog.compare.sidebar" remove="true" />
</referenceContainer>
```

### Contribution checklist

### DIFF
--- a/app/code/Magento/Reports/Setup/InstallData.php
+++ b/app/code/Magento/Reports/Setup/InstallData.php
@@ -83,8 +83,7 @@ class InstallData implements \Magento\Framework\Setup\InstallDataInterface
         // @codingStandardsIgnoreStart
         $reportLayoutUpdate = '<!--
     <referenceContainer name="right">
-        <action method="unsetChild"><argument name="alias" xsi:type="string">right.reports.product.viewed</argument></action>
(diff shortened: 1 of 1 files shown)


dmanners:
Will put this on hold until https://github.com/magento/magento2/pull/11863 has been merged.

dmanners (approved):

PR #11903, opened 2017-10-31
7691: address with saveInAddressBook 0 are still being added to the address book for new customers

RomaKis:
If you place an order as a guest and set the save_in_address_book for an address on 0, that address will still be copied to the customer address book when registering as a new customer on the checkout success page.

1. magento/magento2#7691: address with saveInAddressBook 0 are still being added to the address book for new customers.

### Manual testing scenarios
1. Place an order as a guest and set the save_in_address_book for your shipping address to 0 
2. Click the register link on the checkout success page
3. Addresses with save_in_address_book set to 0 should not be copied to the new customer address book.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Setup/UpgradeData.php
+++ b/app/code/Magento/Sales/Setup/UpgradeData.php
@@ -3,18 +3,25 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+
 namespace Magento\Sales\Setup;
 
 use Magento\Eav\Model\Config;
+use Magento\Framework\App\State;
 use Magento\Framework\DB\AggregatedFieldDataConverter;
 use Magento\Framework\DB\DataConverter\SerializedToJson;
 use Magento\Framework\DB\FieldToConvert;
 use Magento\Framework\Setup\ModuleContextInterface;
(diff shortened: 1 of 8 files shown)


app/code/Magento/Sales/Setup/UpgradeData.php
```diff
@@ -118,4 +162,34 @@ private function convertSerializedDataToJson($setupVersion, SalesSetup $salesSet
...
     }
+
+    /**
+     * Fill quote_address_id in table sales_order_address if it is empty.
+     */
+    public function fillQuoteAddressIdInSalesOrderAddress()
+    {
+        $addressCollection = $this->addressCollectionFactory->create();
+        /** @var \Magento\Sales\Model\Order\Address $orderAddress */
+        foreach ($addressCollection as $orderAddress) {
```
valdislav, line 173: @RomaKis Whole collection will be iterated without any limits and load order and quote for every address.  BatchIteration via collection page limits at least should be used. Or it should be upgraded via query not over iterating every address, loading it and saving. However, this is already merged(

RomaKis: @valdislav you can create pull request with this fixed.

PR #11922, opened 2017-11-01
Unable to save some dropdown attributes

enriquei4:
### Description
Some dropdown product attributes can´t be saved in admin, for example Country of manufacturer attr. This was because the JS table validation. Added here: https://github.com/magento/magento2/commit/023b7925dc648bf09813d839990ef5398254264b
 Had a problem with hidden attr for validation

Unable to change country of manufacture default label value #6879

### Manual testing scenarios

1. Navigate to product attributes in admin panel
2. Edit country of manufacture attribute
3. Try to change default label
4. Hit save
5. Now is saved

### Contribution checklist

### DIFF
--- a/lib/web/mage/validation.js
+++ b/lib/web/mage/validation.js
@@ -194,6 +194,7 @@
         var empty = $(element).closest('table')
             .find('input.required-option:visible')
             .filter(function (i, el) {
+                if($(element).is(":visible"))
                 return $.mage.isEmpty(el.value);
             })
             .length;


orlangur:
Invisible fields (for example, placed in a collapsed panel) must be validated.

https://cloud.githubusercontent.com/assets/677017/19105031/29271556-8ae0-11e6-9cb8-b36212a87150.png why label fields are disabled here?

lib/web/mage/validation.js
```diff
@@ -194,6 +194,7 @@
         var empty = $(element).closest('table')
             .find('input.required-option:visible')
             .filter(function (i, el) {
+                if($(element).is(":visible"))
```
okorshenko, line 197: Why do we need to check visibility of the element while we are iterating the child elements?

enriquei4: Because in attributes like country of manufacture, the attribute "required-dropdown-attribute-entry" (not visible) is failing because the default value of the country list is empty.

omiroshnichenko: "required-dropdown-attribute-entry" this is attribute on element that will be hidden always, it's a "marker" for errors. This element was introduced to decrease validation time. Before commit, that you mentioned in description, validation process collected each field in each row in result it was many objects that validates separately. So, you must check for `$(el).is(':disabled')`, because your fix is incorrect and it's make full skip of validation on required label.

okorshenko:
@omiroshnichenko thank you for review. Closing the PR

OlehDizhakGymgrossisten:
Hi guys. @enriquei4 I'm not sure that the problem is with _visible_ attribute. This function should return false, if we have some elements with an empty value and she work well. I think the problem is in other place. You can check this file https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/view/adminhtml/web/js/options.js#L29 and find **add** function. This function add new row to the table, using the _data_ variable. In some cases _data_ object doesn't have some properties values, such as **id**, **store0**, **store1** or etc. 
When we check this file https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/view/adminhtml/templates/catalog/product/attribute/options.phtml#L73 we can find that the value for this element using constructions such as ```value="<%- data.store<?= (int) $_store->getId() ?> %>"``` and get values from _data_ object (I wrote above). Think, we need to init the default values for javascript object, when object doesn't have required values, because in the table we have the input without value, and this code return value more than 0
```
var empty = $(element).closest('table')
            .find('input.required-option:visible')
            .filter(function (i, el) {
                return $.mage.isEmpty(el.value);
            })
            .length;
```
Will be glad to discuss this problem in future and find the best solution. 
Thank you so much!

PR #11933, opened 2017-11-01
Changed constructor typo in Javascript class

dverkade:
Changed variable with typo consturctor to constructor

### Related Pull Requests
https://github.com/magento/magento2/pull/10755

### DIFF
--- a/app/code/Magento/Ui/view/base/web/js/lib/core/class.js
+++ b/app/code/Magento/Ui/view/base/web/js/lib/core/class.js
@@ -27,7 +27,7 @@ define([
      * Creates constructor function which allows
      * initialization without usage of a 'new' operator.
      *
-     * @param {Object} protoProps - Prototypal propeties of a new constructor.
+     * @param {Object} protoProps - Prototypal properties of a new constructor.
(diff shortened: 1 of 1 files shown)


app/code/Magento/Ui/view/base/web/js/lib/core/class.js
```diff
@@ -61,7 +61,8 @@ define([
     Class = createConstructor({
 
         /**
-         * Entry point to the initialization of consturctors' instance.
+         * Entry point to the initialization of
```
orlangur: Looks like auto-replace failed here, please fix, do commit amend and force push to the same branch.

orlangur (changes requested):

app/code/Magento/Ui/view/base/web/js/lib/core/class.js
dverkade: Thanks, weird that it replaced it with nothing. I made the changes, so this is now done.

ihor-sviziev (approved):

orlangur (approved):

PR #11935, opened 2017-11-01
9990: Product import doesn't allow empty attribute values for non-static attributes

RomaKis:
1. magento/magento2#9990: Product import doesn't allow empty attribute values for non-static attributes

### Manual testing scenarios
1. Change value to empty for some attribute in csv.
2. Run product import.
3. Check this attribute for value. It should be empty.

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
@@ -22,6 +22,12 @@ use Magento\Framework\EntityManager\MetadataPool;
 abstract class AbstractType
 {
     /**
+     * When you want to import product with empty value you need
+     * to change its value to value of this constant.
+     */
+    const EMPTY_VALUE = '__EMPTY__VALUE__';
+
+    /**
      * Common attributes cache
      *
      * @var array
@@ -536,6 +542,10 @@ abstract class AbstractType
         foreach ($this->_getProductAttributes($rowData) as $attrCode => $attrParams) {
             if (!$attrParams['is_static'] && empty($rowData[$attrCode])) {
                 unset($rowData[$attrCode]);
+            }
+
+            if (isset($rowData[$attrCode]) && $rowData[$attrCode] === self::EMPTY_VALUE) {
+                $rowData[$attrCode] = null;
             }
         }
         return $rowData;
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product/Validator.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product/Validator.php
@@ -6,6 +6,7 @@
 namespace Magento\CatalogImportExport\Model\Import\Product;
 
 use Magento\CatalogImportExport\Model\Import\Product;
+use Magento\CatalogImportExport\Model\Import\Product\Type\AbstractType;
 use Magento\Framework\Validator\AbstractValidator;
 
 /**
@@ -188,6 +189,11 @@ class Validator extends AbstractValidator implements RowValidatorInterface
         if (!strlen(trim($rowData[$attrCode]))) {
             return true;
         }
+
+        if ($rowData[$attrCode] === AbstractType::EMPTY_VALUE && !$attrParams['is_required']) {
+            return true;
+        }
+
         switch ($attrParams['type']) {
             case 'varchar':
             case 'text':
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product/Validator/Quantity.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product/Validator/Quantity.php
@@ -6,6 +6,7 @@
 namespace Magento\CatalogImportExport\Model\Import\Product\Validator;
 
 use Magento\CatalogImportExport\Model\Import\Product\RowValidatorInterface;
+use Magento\CatalogImportExport\Model\Import\Product\Type\AbstractType;
 
 /**
  * Class Quantity
@@ -18,7 +19,7 @@ class Quantity extends AbstractImportValidator implements RowValidatorInterface
     public function isValid($value)
     {
         $this->_clearMessages();
-        if (!empty($value['qty']) && !is_numeric($value['qty'])) {
+        if (!empty($value['qty']) && (!is_numeric($value['qty']) && $value['qty'] !== AbstractType::EMPTY_VALUE)) {
             $this->_addMessages(
                 [
                     sprintf(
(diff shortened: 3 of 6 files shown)


app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
```diff
@@ -537,7 +544,11 @@ public function clearEmptyData(array $rowData)
             if (!$attrParams['is_static'] && empty($rowData[$attrCode])) {
                 unset($rowData[$attrCode]);
             }
+            if (isset($rowData[$attrCode]) && $rowData[$attrCode] === self::EMPTY_VALUE) {
+                $rowData[$attrCode] = '';
```
dmanners: I assumption here is that this will set the value to '' or 0 rather than removing the value from the DB. I think that this is ok, but I was wondering if maybe it might be also an option to clean up the database values as well. The value of 0 on an integer is different from not set and I would be worried that this could show up as an odd bug in someone's import. What are your thoughts here?

RomaKis: @dmanners, I think you are right about this. The value 0 and not set for integer is it really big difference. It is also difference for varchar or text if it will be '' or null in db. Seems like it will be better to use instead $rowData[$attrCode] = '';  - $rowData[$attrCode] = null; How do you think ?

dmanners: If null cleans the values from the database then that should be good. I am not sure if/how that part of the import is setup though.

RomaKis: @dmanners, yes,  it will be null. I fixed it and pr desription.

dmanners:
@RomaKis during testing it is noted that you cannot update weight. If you try to set the weight as empty you will get the following error message on the check data step. "Value for 'weight' attribute contains incorrect value"

err

This is because the weight field has it's own validator under `app/code/Magento/CatalogImportExport/Model/Import/Product/Validator/Weight.php`

dmanners (approved):

dmanners:
@RomaKis do you think it would be possible to get this to work without having to add the constant value? What would be great to see would be if I leave a column entry empty for a product then the value is reset. Is there any particular reason you went with the `__EMPTY_VALUE__` constant?

RomaKis:
@dmanners, I tried to change code and made a little bit of testing. Seems like it will be good solution, but in that case we will be not able to set null for required values. For example if value for price will be empty in csv file, so we will get exception while checking data.  And I think it is right that we cann't set null to required fields. 
Should I change code and remove the constant?

dmanners:
Yes I think that is fine. Required fields should always have at least to root level value, they could be not set on a specific store and then it would take the value set on the top level for a multi-store example.

Could you make these changes for me? Thank you,

RomaKis:
@dmanners, done.

app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
```diff
@@ -533,11 +533,6 @@ public function prepareAttributesWithDefaultValueForSave(array $rowData, $withDe
      */
     public function clearEmptyData(array $rowData)
     {
-        foreach ($this->_getProductAttributes($rowData) as $attrCode => $attrParams) {
```
dmanners: @RomaKis I would consider that we could now deprecate this method now if it is not doing anything and also remove the usage of this method from any place we can find.

dmanners (changes requested):

dmanners:
@RomaKis from a code point of view I think it looks a lot cleaner now. I would suggest we can depricate and remove usage of the clearEmpty method. Could you also look into the failing tests? https://travis-ci.org/magento/magento2/jobs/306383232#L955

RomaKis:
@dmanners , oops, some problem appeared when I was founding why tests failed.
When you export product with multistore, it export all values for "All store views" store and for others only  different values. So if you want to import this file  you will get bad values for other stores except "All store views" . That's why we can not write all empty values from csv to db as empty values.
Should I return back to version with constant?

dmanners:
@RomaKis Thank you for the feedback. Does this mean that if there is an empty value on the store view level you actually get an entry in the database with an empty string for that value? In an ideal world it should remove the value if there was one already otherwise if there was no store value in the database before import the delete would just not change a row. If this is not the case then we will need to re-think the options here.

RomaKis:
@dmanners, now it works good, but with this pull request we are deleting next lines:
```
if (!$attrParams['is_static'] && empty($rowData[$attrCode])) {
                unset($rowData[$attrCode]);
}
```
So if in file for specific store are empty values they will became empty in db. That's why these changes aren't good. I think we should get back to the version with constant.

dmanners:
Thanks for the feedback @RomaKis I will get back to you about this.

dmanners:
@RomaKis sorry for the delay in getting back to you on this one. Can you revert back to using the constant. This might also be something to reconsider during the import/export community project. It would be great if we did not need a constant to work with the values.

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=11935) <br/>All committers have signed the CLA.

RomaKis:
@dmanners, I changed the code, pls review. 
And I think this feature should be documented. Should I write to somebody about it?  Or what is the process of documenting something?

RomaKis:
@dmanners, any updates on it?

dmanners:
@RomaKis I would still like to talk with @ishakhsuvarov about this PR. I will update you once I have had a chance to discuss the options with him.

dmanners:
@RomaKis sorry for the delay. I think it would be a good option to migrate this PR into the https://github.com/magento-engcom/import-export-improvements this way we can make choices with the PO and also build the right documentation and testing procedures.

Would you be able to migrate it across or would you like me to do it?

dmanners:
@RomaKis I have migrated this pull request over to the import/export improvement (https://github.com/magento-engcom/import-export-improvements/pull/73) project repository. Feel free to join us in this project and we can take this task on. I would like to talk with some community members and the core architectural team about the implementation and also documentation of the new constant.

Thank you for this code.

dan-advantec:
@dmanners I don't suppose you know when this will be merged into master?

dmanners:
@dan-advantec @PieterCappelle is working on this currently with the aim to be in the 2.3 release.

dan-advantec:
@dmanners Perfect, thank you for the update - I'll keep my fingers crossed!

PR #11944, opened 2017-11-02
Report Handled Exceptions To New Relic

mpchadwick:
### Description

Currently, if you're using New Relic you'll know about fatal errors via the APM instrumentation, but handled exceptions will not be reported. This change reports handled exceptions to New Relic

### Manual testing scenarios

1. Configure the New Relic PHP agent in your environment and enable the New Relic module
2. Purposely enter the wrong database credentials in `app/etc/env.php`.
3. Notice the exception being reported to New Relic

### Contribution checklist


toddbc:
This is a really important change to enable error metrics, alerts, and even correlation within New Relic.

Really looking forward to this support within Magento 2.

lib/internal/Magento/Framework/App/Http.php
```diff
@@ -153,6 +153,8 @@ public function launch()
      */
     public function catchException(Bootstrap $bootstrap, \Exception $exception)
     {
+        $this->_eventManager->dispatch('application_handled_exception', ['exception' => $exception]);
```
orlangur: New events must not be created, you can implement this logic as a before plugin for `catchException` method.

orlangur (changes requested):

lib/internal/Magento/Framework/App/Http.php
mpchadwick: @orlangur i refactored it to use a plugin.

app/code/Magento/NewRelicReporting/Plugin/HttpPlugin.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function __construct(
+        Config $config,
+        NewRelicWrapper $newRelicWrapper
+    ) {
+        $this->config = $config;
+        $this->newRelicWrapper = $newRelicWrapper;
+    }
+
+    public function beforeCatchException(Http $subject, $bootstrap, $exception)
```
ihor-sviziev: Could you add types for $bootstrap and $exception? Also phpdoc block is missing

mpchadwick: @ihor-sviziev this is done.

app/code/Magento/NewRelicReporting/Plugin/HttpPlugin.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /**
+     * @var Config
+     */
+    protected $config;
+
+    /**
+     * @var NewRelicWrapper
+     */
+    protected $newRelicWrapper;
```
ihor-sviziev: In general looks good, btw new properties should be private instead of protected. Sorry for missing if prev. time

ihor-sviziev (changes requested):

app/code/Magento/NewRelicReporting/Plugin/HttpPlugin.php
toddbc: Just wondering - even new properties to new classes?  I understand the BC concerns for existing classes (2.2.0 lacking something that 2.1.11 has, etc.), but does it help for new classes too in some way?

ihor-sviziev: According to technical guidelines: http://devdocs.magento.com/guides/v2.1/coding-standards/technical-guidelines/technical-guidelines.html

2.7. All non-public properties and methods SHOULD be private.

mpchadwick: @ihor-sviziev this is done

ihor-sviziev (approved):

iainhubbard:
Not done my background research on this but does lib/internal/Magento/Framework/App/Http.php encompass all the entry points into magento e.g. cron, cli, XML API etc. We have always done this change in pub/errors/report.php. Still a core change obviously. e.g. 

```php
    $processor->saveReport($reportData);
    if (function_exists('newrelic_notice_error')) {
        if (isset($reportData[1])) {
            if (!isset($e) || $e->getTraceAsString() !== $reportData[1]) {
                newrelic_notice_error($reportData[0] . "\n" . $reportData[1]);
            } else {
                newrelic_notice_error($e->getMessage(), $e);
            }
        } elseif (isset($e)) {
            newrelic_notice_error($e->getMessage(), $e);
        }
    }
```
Not tested etc etc

This has the benefit of not relying on magento for anything especially in the event of a catastrophic failure.

ihor-sviziev:
@iainhubbard if you have any issues with this implementation and know how to improve this module - that's really good! waiting for a PR from you

PR #11947, opened 2017-11-02
Fix json encoded attribute backend type when attribute value is null

tkotosz:
When you create a new product attribute with JsonEncoded backend type then you always get the following error when you try to load a product which doesn't have value for that specific attribute yet:

```
Exception #0 (Magento\Eav\Model\Entity\Attribute\Exception): Unable to unserialize value.
```

The source of the problem is this method:
```
public function unserialize($string)
    {
        $result = json_decode($string, true);
        if (json_last_error() !== JSON_ERROR_NONE) {
            throw new \InvalidArgumentException('Unable to unserialize value.');
        }
        return $result;
    }
```
When you do `json_decode(null, true)`, then `json_last_error()` will return 4 (`JSON_ERROR_SYNTAX`).

With the suggested fix it will set the attribute value to empty array when the attribute value is null.

### DIFF
--- a/app/code/Magento/Eav/Test/Unit/Model/Entity/Attribute/Backend/JsonEncodedTest.php
+++ b/app/code/Magento/Eav/Test/Unit/Model/Entity/Attribute/Backend/JsonEncodedTest.php
@@ -95,4 +95,18 @@ class JsonEncodedTest extends \PHPUnit\Framework\TestCase
         $this->model->afterLoad($product);
         $this->assertEquals([1, 2, 3], $product->getData('json_encoded'));
     }
+
+    /**
+     * Test after load handler with null attribute value
+     */
+    public function testAfterLoadWithNullAttributeValue()
+    {
+        $product = new \Magento\Framework\DataObject(
+            [
+                'json_encoded' => null
+            ]
+        );
+        $this->model->afterLoad($product);
+        $this->assertEquals([], $product->getData('json_encoded'));
+    }
 }
--- a/app/code/Magento/Eav/Model/Entity/Attribute/Backend/JsonEncoded.php
+++ b/app/code/Magento/Eav/Model/Entity/Attribute/Backend/JsonEncoded.php
@@ -58,7 +58,7 @@ class JsonEncoded extends \Magento\Eav\Model\Entity\Attribute\Backend\AbstractBa
     {
         parent::afterLoad($object);
         $attrCode = $this->getAttribute()->getAttributeCode();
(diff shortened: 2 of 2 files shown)


ihor-sviziev:
@tkotosz could you cover your change with unit test in order to prevent regression in this part in future?

dmanners:
Thanks for the PR @tkotosz I am happy with the code but I would love to see a test case to cover this. Currently the method `testAfterLoad` only covers the case when the attribute has been set.

tkotosz:
@dmanners updated

app/code/Magento/Eav/Test/Unit/Model/Entity/Attribute/Backend/JsonEncodedTest.php
```diff
@@ -94,5 +94,13 @@ public function testAfterLoad()
         );
         $this->model->afterLoad($product);
         $this->assertEquals([1, 2, 3], $product->getData('json_encoded'));
+
```
ihor-sviziev: Could you move this new code as separate method (test case)?

ihor-sviziev (changes requested):

app/code/Magento/Eav/Test/Unit/Model/Entity/Attribute/Backend/JsonEncodedTest.php
tkotosz: @ihor-sviziev sure, done. Also I have squashed my commits.

ihor-sviziev (approved):

app/code/Magento/Eav/Test/Unit/Model/Entity/Attribute/Backend/JsonEncodedTest.php
```diff
@@ -95,4 +95,18 @@ public function testAfterLoad()
         $this->model->afterLoad($product);
         $this->assertEquals([1, 2, 3], $product->getData('json_encoded'));
     }
+
+    /**
+     * Test after load handler with null attribute value
+     */
+    function testAfterLoadWithNullAttributeValue()
```
dmanners: guess you also need to add the visibility on this method.

dmanners (changes requested):
Add the visibility on the new test method please.

tkotosz:
@dmanners fix pushed :)

ihor-sviziev (approved):

dmanners (approved):

PR #11952, opened 2017-11-02
11832: Create order (on Customer edit page) - not working from admin environment

RomaKis:
1. magento/magento2#11832: Create order (on Customer edit page) - not working from admin environment

### Manual testing scenarios
1. Go to storefront and log in with that sample customer.
2. Add some products to her wishlist.
3. Login to Admin Panel.
4. Navigate to Customers -> All Customers.
5. Enter Edit page of the customer.
6. Use top Action - Create Order.
7. No js errors in console.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Block/Adminhtml/Order/Create/AbstractCreate.php
+++ b/app/code/Magento/Sales/Block/Adminhtml/Order/Create/AbstractCreate.php
@@ -160,4 +160,22 @@ abstract class AbstractCreate extends \Magento\Backend\Block\Widget
             )
             : $this->priceCurrency->convert($value, $this->getStore());
     }
+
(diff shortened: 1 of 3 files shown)


dmanners:
@RomaKis I am just wondering if there could ever be a case or if we should deal with the case that `$_item` is not of type `Product` or it does not have a `getProduct()` method. What do you think?

RomaKis:
@dmanners, for now it isn't possible case, because $block can be only one of these:
screenshot from 2017-11-13 15-36-39
I filled in every of them. Every block has $_item instance of Product except cart and wishlist. And cart and wishlist need have getProduct() method.

dmanners (approved):

PR #12009, opened 2017-11-03
Better respect plugin disabled attribute

scottsb:
### Description
Prior to this fix, the `disabled` attribute on a plugin was not consistently respected: for example, if a plugin was defined in a specific scope but disabled by another module in global scope, it would generate an error because the `instance` wasn't defined in all scopes.

### Manual testing scenarios
1. Define a plugin in a `frontend` or `adminhtml` scope.
2. Add a reference to the plugin in global scope of another module, disabling it.
3. Ensure that `bin/magento setup:di:compile` completes successfully, without generating an error about `'instance'` being an undefined index.

### DIFF
--- a/lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
+++ b/lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
@@ -175,8 +175,8 @@ class PluginList extends Scoped implements InterceptionPluginList
                         unset($plugins[$key]);
                         continue;
                     }
-                    $this->trimInstanceStartingBackslash($plugin);
 
+                    $plugin['instance'] = ltrim($plugin['instance'], '\\');
                     $pluginType = $this->_omConfig->getOriginalInstanceType($plugin['instance']);
                     if (!class_exists($pluginType)) {
                         throw new \InvalidArgumentException('Plugin class ' . $pluginType . ' doesn\'t exist');
@@ -201,17 +201,6 @@ class PluginList extends Scoped implements InterceptionPluginList
             return $plugins;
         }
         return $this->_inherited[$type];
-    }
-
-    /**
-     * Trims starting backslash from plugin instance name
-     *
-     * @param array $plugin
-     * @return void
-     */
-    private function trimInstanceStartingBackslash(&$plugin)
-    {
-        $plugin['instance'] = ltrim($plugin['instance'], '\\');
     }
 
     /**
--- a/lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
+++ b/lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
@@ -168,8 +168,6 @@ class PluginList extends Scoped implements InterceptionPluginList
             if (is_array($plugins) && count($plugins)) {
                 $this->filterPlugins($plugins);
                 uasort($plugins, [$this, '_sort']);
-                $this->trimInstanceStartingBackslash($plugins);
-                $this->_inherited[$type] = $plugins;
                 $lastPerMethod = [];
                 foreach ($plugins as $key => $plugin) {
                     // skip disabled plugins
@@ -177,6 +175,8 @@ class PluginList extends Scoped implements InterceptionPluginList
                         unset($plugins[$key]);
                         continue;
                     }
+                    $this->trimInstanceStartingBackslash($plugin);
+
                     $pluginType = $this->_omConfig->getOriginalInstanceType($plugin['instance']);
                     if (!class_exists($pluginType)) {
                         throw new \InvalidArgumentException('Plugin class ' . $pluginType . ' doesn\'t exist');
@@ -196,6 +196,7 @@ class PluginList extends Scoped implements InterceptionPluginList
                         }
                     }
                 }
+                $this->_inherited[$type] = $plugins;
             }
             return $plugins;
         }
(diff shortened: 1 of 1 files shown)


lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
```diff
@@ -205,14 +206,12 @@ protected function _inheritPlugins($type)
     /**
      * Trims starting backslash from plugin instance name
      *
-     * @param array $plugins
+     * @param array $plugin
      * @return void
      */
-    private function trimInstanceStartingBackslash(&$plugins)
+    private function trimInstanceStartingBackslash(&$plugin)
```
dmanners: I would be interested if we could update this method so that it is no longer using pass by reference but instead returning the updated value. @scottsb do you feel like making this change?

dmanners (changes requested):

lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
scottsb: I'm fine with that. I only kept it this way to try to avoid changing more than I needed. Do you think  it's worth this being a separate method at all? It's private and not used otherwise, so it might be a good candidate for inlining. https://refactoring.guru/inline-method

dmanners: Feel free to remove this method. I do not think, in this case, a single line method helps anything here.

dmanners (changes requested):

scottsb:
@dmanners Updated per review.

dmanners:
Thanks @scottsb I will have a look at this and get back to you with any feedback.

dmanners (approved):

lib/internal/Magento/Framework/Interception/PluginList/PluginList.php
```diff
@@ -168,15 +168,15 @@ protected function _inheritPlugins($type)
             if (is_array($plugins) && count($plugins)) {
                 $this->filterPlugins($plugins);
                 uasort($plugins, [$this, '_sort']);
-                $this->trimInstanceStartingBackslash($plugins);
-                $this->_inherited[$type] = $plugins;
                 $lastPerMethod = [];
                 foreach ($plugins as $key => $plugin) {
                     // skip disabled plugins
                     if (isset($plugin['disabled']) && $plugin['disabled']) {
```
dmanners, line 174: We have tried to test this but could not recreate the bug on 2.2. Every time you setup a plugin as disabled you will hit this if and then unset the plugin.

@scottsb Is there a testing case that I am missing here.

scottsb: I'll put together a sample test case next week.

drew7721:
Hi, I have a concrete example for this issue. For instance the plugin set in [UrlRewrite](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/UrlRewrite/etc/frontend/di.xml) named _setStoreSpecificRedirectUrl_. If you try to disable it from a global di.xml file _(straight in the `etc` folder)_ it will not work, it has to be in the `frontend`.

`<type name="Magento\Store\Block\Switcher">`
`      <plugin name="setStoreSpecificRedirectUrl" disabled="true"/>`
`</type>`

I've implemented and tested the above pull request last week and it worked fine, this allows non global plugins to be disabled from the global scope. I have only tested this with the above example and **did not test if this works for plugins set in the admin scope only.**

Another test would be that we are still able to disable globally set plugins only in a specific scope (frontend or admin).

dmanners:
Thank you very much for that information @drew7721 I will see what I can do with it.

dmanners:
@scottsb and @drew7721 I have done the following steps and am still unable to reproduce the `generating an error about 'instance' being an undefined index.` step in the PR on the 2.2-develop branch.

Create new module with the following files.

app/code/Magento/PlunginTest/registration.php

```
<?php
/**
 */

    'Magento_PluginTest',
    __DIR__
);
```

app/code/Magento/PlunginTest/composer.json

```
{
    "name": "magento/module-plugin-test",
    "description": "N/A",
    "require": {
        "php": "7.0.2|7.0.4|~7.0.6|~7.1.0"
    },
    "type": "magento2-module",
    "version": "100.2.0",
    "license": [
        "OSL-3.0",
        "AFL-3.0"
    ],
    "autoload": {
        "files": [
            "registration.php"
        ],
        "psr-4": {
            "Magento\\PluginTest\\": ""
        }
    }
}
```

app/code/Magento/PlunginTest/etc/module.xml

```
<?xml version="1.0"?>
<!--
/**
 */
-->
    <module name="Magento_PluginTest" setup_version="2.0.0">
    </module>
</config>
```

app/code/Magento/PlunginTest/etc/di.xml

```
<?xml version="1.0"?>
<!--
/**
 */
-->
    <type name="Magento\Store\Block\Switcher">
        <plugin name="setStoreSpecificRedirectUrl" disabled="true"/>
    </type>
</config>
```

Then I enable the new module and run `bin/magento setup:di:compile` at this point according to the PR description I should see the error appear. But nothing appears in the command line or in any of the log files. Also when running the front end the plugin itself is actually disable and not called.

I am closing this PR as we cannot reproduce the problem. If you do have steps to reproduce or you would like to reopen this please feel free.

PR #12010, opened 2017-11-04
[2.3] Magento 2 Store Code validation regex: doesn't support uppercase letters in store code

manuelson:
### Description
Magento 2 Store Code validation regex: doesn't support uppercase letters in store code

1. magento/magento2#11996: Magento 2 Store Code validation regex: doesn't support uppercase letters in store code

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/Model/Store.php
+++ b/app/code/Magento/Store/Model/Store.php
@@ -12,10 +12,10 @@ use Magento\Framework\App\Filesystem\DirectoryList;
 use Magento\Framework\App\Http\Context;
 use Magento\Framework\App\ObjectManager;
 use Magento\Framework\App\ScopeInterface as AppScopeInterface;
-use Magento\Framework\Filesystem;
 use Magento\Framework\DataObject\IdentityInterface;
-use Magento\Framework\Url\ScopeInterface as UrlScopeInterface;
+use Magento\Framework\Filesystem;
 use Magento\Framework\Model\AbstractExtensibleModel;
+use Magento\Framework\Url\ScopeInterface as UrlScopeInterface;
 use Magento\Framework\UrlInterface;
 use Magento\Store\Api\Data\StoreInterface;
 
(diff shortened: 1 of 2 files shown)


app/code/Magento/Store/Model/Store.php
```diff
@@ -463,10 +463,10 @@ protected function _getValidationRulesBeforeSave()
         $storeLabelRule->setMessage(__('Name is required'), \Zend_Validate_NotEmpty::IS_EMPTY);
         $validator->addRule($storeLabelRule, 'name');
 
-        $storeCodeRule = new \Zend_Validate_Regex('/^[a-z]+[a-z0-9_]*$/');
+        $storeCodeRule = new \Zend_Validate_Regex('/^[a-zA-Z]+[a-z0-9_]*$/');
```
vkublytskyi: This expression will not allow code with uppercase letter after number or underscore (e.g. `a1A`). Instead of modifying regexp it's better to apply modifier `i` ([PCRE_CASELESS](http://php.net/manual/en/reference.pcre.pattern.modifiers.php)) to make case-insensitive comparison.

app/code/Magento/Store/Model/Store.php
```diff
@@ -463,10 +463,10 @@ protected function _getValidationRulesBeforeSave()
...
         $storeCodeRule->setMessage(
             __(
-                'The store code may contain only letters (a-z), numbers (0-9) or underscore (_),'
+                'The store code may contain only letters (a-zA-Z), numbers (0-9) or underscore (_),'
```
vkublytskyi: Please do not modify the description as this will break existed translations (will be backward incompatible change). As string does not say "lowercase letters" it is also applicable for  updated implementation

vkublytskyi (changes requested):
For consistent behavior please provide also similar changes in the `\Magento\Store\Model\ResourceModel\Website::_beforeSave`

app/code/Magento/Store/Model/Store.php
manuelson: Done

app/code/Magento/Store/Model/Store.php
manuelson: Done

app/code/Magento/Store/Model/Store.php
zack6849: I would suggest you change the regex to be 

 `$storeCodeRule = new \Zend_Validate_Regex('/^[A-z]+[a-z0-9_]*$/');`

[A-z] is the same as [a-zA-Z] but shorter and more concise

PR #12034, opened 2017-11-05
Solve problem adding product from wishlist when these require more of one unit

PedroBlascoInteractiv4:
Solve problem adding product from wishlist when these require more of one unit

Description

Problem encountered when obtaining the qty parameter. In this case, the qty is sent by post and url param. The  getParam function in Request, get keyparam from the params array (url params)  before the postparams array. I think the params sending by post,  are  more weight that a param sending by urlparams and querystring

Fixed Issues (if relevant)

magento/magento2#9155: Adding product from wishlist not adding to cart showing warning message.
Manual testing scenarios

1.add a simple product from admin with Minimum Qty Allowed in Shopping Cart 10
2.login in front end
3.Add product to wishlist
4.Click on wishlist
5.Try to add product with qty 1 it will redirect to product detail page
6.try to add product with 10 qty in the same page..

### Contribution checklist


PedroBlascoInteractiv4:
@vrann I have a problem with travis test. Return error when this try to find method "getPost" in a interface, but the class that implement this interface have this method. How i can resolve this error?

PR #12057, opened 2017-11-06
[Backport] magento/magento2#9961: Unused product attributes display with value N/A or NO on storefront.

p-bystritsky:
This is a backport for https://github.com/magento/magento2/pull/10619

### Description
Attributes with no value set should not display on storefront.

1. magento/magento2#9961: Unused product attributes display with value N/A or NO on storefront.
2. magento/magento2#6634: Yes/No attribute value is not shown on a product details page

### Manual testing scenarios
1. Create a Custom Attribute,  set it's "Catalog Input Type for Store Owner" to "Text Area" ,  "Visible on Catalog Pages on Storefront" to "Yes".
2. Assign Custom Attribute to Default Attribute Set.
3. Create a Custom Product with Default Attribute Set, make it visible on Storefront, do not set any value to Custom Attribute.
4. Open Custom Product on Storefront - the Custom Attribute label and it's value should not be visible.
5. Edit the Custom Product, set any value to it's Custom Attribute.
6. Open Custom Product on Storefront - the Custom Attribute label and it's value should be visible.

### Contribution checklist


orlangur:
Hi @p-bystritsky, thanks for your contribution, we can start from processing this PR targeting `2.2-develop`. The only problem in old PR was

```
2) Magento\Test\Integrity\Phrase\ArgumentsTest::testArguments
2 missed phrases were discovered: 
Missed Phrase: File: /home/travis/build/magento/magento2/app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributeTest.php 
Line: 68
Missed Phrase: File: /home/travis/build/magento/magento2/app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributeTest.php 
Line: 133
```

which was not running on Travis at that time.

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
```diff
@@ -0,0 +1,161 @@
...
+    private $priceCurrencyInterface;
+
+    /**
+     * @var \Magento\Catalog\Block\Product\View\Attributes
+     */
+    private $attributesBlock;
+
+    protected function setUp()
+    {
+        // @codingStandardsIgnoreStart
```
orlangur: Can we get rid of such annotations?

orlangur (changes requested):

app/code/Magento/Catalog/Test/Unit/Block/Product/View/AttributesTest.php
p-bystritsky: Removed them.

p-bystritsky:
@orlangur can you please provide more information about requested changes?

orlangur:
@p-bystritsky didn't check after last changes yet, Travis builds are green so I suppose it will be good to go.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
```diff
@@ -85,13 +85,15 @@ public function getAdditionalData(array $excludeAttr = [])
 
                 if (!$product->hasData($attribute->getAttributeCode())) {
                     $value = __('N/A');
+                } elseif ($value instanceof Phrase) {
+                    $value = (string)$value;
                 } elseif ((string)$value == '') {
                     $value = __('No');
```
vkublytskyi: This is weird logic. Why empty string should be transformed to `No`. Looks like this some obsolete workaround that should be removed. If this cause any issue then it definitely should be fixed in another way.
Please remove this `elseif` logic as it has no sense.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
```diff
@@ -85,13 +85,15 @@ public function getAdditionalData(array $excludeAttr = [])
 
                 if (!$product->hasData($attribute->getAttributeCode())) {
                     $value = __('N/A');
```
vkublytskyi: I think we also should remove this code as there are a lot of complaints about such behavior. This code was "dead" for a long time so everyone expects that not used attribute is ignored on storefront.
Please remove this `if`

vkublytskyi (changes requested):
The integrity tests are fixed now.

But I would like to ask you clean up the code and remove obsolete logic to improve code readability and make behavior more explicit. See my inline comments.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
orlangur: Can we change system behavior in a way desired by Community without PO approval?

app/code/Magento/Catalog/Block/Product/View/Attributes.php
orlangur: Most likely this was used to display boolean attributes with Yes/No values.

vkublytskyi: @orlangur Yes/No values at this point already handled in `\Magento\Eav\Model\Entity\Attribute\Frontend\AbstractFrontend::getValue`. And this is dangerous to substitute empty value with "No" as empty value may appear for different reasons (e.g. empty text field, dropdown without selected item, etc.). Magento should rid of such code, decrease coupling between different classes and do not base implementation on current execution flow but build reusable components with high cohesion so changes in one system part will not cause bugs in another.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
vkublytskyi: Actually, behavior was changed when `instanceof Phrase` was added to `if` below. I can see that `__('N/A')` code was dead for a long time (including M1) so expectations in most Magento versions not used attributes are not displayed on a storefornt product page. Support also has multiple tickets to return previous behavior (not show N/A) back.

orlangur: Thanks, I wan't aware about M1, and yes, such useful behavior was introduced accidentally in scope of another bugfix.

app/code/Magento/Catalog/Block/Product/View/Attributes.php
```diff
@@ -81,17 +81,15 @@ public function getAdditionalData(array $excludeAttr = [])
...
-                if (!$product->hasData($attribute->getAttributeCode())) {
-                    $value = __('N/A');
-                } elseif ((string)$value == '') {
-                    $value = __('No');
-                } elseif ($attribute->getFrontendInput() == 'price' && is_string($value)) {
+                $value = $attribute->getFrontend()->getValue($product);
+
+                if ($value instanceof Phrase) {
+                    $value = (string)$value;
+                } if ($attribute->getFrontendInput() == 'price' && is_string($value)) {
```
orlangur: This `if` formatting looks pretty strange, should it be on next line or `elseif`? For the sake of cleanup `===` use would be nice also.

orlangur (changes requested):

app/code/Magento/Catalog/Block/Product/View/Attributes.php
vkublytskyi: Sure. should be `elseif`. Thanks

vkublytskyi (approved):

Ctucker9233:
@magento-engcom-team Are these changes safe for 2.1.x? Is there a backport in progress? If not, I'll take it.

PR #12060, opened 2017-11-06
Missing dependency of Google AdWords

phoenix-bjoern:
If you disable Magento Google Analytics (bin/magento module:disable Magento_GoogleAnalytics) the configuration section "Google API" is missing the Magento backend and Google AdWords can't be configured. However, Magento_GoogleAnalytics is not a dependency of Magento_GoogleAdWords.
To fix this Magento_GoogleAnalytics should be declared as dependency of Magento_GoogleAdWords so that the configuration is possible.

### DIFF
--- a/app/code/Magento/GoogleAdwords/composer.json
+++ b/app/code/Magento/GoogleAdwords/composer.json
@@ -5,8 +5,8 @@
         "php": "7.0.2|7.0.4|~7.0.6|~7.1.0",
         "magento/module-store": "100.2.*",
         "magento/module-sales": "101.0.*",
-        "magento/module-google-analytics": "100.2.*",
-        "magento/framework": "101.0.*"
+        "magento/framework": "101.0.*",
+        "magento/module-google-analytics": "100.2.*"
     },
     "type": "magento2-module",
     "version": "100.2.0",
--- a/app/code/Magento/GoogleAdwords/composer.json
+++ b/app/code/Magento/GoogleAdwords/composer.json
@@ -5,6 +5,7 @@
         "php": "7.0.2|7.0.4|~7.0.6|~7.1.0",
         "magento/module-store": "100.2.*",
         "magento/module-sales": "101.0.*",
+        "magento/module-google-analytics": "100.2.*",
         "magento/framework": "101.0.*"
     },
     "type": "magento2-module",


miguelbalparda:
Thanks for your contribution @phoenix-bjoern! Can you please reformat the PR to use the correct template?

miguelbalparda:
Also, can you please check the failing tests? Specifically [this](https://travis-ci.org/magento/magento2/jobs/299073135) one.

phoenix-bjoern:
@miguelbalparda Thanks for heads up! I've checked the build errors several times but honestly I don't understand the error. There is no dependency between the Google_AdWords and the Google_Analytics module. So I don't understand why it should be redundant. Or is the error because both modules have these dependencies?

`        "php": "~5.6.5|7.0.2|7.0.4|~7.0.6",
        "magento/module-store": "100.1.*",
        "magento/module-sales": "100.1.*",
        "magento/framework": "100.1.*"`

> Can you please reformat the PR to use the correct template?

Do you have a link for me? Didn't find any info here http://devdocs.magento.com/guides/v2.0/contributor-guide/contributing.html

phoenix-bjoern:
@miguelbalparda would appreciate if you could follow up on my comments.

miguelbalparda:
Hi @phoenix-bjoern! Sorry for the late reply. When you create a new PR there is a template you probably deleted. You can see a PR with the correct template [here](https://github.com/magento/magento2/pull/12137) or see the actual template [here](https://github.com/magento/magento2/blob/2.2-develop/PULL_REQUEST_TEMPLATE.md). Regarding the error, let me check that internally and report back.

okorshenko:
Hello @phoenix-bjoern @miguelbalparda 
Thank you for the research on the issue. The test fails correctly. We should not have dependency between `Magento_GoogleAnalytics` and `Magento_GoogleAdwords`
These two modules have a dependency on the menu level and on the ACL resource. 
In order to resolve this dependency, we need to introduce the new module `Magento_GoogleApi` and move shared resources to the new module. `Magento_GoogleAnalytics` and `Magento_GoogleAdwords` will depend on `Magento_GoogleApi`

The most important part of this fix will be backward compatibility. We should find a way how we can preserve backward compatibility for the users after the upgrade.
Please let me know if you need more details. 

**note**
Adding the dependency on existing module is BiC

okorshenko:
Closing this PR. @phoenix-bjoern feel free to submit a pull request following my previous comment.

PR #12103, opened 2017-11-08
Fix coupon code failing to apply discount to shipping amount (2.1-develop)

driskell:
### Issue

Create new coupon code that applies to shipping amount, has fixed discount, and has a product condition that is never met (qty>999 for example) so it only ever applies to shipping.
This is a "£5 off shipping for all orders".
Use table rates to apply a cost to shipping of more than £5 or exactly £5.
Create new backend order with this coupon code and apply to an order, so it should only discount the shipping. You will see the totals update correspondingly.
Ensure "Set shipping as billing" is ticked.

Submit the order.

### Expected

Order is created with same totals and the £5 discount to shipping.

### Actual

Order is created and there is no discount. The shipping is charged full and the total amount charged to the customer is higher than what was on the previous screen.

### Reason

When the order is submitted, the Magento\Sales\Model\AdminOrder\Create::setShippingAsBilling function overwrites the shipping amounts in the shipping address. It copies billing to shipping but the data contains not only a blank method but also 0 against all shipping amounts. This data is not recalculated. When Magento\SalesRule\Model\Quote\Discount comes along to calculate discount it sees 0 shipping amounts on the address and fails to apply any discount.

### Fix

Inside Magento\Sales\Model\AdminOrder\Create::setShippingAsBilling trigger a recollect of shipping rates.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/AdminOrder/Create.php
+++ b/app/code/Magento/Sales/Model/AdminOrder/Create.php
@@ -417,6 +417,8 @@ class Create extends \Magento\Framework\DataObject implements \Magento\Checkout\
         }
 
         if ($this->_needCollect) {
+            $this->getQuote()->getShippingAddress()->unsCachedItemsAll();
+            $this->getQuote()->setTotalsCollectedFlag(false);
             $this->getQuote()->collectTotals();
         }
 
--- a/app/code/Magento/Sales/Model/AdminOrder/Create.php
+++ b/app/code/Magento/Sales/Model/AdminOrder/Create.php
@@ -1413,11 +1413,14 @@ class Create extends \Magento\Framework\DataObject implements \Magento\Checkout\
             $tmpAddress->unsAddressId()->unsAddressType();
             $data = $tmpAddress->getData();
             $data['save_in_address_book'] = 0;
+            // Don't lose the shipping method
+            unset($data['shipping_method']);
             // Do not duplicate address (billing address will do saving too)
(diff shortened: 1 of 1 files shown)


sidolov:
Hi @driskell, I can't reproduce described issue. 

My Rule test data: 
- Coupon: Specific coupon
- Coupon Code: test_coupon

Actions: 
- Apply: Fixed amount discount
- Discount Amount: 5
- Apply to Shipping Amount: Yes
- Actions Condition: Quantity in cart  equals or greater than  999

Steps: 

1. Proceed to Sales -> Orders -> Create New Order 
2. Select existing customer 
3. Add simple product to cart with price $10 
4. Verify that checkbox "Same As Billing Address" is checked
5. Click on link "Get shipping methods and rates"
6. Select shipping method with price $5
7. Order Grand Total: $15.00
8. Apply Coupon Code: test_coupon
9. Message "The coupon code has been accepted." appears
10. Order Grand Total: $10.00
11. Press Submit Order button
12. Order created

Order totals the same as in the previous step, discount is present in totals summary.
ea263a591f

Maybe you can provide additional steps to reproduce this issue, for now, all works as expected on the latest commit (2.1-develop:bf06306f4ec8d9ba2335860c91b9642586ce1351)

sidolov:
Hi @driskell , will you continue work with PR?

sidolov:
@driskell , I am closing this PR now due to inactivity.
Thank you for collaboration

PR #12050, opened 2017-11-05
[2.1] - Add command to view mview state and queue

convenient:
- PR for 2.1 = #12050
- PR for 2.2 = #12122
- PR for 2.3 = #12592

### Description

This is similar to the magerun1 command here: https://github.com/netz98/n98-magerun/pull/891

I like the ability to view the mview queue in realtime as its being processed, it can be quite helpful when debugging indexing issues.

This command will actually show how many items are in the list pending processing, as well information from the `mview_state` table.

```
php bin/magento indexer:status:mview
+---------------------------+----------+--------+---------------------+------------+---------+
| ID                        | Mode     | Status | Updated             | Version ID | Backlog |
+---------------------------+----------+--------+---------------------+------------+---------+
| catalog_category_product  | enabled  | idle   | 2017-11-02 10:00:00 | 1          | 0       |
| catalog_product_attribute | enabled  | idle   | 2017-11-02 10:00:00 | 1          | 1       |
| catalog_product_category  | disabled | idle   | 2017-11-02 10:00:00 | 1          | 0       |
| catalog_product_price     | enabled  | idle   | 2017-11-02 10:00:00 | 1          | 0       |
+---------------------------+----------+--------+---------------------+------------+---------+
```

screen shot 2017-11-05 at 20 47 47

### Manual testing scenarios
1. Enable an indexer `php bin/magento indexer:set-mode schedule catalog_product_price`
2. Trigger an action on the product price that will update the `_cl` table
3. Run `php bin/magento indexer:status:mview` to see the backlog increase
4. `magerun2 sys:cron:run indexer_update_all_views` will trigger the mview processes
5. See the backlog decrease 

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

PR #12122, opened 2017-11-08
[2.2] - Add command to view mview state and queue

convenient:
- PR for 2.1 = https://github.com/magento/magento2/pull/12050
- PR for 2.2 = https://github.com/magento/magento2/pull/12122
- PR for 2.3 = https://github.com/magento/magento2/pull/12592

### Description

Update `indexer:status` to output information about the schedule mview backlog

screen shot 2017-12-04 at 16 55 50

### Manual testing scenarios
1. Enable an indexer `php bin/magento indexer:set-mode schedule catalog_product_price`
2. Trigger an action on the product price that will update the `_cl` table
3. Run `php bin/magento indexer:status` to see the backlog increase
4. `magerun2 sys:cron:run indexer_update_all_views` will trigger the mview processes
5. See the backlog decrease 

### ~~Original Description~~ 

~~This is similar to the magerun1 command here: https://github.com/netz98/n98-magerun/pull/891~~

~~I like the ability to view the mview queue in realtime as its being processed, it can be quite helpful when debugging indexing issues.~~

~~This command will actually show how many items are in the list pending processing, as well information from the `mview_state` table.~~

```
php bin/magento indexer:status:mview
```

screen shot 2017-11-05 at 20 47 47

### Contribution checklist


ihor-sviziev (approved):

ihor-sviziev (changes requested):
Unit tests are failed. Could you adjust them according to new phpunit version?

kandy:
I personally more like the idea to update indexer:status command instead of introducing of new.

app/code/Magento/Indexer/Console/Command/IndexerStatusMviewCommand.php
```diff
@@ -0,0 +1,98 @@
...
+                $state = $view->getState();
+                $changelog = $view->getChangelog();
+
+                try {
+                    $currentVersionId = $changelog->getVersion();
+                } catch (View\ChangelogTableNotExistsException $e) {
+                    continue;
+                }
+
+                $pendingCount = count($changelog->getList($state->getVersionId(), $currentVersionId));
```
kandy: I believe that this code will have memory usage  issue in case of huge number of item in log

convenient: Good point, it's defined as 

```php
    public function getList($fromVersionId, $toVersionId)
    {
        $changelogTableName = $this->resource->getTableName($this->getName());
        if (!$this->connection->isTableExists($changelogTableName)) {
            throw new ChangelogTableNotExistsException(new Phrase("Table %1 does not exist", [$changelogTableName]));
        }

        $select = $this->connection->select()->distinct(
            true
        )->from(
            $changelogTableName,
            [$this->getColumnName()]
        )->where(
            'version_id > ?',
            (int)$fromVersionId
        )->where(
            'version_id <= ?',
            (int)$toVersionId
        );

        return $this->connection->fetchCol($select);
    }
```

I'll address.

ishakhsuvarov:
@convenient Please let us know when it's updated.
Thanks!

convenient:
@ishakhsuvarov can i have some clarification please.

Is [this comment](https://github.com/magento/magento2/pull/12122#issuecomment-343116800 ) a requirement or a suggestion?  
>I personally more like the idea to update indexer:status command instead of introducing of new.

I'd prefer to keep them separate tbh, but let me know please.

ihor-sviziev:
@convenient from my perspective would be great to improve existing indexer:status command because mview status it's like part of indexer status (see as highly recommended), but if you can't do it right now - probably we could accept your implementation as MVP.
Could you share your idea why do you prefer to separate these two commans?

convenient:
I've given this a quick hack locally and I can't find a pretty way of outputting all the data I'd need to vary between the scheduled indexers, and the non scheduled indexers.

```
bin/magento indexer:status
+----------------------+------------------+-----------+-----------------+---------------------+------------+---------+
| Title                | Status           | Update On | Schedule Status | Updated             | Version ID | Backlog |
+----------------------+------------------+-----------+-----------------+---------------------+------------+---------+
| Design Config Grid   | Ready            | Save      |                 |                     |            |         |
| Catalog Product Rule | Reindex required | Schedule  | idle            | 2017-11-08 20:33:05 | 1          | 0       |
| Catalog Rule Product | Reindex required | Schedule  | idle            | 2017-11-08 20:33:04 | 1          | 0       |
| Catalog Search       | Ready            | Schedule  | idle            | 2017-11-08 20:33:07 | 1          | 4       |
| Category Products    | Reindex required | Schedule  | idle            | 2017-11-08 20:33:02 | 1          | 0       |
| Customer Grid        | Ready            | Schedule  | idle            | 2017-11-08 20:33:02 | 1          | 0       |
| Product Categories   | Reindex required | Schedule  | idle            | 2017-11-08 20:33:02 | 1          | 0       |
| Product EAV          | Reindex required | Schedule  | idle            | 2017-11-08 20:33:04 | 1          | 0       |
| Product Price        | Reindex required | Schedule  | idle            | 2017-11-08 20:33:03 | 1          | 0       |
| Stock                | Reindex required | Schedule  | idle            | 2017-11-08 20:33:04 | 1          | 0       |
+----------------------+------------------+-----------+-----------------+---------------------+------------+---------+
```

Having the `$indexer->getStatus()` as well as the `$view->getStatus()` in the same table just seems messy to me, especially for "Update on Save" indexers.

Either way I'm happy to do whatever is required, I just want to know how it should look at the end.

Bundling both the scheduled and the non-scheduled indexers into the same table as above really makes the `indexer:show-mode` command a bit superfluous as well.

What does everyone think?

ihor-sviziev:
@convenient probably show all information that we have isn't really good idea. I would like to remove "version id" column, and data from "backlog" column show as part of "schedule status", like "Idle (4 in backlog)", I think it will be better in this way.

convenient:
@ihor-sviziev 
```
+----------------------+------------------+-----------+---------------------+---------------------+
| Title                | Status           | Update On | Schedule Status     | Schedule Updated    |
+----------------------+------------------+-----------+---------------------+---------------------+
| Design Config Grid   | Ready            | Save      |                     |                     |
| Catalog Product Rule | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:05 |
| Catalog Rule Product | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:04 |
| Catalog Search       | Ready            | Schedule  | idle (4 in backlog) | 2017-11-08 20:33:07 |
| Category Products    | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:02 |
| Customer Grid        | Ready            | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:02 |
| Product Categories   | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:02 |
| Product EAV          | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:04 |
| Product Price        | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:03 |
| Stock                | Reindex required | Schedule  | idle (0 in backlog) | 2017-11-08 20:33:04 |
+----------------------+------------------+-----------+---------------------+---------------------+
```

It looks a little neater, but i'm usually a opposed to bundling semantic data into a string like this which would be a bit harder to `awk` out etc. But that might just be my preference :)

This data would still make `indexer:show-mode` a bit redundant. What do you think?

ihor-sviziev:
@convenient from my perspective - looks good.

@ishakhsuvarov could you review latest comment. Would it be ok to improve this command in this way?

convenient:
Once we've decided on how the output looks i'll update the test suite and push up my commits, just delaying doing any further work until the output is agreed.

vkublytskyi:
@convenient I had a conversation with @kandy and he is ok with the proposed output.

> This data would still make indexer:show-mode a bit redundant. What do you think?

It still may be usable if someone need only to check when indexer will be triggered. Plus removing of CLI commands are not allowed due to backward compatibility policy.

convenient:
@vkublytskyi thanks, i'm working on this and will update soon.

app/code/Magento/Indexer/Console/Command/IndexerStatusCommand.php
```diff
@@ -30,21 +32,84 @@ protected function configure()
...
+
+        $table->addRows($rows);
+        $table->render($output);
+    }
+
+    /**
+     * @param Indexer\IndexerInterface $indexer
+     * @return string
+     */
+    protected function getStatus(Indexer\IndexerInterface $indexer)
```
ihor-sviziev: Could you use private method visibility instead of protected?
Same for getPendingCount method

app/code/Magento/Indexer/Test/Unit/Console/Command/IndexerStatusCommandTest.php
```diff
@@ -18,35 +20,132 @@ class IndexerStatusCommandTest extends AbstractIndexerCommandCommonSetup
...
-            . sprintf('%-50s ', 'Title_indexerTwo' . ':') . 'Reindex required' . PHP_EOL
-            . sprintf('%-50s ', 'Title_indexerThree' . ':') . 'Processing' . PHP_EOL
-            . sprintf('%-50s ', 'Title_indexerFour' . ':') . 'unknown' . PHP_EOL;
 
-        $this->assertStringStartsWith($expectedValue, $actualValue);
+        $linesOutput = array_filter(explode(PHP_EOL, $commandTester->getDisplay()));
+
+        $this->assertCount(8, $linesOutput, 'There should be 8 lines output. 3 Spacers, 1 header, 4 content.');
+        $this->assertEquals($linesOutput[0], $linesOutput[2], "Lines 0, 2, 7 should be spacer lines");
+        $this->assertEquals($linesOutput[2], $linesOutput[7], "Lines 0, 2, 6 should be spacer lines");
```
ihor-sviziev: Probably there is mistake

lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -127,14 +127,12 @@ public function clear($versionId)
...
-     * Retrieve entity ids by range [$fromVersionId..$toVersionId]
-     *
      * @param int $fromVersionId
      * @param int $toVersionId
-     * @return int[]
+     * @return \Magento\Framework\DB\Select
      * @throws ChangelogTableNotExistsException
      */
-    public function getList($fromVersionId, $toVersionId)
+    protected function getListSelect($fromVersionId, $toVersionId)
```
ihor-sviziev: All new added methods should have private or public visibility

lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -154,9 +152,39 @@ public function getList($fromVersionId, $toVersionId)
...
 
+    /**
+     * Retrieve the count of entity ids in the range [$fromVersionId..$toVersionId]
+     *
+     * @param int $fromVersionId
+     * @param int $toVersionId
+     * @return int[]
+     * @throws ChangelogTableNotExistsException
+     */
+    public function getListSize($fromVersionId, $toVersionId)
```
ihor-sviziev: This method was just added, this change is not backward compatible. Unfortunately we can't accept  such change. Couldn't it be done without adding new methods?
If no - then we need to introduce new interface that will include new method.

ihor-sviziev (changes requested):

lib/internal/Magento/Framework/Mview/View/Changelog.php
convenient: Thanks I was waiting for the suite to go green before raising questions about this function, this was my attempt to address this comment: https://github.com/magento/magento2/pull/12122#discussion_r149927089

I don't believe this can be achieved without a similar query, I was trying to prevent the `SELECT` query from being generated in two places, but if the interface is not up for modification I can either do it in a new interface or directly within the `Command` class itself. How do you think this should be architected? If you'd suggest a new interface how do you think it should be named etc?

I'm very much appreciating all your feedback on this PR by the way.

convenient: Is this 4af04b1c43ea8b1198ba5027500552c88a09737c what you meant?

app/code/Magento/Indexer/Console/Command/IndexerStatusCommand.php
```diff
@@ -30,21 +32,84 @@ protected function configure()
...
+
+        try {
+            $currentVersionId = $changelog->getVersion();
+        } catch (Mview\View\ChangelogTableNotExistsException $e) {
+            return '';
+        }
+
+        $state = $view->getState();
+
+        $pendingCount = $changelog->getListSize($state->getVersionId(), $currentVersionId);
```
vkublytskyi: `\Magento\Framework\Mview\ViewInterface::getChangelog` returns `\Magento\Framework\Mview\View\ChangelogInterface` so method `getListSize` cannot be used here as it is declared in `ChangelogCounterInterface` and is unkown in this code.

Seems there is no *clean* and *backward compatible* solution to implement counting without list load. I propose in the scope of this PR implement counting as was done in the initial version with `count($changelog->getList())` and process this PR to deliver this feature to M2. Once PR will be merged new PR may be created with a refactoring of changelog and performance optimization. 

@convenient, @ihor-sviziev, @kandy are you agree with this plan?

ihor-sviziev:
Sure

app/code/Magento/Indexer/Console/Command/IndexerStatusCommand.php
convenient: Thanks @vkublytskyi I was really struggling to balance out clean/backwards-compatability as you noted. 

If @kandy agrees my plan of action will be
1. Roll back to `count($changelog->getList())` with associated tests
2. Update description/title to properly highlight what is occurring in this PR
3. Wait for approval from all parties (or potentially for merge?)
4. Backport changes to #12050 

Sound good?

kandy: Ok, looks good for me

Issue #6239, opened 2016-08-18
[Magento_Version] Do not publicly disclose store version


kassner:
Hi,

Looks like we have this particular [Magento_Version](https://github.com/magento/magento2/blob/develop/app/code/Magento/Version/Controller/Index/Index.php) module, that the unique purpose is to disclose publicly which Magento version and edition the store is running. Although is a good idea to allow other systems that integrate to Magento retrieve the current version to do proper API calls, I strongly believe this should not be available wide open, but rather inside secure/authenticated endpoint.

Can we make this call authenticated? Or maybe remove the module at all? Why outsiders need to know which minor version a particular store is running?

Thanks!

avoelkl:
I agree, I would not show this as a public available information per default.

veloraven:
@kassner your proposition looks like an improvement.
Could you please transfer it to the new  [Magento 2 Feature Requests and Improvements forum](https://community.magento.com/t5/Magento-2-Feature-Requests-and/idb-p/feature-requests) (see details [here](https://community.magento.com/t5/News-Announcements/Improvements-to-GitHub-Management/m-p/44572#M96))?

kassner:
@veloraven no, I think it is a security issue, not a feature request.

veloraven:
@kassner than please format it according to the [Issue reporting guidelines](https://github.com/magento/magento2/wiki/Issue-reporting-guidelines): with steps to reproduce, actual result and expected result.
Please, also identify which version of Magento you are running.

veloraven:
According to [contributor guide](http://devdocs.magento.com/guides/v2.1/contributor-guide/contributing.html), tickets without response for two weeks should be closed.
If this issue still reproducible please feel free to create the new one: format new issue according to the [Issue reporting guidelines](https://github.com/magento/magento2/wiki/Issue-reporting-guidelines): with steps to reproduce, actual result and expected result and specify Magento version.

kassner:
@benmarks @sherrierohde What is the point of having community moderators if the only thing they do is poking about a damn ticket structure?

miguelbalparda:
You were told to go to the forums or reformat the issue according to the guidelines and you failed to do it. These rules are probably here to keep things in order and have a clear workflow (and make things reproducible for others FWIW). 
If everybody thinks the issue they are reporting is more important than others and that they don't need to follow the guidelines because "the only thing they (community moderators) do is poking about a damn ticket structure" this would be a total chaos.
[...]

PR #12134, opened 2017-11-09
[Magento 2.2] - Do not allow to see the version of magento in production mode

crissanclick:
### Description
When an user enter to domain.com/magento_version can see magento version.

### Manual testing scenarios
1. Set Magento to Production mode.
2. Enter to domain.com/magento_version

### Contribution checklist

### DIFF
--- a/app/code/Magento/Version/Controller/Index/Index.php
+++ b/app/code/Magento/Version/Controller/Index/Index.php
@@ -9,7 +9,7 @@ namespace Magento\Version\Controller\Index;
 use Magento\Framework\App\Action\Action;
 use Magento\Framework\App\Action\Context;
 use Magento\Framework\App\ProductMetadataInterface;
-use Magento\Framework\Exception\StateException;
+use Magento\Framework\App\State;
 
 /**
  * Magento Version controller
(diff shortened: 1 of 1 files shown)


miguelbalparda:
This was already discussed [here](https://github.com/magento/magento2/issues/6239). We can disagree on the original motive to include this in the core but this PR seems to be a feature request instead of a bugfix.

app/code/Magento/Version/Controller/Index/Index.php
```diff
@@ -22,12 +22,22 @@ class Index extends Action
...
      * @param Context $context
      * @param ProductMetadataInterface $productMetadata
+     * @param State $state
      */
-    public function __construct(Context $context, ProductMetadataInterface $productMetadata)
-    {
+    public function __construct(
+        Context $context,
+        ProductMetadataInterface $productMetadata,
+        State $state
```
dmanners: This will need to be injected in with an optional dependency. See http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#adding-a-constructor-parameter

app/code/Magento/Version/Controller/Index/Index.php
```diff
@@ -22,12 +22,22 @@ class Index extends Action
     protected $productMetadata;
 
     /**
+     * @var State
+     */
+    protected $state;
```
dmanners: New properties can be private by default.

app/code/Magento/Version/Controller/Index/Index.php
```diff
@@ -39,9 +49,14 @@ public function __construct(Context $context, ProductMetadataInterface $productM
      */
     public function execute()
     {
+        if ($this->state->getMode() == State::MODE_PRODUCTION) {
```
dmanners: Would an exact match work here also?

dmanners (changes requested):

MaxSavich:
This feature has been implemented intentionally.
We will work on it's improvement in order not to disclose Magento version to public.

dmanners:
I will close this as it will be improved by the security team.

dragoonis:
@dmanners what did you (or another magento team) do about this? You gotta keep people secure. Looking forward to seeing the implementation.

Styopchik:
Want to see it too

PR #12166, opened 2017-11-10
Removed

p-bystritsky:
### Description
On frontend framework will update session expiration time on every request after session cookie is set.

1. magento/magento2##7627: Cookie expiration time is not extended when browsing the frontend.

### Manual testing scenarios
1. Browse frontend until PHPSESSID is set. Take a look on PHPSESSID expiration time.
2. Open any page.
3. Make sure PHPSESSID expiration time is updated.

### Contribution checklist


app/code/Magento/Cookie/Model/App/Response/HttpPlugin.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * We need this method because else __construct would not be called.
+     *
+     * @param Http $subject
+     * @return void
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function beforeSendResponse(Http $subject)
+    {
+        // Blank. Creating object that implements SessionManagerInterface is enough.
```
ishakhsuvarov: Please provide more detailed background for this kind of fix, adding a plugin to Http Response seems to be an overhead.

app/code/Magento/Cookie/etc/frontend/di.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
```
ishakhsuvarov: Module `Cookie` is not responsible for managing Session Cookies.
It's only purpose is to display a warning, that website uses cookies, when configured accordingly.

app/code/Magento/Cookie/Model/App/Response/HttpPlugin.php
```diff
@@ -0,0 +1,39 @@
...
+ * HTTP response plugin for frontend.
+ */
+class HttpPlugin
+{
+    /**
+     * @var SessionManagerInterface
+     */
+    private $sessionManager;
+
+    public function __construct(
```
ishakhsuvarov: Doc Block is missing

app/code/Magento/Cookie/Model/App/Response/HttpPlugin.php
```diff
@@ -0,0 +1,39 @@
...
+        SessionManagerInterface $sessionManager
+    ) {
+        $this->sessionManager = $sessionManager;
```
ishakhsuvarov: Dependency which is not used anywhere in the class should not be injected.
It is a good idea to look for an alternative solution.

ishakhsuvarov (changes requested):
Please review the comments and rework this PR if you wish to continue.

p-bystritsky:
@ishakhsuvarov  Moved changes to /lib/internal/Magento/Framework/App/Response/Http.php

omiroshnichenko:
Hi, @p-bystritsky. This fix will not work with Varnish or another cache server. Correct fix is to implement mechanism where some JS will make a request on Magento controller that will check session and prolong it. So, will you plan to make such implementation or I can close this PR?

p-bystritsky:
@omiroshnichenko, yes, this PR should be closed.

PR #12169, opened 2017-11-10
12073: ProductRepository fails to load product with camelcase SKU

RomaKis:
ProductRepository fails to load product with camelcase SKU

1. magento/magento2#12073: ProductRepository fails to load product with camelcase SKU

### Manual testing scenarios
1. Create a simple product with the camel case sku (for example "Test")
2. Use get() method of \Magento\Catalog\Model\ProductRepository to get this product by sku, and specify the sku in lower case.

### Expected result
1. NoSuchEntityException is returned

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ProductRepository.php
+++ b/app/code/Magento/Catalog/Model/ProductRepository.php
@@ -251,6 +251,11 @@ class ProductRepository implements \Magento\Catalog\Api\ProductRepositoryInterfa
         if (!isset($this->instances[$sku])) {
             $sku = trim($sku);
         }
+
+        if (!isset($this->instances[$sku])) {
+            throw new NoSuchEntityException(__('Requested product doesn\'t exist'));
+        }
+
         return $this->instances[$sku][$cacheKey];
     }
 
--- a/app/code/Magento/Catalog/Test/Unit/Model/ProductRepositoryTest.php
+++ b/app/code/Magento/Catalog/Test/Unit/Model/ProductRepositoryTest.php
@@ -527,6 +527,25 @@ class ProductRepositoryTest extends \PHPUnit\Framework\TestCase
         $this->assertEquals($this->productMock, $this->model->get($productSku));
     }
 
+    /**
+     * @expectedException \Magento\Framework\Exception\NoSuchEntityException
+     */
+    public function testGetBySkuWithCamelCaseException()
+    {
+        $productId = 22;
+        $fakeProductSku = 'testproduct';
+        $realProductSku = 'testProduct';
+        $this->productFactoryMock->expects($this->once())->method('create')
+            ->will($this->returnValue($this->productMock));
+        $this->resourceModelMock->expects($this->once())->method('getIdBySku')
+            ->with($fakeProductSku)->willReturn($productId);
+        $this->productMock->expects($this->once())->method('load')->with($productId);
+        $this->productMock->expects($this->atLeastOnce())->method('getId')->willReturn($productId);
+        $this->productMock->expects($this->once())->method('getSku')->willReturn($realProductSku);
+
+        $this->assertEquals($this->productMock, $this->model->get($fakeProductSku));
+    }
+
     public function testSaveExisting()
     {
         $this->storeManagerMock->expects($this->any())->method('getWebsites')->willReturn([1 => 'default']);
--- a/dev/tests/integration/testsuite/Magento/Catalog/Model/ProductRepositoryTest.php
+++ b/dev/tests/integration/testsuite/Magento/Catalog/Model/ProductRepositoryTest.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Catalog\Model;
+
+use Magento\TestFramework\Helper\Bootstrap;
+
+/**
+ * Test for Magento\Catalog\Model\ProductRepository
+ */
+class ProductRepositoryTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @var ProductRepository
+     */
+    protected $model;
+
+    protected function setUp()
+    {
+        $this->model = Bootstrap::getObjectManager()->create(ProductRepository::class);
+    }
+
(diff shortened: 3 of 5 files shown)


app/code/Magento/Catalog/Model/ProductRepository.php
```diff
@@ -251,6 +251,11 @@ public function get($sku, $editMode = false, $storeId = null, $forceReload = fal
         if (!isset($this->instances[$sku])) {
             $sku = trim($sku);
         }
+
+        if (!isset($this->instances[$sku])) {
+            throw new NoSuchEntityException(__('Requested product doesn\'t exist'));
```
orlangur, line 256: This code completely duplicates previous condition.

What is this issue all about? I see a `trim` but don't see any `strtolower` in logic like `$this->instances[$product->getSku()][$cacheKey] = $product;`.

orlangur (changes requested):

app/code/Magento/Catalog/Model/ProductRepository.php
RomaKis: @orlangur, ok, let me explain. We have product with sku "Test". We try to use "get" method of repository to get product with sku "test". Method resourceModel->getIdBySku($sku) will return to us id of "Test" and in $this->instances it put the sku "Test". So after trim we are checking if we have $this->instances["test"] and we don't have it, because we have $this->instances["Test"].

orlangur: How? https://3v4l.org/smbkn

RomaKis: @orlangur, what how? Trimming isn't my logic). It was before my changes. I only check if product sku("Test") is the same as requested sku("test"). And if it isn't the same exception will be thrown.

RomaKis:
@orlangur, any updates on it?

RomaKis:
@orlangur, ping.

app/code/Magento/Catalog/Model/ProductRepository.php
orlangur: Ok, let me reword. Why product is stored with lower-cased SKU in the first place?

We must find a root cause so that exception will be thrown from line 240 instead of adding more exceptions throwing.

orlangur (changes requested):

erikhansen:
For the record, this PR will _not_ fix the issue that it purports to fix (#12073). It will simply result in an exception rather than an error if the condition described in #12073 is met. However it will not allow a SKU with a mixed-case to get loaded via the product repository.

See [this comment](https://github.com/magento/magento2/issues/12073#issuecomment-361402439) for more context.

orlangur:
@erikhansen,
> this PR will not fix the issue that it purports to fix (#12073)

> Expected result
> NoSuchEntityException is returned

Actually, not, but I do agree with comment you referring to that it is better to fix regression introduced.

orlangur (changes requested):
@RomaKis,

> It's a regression in 2.2.0, still present as of 2.2.2. The problem was introduced right here: cb3b5d1#diff-ff4893a5628d34910295bae32fbcddfdL241

Just change implementation of `cacheProduct` in such a way that we pass both `productId` and `sku` instead of obtaining them from product instance.

The test should enforce that we can actually load product by such SKU.

RomaKis:
@orlangur
>Just change implementation of cacheProduct in such a way that we pass both productId and sku instead of obtaining them from product instance.

This sounds obscure, because we have two different functions `get()` and `getById()` which have only `sku` or only `id` as argument. Or do you mean to create two separate methods `cacheBySku` and `cacheById` ? In that way product will be cashed after first load only by `by id` or only `by sku`, but not by both.
Also I think this is not good idea, because we can ask product by sku `test` and `Test`. First it will cache it with the key `test` and the second time it will cache it again with the key `Test`, so we will have two same products in cache with different keys.

orlangur:
@RomaKis,

> so we will have two same products in cache with different keys.

That's the point. It should work this way.

Just rework implementation of `cacheProduct` in https://github.com/magento/magento2/commit/cb3b5d1e0715632c15d8a9697538afca4083a859#diff-ff4893a5628d34910295bae32fbcddfdL241 so that code behavior remain unchanged.

RomaKis:
@orlangur, Why do we need to cache "test" and "Test" if we have only only one product? It is obscure. This is misleading and I don't think this is right. 
I think we have two ways to resolve this issue:
1. make all skus strtolower();
2. throw exception if request sku is wrong.

orlangur:
> It is obscure. This is misleading and I don't think this is right.

Please reword a bit. `strtolower` implementation is a bit fragile and counter-intuitive. Proposed approach does not lead to any negative side effect as there is already mechanism to prevent cache overflow.

RomaKis:
You are right that `strtolower` isn't a good solution.

So, I think it should throw exception. If I have product with sku "Test", so why when I ask product with sku "teSt" it should return the same product? We don't have product with sku "teSt", we have only with sku "Test". This is misleading for me.

orlangur:
@RomaKis because that's how it worked before we introduced caching in product repository.

Another case I didn't mention is that MySQL does not distinguish not just lower/upper case but also е and ё, o and ö, a and ã and so on.

There is no need to introduce any additional restrictions if we can keep the things consistent without them.

RomaKis:
If it worked in such way it didn't mean that it worked right. I remain at my mind - Magento shouldn't cache products with non-existent sku. As I mentioned it is misleading.

You can `SELECT` from SQL case sensitive with `BINARY`.

If you don't accept this implementation - so close this PR )

orlangur:
Such case is not nonexistent SKU in fact and is a behavior explicitly requested in https://github.com/magento/magento2/pull/12169#issuecomment-366882438.

mcspronko:
@orlangur What is the reason for rejecting the pull request? The issue is from Nov 2017 and still not fixed. Is there a plan in place to address this issue or it is up to the community and their availability to fix it?

orlangur:
@mcspronko because a proper implementation approach is performed in https://github.com/magento/magento2/pull/14019. It needs some assistance with test but no time to look into it yet.

Cc: @maximgubar

PR #12170, opened 2017-11-10
9990: Product import doesn't allow empty attribute values for non-static attributes(backport to 2.2)

RomaKis:
Backport from https://github.com/magento/magento2/pull/11935.
1. magento/magento2#9990: Product import doesn't allow empty attribute values for non-static attributes

### Manual testing scenarios
1. Change value to empty for some attribute in csv.
2. Run product import.
3. Check this attribute for value. It should be empty.

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
@@ -533,11 +533,6 @@ abstract class AbstractType
      */
     public function clearEmptyData(array $rowData)
     {
-        foreach ($this->_getProductAttributes($rowData) as $attrCode => $attrParams) {
-            if (!$attrParams['is_static'] && !isset($rowData[$attrCode])) {
-                unset($rowData[$attrCode]);
-            }
-        }
         return $rowData;
     }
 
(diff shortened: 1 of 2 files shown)


app/code/Magento/CatalogImportExport/Model/Import/Product/Type/AbstractType.php
```diff
@@ -533,11 +533,6 @@ public function prepareAttributesWithDefaultValueForSave(array $rowData, $withDe
      */
     public function clearEmptyData(array $rowData)
     {
-        foreach ($this->_getProductAttributes($rowData) as $attrCode => $attrParams) {
```
dmanners, line 536: @RomaKis I would consider that we could now deprecate this method now if it is not doing anything and also remove the usage of this method from any place we can find.

dmanners (changes requested):

dmanners:
@RomaKis I have migrated this pull request over to the import/export improvement (https://github.com/magento-engcom/import-export-improvements/pull/73) project repository. Feel free to join us in this project and we can take this task on. I would like to talk with some community members and the core architectural team about the implementation and also documentation of the new constant.

Thank you for this code.

PR #9367, opened 2017-04-23
[PSR-2 Compliance] The Great @codingStandardsIgnoreFile Massacre

orlangur:
This PR fixes mistake introduced in https://github.com/magento/magento2/commit/004c9bb917b250ae2d6515ec3b96d4099b6f9c2e which resulted in hundreds of files with coding style violations.

### Description

As pointed out in https://github.com/magento/magento2/pull/9251#issuecomment-294478242 manual fixing would take too much time and developer/reviewer efforts. Thus I did exactly what was described in https://github.com/magento/magento2/pull/7166#issuecomment-287712529 to get maximum profit out of existing automated tools.

### Review Notes
Manual commits: https://github.com/orlangur/magento2/commit/a1f9e1adbafc54c981a5d1e18bf68d87db9a1a8c, https://github.com/magento/magento2/commit/40ba673f56e2f45cd05564a0e01b127a9ad4476a, https://github.com/magento/magento2/commit/c87c521bb06696b35f789d595b3e8a4f452e29bc
Automated commits (just repeat steps from commit message instead of manual review): https://github.com/orlangur/magento2/commit/b51681c5db4b639c6996fc1c22fdc642391165ad, https://github.com/magento/magento2/commit/4dc56e0b1d3116821a5e81545f41686c59277be1, https://github.com/magento/magento2/commit/c6194044c158abf8747ea3c5e9699439cbae8278, https://github.com/magento/magento2/commit/d8139743b4888a9d95c3e7a0951f09fe392adf8e
PHPMD on Travis CI was not able to complete within 10 minutes limit thus I checked static tests in two chunks (modules only and all but modules in whitelist): https://travis-ci.org/orlangur/magento2/builds/224861865, https://travis-ci.org/orlangur/magento2/builds/224861991

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

PR #12192, opened 2017-11-11
CS: reapply fixers

keradus:
Hi !
I wanted to do sth with repo, and then executed boundled PHP CS Fixer config. Yet, it crashes on a lot of violation. So let me extract the CS changes to initial PR from my side containing only them.

Probably it would be good to add PHP CS Fixer integration into Travis later (I could help with it if it would be wanted).

This PR is only automated result of "php-cs-fixer fix .` using config file


keradus:
`static` job is failing as it's too slow for Travis this time

orlangur:
Hi @keradus, glad to see you again!

Back to the https://github.com/magento/magento2/pull/10770#issuecomment-332445629, are you going to contribute a mechanism enforcing these rules? Without such mechanism applying coding style fixes does not make a lot of sense. I don't think wrapping it as PHPUnit-based test is the only option, so, if you see a better approach which can work fine with Travis+Bamboo+Jenkins, please share some thoughts.

keradus:
Hi ;) Thank you for warm welcome!

I would suggest to integrate it with one (no need for multiple) CI service that is there. I could send a PR to implement it into Travis. Totally outside of phpunit.
So, if one would run it locally, first run is slow, but it builds cache, next run is fast. On Travis, each run would touch only modified files.
If interoperability between different CI is must have, we could wrap logic of gathering files to check and running the command into few-liner shell script. 
If you wish, I could prepare a dedicated PR.

But for that, let us first fix current violations.

orlangur:
Unassigning from myself for now, I'm quite interested in such cleanup and enforcing it with test (probably list of rules should be revised before apply).

Just that I need to round out a bunch of other PRs. Will raise a maintainers discussion on appropriate implementation approach after this if nobody do it before.

keradus:
OK, thanks for info.
So now, please guide me, is there anything for now I could help with?

orlangur:
I think all is good from your side, approach upon merging such changes should be agreed first, I did an unsuccessful attempt of reapply in scope of https://github.com/magento/magento2/pull/9367#issuecomment-307506978 earlier.

ishakhsuvarov:
@keradus Unfortunately we can not accept such big amount of changes in a single PR as it requires manual review of 4300+ files, which is nearly impossible.
Additionally, this PR requires all queue to be stopped, as merge conflicts will arise nearly for every other PR merged.
I am closing this for now, until we discuss the flow to merge this kind of PR's. We will reopen and let you know if we manage to work out a flow for this.

PR #12220, opened 2017-11-13
12180 Remove unnecessary use operator for Context, causes 503 error i…

chris-pook:
…n account address book.

Removing an unnecessary "use" to prevent a 503 error in account address book page.

### Description
Remove the unnecessary "use".

1. magento/magento2#12180: M2.2.1 Unable to open Address book after account creation

### Manual testing scenarios

- Set up a Magento v2.2.1 environment (prod mode + composer dump-autoload -o)
- Create an account as a customer.
- Immediately after creating your customer, try to open your address book.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Model/AttributeChecker.php
+++ b/app/code/Magento/Customer/Model/AttributeChecker.php
@@ -8,7 +8,6 @@ namespace Magento\Customer\Model;
 use Magento\Customer\Api\AddressMetadataInterface;
 use Magento\Customer\Api\AddressMetadataManagementInterface;
 use Magento\Customer\Model\Metadata\AttributeResolver;
-use Magento\Framework\App\Helper\Context;
 
 /**
  * Customer attribute checker.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12220) <br/>All committers have signed the CLA.

orlangur:
@chris-pook could you please sign CLA?

chris-pook:
CLA signed. Thanks

app/code/Magento/Customer/Model/AttributeChecker.php
```diff
@@ -8,7 +8,6 @@
 use Magento\Customer\Api\AddressMetadataInterface;
 use Magento\Customer\Api\AddressMetadataManagementInterface;
 use Magento\Customer\Model\Metadata\AttributeResolver;
-use Magento\Framework\App\Helper\Context;
```
orlangur, line 11: Hi @chris-pook, please check https://github.com/magento/magento2/commit/a872f4061695b91e923d1c1abb87a0a50e18c77d#diff-fe978939bb978f6074e10338e0b14791

I believe the fix is not correct, helper context is supposed to be injected, not customer model context.

At least parent class expects `Magento\Framework\App\Helper\Context`.

Could you please replace import with `Magento\Framework\App\Helper\Context as HelperContext` instead and change other places in class accordingly?

orlangur (changes requested):

orlangur:
Hello @chris-pook, please check my comment https://github.com/magento/magento2/pull/12220#discussion_r152822053.

app/code/Magento/Customer/Model/AttributeChecker.php
chris-pook: Sorry @orlangur , you've lost me.. `Customer\Model\AttributeChecker` (where my change is) has no parent class. This change I have made is in `Customer\Model\AttributeChecker` where there is a `use` statement for `Magento\Framework\App\Helper\Context` which is never used in the class. I could alias this to `HelperContext` but it is not used so better to remove I think? Unless I'm missing something here?

orlangur: @chris-pook, my bad, I was opening `AttributeChecker.php` in my IDE and didn't notice I'm checking helper and not model with the same name.

Looks good to me now after your explanation :)

orlangur (approved):

app/code/Magento/Customer/Model/AttributeChecker.php
chris-pook: Ah good stuff! Thanks

PR #12237, opened 2017-11-14
Correctly provide list of swatch attributes for product #11703

therool:
### Description
Correctly provide list of swatch attributes for product. This is an fix for #11703

The `populateAdditionalDataEavAttribute`, `isSwatchAttribute`, `isVisualSwatch`, `isTextSwatch` method definitions moved from `Magento\Swatches\Helper\Data` class to `Magento\Swatches\Model\SwatchAttributesProvider` class but methods are still callable from `Magento\Swatches\Helper\Data` which just calls the `Magento\Swatches\Model\SwatchAttributesProvider` methods instead so the backward compatibility is preserved. Method definitions are moved so they can be used in the `Magento\Swatches\Model\SwatchAttributesProvider::provide` method to correctly list of swatch attributes for product.

1. magento/magento2#11703: Changing Swatches to Drop-down does not remove swatches from existing products

### Manual testing scenarios
1. Create swatch text attribute
1. Create configurable product using this attribute
1. Open product on frontend - attribute is being outputted as swatch attribute and you can select all the product configurations which is ok
1. Change the swatch attribute back to dropdown
1. Open product on the frontend - attribute is being still outputted as swatch attribute which is wrong but you still can select the product configurations
1. Add additional option for the dropdown attribute and add additional configuration for this option to the configurable product
1. Open the product on the frontend - the attribute is still being still outputted as swatch attribute and the newly added configuration can't be selected on the frontend because the added attribute does not contain necessary data for it to be rendered as swatch option as the attribute should be rendered as dropdown option instead of swatch.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Swatches/Helper/Attribute.php
+++ b/app/code/Magento/Swatches/Helper/Attribute.php
@@ -7,6 +7,7 @@ namespace Magento\Swatches\Helper;
 
 use Magento\Framework\Serialize\SerializerInterface;
 use Magento\Eav\Model\Entity\Attribute\AbstractAttribute;
+use Magento\Swatches\Model\Swatch;
 
 /**
  * Provide list of swatch attributes for product.
@@ -35,7 +36,7 @@ class Attribute
     public function getEavAttributeAdditionalDataKeys()
     {
         return [
-            \Magento\Swatches\Model\Swatch::SWATCH_INPUT_TYPE_KEY,
+            Swatch::SWATCH_INPUT_TYPE_KEY,
             'update_product_preview_image',
             'use_product_image_for_swatch'
         ];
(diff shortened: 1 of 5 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12237) <br/>All committers have signed the CLA.

omiroshnichenko:
Hi, @therool. The main problem that after changing type of attribute from swatch to dropdown, swatch options stay in `eav_attribute_option_swatch`. In result of this `\Magento\Swatches\Model\SwatchAttributeCodes::getSwatchAttributeCodes` method returns swatch options. Correct fix will be to delete options from `eav_attribute_option_swatch` when we converting attribute. Will you continue working on this PR?

therool:
@omiroshnichenko yes the problem is with `\Magento\Swatches\Model\SwatchAttributeCodes::getSwatchAttributeCodes` but then if the options are deleted from `eav_attribute_option_swatch` and in case if the attribute is being changed back to swatch then all the previously entered swatch options data is lost but it might be good to preserve them.
So if the current changes are not good then it might be better to just add a new `is_swatch` column to `catalog_eav_attribute` so the `\Magento\Swatches\Model\SwatchAttributeCodes::getSwatchAttributeCodes` can use as simple sql as: 

```sql
SELECT a.attribute_code
FROM catalog_eav_attribute c
JOIN eav_attribute a ON a.attribute_id = c.attribute_id
WHERE c.is_swatch = 1
```

omiroshnichenko:
@therool In case when you will change back, you will need to enter new data for swatches. I don't see any reason to keep swatch options data if it was converted to dropdown.
Your fix just masks problem, unused swatch options must be removed in convert process.

omiroshnichenko:
I will close this PR, because https://github.com/magento/magento2/pull/12771 provides correct fix. @therool Thank you for your contribution.

PR #12276, opened 2017-11-15
Webshop throws an exception when sharing wishlist with RSS enabled

mediactbv:
### Description
In this pull request a bug that was discovered while the "Share Wishlist" functionality was executed. When a logged in user adds products to the wishlist and then tries to share them using RSS, an exception is thrown: `report.INFO: Broken reference: the 'wishlist.email.rss' element cannot be added as child to 'root', because the latter doesn't exist []`

### Manual testing scenarios
0. Enable the RSS feed from wishlist in the backend
1. Open the front-end of a magento 2 installation.
2. Log in a user on the frontend
3. Add one or more products to the wishlist
4. Go to My Account -> My wishlist
5. Click on the "Share wishlist" button
6. Ensure the RSS feed checkbox is checked
7. Click on Share wishlist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12276) <br/>All committers have signed the CLA.

ishakhsuvarov:
Hi @mediactbv 
Please sign the CLA with the email you used for the commit, so that we can proceed with the review and merge. 
Thank you

Issue #12278, opened 2017-11-15
Error while fetching bundle item with repository


hatimeria-artur-jewula:
### Preconditions
1. Magento 2.1.7 
2. PHP 7.0.22

### Steps to reproduce
1. Create a bundle product with at least 1 option of simple product
2. Create a module which defines any extension attributes for `Magento\Catalog\Api\Data\ProductInterface`
3. Create an observer for `catalog_product_collection_load_after` event and in that observer set extension attribute defined in step 2 on each item
4. Try to retrieve bundle product using `Magento\Catalog\Api\ProductRepositoryInterface::getById()` method providing only id of the bundle product created in step 1. 

### Expected result
1. Receive an instance of `Magento\Catalog\Api\Data\ProductInterface` with data of bundle product

### Actual result
1. An error is thrown:
Argument 1 passed to Magento\Bundle\Model\Link::setExtensionAttributes() must be an instance of Magento\Bundle\Api\Data\LinkExtensionInterface, instance of Magento\Catalog\Api\Data\ProductExtension given, called in vendor/magento/framework/Api/DataObjectHelper.php on line 125

After inspection I found that bundle module defines its own extension attribute `bundle_product_options` for `Magento\Catalog\Api\Data\ProductInterface`. When it tries to set it for loaded product it fetches collection of products defined as children in the bundle and converts those products to array of `Magento\Bundle\Api\Data\OptionInterface` classes. The problem is those products have `extension_attributes` set in `$_data` and the class there is `Magento\Catalog\Api\Data\ProductExtension` as it should be for product class.

magento-engcom-team:
@hatimeria-artur-jewula, thank you for your report.
We've created internal ticket(s) MAGETWO-84142 to track progress on the issue.

AlexandreKhayrullin:
Hello.

We are currently having the same issue with Magento 2.3.0.
I hope you are aware that problems of this kind compromise the usability of service contracts, leading to the rejection of this practice and a return to using Models, ResourceModels and Collections.

sheepfy:
Hei, are there any updates on this? 
I saw that PR that fix the issue, but the PR is closed and is not merged since 2017.

lfolco:
Any chance someone can take a look at this???

Aditya-Prakash-Talluru:
@magento I am working on this

simonrl:
Hi, does anybody have an ETA on this or a fix?

sheepfy:
@simonrl 
[...]

PR #12279, opened 2017-11-15
#12278 remove existing extension attributes during extraction of exte…

Zefiryn:
…nsion attribues when existing class does not extend correct interface

When extracting extension attribute from data of one class make sure the result will not have extension attributes defined for that class if the two entities do not share same extension class. Otherwise it may result in error when trying copy one class data to another.

### Description
Changes are designed to handle the case when input `$data` contains `extension_attributes` field but the class for which extension attributes need to be extracted does not have any extension attributes defined. In such case it is possible that extension attribute from source object will cause type error when passed to constructor of the `$extensibleEntityClass`

1. magento/magento2#12278: Error while fetching bundle item with repository

### Manual testing scenarios
1. Create a bundle product with at least 1 option of simple product
2. Create a module which defines any extension attributes for `Magento\Catalog\Api\Data\ProductInterface`
3. Create an observer for `catalog_product_collection_load_after` event and in that observer set extension attribute defined in step 2 on each item
4. Try to retrieve bundle product using `Magento\Catalog\Api\ProductRepositoryInterface::getById()` method providing only id of the bundle product created in step 1. 
5. No error should be raised. 

### Contribution checklist

### Ports to other versions
* `2.1-develop` https://github.com/magento/magento2/pull/12280
* `2.3-develop` https://github.com/magento/magento2/pull/12281


dmanners:
Hello @Zefiryn, sorry for the delay in getting to this pull request. There are a few failures in the travis test suite. Could you please have a look into these before I process this pull request.

Thanks.

Zefiryn:
@dmanners sure, I will take a look at it during the weekend

Zefiryn:
@dmanners I have checked the errors and decided to redo the approach and to only fix the problem with loading bundle product originally posted in the issue. I noticed that several tests uses regular arrays when setting `extension_attributes` data in the object. I am not sure why this has been done this way as `extension_attributes` is always an object. Anyway updated PRs for branches 2.1-develop, 2.2-develop and 2.3-develop are ready.

app/code/Magento/Bundle/Model/Product/LinksList.php
```diff
@@ -70,4 +77,20 @@ public function getItems(\Magento\Catalog\Api\Data\ProductInterface $product, $o
...
+     * Get selection product data and remove extension attributes if necessary
+     *
+     * @param ProductInterface $selection
+     * @return array
+     */
+    private function getSelectionData(ProductInterface $selection)
+    {
+        $selectionData = $selection->getData();
+        if (array_key_exists(ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY, $selectionData)) {
+            unset($selectionData[ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY]);
```
orlangur, line 91: Please replace with unconditional `unset`.

lib/internal/Magento/Framework/Api/ExtensionAttribute/JoinProcessor.php
```diff
@@ -100,7 +101,10 @@ private function getReferenceTableAlias($attributeCode)
     public function extractExtensionAttributes($extensibleEntityClass, array $data)
     {
         if (!$this->isExtensibleAttributesImplemented($extensibleEntityClass)) {
-            /* do nothing as there are no extension attributes */
+            /* remove any extension attributes from source data as destination entity has no extension attributes */
+            if (array_key_exists(ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY, $data)) {
+                unset($data[ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY]);
```
orlangur, line 106: Please replace with unconditional `unset`.

app/code/Magento/Bundle/Model/Product/LinksList.php
```diff
@@ -70,4 +77,20 @@ public function getItems(\Magento\Catalog\Api\Data\ProductInterface $product, $o
...
         return $productLinks;
     }
+
+    /**
+     * Get selection product data and remove extension attributes if necessary
+     *
+     * @param ProductInterface $selection
+     * @return array
+     */
+    private function getSelectionData(ProductInterface $selection)
```
orlangur, line 87: Please rename to `getSelectionDataWithoutExtensionAttributes`.

orlangur (changes requested):

ihor-sviziev:
Hi @Zefiryn,
I'm closing this Pull Request due to inactivity. Feel free to fix all requested changes and reopen your PR or create new one.

sheepfy:
@ihor-sviziev 
Hei, is this up-to-date? I mean i currently have the same issue with the extension attributes.
You just cannot use both extension_attributes and bundle products at the same time. 

This PR seems to fix this issue. Is it OK if i continue with this PR ?

PR #12280, opened 2017-11-15
[Backport] #12278 remove existing extension attributes during extraction (version 2.1)

Zefiryn:
…nsion attribues when existing class does not extend correct interface

When extracting extension attribute from data of one class make sure the result will not have extension attributes defined for that class if the two entities do not share same extension class. Otherwise it may result in error when trying copy one class data to another.

### Description
Changes are designed to handle the case when input `$data` contains `extension_attributes` field but the class for which extension attributes need to be extracted does not have any extension attributes defined. In such case it is possible that extension attribute from source object will cause type error when passed to constructor of the `$extensibleEntityClass`

1. magento/magento2#12278: Error while fetching bundle item with repository

### Manual testing scenarios
1. Create a bundle product with at least 1 option of simple product
2. Create a module which defines any extension attributes for `Magento\Catalog\Api\Data\ProductInterface`
3. Create an observer for `catalog_product_collection_load_after` event and in that observer set extension attribute defined in step 2 on each item
4. Try to retrieve bundle product using `Magento\Catalog\Api\ProductRepositoryInterface::getById()` method providing only id of the bundle product created in step 1. 
5. No error should be raised. 

### Contribution checklist

### Ports to other versions
* `2.2-develop` https://github.com/magento/magento2/pull/12279
* `2.3-develop` https://github.com/magento/magento2/pull/12281


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12280) <br/>All committers have signed the CLA.

ihor-sviziev:
Original Pull Request was closed due to inactivity. As this is port of original PR - i'm closing this PR too.

PR #12328, opened 2017-11-17
9742: Default welcome message returns after being deleted #9742

RomaKis:
Default welcome message returns after being deleted #9742

### Description
Column value in core_config_data was nullable. When we saved '' as column was nullable in lib/internal/Magento/Framework/DB/Adapter/Pdo/Mysql.php line 3001 the value became null, so it was null in db. If value is null it will take the default value, so we couldn't save '' and see '' at the form.
Now column value isn't nullable, so when we save '' it will save in db as '' and show at the form as ''.

1. magento/magento2#9742:Default welcome message returns after being deleted

### Manual testing scenarios
1. Go to Content -> Configuration -> Click Edit on theme being used.
2. Open header tab -> clear Welcome Text.
3. Click save and continue.
4. Open header tab -> Welcome Text field should be empty.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Theme/Model/Design/Config/Storage.php
+++ b/app/code/Magento/Theme/Model/Design/Config/Storage.php
@@ -87,10 +87,13 @@ class Storage
                 $scopeId,
                 $fieldData->getFieldConfig()
             );
-            if ($value !== null) {
-                $fieldData->setValue($value);
+
+            if ($value === null) {
+                $value = '';
             }
+            $fieldData->setValue($value);
         }
+
         return $designConfig;
     }
 
--- a/dev/tests/integration/testsuite/Magento/Theme/Model/Design/ConfigTest.php
+++ b/dev/tests/integration/testsuite/Magento/Theme/Model/Design/ConfigTest.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Theme\Model\Design;
+
+/**
+ * Test for \Magento\Theme\Model\Design\Config\Storage.
+ */
+class ConfigTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @var \Magento\Theme\Model\Design\Config\Storage
+     */
+    private $storage;
+
+    protected function setUp()
+    {
+        $this->storage = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()->create(
+            \Magento\Theme\Model\Design\Config\Storage::class
+        );
+    }
+
+    /**
+     * Test design/header/welcome if it is saved in db as empty(null) it should be shown on backend as empty.
+     *
+     * @magentoDataFixture Magento/Theme/_files/config_data.php
+     */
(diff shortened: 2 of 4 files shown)


sidolov:
Hi @RomaKis, there is no bug with configuration persistence, you can save empty welcome message, field is empty in db (screenshot added).
17607381ee

Issue caused by incorrect form data population (fallback to default value) in \Magento\Theme\Model\Design\Config\DataProvider\MetadataLoader::getData() method. 
The same behavior with Copyright field. 

On frontend both fields are empty
94a239e4a4

RomaKis:
Hi @sidolov, the bug is that they are not empty in backend. When you tried to delete it and save. It will reappear after save. 
You are right that it is getting data from \Magento\Theme\Model\Design\Config\DataProvider\MetadataLoader::getData(), but the problem is deeper. 
The problem is that if we save (for example) Footer -> Copyright  = '' (as empty line) it will be converted as null and saved to db. As Magento see that value is null, so it show the default copyright from app/code/Magento/Theme/etc/config.xml. So it is the same: if Magento has clean installation - it is null or we save it as '' it is null too.

sidolov:
@RomaKis after clean Magento installation records about welcome message and copyrights are absent on db. They will be created if saved values were different then default. It's expected result that empty values stored in db as null, doesn't make sense store empty strings, we don't consider this to be a good design. Described problem exists only on backend form, frontend works as expected, so, issue should be fixed in form data population flow (for example system configuration show empty values correctly).

RomaKis:
@sidolov, ok, I agree with you. I change code. Please, review it again.

app/code/Magento/Theme/Model/Design/Config/Storage.php
```diff
@@ -87,10 +87,14 @@ public function load($scope, $scopeId)
                 $scopeId,
                 $fieldData->getFieldConfig()
             );
-            if ($value !== null) {
-                $fieldData->setValue($value);
+
+            if ($value == null) {
```
sidolov: Use strict comparison

sidolov (changes requested):

app/code/Magento/Theme/Model/Design/Config/Storage.php
RomaKis: @sidolov, done.

app/code/Magento/Theme/Model/Design/Config/Storage.php
```diff
@@ -87,10 +87,14 @@ public function load($scope, $scopeId)
                 $scopeId,
                 $fieldData->getFieldConfig()
             );
-            if ($value !== null) {
-                $fieldData->setValue($value);
+
+            if ($value === null) {
+                $value = '';
+
```
sidolov: Remove blank line at end of control structure

sidolov (changes requested):

app/code/Magento/Theme/Model/Design/Config/Storage.php
RomaKis: @sidolov, done and squashed.

sidolov (approved):

PR #12379, opened 2017-11-21
Dissociate data and schema version

j-willette:
### Description
Adding ability to use different version for data and schema parts (see…setup_module table).
In current version, Magento use setup_version for these two different versions (data and schema) in setup_module table.

### Manual testing scenarios
Create a module with following module.xml file :

```<?xml version="1.0"?>
    <module name="Synolia_Example" data_version="1.0.0" schema_version="1.0.5">
    </module>
</config>
```

Obviously, this code still work:
```<?xml version="1.0"?>
    <module name="Synolia_Example" setup_version="1.0.0">
    </module>
</config>
```
Maybe it's unnecessary but if so, we don't undestand the purpose of two colums in setup_module table (maybe Magento 1 residue ? ;-)).

Thanx in advance.

Josselin

### DIFF
--- a/lib/internal/Magento/Framework/Module/Declaration/Converter/Dom.php
+++ b/lib/internal/Magento/Framework/Module/Declaration/Converter/Dom.php
@@ -32,6 +32,19 @@ class Dom implements \Magento\Framework\Config\ConverterInterface
                 throw new \Exception("Attribute 'setup_version' is missing for module '{$name}'.");
             }
             $moduleData['setup_version'] = $versionNode->nodeValue;
+
+            $schemaVersionNode = $moduleAttributes->getNamedItem('schema_version');
+
+            if ($schemaVersionNode !== null) {
+                $moduleData['schema_version'] = $schemaVersionNode->nodeValue;
+            }
+
+            $dataVersionNode = $moduleAttributes->getNamedItem('data_version');
+
+            if ($dataVersionNode !== null) {
+                $moduleData['data_version'] = $dataVersionNode->nodeValue;
+            }
+
             $moduleData['sequence'] = [];
             /** @var $childNode \DOMNode */
             foreach ($moduleNode->childNodes as $childNode) {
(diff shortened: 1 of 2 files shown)


orlangur:
`<module name="Synolia_Example" data_version="1.0.0" schema_version="1.0.5">` how this can be useful? Data and schema version should always be in sync.

j-willette:
> Data and schema version should always be in sync

This is precisely the point. Why there are two columns in setup_module table (data_version and schema_version) if the datas are always in sync?
We are purposing to make evolve them separatly, for example, when you make an upgradeData like CMS page or any config which not impact the schema, to really use these different columns.

orlangur:
Because they are intended for different things. I didn't look into implementation and whether it is feasible to combine such two columns into one.

> when you make an upgradeData like CMS page or any config which not impact the schema, to really use these different columns.

It does not matter, data upgrade must use concrete schema version.

> We are purposing to make evolve them separatly

Again, this is not how Magento is supposed to work. They are supposed to be always in sync.

orlangur:
@j-willette could you please clarify what would be the value of "We are purposing to make evolve them separatly"?

syno-dlr:
Hello @orlangur,

Infact, in Magento 1, the two versions (Data and Schema) could be different (for example if the countries table data change, the schema is not impacted).
This Pull request reproduce this system.

If Data should be equal to Schema, what is the purpose of "data_version" column in setup_module table ?

Damien.

orlangur:
@damien-synolia, actually `data_version` is not always present: https://github.com/magento/magento2/blob/2.2-develop/lib/internal/Magento/Framework/Module/ModuleResource.php#L65

@j-willette I've learned something new from a recent Mageconf, in 2.3 schema/data upgrades are significantly reworked and only a module version from `composer.json` will be effective. So, there is no need to evolve functionality related to schema_version/data_version. Thanks for collaboration!

PR #12383, opened 2017-11-21
Update CAPTCHA labels to reflect the symbols in the CAPTCHA image

RhodriOwainDavies:
### Description
The text labels referring to the CAPTCHA image only refer to letters, when there are also numbers in the CAPTCHA images.  Meaning the instruction to the user is ambiguous.  The message implies to NOT enter the numbers, thus the CAPTCHA doesn't match.

I've changed the labels in a few places and updated the dictionary file

### Manual testing scenarios
1. As a customer visit the sign-in page and enter the wrong password for the same email address several times until the captcha image & label appears
2. As a merchant visit the admin login page and enter the wrong password for the same user account several times until the captcha image & label appears
3. Install a language pack which includes a dictionary key for the new phrases, change the language of the store and repeat steps 1 & 2 ensuring the phrases appear in their translated forms.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Captcha/composer.json
+++ b/app/code/Magento/Captcha/composer.json
@@ -13,7 +13,7 @@
         "zendframework/zend-session": "^2.7.3"
     },
     "type": "magento2-module",
-    "version": "100.2.0",
+    "version": "100.2.1",
     "license": [
         "OSL-3.0",
         "AFL-3.0"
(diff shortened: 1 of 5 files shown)


app/code/Magento/Captcha/composer.json
```diff
@@ -13,7 +13,7 @@
         "zendframework/zend-session": "^2.7.3"
     },
     "type": "magento2-module",
-    "version": "100.2.0",
+    "version": "100.2.1",
```
okorshenko, line 16: please, revert this fix. We will update the package versions on our side when needed

PR #12407, opened 2017-11-22
Added missing dependencies of the Magento_Config module.

hostep:
### Description
This was discovered in scope of https://github.com/magento/magento2/issues/12405 (which I closed because the initial post didn't contain the correct steps to reproduce, the correct ones are listed below).

I took the liberty of including more then only the `Magento_Store` module in the dependencies, because the `Magento_Config` module uses 8 other modules, so I think they should all be defined. These are the same dependencies as defined in the [`composer.json` file](https://github.com/magento/magento2/blob/2.2.1/app/code/Magento/Config/composer.json#L7-L13), but the `composer.json` file doesn't list `Magento_User`. This is I think correct, since `Magento_User` is only used in the tests. So it shouldn't be a hard dependency in the `composer.json` file, but still needs to be included in the `module.xml` file I think?

1. https://github.com/magento/magento2/issues/12405: Magento 2.2.1 - Impossible to create a new storeview

### Manual testing scenarios
1. Setup a clean Magento 2.2.1
2. Install https://github.com/baldwin-agency/magento2-module-less-js-compiler
3. After running `setup:upgrade` check your `app/etc/config.php` file, it will have `Magento_Config` as a higher priority to load then `Magento_Store`
4. Login into the backend
5. Go to Stores => All Stores
6. Click 'Create Store View'
7. Enter some dummy content in the fields and click 'Save Store View'
8. An error `Requested store is not found` is shown and no storeview is added.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Config/etc/module.xml
+++ b/app/code/Magento/Config/etc/module.xml
@@ -6,5 +6,16 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Magento_Config" setup_version="2.0.0"/>
+    <module name="Magento_Config" setup_version="2.0.0">
+        <sequence>
+            <module name="Magento_Backend"/>
+            <module name="Magento_Cron"/>
+            <module name="Magento_Deploy"/>
+            <module name="Magento_Directory"/>
+            <module name="Magento_Email"/>
+            <module name="Magento_MediaStorage"/>
+            <module name="Magento_Store"/>
+            <module name="Magento_User"/>
+        </sequence>
+    </module>
 </config>


orlangur:
> because the Magento_Config module uses 8 other modules, so I think they should all be defined. These are the same dependencies as defined in the composer.json file, but the composer.json file doesn't list Magento_User

Are all those modules actually essential from data integrity perspective?

So, in sequence we need to specify only those modules which define entities in database which are used in `Magento_Config`.

hostep:
Hmm, I think this order also determines other things, quote from the [documentation](http://devdocs.magento.com/guides/v2.2/extension-dev-guide/build/module-load-order.html):
> `<sequence>` declares the list of components that must be loaded before the current component is loaded. It’s used for loading different kind of files: configuration files, view files (including CSS, LESS, and template files), or setup classes.

Any idea how I can determine what modules should be included in the `<sequence>` and which ones not?

orlangur:
@hostep yes, it determines order of all these things, just that in most cases it is needed for DB stuff.

Only those modules for which order is proven to be important should be included. I don't know an easy way, just check the occurrences of each module in Magento_Config should be analyzed and then understood whether the things will break in case of opposite module loading order.

hostep:
I'm closing this, no longer convinced this is the correct solution. I reopened https://github.com/magento/magento2/issues/12405 with the proper steps of how to reproduce the bug and added a new comment.

PR #12441, opened 2017-11-25
Add command "app:config:status" to check if "app:config:import" needed

jalogut:
### Description
Add command `app:config:status` to check whether config propagation is up to date. This command can be used to decide whether `app:config:import` execution is needed during deployments.

### Fixed Issues
On version 2.2 a new command `setup:db:status` was added to decide whether executing `setup:upgrade` is required. The purpose of this command was to add the possibility of deploying without executing `setup:upgrade` if module versions are up to date.

With the same objective, a command to check the config status is needed to decide whether `app:config:import` is needed.

**Issue**
If `setup:upgrade` is not executed, we need to always execute `app:config:import` because there is no way to know, if config files are up to date or not. This will mean, we still need to set the maintenance and we cannot simply switch the code to the new version.

**Solution**
Using `app:config:status`, we would be able to switch to a new version without the need of a maintenance, if `setup:db:status` also returns that everything is up to date 

### Manual testing scenarios
Thanks to this command, we would be able to deploy without maintenance as follows:

1.- Check `setup:db:status` returns `All modules up to date`
2.- Check `app:config:status` returns `Configuration files are up to date.`
3.- No maintenance is required because we do not need to execute `setup:upgrade` or `app:config:import` 

### Contribution checklist


jalogut:
This PR adds improvements in same area as:

* #12410
* #12361
* #12178

hostep:
Quick heads-up: This was supposed to be added to Magento 2.2.4, and it is, but unfortunately the line which was added to the `di.xml` file has been (accidentally?) removed by this commit: https://github.com/magento/magento2/commit/e6b6c652edf7810cf1e3a69a5b2294a31bba78e8#r28828688

Can someone try to figure out how this happened and if something needs to happen around this? In Magento 2.2.4 the line is gone, in the 2.2-develop branch it still exists (at the moment of writing) ...

PR #12495, opened 2017-11-30
Fixed invalid parameter type in phpdoc block in Topmenu class

vovayatsyuk:
### Description
`addIdentity` method accepts strings, not an array.

See the usage: 

1. https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Plugin/Block/Topmenu.php#L119
2. https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Plugin/Block/Topmenu.php#L129

### Contribution checklist

### DIFF
--- a/app/code/Magento/Theme/Block/Html/Topmenu.php
+++ b/app/code/Magento/Theme/Block/Html/Topmenu.php
@@ -331,7 +331,7 @@ class Topmenu extends Template implements IdentityInterface
     /**
      * Add identity
      *
-     * @param array $identity
+     * @param string $identity
      * @return void
      */
     public function addIdentity($identity)
--- a/app/code/Magento/Theme/Block/Html/Topmenu.php
+++ b/app/code/Magento/Theme/Block/Html/Topmenu.php
@@ -331,7 +331,7 @@ class Topmenu extends Template implements IdentityInterface
     /**
      * Add identity
      *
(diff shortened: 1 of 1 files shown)


app/code/Magento/Theme/Block/Html/Topmenu.php
```diff
@@ -331,7 +331,7 @@ protected function _getMenuItemClasses(\Magento\Framework\Data\Tree\Node $item)
     /**
      * Add identity
      *
-     * @param array $identity
+     * @param string $identity
```
dmanners: technically this method accepts both strings and arrays as `in_array` will work with both and we never actually check the type. Since this class is noted with the `@api` I think it would be best to update the typehint to be `string|array` as we do not know how people use it.

dmanners (changes requested):

app/code/Magento/Theme/Block/Html/Topmenu.php
vovayatsyuk: Thanks, never know that I can use `array` as identity.

dmanners: I mean technically you can, not sure you should do though.

dmanners (approved):

PR #12498, opened 2017-11-30
Prevent empty label when element has no label set

sanderjongsma:
### Description
Knockout checks if element has a label. When it has not a label set, the emtpy <label></label> tags are still generated in the dom. This is unnecessary, so the knockout if statement can be placed around the label itself.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Ui/view/frontend/web/templates/form/field.html
+++ b/app/code/Magento/Ui/view/frontend/web/templates/form/field.html
@@ -6,9 +6,9 @@
 -->
 <div class="field" data-bind="visible: visible, attr: {'name': element.dataScope}, css: additionalClasses">
 
-    <!-- ko if: element.label -->
-        <label class="label" data-bind="attr: { for: element.uid }"><span translate="element.label"></span></label>
-    <!-- /ko -->
+    <label if="element.label" class="label" data-bind="attr: { for: element.uid }">
+        <span translate="element.label"></span>
+    </label>
 
     <div class="control" data-bind="css: {'_with-tooltip': element.tooltip}">
         <!-- ko ifnot: element.hasAddons() -->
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12498) <br/>All committers have signed the CLA.

fooman:
Thanks @sanderjongsma for this contribution. Do you think that some people might rely on the presence of the label attribute, even if it's empty, to get their layout right?

app/code/Magento/Ui/view/frontend/web/templates/form/field.html
```diff
@@ -6,11 +6,9 @@
 -->
 <div class="field" data-bind="visible: visible, attr: {'name': element.dataScope}, css: additionalClasses">
 
-    <label class="label" data-bind="attr: { for: element.uid }">
-        <!-- ko if: element.label -->
-        <span translate="element.label"></span>
-        <!-- /ko -->
-    </label>
+    <!-- ko if: element.label -->
```
sidolov: Replace old style logic operator to new. 
Example:  
```
<label if="element.label" class="label" data-bind="attr: { for: element.uid }">
    ...
</label>
```

sidolov (changes requested):

sidolov (approved):

sidolov:
Hi @sanderjongsma, I fixed minor issue, thank you for contribution!

sanderjongsma:
@fooman yes I think there are a few people relying on it (if they float the label left from the input, and don't use float right for the input itself). In my opinion it is always better to fix the styling without a label attribute. The reason for this pr is because now the streetaddress field(s) renders an extra label tag, which can cause unwanted html. I know the default checkout isn't affected, because labels are above the inputs. I know it's about preferences, but let me know what you think.

sidolov:
Hi @sanderjongsma , after the long discussion, according to web accessibility basics, we decide that we always need label tag for all inputs (including cases with empty label value). 
For example, simple case: designer can style label with icon but without text, with provided solution label will not be rendered. 

PR #12499, opened 2017-11-30
Format generated config files using the short array syntax

cykirsch:
### Description

1. Magento is following a standard of new code using short array syntax (https://github.com/magento/magento2/issues/758).
2. Magento has an automated checker to enforce the short array syntax precommit (https://github.com/magento/magento2/blob/2.2/.php_cs.dist#L31). 
3. Magento recommends checking in `app/etc/config.php` (http://devdocs.magento.com/guides/v2.2/config-guide/config/config-php.html)
>Check this file in to source control and use it in your development, staging, and production systems.
4. Magento's auto-generated `app/etc/config.php` uses the long array syntax.

Put that all together and the auto-generated config doesn't match the required standards to be checked in. The change made in this PR is to modify `PhpFormatter` to recursively return the array representation using short array syntax `[]` instead of long `array()`. If the given variable is not an array, it uses the standard `var_export` behavior.

### Manual testing scenarios

1. bin/magento setup:upgrade
2. app/etc/config.php should use the short array syntax

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/App/DeploymentConfig/Writer/PhpFormatter.php
+++ b/lib/internal/Magento/Framework/App/DeploymentConfig/Writer/PhpFormatter.php
@@ -11,6 +11,8 @@ namespace Magento\Framework\App\DeploymentConfig\Writer;
  */
 class PhpFormatter implements FormatterInterface
 {
+    const INDENT = '  ';
+
     /**
      * Format deployment configuration.
      * If $comments is present, each item will be added
@@ -23,7 +25,7 @@ class PhpFormatter implements FormatterInterface
         if (!empty($comments) && is_array($data)) {
             return "<?php\nreturn array (\n" . $this->formatData($data, $comments) . "\n);\n";
         }
-        return "<?php\nreturn " . var_export($data, true) . ";\n";
+        return "<?php\nreturn " . $this->varExportShort($data, true) . ";\n";
     }
 
     /**
@@ -64,5 +66,28 @@ class PhpFormatter implements FormatterInterface
         }
 
         return var_export($data, true);
+    }
+
+    /**
+     * If variable to export is an array, format with the php >= 5.4 short array syntax. Otherwise use
+     * default var_export functionality.
+     *
+     * @param mixed   $var
+     * @param integer $depth
+     * @return string
+     */
+    private function varExportShort($var, $depth=0) {
+        if (gettype($var) === 'array') {
+            $indexed = array_keys($var) === range(0, count($var) - 1);
+            $r = [];
+            foreach ($var as $key => $value) {
+                $r[] = str_repeat(self::INDENT, $depth)
+                    . ($indexed ? '' : $this->varExportShort($key) . ' => ')
+                    . $this->varExportShort($value, $depth + 1);
+            }
+            return sprintf("[\n%s\n%s]", implode(",\n", $r), str_repeat(self::INDENT, $depth - 1));
+        }
+
+        return var_export($var, TRUE);
     }
 }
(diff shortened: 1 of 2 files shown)


cykirsch:
As an added note, I did attempt to do this using `Zend\Code\Generator\ValueGenerator` instead of `var_export`, but the array depth wasn't being tracked properly. The unit test array would look like this

    [
      'ns1' => [
            's1' => [
                's11',
                's12',
            ],
            's2' => [
                's21',
                's22',
            ],
        ],
      'ns2' => [
            's1' => [
                's11',
            ],
        ],
      'ns3' => 'just text',
      'ns4' => 'just text',
    ]

ghost (approved):

lib/internal/Magento/Framework/App/DeploymentConfig/Writer/PhpFormatter.php
```diff
@@ -51,18 +53,42 @@ private function formatData($data, $comments = [], $prefix = '  ')
...
+     *
+     * @param mixed   $var
+     * @param integer $depth
+     * @return string
+     */
+    private function varExportShort($var, $depth = 0)
+    {
+        if (gettype($var) === 'array') {
+            $indexed = array_keys($var) === range(0, count($var) - 1);
+            $r = [];
```
ishakhsuvarov: Please avoid variables with unclear names. `$r` does not really describe what is it for.

lib/internal/Magento/Framework/App/DeploymentConfig/Writer/PhpFormatter.php
```diff
@@ -51,18 +53,42 @@ private function formatData($data, $comments = [], $prefix = '  ')
...
+
+    /**
+     * If variable to export is an array, format with the php >= 5.4 short array syntax. Otherwise use
+     * default var_export functionality.
+     *
+     * @param mixed   $var
+     * @param integer $depth
+     * @return string
+     */
+    private function varExportShort($var, $depth = 0)
```
ishakhsuvarov: Please take advantage of PHP 7.0 syntax (type hints and return types for methods)

lib/internal/Magento/Framework/App/DeploymentConfig/Writer/PhpFormatter.php
```diff
@@ -51,18 +53,42 @@ private function formatData($data, $comments = [], $prefix = '  ')
...
+    {
+        if (gettype($var) === 'array') {
```
ishakhsuvarov: Is there any specific reason not to use `is_array`?
Additionally, readability may be slightly improved by reversing the logic like

```php
if (!is_array($var)) {
    return var_export($var, true);
}
/// Transforming to short array syntax here
...
```

lib/internal/Magento/Framework/App/DeploymentConfig/Writer/PhpFormatter.php
```diff
@@ -11,6 +11,8 @@
  */
 class PhpFormatter implements FormatterInterface
 {
+    const INDENT = '  ';
```
ishakhsuvarov: Please add a description for the constant (docBlock)

ishakhsuvarov (changes requested):

ishakhsuvarov (approved):

cykirsch:
@ishakhsuvarov @okorshenko I was more or less expecting these changes would be in 2.2.2. Is it odd to you that it is on the 2.2.2-preview branch but not on the 2.2.2 tag? Or does that happen often?

clemblanco:
What if we want 4 spaces tab indentation? Shouldn't this follow a PSR?

cykirsch:
@clementblanco That did cross my mind, and I would prefer it as well, but I neglected to do it...I think because I was continuing to use `var_export` and I didn't want to mix 4 and 2.

PR #12521, opened 2017-12-02
Match flexible static file version in nginx sample config

scottsb:
### Description
Magento 2.2 allows you to specify a custom version for static files being deployed. That means it may not necessarily be numeric like it is when using the default timestamp version numbers. This generalizes the nginx sample config to match custom versions as well.

### DIFF
--- a/nginx.conf.sample
+++ b/nginx.conf.sample
@@ -100,7 +100,7 @@ location /static/ {
 
     # Remove signature of the static files that is used to overcome the browser cache
     location ~ ^/static/version {
-        rewrite ^/static/(version\d*/)?(.*)$ /static/$2 last;
+        rewrite ^/static/(version[^/]+/)?(.*)$ /static/$2 last;
     }
 
     location ~* \.(ico|jpg|jpeg|png|gif|svg|js|css|swf|eot|ttf|otf|woff|woff2)$ {


scottsb:
FYI, I had a typo in the original commit. I just force-pushed an update to my branch that corrects it.

nginx.conf.sample
```diff
@@ -100,7 +100,7 @@ location /static/ {
 
     # Remove signature of the static files that is used to overcome the browser cache
     location ~ ^/static/version {
-        rewrite ^/static/(version\d*/)?(.*)$ /static/$2 last;
+        rewrite ^/static/(version[^/]+/)?(.*)$ /static/$2 last;
```
dmanners, line 103: I think that this regex change will have the desired effect but what if we updated it to keep in the same style as before? Something like `rewrite ^/static/(version[\w\d]*/)?(.*)$ /static/$2 last;` this would then match any word or digit character. Or was there another reason for your choice @scottsb

scottsb: Because of the fact that the version can be specified as a command line argument and no characters are filtered out, I don't think `\w\d` is sufficient. For example, that would not match a hyphen, which is allowed by the `--content-version` argument.

dmanners: Do you know if there is validation on the `--content-version` call? I would worry that anything but / would be a bit too wide and could cause issues.

scottsb: There is no validation. Whatever is passed in is used:
https://github.com/magento/magento2/blob/2275b8911c3ed2a9b70b1510c8f476c18e1915c7/app/code/Magento/Deploy/Service/DeployStaticContent.php#L78

I tested, and it will even accept characters that really shouldn't be allowed like `/` and even `?`. Validation should be added to the deploy command to block those, but here in nginx, I can't see any issue with the broad match.

I'll open a separate issue for the need for validation on the command.

dmanners: Thanks for the feedback @scottsb good to know that you will open an issue about validation on the command side. I would also suggest that in this new issue it might be a good idea to also update the regex to fit the requirements of the new validation.

avoelkl: Thanks for the great discussion and feedback @dmanners and @scottsb. I will process this PR then.

avoelkl:
Hi @scottsb,
Thanks for the PR and sorry for the silence in here. This PR should soon be merged!

PR #12528, opened 2017-12-03
Bugfix/magetwo 61422 category max depth configuration

arnoudhgz:
### Description
The setting `Stores > Configuration > Catalog > Catalog > Category Top
Navigation > Maximal Depth` was actually never being used to limit the
topmenu.

Now the setting will be used as a condition for the category collection when it is fetched.

### Fixed Issues
1. magento/magento2#7543: Category Top Navigation / Maximal Depth configuration not working

### Manual testing scenarios
1. Change the setting in the backend
2. Refresh the invalidated cache
3. See that the setting actually is being used

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
@@ -6,6 +6,7 @@
 namespace Magento\Catalog\Model\ResourceModel\Category;
 
 use Magento\CatalogUrlRewrite\Model\CategoryUrlRewriteGenerator;
+use Magento\Store\Model\ScopeInterface;
 
 /**
  * Category resource collection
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12528) <br/>All committers have signed the CLA.

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
```diff
@@ -6,12 +6,14 @@
...
 
 use Magento\CatalogUrlRewrite\Model\CategoryUrlRewriteGenerator;
+use Magento\Store\Model\ScopeInterface;
 
 /**
  * Category resource collection
  *
  * @api
  * @author      Magento Core Team <core@magentocommerce.com>
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
```
mzeis, line 16: Adding suppressions for CouplingBetweenObjects warnings is not allowed. The reason is that this increases Technical Debt. We understand this means more work, but refactoring this will help improve the code. Also, just adding an additional parameter would change the signature of the `@api` annotated class. Both changes mean that we probably should rebase this PR against `2.3-develop` so we can add breaking changes.

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
```diff
@@ -58,6 +60,60 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Collection\Abstrac
      */
     protected $_loadWithProductCount = false;
 
+    /**
+     * Core store config
+     *
+     * @var \Magento\Framework\App\Config\ScopeConfigInterface
+     */
+    protected $_scopeConfig;
```
mzeis, line 68: Please use `private` visibility. The core team avoids `protected` for new properties (see [2.7](http://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html)). Also, new properties shouldn't have a name starting with an underscore (this is part of [PSR-2](http://www.php-fig.org/psr/psr-2/)).

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
```diff
@@ -58,6 +60,60 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Collection\Abstrac
...
+        \Psr\Log\LoggerInterface $logger,
+        \Magento\Framework\Data\Collection\Db\FetchStrategyInterface $fetchStrategy,
+        \Magento\Framework\Event\ManagerInterface $eventManager,
+        \Magento\Eav\Model\Config $eavConfig,
+        \Magento\Framework\App\ResourceConnection $resource,
+        \Magento\Eav\Model\EntityFactory $eavEntityFactory,
+        \Magento\Eav\Model\ResourceModel\Helper $resourceHelper,
+        \Magento\Framework\Validator\UniversalFactory $universalFactory,
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
+        \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
```
mzeis, line 98: The rules for [backward compatible development](http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/) only allow us to add a new argument to the constructor if it is optional (nullable) for changes to the same minor release. This means that adding the new parameter is okay if we re-base the PR against `2.3-develop` but for `2.2-develop` we would introduce breaking changes which developers are not expecting.

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
```diff
@@ -404,6 +460,23 @@ public function addRootLevelFilter()
         return $this;
     }
 
+    /**
+     * Add root category filter
+     *
+     * @return $this
+     */
+    public function addNavigationMaxDepthFilter()
```
mzeis, line 468: Adding a public method to a class or an interface is another thing which we need to get approval for from a Magento architect as it changes the public signature of the `@api`-annotated class.

mzeis (changes requested):
Hi @arnoudhgz,

thanks for your PR! You find a few comments from me inside the changelist. There are some parts that have to be refactored so they follow the Magento coding standards. It would be great if you can follow up on them.

arnoudhgz:
@mzeis I accidentally closed the PR, I will create a new one and link it here

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
arnoudhgz: I will do a PR on 2.3-develop. 

The suppression is the same as the one from it's parent, so I'll keep that one.

mzeis:
@arnoudhgz Thank you very much!

arnoudhgz:
New PR on `2.3-develop`: https://github.com/magento/magento2/pull/12640

Issue #12450, opened 2017-11-27
Store not found when adding a ? to site URL.


Cinawork:
### Preconditions
1. Magento 2.2.0
2. PHP 7.0.25
3. MariaDB 10.1.29

### Steps to reproduce
1. Turn on "Add storecode to URL's" in "Configuration > General > Web"
2. Navigate in browser to "BASE_URL/?"
3. Error thrown

### Expected result
1.  With just a ? after the BASE_URL site should function normally

### Actual result
1. Store throws error.

### Additional information
`1 exception(s):
Exception #0 (Magento\Framework\Exception\NoSuchEntityException): Requested store is not found
Exception #0 (Magento\Framework\Exception\NoSuchEntityException): Requested store is not found
#0 /Users/*/Development/*/vendor/magento/module-store/Model/StoreManager.php(168): Magento\Store\Model\StoreRepository->get('')
#1 /Users/*/Development/*/generated/code/Magento/Store/Model/StoreManagerInterface/Proxy.php(119): Magento\Store\Model\StoreManager->getStore('')
#2 /Users/*/Development/*/vendor/magento/module-store/Model/Resolver/Store.php(29): Magento\Store\Model\StoreManagerInterface\Proxy->getStore(NULL)
#3 /Users/*/Development/*/vendor/magento/framework/App/Config/ScopeCodeResolver.php(49): Magento\Store\Model\Resolver\Store->getScope(NULL)
#4 /Users/*/Development/*/vendor/magento/framework/App/Config.php(69): Magento\Framework\App\Config\ScopeCodeResolver->resolve('stores', NULL)
#5 /Users/*/Development/*/vendor/magento/framework/App/MutableScopeConfig.php(36): Magento\Framework\App\Config->getValue('web/secure/base...', 'stores', NULL)
#6 /Users/*/Development/*/vendor/magento/module-store/Model/Store.php(797): Magento\Framework\App\MutableScopeConfig->getValue('web/secure/base...', 'store')
#7 /Users/*/Development/*/vendor/magento/module-store/Model/Store.php(601): Magento\Store\Model\Store->isCurrentlySecure()
#8 /Users/*/Development/*/vendor/magento/framework/Interception/Interceptor.php(58): Magento\Store\Model\Store->getBaseUrl()
#9 /Users/*/Development/*/vendor/magento/framework/Interception/Interceptor.php(138): Magento\Store\Model\Store\Interceptor->___callParent('getBaseUrl', Array)
[...]

PR #12529, opened 2017-12-03
#12450: Set Current Store from Store Code if isUseStoreInUrl

osrecio:
Set Current Store from Store Code
### Description
If use Store Code in Url but for some reason is empty after process, set Store_Code from Store

1. magento/magento2#12450: Store not found when adding a ? to site URL

### Manual testing scenarios

1. Turn on "Add storecode to URL's" in "Configuration > General > Web"
1. Navigate in browser to `{BASE_URL}/?`
1. Error thrown

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/App/Request/PathInfoProcessor.php
+++ b/app/code/Magento/Store/App/Request/PathInfoProcessor.php
@@ -44,7 +44,7 @@ class PathInfoProcessor implements \Magento\Framework\App\Request\PathInfoProces
 
         if ($store->isUseStoreInUrl()) {
             if (!$request->isDirectAccessFrontendName($storeCode) && $storeCode != Store::ADMIN_CODE) {
-                $this->storeManager->setCurrentStore($storeCode);
+                $this->storeManager->setCurrentStore($store->getCode());
                 $pathInfo = '/' . (isset($pathParts[1]) ? $pathParts[1] : '');
                 return $pathInfo;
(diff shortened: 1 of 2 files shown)


app/code/Magento/Store/App/Request/PathInfoProcessor.php
```diff
@@ -44,7 +44,9 @@ public function process(\Magento\Framework\App\RequestInterface $request, $pathI
 
         if ($store->isUseStoreInUrl()) {
             if (!$request->isDirectAccessFrontendName($storeCode) && $storeCode != Store::ADMIN_CODE) {
-                $this->storeManager->setCurrentStore($storeCode);
+                if ($storeCode !== '') {
```
orlangur: There is ` } elseif (!empty($storeCode)) {` condition in the bottom, can the whole logic be simplified maybe?

orlangur (changes requested):

app/code/Magento/Store/App/Request/PathInfoProcessor.php
osrecio: @orlangur finally I added `setCurrentStore` from `$store->getCode()` and I adjust tests for this logic

orlangur (approved):

osrecio:
Thanks @orlangur I will make Backport for 2.3 in case my tests fail for same scenario.

PR #12606, opened 2017-12-08
Fix error loading theme configuration on PHP 7.2

Alanaktion:
This fixes an issue loading theme configuration on PHP 7.2 by checking that `$parentPathPieces` is an array before counting it, avoiding an error when the value is `NULL`. PHP [no longer allows](http://php.net/manual/en/migration72.incompatible.php#migration72.incompatible.warn-on-non-countable-types) running `count()` on non-countable types.

An easy way to trigger this bug in the current 2.2 branch is to run the `setup:upgrade` command:
```
$ php bin/magento setup:upgrade
...
Running data recurring...
Warning: count(): Parameter must be an array or an object that implements Countable in vendor/magento/framework/View/Design/Theme/ThemeList.php on line 237
```

This might be better to merge into 2.3, but I'm not sure what the planned version support/release scheduled for 2.2 vs 2.3 is.

### DIFF
--- a/lib/internal/Magento/Framework/View/Design/Theme/ThemeList.php
+++ b/lib/internal/Magento/Framework/View/Design/Theme/ThemeList.php
@@ -234,7 +234,7 @@ class ThemeList extends \Magento\Framework\Data\Collection implements ListInterf
         $media = $themeConfig->getMedia();
 
         $parentPathPieces = $themeConfig->getParentTheme();
-        if (count($parentPathPieces) == 1) {
+        if (is_array($parentPathPieces) && count($parentPathPieces) == 1) {
             $pathPieces = $pathData['theme_path_pieces'];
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12606) <br/>All committers have signed the CLA.

avoelkl:
@Alanaktion Even though this helps PHP 7.2 compatibility, performing the additional `is_array` check is good anyway so i think it's ok go into 2.2 :)
It can be forward ported to 2.3 after we merged this one (either you provide an additional PR or the core team can do it). First we process this one.

avoelkl (approved):

avoelkl:
@Alanaktion Thanks for the PR!

PR #12640, opened 2017-12-11
MAGETWO-61422 Respect Category Top Navigation Max Depth setting

arnoudhgz:
### Description
The setting `Stores > Configuration > Catalog > Catalog > Category Top Navigation > Maximal Depth` was actually never being used to limit the top menu maximum depth.

Now the setting will be used as a condition for the category collection when it is fetched.

### Fixed Issues
1. magento/magento2#7543: Category Top Navigation / Maximal Depth configuration not working

### Manual testing scenarios
1. Change the setting in the backend
2. Refresh the invalidated cache
3. See that the setting actually is being used

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
@@ -79,8 +79,8 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Collection\Abstrac
      * @param \Magento\Eav\Model\ResourceModel\Helper $resourceHelper
      * @param \Magento\Framework\Validator\UniversalFactory $universalFactory
      * @param \Magento\Store\Model\StoreManagerInterface $storeManager
+     * @param \Magento\Framework\DB\Adapter\AdapterInterface $connection
      * @param \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig
-     * @param \Magento\Framework\DB\Adapter\AdapterInterface|null $connection
      *
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12640) <br/>All committers have signed the CLA.

arnoudhgz:
@lenlorijn can you sign the 'Contributor License Agreement'?

lenlorijn:
I signed the Contributor License Agreement :)

mzeis:
FYI: I will fix the Travis error regarding maximum line length after the holidays. Suppressing CouplingBetweenObjects in this case is ok because it would mean a major refactoring for the frequently used parent classes.

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
```diff
@@ -404,6 +462,23 @@ public function addRootLevelFilter()
...
+     * @return $this
+     */
+    public function addNavigationMaxDepthFilter()
+    {
+        $navigationMaxDepth = (int)$this->scopeConfig->getValue(
+            'catalog/navigation/max_depth',
+            ScopeInterface::SCOPE_STORE
+        );
+        if ($navigationMaxDepth > 0) {
+            $this->addLevelFilter($navigationMaxDepth);
```
mzeis, line 477: In my test, this looks like it doesn't work as intended. If I set a maximum depth of 1, no category is shown. If I set the value to 2, the top level category is displayed. If we add 1 to `$navigationMaxDepth` like `$this->addLevelFilter($navigationMaxDepth + 1)`, it works like I would expect it to. I guess that's because the root category level also has to be considered.

mzeis (changes requested):
@arnoudhgz I hope you had a good start into the new year! I finally got around to check your PR and think there might be a change necessary to make this work as intended.

Please can you have a look at the code review?

Thanks!

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
arnoudhgz: No, that seems not like a good solution to me. I think we should adhere to the levels on http://docs.magento.com/m2/ee/user_guide/catalog/navigation-top.html

So in the admin configuration you have to use 2 for the first level.

mzeis (approved):

app/code/Magento/Catalog/Model/ResourceModel/Category/Collection.php
mzeis: You are absolutely right. Thank you for pointing this out. The PR works as intended.

mzeis:
Thanks for your contribution @arnoudhgz! It has been merged into `2.3-develop`.

PR #12650, opened 2017-12-12
Add fallback for Product_links position attribute if not set in request

mohammedsalem:
By setting product_links index on a product via API call, only the first two of each link type are shown in the backend or a in the response of a GET request, although they are correctly added to the database.
The problem was that if the position wasn't set there's no default value or a fallback set in the table "catalog_product_link_attribute_int" for the position attribute.

### Description
in SaveHandler class from Magento\Catalog\Model\Product\Link 
A check goes in all Products link by type to check if the position were set for all products of this type, else if any wasn't set then do a fallback by array position. 

### Manual testing scenarios
Before applying PR:
1. POST/PUT request to following API endpoint with more than 2 linked products (product_links) per link_type without a position: http://dev.domian.com/rest/V1/products/{sku}  
2. Perform a GET request to http://dev.domian.com/rest/V1/products/{sku}  - you should only see the first two linked products per link type

After applying PR:
1. POST/PUT request to following API endpoint with more than 2 linked products (product_links) per link_type without a position: http://dev.domian.com/rest/V1/products/{sku}  
2. Perform a GET request to http://dev.domian.com/rest/V1/products/{sku}  - you should see all linked products per link type with a position set.

If the position is set manually in the POST/PUT request for all products of a specific link type, this positions should be taken into account.

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
+++ b/app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
@@ -66,8 +66,8 @@ class SaveHandler
         $link = $entity->getData($this->metadataPool->getMetadata($entityType)->getLinkField());
         if ($this->linkResource->hasProductLinks($link)) {
             /** @var \Magento\Catalog\Api\Data\ProductInterface $entity */
-            foreach ($this->productLinkRepository->getList($entity) as $link) {
-                $this->productLinkRepository->delete($link);
+            foreach ($this->productLinkRepo->getList($entity) as $link) {
+                $this->productLinkRepo->delete($link);
             }
         }
         $productLinks = $entity->getProductLinks();
@@ -97,7 +97,7 @@ class SaveHandler
 
         if (count($productLinks) > 0) {
             foreach ($entity->getProductLinks() as $link) {
-                $this->productLinkRepository->save($link);
+                $this->productLinkRepo->save($link);
             }
         }
         return $entity;
--- a/app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
+++ b/app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
@@ -119,7 +119,7 @@ class SaveHandler
             }
         }
 
-        // Check if at least on link without position exists per Link type
+        // Check if at least one link without position exists per Link type
         foreach ($linksByType as $type => $links) {
             foreach ($links as $link) {
                 if (!array_key_exists('position', $link->getData())) {
--- a/app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
+++ b/app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
@@ -31,18 +31,28 @@ class SaveHandler
     private $linkResource;
 
     /**
+     * @var linkTypeProvider
+     */
+    private $linkTypeProvider;
+
+    /**
+     * SaveHandler constructor.
      * @param MetadataPool $metadataPool
      * @param Link $linkResource
      * @param ProductLinkRepositoryInterface $productLinkRepository
+     * @param \Magento\Catalog\Model\Product\LinkTypeProvider $linkTypeProvider
      */
     public function __construct(
         MetadataPool $metadataPool,
         Link $linkResource,
-        ProductLinkRepositoryInterface $productLinkRepository
-    ) {
+        ProductLinkRepositoryInterface $productLinkRepository,
+        \Magento\Catalog\Model\Product\LinkTypeProvider $linkTypeProvider
+    )
+    {
         $this->metadataPool = $metadataPool;
         $this->linkResource = $linkResource;
         $this->productLinkRepository = $productLinkRepository;
+        $this->linkTypeProvider = $linkTypeProvider;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12650) <br/>All committers have signed the CLA.

dmanners:
@mohammedsalem looks like @mzeis is going to process this pull request for you. If you need anything else just drop me or Matthias a message here. Thanks.

mohammedsalem:
@dmanners many thanks, for sure gonna do :)

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
```diff
@@ -18,7 +18,7 @@ class SaveHandler
     /**
      * @var ProductLinkRepositoryInterface
      */
-    protected $productLinkRepository;
+    protected $productLinkRepo;
```
mzeis: Please don't rename a protected property as this breaks backwards compatibility ([source](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html)). I know you get a warning regarding the variable name length, but backwards compatibility is more important than code style.

mzeis (changes requested):
Hi @mohammedsalem,

thank you very much for your PR! Please have a look at my code review, there is a minor change I requested.

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
```diff
@@ -55,17 +64,71 @@ public function execute($entityType, $entity)
...
     }
+
+    /**
+     * Check if the position is set for all product links per link type.
+     * array with boolean per type
+     *
+     * @param $linksByType
+     * @return array
+     */
+    private function isPositionsSet($linksByType)
```
mzeis: Please rename this to `isPositionSet()` for consistency.

mohammedsalem: @mzeis Changes made 
the idea of the plural positions naming instead of singular was that the function returns an array of multi product link type, on the same time wanted to keep the bool naming convention using 'is' or 'has'.

mzeis (approved):

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
```diff
@@ -55,17 +64,71 @@ public function execute($entityType, $entity)
...
     }
+
+    /**
+     * Check if the position is set for all product links per link type.
+     * array with boolean per type
+     *
+     * @param $linksByType
+     * @return array
+     */
+    private function isPositionSet($linksByType)
```
mzeis: Thinking some more about it, it's confusing that a method name starting with `is` returns an array. It suggests the result should be boolean.

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
```diff
@@ -55,17 +64,71 @@ public function execute($entityType, $entity)
...
         $productLinks = $entity->getProductLinks();
+
+        // Build links per type
+        $linksByType = [];
+        foreach ($productLinks as $link) {
+            $linksByType[$link->getLinkType()][] = $link;
+        }
+
+        // Do check
+        $hasPositionLinkType = $this->isPositionSet($linksByType);
```
mzeis: It's confusing that a variable name starting with `has` contains an array. The name suggests that it contains a boolean value.

mzeis (changes requested):
I have the feeling this code could be a little bit more simplified. I'm wondering whether we really need so many iterations to find the links that need their position set.

Please can you consider this change? Also see my comments regarding variable naming.

Thanks!

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
mohammedsalem: it contains an array of boolean values per link type, I had the case that the positions were set for the products in "Related" but not set for products under "Cross sell" and so on, 

the same that under on Link type got the case of some products had position attribute and others don't, therefor I had to check if at least one product doesn't have a position under a product Link type all positions under this type will be ignored and user array index instead. 

regarding the question whether we need that amount of iterations, 
This came from too many use cases and variants of sending data, as I said about some link types aren't used at all others have positions but not for all products, another type with full defined products positions.

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
mohammedsalem: Naming convention will be improved

mohammedsalem:
I'll continue working in January because of the holidays, 
Merry Christmas 🎄🌟

mzeis:
Hi @mohammedsalem, please can you provide a way to re-produce the initial problem by posting the calls starting from a fresh 2.2 install? Sorry I only come up with this now.

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
```diff
@@ -55,17 +64,51 @@ public function execute($entityType, $entity)
...
+            /** @var \Magento\Catalog\Api\Data\ProductInterface $entity */
             foreach ($this->productLinkRepository->getList($entity) as $link) {
                 $this->productLinkRepository->delete($link);
             }
         }
         $productLinks = $entity->getProductLinks();
+
+        // Build links per type
+        $linksByType = [];
+        foreach ($productLinks as $link) {
```
mzeis: After your last refactoring, we don't need the variable `$productLinks` in line 72 anymore. Call `$entity->getProductLinks()` directly in line 72. By doing this, we adhere to the best practice that a variable shouldn't be changed anymore after initialisation.

app/code/Magento/Catalog/Model/Product/Link/SaveHandler.php
```diff
@@ -55,17 +64,51 @@ public function execute($entityType, $entity)
...
         }
         return $entity;
     }
+
+    /**
+     * Check if at least one link without position
+     * @param $links
+     * @return bool
+     */
+    public function hasPosition($links)
```
mzeis: Please change the scope - methods should be `private` whenever possible. Also, we now that `$links` will be an array so we can use type hints.

mzeis (changes requested):
Thanks for your refactoring! I have two more recommendations, please have a look at the review.

mohammedsalem:
[PR-usecase.json.txt](https://github.com/magento/magento2/files/1649718/PR-usecase.json.txt)
@mzeis here you are sample request of adding product using Rest API.

magento-engcom-team:
@mohammedsalem thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

mzeis (approved):

ihor-sviziev:
FYI In this PR introduced new dependency to constructor, that wasn't used. I removed it inside https://github.com/magento/magento2/pull/13436

mzeis:
@mohammedsalem Thanks again for your contribution! Your PR has been merged into `2.2-develop`.

@ihor-sviziev Thanks for catching this! The dependency was needed first, but not anymore after the refactoring.

PR #12677, opened 2017-12-13
Add a ModuleManagerInterface for determining if a module is enabled

navarr:
This is a useful addition for feature detection and enablement for third party modules (such as "activate this functionality if such and such GiftCard extension exists"

### Contribution checklist

### DIFF
--- a/app/etc/di.xml
+++ b/app/etc/di.xml
@@ -61,6 +61,7 @@
     <preference for="Magento\Framework\Config\CacheInterface" type="Magento\Framework\App\Cache\Type\Config" />
     <preference for="Magento\Framework\Config\ValidationStateInterface" type="Magento\Framework\App\Arguments\ValidationState" />
     <preference for="Magento\Framework\Module\ModuleListInterface" type="Magento\Framework\Module\ModuleList" />
+    <preference for="Magento\Framework\Module\ModuleManagerInterface" type="Magento\Framework\Module\ManagerInterface" />
     <preference for="Magento\Framework\Component\ComponentRegistrarInterface" type="Magento\Framework\Component\ComponentRegistrar"/>
     <preference for="Magento\Framework\Event\ConfigInterface" type="Magento\Framework\Event\Config" />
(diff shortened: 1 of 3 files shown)


orlangur:
```
public function isEnabled($moduleName)
    {
        return $this->moduleList->has($moduleName);
    }
```

This is the only method of mentioned class. What is the purpose of proposed interface?

navarr:
Purpose is to add an easy ability to determine if a module is enabled to the API.  Currently neither the ModuleManager nor the ModuleListInterface are @api annotated

@orlangur

orlangur:
Interface is only useful when implementation may be potentially switched. Why not mark class method with `@api` instead?

navarr:
I can do that.  I defaulted to an interface b/c I figured it'd be what'd be desired - after all, just b/c I can't think of an alternative implementation doesn't mean one can't exist!

Should I PR the addition on the model itself?  Doing so would add a couple non-api deprecated methods to the API though

orlangur:
I was thinking about marking only one method as `@api`.

Let me check in Slack whether there was some reasoning behind not marking this class as `@api` and whether we should mark class or method.

orlangur:
Raised a question in Slack

> Is there any reason of not marking https://github.com/magento/magento2/blob/2.2-develop/lib/internal/Magento/Framework/Module/Manager.php as `@api`?

navarr:
@orlangur I'm guessing you never got any feedback in Slack?

orlangur:
@navarr yeah, but now we have a better process.

orlangur:
@antonkril @buskamuza please take a look on this interface suggestion. Of course it should be done against `2.3-develop` but first it would be nice to understand whether such feature detection is an acceptable approach.

antonkril:
Explicit `isModuleEnabled` checks are usually a sign of bad modularity.

But yes, such interface with @api annotation makes sense.

antonkril (approved):
Makes sense

navarr:
I'm closing this PR to make a new one against 2.3

PR #12755, opened 2017-12-15
magento/magento2#12294: Bug: Adding Custom Attribute - The value of A…

virtual97:
1. magento/magento2#12294: Bug: Adding Custom Attribute - The value of Admin scope can't 
be empty.- Made hidden fields disabled.

### Manual testing scenarios
1. Add customer Product Attribute
2. Choose Dropdown
3. Under Manage Attributes, click "Add Option" and then "Delete" empty boxes
4. Attempt to save.

### Contribution checklist


omiroshnichenko:
Hi, @virtual97. Unfortunately, your fix is incorrect. When you makes input element disabled it will not present in serialized object. For case with new option it's ok. But when attribute will have saved option, this option will not be present in serialized object and in result will not be removed from attribute on backend. To fix this problem you can make next:
1. Mark new option with some class
2. On remove, check if option has this class.

Example:
In `app/code/Magento/Catalog/view/adminhtml/web/js/options.js`
```js
...
    rendered: 0,
    template: mageTemplate('#row-template'),
    newOptionClass: 'new-option',// Add this property
    isReadOnly: config.isReadOnly,
    add: function (data, render) {
        var isNewOption = false,
            element;

        if (typeof data.id == 'undefined') {
            data = {
                'id': 'option_' + this.itemCount,
                'sort_order': this.itemCount + 1,
                'rowClasses': this.newOptionClass// Add class if option is new
            };
            isNewOption = true;
        },
        remove: function (event) {
            var element = $(Event.findElement(event, 'tr'));

            element.ancestors().each(function (parentItem) {...});

            if (element) {...}

            if (element.hasClassName(this.newOptionClass)) {// Check if element has class newOption
                element.remove();// Remove this element
            }
        },
...
```
In `app/code/Magento/Catalog/view/adminhtml/templates/catalog/product/attribute/options.phtml`
```js
...
<script id="row-template" type="text/x-magento-template">
         <tr <% if (data.rowClasses) { %>class="<%- data.rowClasses %>"<% } %>>//Add class to row template
            <td class="col-draggable">
...
```

PR #12768, opened 2017-12-15
magento/magento2: Missing ext-bcmath dependency added

Mobecls:
Composer does not check availability of the "bcmath" php module. But Magento framework uses "bccomp" function in getBatchSize method of the Magento\Framework\DB\FieldDataConverter class.

### Description
This request proposes to change the composer.json file by adding requirement for the "bcmath" module.

No issues exist.

### Manual testing scenarios
1. remove bcmath php module if installed.
2. run "composer update"
3. run phpunit tests: "vendor/bin/phpunit -c dev/tests/unit/phpunit.xml.dist"
4. error will appear about missing "bccomp" method.

### Contribution checklist


composer.lock
```diff
@@ -51,7 +52,7 @@
                 }
             ],
             "description": "Braintree PHP Client Library",
-            "time": "2017-08-25T19:38:09+00:00"
+            "time": "2017-08-25 19:38:09"
```
ihor-sviziev: Such changes are coming when old composer version used. Could you update your composer to latest version and run composer update?

ihor-sviziev (changes requested):

composer.lock
ishakhsuvarov: Also, it is worth updating `composer.json` at the framework level.
I'll do these changes on our side.

ihor-sviziev (approved):

PR #12773, opened 2017-12-15
Removing option_id from a duplicated option

JosephMaxwell:
### Description
When duplicating bundled product, the option_id was not unset. The new option row is inserted, but the option id is not changed. As such, any deletions to the deleted product's options resulted in those same deletions happening on the original product.

### Manual testing scenarios
1. Create a bundle product and assign some products.
2. Save that bundle product.
3. Note the `option_id` in the `catalog_product_bundle_option` table.
4. Save and Duplicate the bundle product.
5. After the bugfix, the `option_id` value should be different than the `option_id` for the original product.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Bundle/Model/Product/CopyConstructor/Bundle.php
+++ b/app/code/Magento/Bundle/Model/Product/CopyConstructor/Bundle.php
@@ -26,8 +26,16 @@ class Bundle implements \Magento\Catalog\Model\Product\CopyConstructorInterface
 
         $bundleOptions = $product->getExtensionAttributes()->getBundleProductOptions() ?: [];
         $duplicatedBundleOptions = [];
+        
+        /**
+         * @var string $key
+         * @var \Magento\Bundle\Model\Option $bundleOption
+         */
         foreach ($bundleOptions as $key => $bundleOption) {
-            $duplicatedBundleOptions[$key] = clone $bundleOption;
+            $duplicatedBundleOption = clone $bundleOption;
+            $duplicatedBundleOption->unsetData('option_id');
+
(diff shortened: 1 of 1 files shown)


app/code/Magento/Bundle/Model/Product/CopyConstructor/Bundle.php
```diff
@@ -26,8 +26,16 @@ public function build(Product $product, Product $duplicate)
...
+        /**
+         * @var string $key
+         * @var \Magento\Bundle\Model\Option $bundleOption
+         */
         foreach ($bundleOptions as $key => $bundleOption) {
-            $duplicatedBundleOptions[$key] = clone $bundleOption;
+            $duplicatedBundleOption = clone $bundleOption;
+            $duplicatedBundleOption->unsetData('option_id');
+
+            $duplicatedBundleOptions[$key] = $duplicatedBundleOption;
```
orlangur, line 38: 1) `Error: Call to undefined method Mock_BundleOptionInterface_3055e245::unsetData()` only methods from interface should be used, probably `setOptionId(null)` or something can do the trick
2) Please specify typehint for `$bundleOptions` instead (`\Magento\Bundle\Model\Option[]`)
3) Codacy is complaining about long names, probably `clonedBundleOptions` will work
4) Adopt tests to changed code, write a separate test method to enforce `option_id` is unset
5) Use commit amend + force push so that after all changes there is only one commit

orlangur (changes requested):

orlangur:
Hi @JosephMaxwell, please check my earlier comment and apply necessary changes.

orlangur:

PR #12786, opened 2017-12-18
Fix typo in SINGLE_PRODUCT_LAYOUT_HANLDE

aschrammel:
The constant `SINGLE_PRODUCT_LAYOUT_HANLDE` contains a typo. It should be `SINGLE_PRODUCT_LAYOUT_HANDLE`.

### Description
Correct spelling of "handle" in `SINGLE_PRODUCT_LAYOUT_HANLDE` constant in `Magento\Widget\Model\Widget\Instance`

--

### Manual testing scenarios
--

### Contribution checklist

### DIFF
--- a/app/code/Magento/Widget/Model/Widget/Instance.php
+++ b/app/code/Magento/Widget/Model/Widget/Instance.php
@@ -34,7 +34,7 @@ class Instance extends \Magento\Framework\Model\AbstractModel
 
     const PRODUCT_LAYOUT_HANDLE = 'catalog_product_view';
 
-    const SINGLE_PRODUCT_LAYOUT_HANLDE = 'catalog_product_view_id_{{ID}}';
+    const SINGLE_PRODUCT_LAYOUT_HANDLE = 'catalog_product_view_id_{{ID}}';
 
     const PRODUCT_TYPE_LAYOUT_HANDLE = 'catalog_product_view_type_{{TYPE}}';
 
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12786) <br/>All committers have signed the CLA.

dmanners:
Hi @aschrammel. Thanks for the pull request. First thing is first can you sign the CLA and then we can work on accepting this pull request.

app/code/Magento/Widget/Model/Widget/Instance.php
```diff
@@ -34,7 +34,7 @@ class Instance extends \Magento\Framework\Model\AbstractModel
 
     const PRODUCT_LAYOUT_HANDLE = 'catalog_product_view';
 
-    const SINGLE_PRODUCT_LAYOUT_HANLDE = 'catalog_product_view_id_{{ID}}';
+    const SINGLE_PRODUCT_LAYOUT_HANDLE = 'catalog_product_view_id_{{ID}}';
```
dmanners: Unfortunately we cannot rename or remove constants in a PR to `2.2-develop`. I would suggest the best case here would be to add a new constant with the correct spelling and then keep the old constant, deprecate it and then use the new constant to give the old constant it's value. Something like:

```
/**
 * @deprecated see self::SINGLE_PRODUCT_LAYOUT_HANDLE
 */
const SINGLE_PRODUCT_LAYOUT_HANLDE = self::SINGLE_PRODUCT_LAYOUT_HANDLE;

const SINGLE_PRODUCT_LAYOUT_HANDLE = 'catalog_product_view_id_{{ID}}';
```

This way anyone relying on the old constant will still have the constant but all the magento core code that uses it will have been updated to use the new constant.

dmanners (changes requested):

aschrammel:
Hi @dmanners. Thanks for your feedback - I've now deprecated the old constant like you've suggested.

I've already tried to sign the CLA three times - I don't know, why it isn't recognized as signed 😞 

EDIT: CLA problem is solved. I missed to set the correct name used for the PR in the profile.

dmanners:
@aschrammel common issues with the CLA signing are that the email address attached to your github is not public or it does not match your user that you committed with.

dmanners (approved):

ihor-sviziev (approved):

PR #12799, opened 2017-12-19
Removed not-needed scrollbar from checkout sidebar

vovayatsyuk:
### Description
This commit fixes empty gap below the image in checkout sidebar panel that could produce not-needed scrollbar.

Tested on `Chrome Version 63.0.3239.108 (Official Build) (64-bit)`

### Manual testing scenarios
1. Add product to the cart. I've used "Overnight Duffle" from sample data module.
2. Proceed to checkout and see if scrollbar is rendered is cart sidebar aside the product

Before | After
----------|-------
before | after

### Contribution checklist

### DIFF
--- a/app/design/frontend/Magento/blank/Magento_Checkout/web/css/source/module/_minicart.less
+++ b/app/design/frontend/Magento/blank/Magento_Checkout/web/css/source/module/_minicart.less
@@ -242,6 +242,10 @@
 
         .product-image-wrapper {
             &:extend(.abs-reset-image-wrapper all);
+
+            img {
+                vertical-align: top;
+            }
         }
 
         .product-item-pricing {
--- a/app/design/frontend/Magento/luma/Magento_Checkout/web/css/source/module/_minicart.less
+++ b/app/design/frontend/Magento/luma/Magento_Checkout/web/css/source/module/_minicart.less
@@ -252,6 +252,10 @@
 
         .product-image-wrapper {
             &:extend(.abs-reset-image-wrapper all);
+
+            img {
+                vertical-align: top;
+            }
         }
 
         .product-item-pricing {


rogyar:
Hello @vovayatsyuk. Thank you for the PR. Could you specify the OS version where you have this issue? 
I cannot reproduce it on Mac either on Chrome v 63.0.3239.132 (Official Build) (64-bit) or on other browsers (Safari, Firefox). Also, the `.product-image-wrapper` block has `overflow: hidden` style. I'm not sure how a browser can render the scroll in this case there.
Thanks

vovayatsyuk:
Indeed, it does not appear on Luma.

Just tried again and found that this issue is not reproducible because of `padding-bottom: 20px` on the `.minicart-items .product-item` element. If you decrease it - the scrollbar will appear. (So, maybe it was a theme override that caused scrollbar)

Here is a online demo: https://codepen.io/vovayatsyuk/pen/zRvxMo?editors=1100

As you can see from the demo `product-image-wrapper img` has 75px height and `product-image-wrapper` has 79px.

```
OS: Ubuntu 17.10
Chrome: Version 64.0.3282.119 (Official Build) (64-bit)
```

rogyar:
Thanks for the update. I have checked the blank theme as well and found no issues with the scrollbar. So, you said that the issue might be caused by an override. Did you mean the custom override? Just want to ensure that we have this issue in the core.

vovayatsyuk:
Yes, I meant custom template. I didn't realize this when I created an issue.

But I still think that this modification could be added into magento core, as it fixes the possible issue for
third-party themes and modules.

I mean you don't expect to get a scrollbar just because you removed a padding from the `.minicart-items .product-item` element.

rogyar:
Thanks, @vovayatsyuk. We cannot consider it either as a bugfix or as an improvement. There are many places where we can have potential issues after changing the original styles. Unfortunately, it's really hard to cover them all. The main goal is to stick with a golden meal between minimum styles properties and the proper styling. 
Thank you for understanding.

vovayatsyuk:
Okay. As I said, I didn't realize from the beginning that custom modification caused this issue.

I'll close this PR.

PR #12801, opened 2017-12-19
#7627 Cookie expiration time is not extended when browsing the frontend

eugene-shramko:
Current pull request fixes specified issue by extending Magento_PageCache module functionality.

### Description
The issue is related to enabled FPC and both its modes: "Built-in Cache" & "Varnish Cache".
Implemented different solution for each caching application:
* Extended front controller build-in plugin for "Built-in Cache" mode.
* Added a JS widget to send post ajax request to prolong cookie action for "Varnish Cache" mode, because the request may be processed by varnish completely.

The following things were done:
* Added cookie prolongation JS widget with corresponding block, template & layout update.
* Added cookie prolongation interface with corresponding model for frontend cookie.
* Added prolong cookie controller action.
* Updated front controller build-in plugin.
* Added tests for implemented functionality.

1. magento/magento2#7627: Cookie expiration time is not extended when browsing the frontend

### Manual testing scenarios
Scenario 1:
1. Check "Page Cache" cache type is enabled in Backend.
Follow: "System" / "Cache Management"
2. Check "Build-in Cache" caching application for "Page Cache" cache type is selected in Backend.
Follow: "Stores" / "Configuration" / "Advanced" / "System" / "Full Page Cache"
3. Clean cache.
4. Browse the Frontend.
5. Check cookie expiration time is prolonged on every page.

Scenario 2:
1. Check "Page Cache" cache type is enabled in Backend:
Follow: "System" / "Cache Management"
2. Check "Varnish Cache" caching application for "Page Cache" cache type is selected in Backend.
Follow: "Stores" / "Configuration" / "Advanced" / "System" / "Full Page Cache"
3. Clean cache.
4. Browse the Frontend.
5. Check cookie expiration time is prolonged on every page.

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12801) <br/>All committers have signed the CLA.

app/code/Magento/PageCache/view/frontend/web/js/prolong-cookie.js
```diff
@@ -0,0 +1,38 @@
...
+        /**
+         * Prolongs cookie(s) lifetime by performing post ajax request.
+         * @private
+         */
+        _prolongCookie: function () {
+            if (this.options.prolongActionUrl.length === 0) {
+                return;
+            }
+
+            $.post(this.options.prolongActionUrl);
```
kandy, line 35: I don't think that is a good idea to add one more request on every page. We should send post request only in x  second before session end.

eugene-shramko: Thanks for your comment.

I thought about this too. But in the case Varnish is used as caching application, we may not reach Backend at all during configured cookie lifetime. So we may increase configured cookie lifetime up to 2 times by sending such request just before session is exceeded, for instance, when customer is reading the same page all this time. I guess such behavior is not expected.

There is one more idea that just has come to my mind, but it's more tricky. We can introduce another cookie to store expected lifetime of frontend cookie. It may be updated by JS every time the customer opens new page, even fully cached. And when the expected lifetime of frontend cookie will be much greater than its real lifetime in browser, we will send such prolongation request and synchronize the lifetime of frontend cookie (by calculated delta).

There are some other nuances that should be kept in mind, but what do you think in general, should we implement a tricky logic in this case?

kandy: I believe that making a request on every page (or only on cached) is not acceptable solution compared with a workaround like significant increase of session/cookies or switch to built-in FPC. So we need more advanced/tricky implementation. 
Also cookies expiration data can be added to private data. See http://devdocs.magento.com/guides/v2.2/extension-dev-guide/cache/page-caching/private-content.html for more details

eugene-shramko: Agreed, will come back with updated solution soon.
Thanks for the advice regarding private data.

kandy: Ok, thank you for your cooperation

sidolov:
Hi @eugene-shramko , any updates on this issue? It's a good feature, it would be great if you continue!

eugene-shramko:
Hi @sidolov, there are no updates for now due to some other priorities in my work, but I know that the issue may be important and I am going to proceed with it in the very near future.
Thanks for reminding )

sidolov:
@eugene-shramko , I am closing this PR now due to inactivity.
It would be great if you continue work on it.
Please, reopen PR if you wish to continue
Thank you for collaboration!

PR #12806, opened 2017-12-19
Update checkout controller json usage

dmanners:
### Description
The controller `app/code/Magento/Checkout/Controller/Account/Create.php` uses the object manager directly when building it's json response. I have replaced the usage of object manager with the usage of `$this->resultFactory->create(ResultFactory::TYPE_JSON);`

### Manual testing scenarios

1. Creating an account via checkout using this branch and the clean mainline branch should result in the exact same behavior.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/Test/Unit/Controller/Account/CreateTest.php
+++ b/app/code/Magento/Checkout/Test/Unit/Controller/Account/CreateTest.php
@@ -38,12 +38,14 @@ class CreateTest extends \PHPUnit\Framework\TestCase
     protected $orderCustomerService;
 
     /**
-     * @var \PHPUnit_Framework_MockObject_MockObject
+     * @var \Magento\Framework\Controller\ResultFactory|\PHPUnit_Framework_MockObject_MockObject
      */
-    protected $objectManagerMock;
+    private $resultFactory;
 
-    protected $resultFactory;
-    protected $resultPage;
+    /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/Checkout/Test/Unit/Controller/Account/CreateTest.php
```diff
@@ -66,53 +83,48 @@ protected function setUp()
...
-            $this->returnValue($customer)
-        );
-
-        $jsonMock->expects($this->once())
+        $this->orderCustomerService->expects($this->once())
+            ->method('create')
+            ->with(100)
+            ->will($this->returnValue($customer));
+
+        $resultJson = '{"errors": "false", "message": "A letter with further instructions will be sent to your email."}';
```
mzeis: @dmanners Please can you split this up so the line doesn't exceed 120 characters? I'd do it myself but my VM pretends to by dying atm so I cannot run my tests locally *sigh*. ;)

mzeis (changes requested):

app/code/Magento/Checkout/Test/Unit/Controller/Account/CreateTest.php
dmanners: sure I can sort that out today

dmanners:
@mzeis updated from your feedback.

mzeis (approved):

mzeis:
Thank you very much for your contribution, the change will be merged!

Issue #6401, opened 2016-08-30
X-Magento-Tags header too large


katef:
### Preconditions
I've been reading about people using Magento2 and the size of the `X-Magento-Tags` header.
### Steps to reproduce
For example http://www.maxbucknell.com/blog/2016/2/10/troubleshooting-varnish-with-magento-2 writes:

> The rub here is that Magento adds a tag for every product in a category. And the tag is of the form "catalog_product_{{PRODUCT_ID}}".
### Actual result
That's an integer and 20 bytes of string per product, which gives some users headers several kB in size, when many products are in one category. In that case, the author describes hitting Varnish's limit at 400 products.
### Expected result
I work for a CDN where some of our users use Magento, and we cache their content with Varnish. Increasing Varnish's `http_resp_hdr_len` size may suit that one person's situation, but it is infeasible for us, because it would affect all of our users.

### Proposed solution
I have a suggestion to offer, for how to use header space more efficiently:

Each product ID is an integer. The `X-Magento-Tags` header describes the presence or absence of products only. That information can be done with a single bit per product. The index of the bit would give the product ID.

That's usually known as a bit vector, or a bitfield, or a bitmap. My suggestion is to switch the header to use a bitmap, encoded in hex. That'd take one bit per product, instead of over 20 _bytes_.

More on bitmaps: https://en.wikipedia.org/wiki/Bit_array

stamster:
Excellent point!
But I'm afraid such change is not going to happen anytime soon, maybe for Magento 3.0... and this is only one of the topics out there to discuss when it comes to optimization.

katef:
Thank you - Although I'd request that you consider this a blocking item, rather than an optimisation, because this header is so large that it prevents certain users from using Varnish.

melnikovi:
@katef thank you for reporting this issue, internal ticket created MAGETWO-57851

stamster:
@katef : I'm not representative of Magento Commerce Inc. ...  :)  I completely agree with you, of course it is blocking for certain use cases  out there.

pynej:
Some more detail:
The `catalog_category_product_` and `catalog_category_` could easily be changed to 2 characters that would still be unique and cut down the length of this header by ~80%.

[...]

PR #12831, opened 2017-12-20
Split X-Magento-Tags HTTP header into multiple headers

sambolek:
On Magento sites with large catalogs, it's easy to hit HTTP header length limit on Varnish or HTTP server, for categories that have large amount of products and are set as anchor. This is because all of the product tags are sent in initial backend_fetch request. 

This is usually addressed by implementing a workaround from the documentation http://devdocs.magento.com/guides/v2.2/config-guide/varnish/tshoot-varnish-503.html . Thing is, that workaround doesn't solve the issue, as you end up increasing the limit constantly as categories grow larger. 

Furthermore, you might hit another limit on your Apache/nginx HTTP server, as [both](http://httpd.apache.org/docs/2.2/mod/core.html#limitrequestfieldsize) [have](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers) 8kB max header value by default.

In M2.2 https://github.com/magento/magento2/commit/2529ec4fc035eec8c524e94c0b73efd8c3efbb28 , cache tags have been refactored to be a bit shorter (e.g. cat_p_xx instead of catalog_product_xx), which is another workaround to address the same issue.

According to RFC2616 https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html:

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma.

### Description

This pull request proposes splitting the X-Magento-Tags header into multiple X-Magento-Tags headers that would stay roughly under 8kB limit each.

1. Fixes magento/magento2#8696
2. Fixes magento/magento2#6401

### Manual testing scenarios

1. Set category as "anchor" and assign about 2500 products to it (it's important that X-Magento-Tags header goes over 8kB limit)
2. Set up Magento store to use Varnish, with default Magento-generated VCL and no other parameter configured (ignore workaround/recommendations from this page http://devdocs.magento.com/guides/v2.2/config-guide/varnish/tshoot-varnish-503.html or just assign 2500 more products to the same category)
3. Attempt to open a category on frontend
4. Category returns 503 error from Varnish
5. If you have varnishlog active at the time this occurs, observe the error `FetchError     http read error: overflow` on `bereq 2 fetch` request

### Contribution checklist

### DIFF
--- a/app/code/Magento/PageCache/Model/HTTP/Header/XMagentoTags.php
+++ b/app/code/Magento/PageCache/Model/HTTP/Header/XMagentoTags.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace Magento\PageCache\Model\HTTP\Header;
+
+use Zend\Http\Header\MultipleHeaderInterface;
+use Zend\Http\Header\GenericHeader;
+use Zend\Http\Header\HeaderValue;
+use Zend\Http\Header\Exception;
+
+class XMagentoTags implements MultipleHeaderInterface
+{
+    /**
+     * @var string
+     */
+    protected $value;
+
+    /**
+     * XMagentoTags constructor.
+     *
+     * @param string|null $value
+     */
+    public function __construct($value = null)
+    {
+        if ($value) {
+            HeaderValue::assertValid($value);
+            $this->value = $value;
+        }
+    }
+
+    /**
+     * Create X-Magento-Tags header from a given header line
+     *
+     * @param string $headerLine The header line to parse.
+     *
+     * @return self
+     * @throws Exception\InvalidArgumentException If the name field in the given header line does not match.
+     */
+    public static function fromString($headerLine)
+    {
+        list($name, $value) = GenericHeader::splitHeaderLine($headerLine);
+
+        // check to ensure proper header type for this factory
+        if (strtolower($name) !== 'x-magento-tags') {
+            throw new Exception\InvalidArgumentException(
+                sprintf(
+                    'Invalid header line for X-Magento-Tags string: "%s"',
+                    $name
+                )
+            );
+        }
+
+        // @todo implementation details
+        $header = new static($value);
+
+        return $header;
+    }
+
+    /**
+     * Cast multiple header objects to a single string header
+     *
+     * @param  array $headers
+     * @throws Exception\InvalidArgumentException
+     * @return string
+     */
+    public function toStringMultipleHeaders(array $headers)
+    {
+        $name = $this->getFieldName();
+        $values = array($this->getFieldValue());
+        foreach ($headers as $header) {
+            if (!$header instanceof static) {
+                throw new Exception\InvalidArgumentException(
+                    'This method toStringMultipleHeaders was expecting an array of headers of the same type'
+                );
+            }
+            $values[] = $header->getFieldValue();
+        }
+
+        return $name . ': ' . implode(',', $values) . "\r\n";
+    }
+
+    /**
+     * Get the header name
+     *
+     * @return string
+     */
+    public function getFieldName()
+    {
+        return 'X-Magento-Tags';
+    }
+
+    /**
+     * Get the header value
+     *
+     * @return string
+     */
+    public function getFieldValue()
+    {
+        return $this->value;
+    }
+
+    /**
+     * Return the header as a string
+     *
+     * @return string
+     */
+    public function toString()
+    {
+        return 'X-Magento-Tags: ' . $this->getFieldValue();
+    }
+}
--- a/app/code/Magento/PageCache/Model/HTTP/PhpEnvironment/ResponsePlugin.php
+++ b/app/code/Magento/PageCache/Model/HTTP/PhpEnvironment/ResponsePlugin.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Magento\PageCache\Model\HTTP\PhpEnvironment;
+
+use Magento\Framework\HTTP\PhpEnvironment\Response as Subject;
+use Magento\PageCache\Model\Config;
+use Zend\Http\HeaderLoader;
+use Magento\PageCache\Model\HTTP\Header\XMagentoTags;
+
+class ResponsePlugin
+{
+    /**
+     * Approximately 8kb in length
+     *
+     * @var int
+     */
+    private $requestSize = 8000;
+
+    /**
+     * PageCache configuration
+     *
+     * @var Config
+     */
+    private $config;
+
+    /**
+     * ResponsePlugin constructor.
+     *
+     * @param Config $config
+     */
+    public function __construct(
+        Config $config
+    ) {
+        $this->config = $config;
+    }
+
+    /**
+     * Special case for handling X-Magento-Tags header
+     * splits very long header into multiple headers
+     *
+     * @param Subject  $subject
+     * @param \Closure $proceed
+     * @param string   $name
+     * @param string   $value
+     * @param bool     $replace
+     *
+     * @return Subject|mixed
+     */
+    public function aroundSetHeader(Subject $subject, \Closure $proceed, $name, $value, $replace = false)
+    {
+        //if varnish isn't enabled, don't do anything
+        if (!$this->config->isEnabled() || $this->config->getType() != Config::VARNISH) {
+            return $proceed($name, $value, $replace);
+        }
+
+        $this->addHeaderToStaticMap();
+
+        if ($name == 'X-Magento-Tags') {
+            $headerLength = 0;
+            $value = (string)$value;
+            $tags = explode(',', $value);
+
+            $newTags = [];
+            foreach ($tags as $tag) {
+                if ($headerLength + strlen($tag) > $this->requestSize - count($tags) - 1) {
+                    $tagString = implode(',', $tags);
+                    $subject->getHeaders()->addHeaderLine($name, $tagString);
+                    $newTags = [];
+                    $headerLength = 0;
+                }
+                $headerLength += strlen($tag);
+                $newTags[] = $tag;
+            }
+
+            return $subject;
+        }
+
+        return $proceed($name, $value, $replace);
+    }
+
+    /**
+     * Add X-Magento-Tags header to HeaderLoader static map
+     */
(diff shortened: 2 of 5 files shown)


sambolek:
I'm open to recommendations for this one - especially the part adding header to StaticMap. It appears a bit difficult to add another header mapping to `Zend\Http\HeaderLoader`, as this has to be done before first header is set on response.

Header splitting might also be improved, but curious to see if this is the right approach to solve this issue.

sambolek:
Any news about this one?

sambolek:
@magento-team @magento-engcom-team 

Anyone caring enough to discuss this? Don't think we should be sending MBs inside a single X-Magento-Tags header, but keep to hear what you think.

ihor-sviziev:
@kokoc @paliarush, you're responsible for page cache module. Could you review this PR?

udovicic (approved):

mropanen:
https://github.com/magento/magento2/pull/12831/files/05ff2db5923c61d3a6a0efa374ce2ded3af8cc93#diff-8824a34ec801e0f52b1299f54ac9104cR66

I think you mean to implode $newTags, not $tags there :)

ihor-sviziev:
@sidolov could you review these changes?

ishakhsuvarov:
Hi @sambolek 
Sorry for a late review. I've got a few questions regarding the implementation:
* What is the specific reason to split only one header? If there is a real-world example and described in RFC this may be implemented for any headers or for some configurable.
* If this problem is applicable exclusively to specific catalog setups why not reconfigure varnish?

sambolek:
Hi @ishakhsuvarov ,
the issue we actually had couldn't be easily traced, but turned out to be related to X-Magento-Tags header being too large for some part of the system. It might be due to some other limit that isn't documented in Magento docs, but we (nor the hosting company) have figured it out.

As per my initial message, if a category has about 2500 products, a single header X-Magento-Tags gets too large, causing issues with Varnish (having to adjust parameters all around), web server (Apache and nginx have 8kB header limit), and LoadBalancers (HAProxy and nginx have 8kB limit as well). Even if there is a limit to adjust, the solution isn't to keep adjusting the limit as you add more products to the system and assign them to categories.

Many merchants could easily have categories with large amount of products, or anchor categories with plenty subcategories - since Magento is targeting mid/large merchants, catalog should work regardless of these factors.

Since RFC allows for multiple headers with the same name, the logical way forward would be do build the application (Magento) in a way that headers are split properly, without a need to adjust limits on the entire system. 

This pull request solves the problem at hand, without going too deep into architectural choices of Magento system/platform. Of course, it might be wise to explore splitting all cookies that can be comma-separated (as per RFC) on platform level, but, in the meantime implement this pull request, which shouldn't have any BC to help other merchants that have the same issue we did.

ishakhsuvarov:
Hi @sambolek 
> This pull request solves the problem at hand, without going too deep into architectural choices of Magento system/platform. Of course, it might be wise to explore splitting all cookies that can be comma-separated (as per RFC) on platform level, but, in the meantime implement this pull request, which shouldn't have any BC to help other merchants that have the same issue we did.

This PR indeed, seems to solve the problem. However, it could be better to try and implement the desired state approach right away, as fixing only one header type just masks the problem.

I'll submit more thoughts on it via the code review to the Pull Request to continue the discussion.

app/code/Magento/PageCache/Model/HTTP/PhpEnvironment/ResponsePlugin.php
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * @param Subject  $subject
+     * @param \Closure $proceed
+     * @param string   $name
+     * @param string   $value
+     * @param bool     $replace
+     *
+     * @return Subject|mixed
+     */
+    public function aroundSetHeader(Subject $subject, \Closure $proceed, $name, $value, $replace = false)
```
ishakhsuvarov, line 49: It's not recommended to add new around plugins, as it has a greater performance penalty, specifically, in such a critical part of the system.

app/code/Magento/PageCache/Model/HTTP/PhpEnvironment/ResponsePlugin.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        //if varnish isn't enabled, don't do anything
+        if (!$this->config->isEnabled() || $this->config->getType() != Config::VARNISH) {
+            return $proceed($name, $value, $replace);
+        }
+
+        $this->addHeaderToStaticMap();
+
+        if ($name == 'X-Magento-Tags') {
```
ishakhsuvarov, line 58: There are already plugins in the system which set `X-Magento-Tags`. Did you consider updating those instead of processing same header again?

app/code/Magento/PageCache/Model/HTTP/PhpEnvironment/ResponsePlugin.php
```diff
@@ -0,0 +1,93 @@
...
+     * @param string   $name
+     * @param string   $value
+     * @param bool     $replace
+     *
+     * @return Subject|mixed
+     */
+    public function aroundSetHeader(Subject $subject, \Closure $proceed, $name, $value, $replace = false)
+    {
+        //if varnish isn't enabled, don't do anything
+        if (!$this->config->isEnabled() || $this->config->getType() != Config::VARNISH) {
```
ishakhsuvarov, line 52: While the issue described in the Pull Request only relates to Varnish, this header is also used by Builtin Implementation. Would it make sense to make it more generic?

app/code/Magento/PageCache/Model/HTTP/PhpEnvironment/ResponsePlugin.php
```diff
@@ -0,0 +1,93 @@
...
+                    $tagString = implode(',', $tags);
+                    $subject->getHeaders()->addHeaderLine($name, $tagString);
+                    $newTags = [];
+                    $headerLength = 0;
+                }
+                $headerLength += strlen($tag);
+                $newTags[] = $tag;
+            }
+
+            return $subject;
```
ishakhsuvarov, line 75: Different plugins, for example solving similar issue for other header types may be theoretically broken in this way. Does it make sense to tune the actual Response implementation instead?

app/code/Magento/PageCache/Model/HTTP/Header/XMagentoTags.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Cast multiple header objects to a single string header
+     *
+     * @param  array $headers
+     * @throws Exception\InvalidArgumentException
+     * @return string
+     */
+    public function toStringMultipleHeaders(array $headers)
```
ishakhsuvarov, line 65: This method seems to be copied from the Zend's default implementation of `\Zend\Http\Header\GenericMultiHeader::toStringMultipleHeaders` which could make it more challenging to support in the future

app/code/Magento/PageCache/Model/HTTP/Header/XMagentoTags.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Get the header name
+     *
+     * @return string
+     */
+    public function getFieldName()
+    {
+        return 'X-Magento-Tags';
```
ishakhsuvarov, line 88: `X-Magento-Tags` literal occurs many times across this this PR, would it make sense to move it to a constant?

ishakhsuvarov (changes requested):
To accept this solution to the core I would suggest making it more generic, as current state of it seems to be a perfect solution for a third-party module, but not a core component

sidolov:
Hi @sambolek , I am closing this PR now due to inactivity.

kurtdeserter:
Are there any updates for this issue? This problem still exists in M2.3.3 
All "fixes" found by now, are not resolving this issue.

ihor-sviziev:
Would like to add some summary on this PR:

This solution is quite good, but it's not finalized in order to be accepted. 
You could create new PR based on this PR and apply requested changes. 

I have only one concern - such change might break some services like Fastly (for instance used in Magento Cloud), so probably it should be double checked additionally

hostep:
It would still be nice if a solution for this could get implemented anywhere in the future for this problem.

We currently run against this huge amount `X-Magento-Tags` header on a Magento 2.4.3-p2 installation with the search results page when it returns > 5000 products or something like that in its resultset.

@ishakhsuvarov, @sidolov, are there plans to ever support this somehow? Not necessarily by splitting the header, but maybe by using some other (and fewer) cache tags in search results so it doesn't list every single product id from the resultset?

hostep:
Sorry, ignore comment above, we had this issue because the pagination on search results pages didn't work, all products were always being displayed instead of only one page. So, once we solve this the pagination of search result pages, we won't run into this anymore.

Sorry for the noise 🙂

hostep:
@ihor-sviziev: in our case it was a custom module that loaded the product collection too soon before the pagination was getting applied (or something like that, still trying to figure out exactly is going on and how to fix it).

If you have this issue with all products being shown in the search result page, it's most likely not a core Magento bug, but some customisation done to your shop, either in the custom theme or in custom/3rd party modules. Try switching to a default theme and/or disabling modules and test again, until you find the culprit, that's how I found out about which module caused the issue in our case.

hostep:
Update @ihor-sviziev, seems to have something to do with this change: https://github.com/magento/magento2/pull/27263, when I add `$this->addToolbarBlock($collection);` back, it works again in our case.

Also possibly related issue: https://github.com/magento/magento2/issues/34292

We use the `catalog_block_product_list_collection` event in our custom module to get the product collection, and after the PR change from above, the toolbar filtering is no longer applied, so we get the full collection and by looping over it we load the collection before the toolbar filters have been applied.

----

Sorry to the readers of this thread, this has nothing to do with it, I'll try to shut up now 😄

PR #12875, opened 2017-12-24
Add more parameters to ajax:addToCart

srenon:
### Description

The SKU by itself is not very useful because for most third-party integration you will need at minimum SKU and qty. 

The 'res' is also important because you could write a plugin for Magento\Checkout\Controller\Cart\Add to add more information to the return JSON.

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/web/js/catalog-add-to-cart.js
+++ b/app/code/Magento/Catalog/view/frontend/web/js/catalog-add-to-cart.js
@@ -97,7 +97,7 @@ define([
                 success: function (res) {
                     var eventData, parameters;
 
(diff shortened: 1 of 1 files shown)


koenner01:
Would be handy if this was backported to 2.1 too :)

koenner01:
However after looking at other places Magento fires events through the trigger, I'm not a fan of dumping all the parameters after each other in the trigger. It would be better to add 1 parameter 'eventData' that contains the seperate variables.
This way you can more easily create a generic handler for the triggered events and the codebase is staying more closely to how the current implementations work

Ctucker9233:
Created Backport for this #14898

Issue #12866, opened 2017-12-23
Can't Disable Advanced Reporting


nfourteen:
### Preconditions

1. Magento 2.2.2

### Steps to reproduce

1. After a fresh install of version 2.2.2, a system message was received that Synchronization Failed with Magento Business Intelligence (I don't remember the exact message).
2. To prevent further synchronization error messages, go to `Stores > Configuration > General > Advanced Reporting` and change **Advanced Reporting Service** to Disable
3. Press **Save Config**

### Expected result

1. The advanced reporting capabilities should be disabled.

### Actual result

1. The following weird error message is received: `Please select a vertical.`
2. The advanced reporting capabilities revert back to being enabled.

### Reason and Proposed Solutions

The error is thrown from `Magento\Analytics\Model\Config\Backend\Vertical::beforeSave()` method because I did not select an **Industry** under **Industry Data**. That defaults to `--Please Select--` and is an empty value, so `beforeSave()` behaves as it should.

Since the **Advanced Reporting** needs an external account and service, it should default to Disable on new installs. However, if it was Magento's intention to have it enabled for all new installs, then 1) the logic should be changed so an empty value for the **Industry** can be provided when **Advanced Reporting Service** is set to Disable and **Save Config** is pressed or 2) a more detailed error message should be displayed so merchants understand that an **Industry** must be provided, even to Disable the **Advanced Reporting Service**.

PieterCappelle:
Hi @nfourteen, could you create PR? Thanks!

nfourteen:
Hi @PieterCappelle, I just submitted the [pull request](https://github.com/magento/magento2/pull/12879).

magento-engcom-team:
@nfourteen, thank you for your report.
We've created internal ticket(s) MAGETWO-86006 to track progress on the issue.

tony-sar:
Select Disable and also select Industry data and click on save it should save . I think vertical message relate to the industry data

jzahedieh:
![image](https://user-images.githubusercontent.com/2098908/34492460-93f09c8a-efdf-11e7-8ad3-b666b18fa054.png)

And even when you disable it the message shows in the dashboard, this is in really bad taste Magento, promoting 3rd party services on the homepage of the admin interface without being able to disable it!

All you need to do is add `ifconfig="analytics/general/enabled"` to `vendor/magento/module-analytics/view/adminhtml/layout/adminhtml_dashboard_index.xml`

adamlavery:
[...]

PR #12879, opened 2017-12-25
Fix issue #12866: Can't Disable Advanced Reporting

nfourteen:
### Description

Disable Magento_Analytics (aka Advanced Reporting) by default and provide clearer message if an industry is not provided.

1. [magento/magento2#12866](https://github.com/magento/magento2/issues/12866): Can't Disable Advanced Reporting

### Manual testing scenarios

1. Install new Magento 2.2 instance and `Stores > Configuration > General > Advanced Reporting > Advanced Reporting Service` should be set to Disable after install
2. Go to `Stores > Configuration > General > Advanced Reporting` and change **Advanced Reporting Service** to Enable/Disable and press **Save Config**. If an **Industry** is not provided, then the `Please select an industry.` error message is shown rather than `Please select a vertical.` 

### Contribution checklist

### DIFF
--- a/app/code/Magento/Analytics/Setup/InstallData.php
+++ b/app/code/Magento/Analytics/Setup/InstallData.php
@@ -30,7 +30,7 @@ class InstallData implements InstallDataInterface
                     'scope' => 'default',
                     'scope_id' => 0,
                     'path' => 'analytics/subscription/enabled',
-                    'value' => 1
+                    'value' => 0
                 ],
                 [
                     'scope' => 'default',
--- a/app/code/Magento/Analytics/Model/Config/Backend/Vertical.php
+++ b/app/code/Magento/Analytics/Model/Config/Backend/Vertical.php
@@ -24,7 +24,7 @@ class Vertical extends \Magento\Framework\App\Config\Value
     public function beforeSave()
     {
         if (empty($this->getValue())) {
-            throw new LocalizedException(__('Please select a vertical.'));
+            throw new LocalizedException(__('Please select an industry.'));
         }
 
         return $this;


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12879) <br/>All committers have signed the CLA.

app/code/Magento/Analytics/Setup/InstallData.php
```diff
@@ -30,7 +30,7 @@ public function install(ModuleDataSetupInterface $setup, ModuleContextInterface
                     'scope' => 'default',
                     'scope_id' => 0,
                     'path' => 'analytics/subscription/enabled',
-                    'value' => 1
+                    'value' => 0
```
ishakhsuvarov, line 33: I would suggest changing logic the validation, to better support the case with disabling validation, instead of disabling the feature on every new install. It was intended to be that way.

ishakhsuvarov (changes requested):

app/code/Magento/Analytics/Setup/InstallData.php
nfourteen: @ishakhsuvarov I can do that, but then what industry should be set as a default? Otherwise, on fresh installs the reporting service will be enabled, but the **Industry** will be empty and "will cause distortion of the analytics reports", thus breaking the validation logic of `Magento\Analytics\Model\Config\Backend\Vertical::beforeSave()`. If the module is enabled by default and someone signs up for the service, will the Magento BI alert them that they need to select an **Industry**?

If that's handled, I can update the pull request so the module is enabled by default and update the disabling validation logic with:

```
// Magento\Analytics\Model\Config\Backend\Vertical
public function beforeSave()
{
    if (empty($this->getValue()) && ($this->getFieldsetDataValue('enabled') === '1')) {
        throw new LocalizedException(__('Please select an industry.'));
    }

    return $this;
}
```

which will allow disabling of the module, but make sure an **Industry** is supplied when trying to save the config with the module enabled.

ishakhsuvarov: @nfourteen Currently waiting for someone from product organization to describe the desired state of the feature.
Thank you for collaboration

nfourteen: @ishakhsuvarov Thanks for the update. Let me know when you hear something as I'd be happy to edit this PR to reflect the desired state.

slavvka:
@nfourteen @ishakhsuvarov As far as I know Advanced Reporting should be on for 2.2 and 2.3 by default. But I am waiting for the exact response from PM

slavvka:
@nfourteen According to the info from our PM, it should be on by default. So I am closing this PR

nfourteen:
@slavvka Ok, then there is another issue in that Advanced Reporting can't be disabled when it is on by default. See the related issue here: https://github.com/magento/magento2/issues/12866. I think this PR should be reopened and the code changed to address the issue using one of the proposed solutions: see **Reason and Proposed Solutions** within the issue report.

slavvka:
@nfourteen I agree. Please provide the changed code

slavvka:
@nfourteen since the code you provided fixes the issue in a wrong way I am closing this PR. Please feel free to create a new one with a correct solution. Thank you!

nfourteen:
@slavvka This is a Magento core problem and an issue that needs to be fixed. However, since it's a Magento problem, I don't know how the PM wants to proceed. So, I'd be happy to fix Magento core code, but I need some direction as to which way Magento would like to proceed. Please stop closing this PR because it is a issue that needs to be resolved.

Please have the PM review issue [#12866](https://github.com/magento/magento2/issues/12866) and advise which proposed solution would be acceptable or if another solution is desired.

okorshenko:
Hi @nfourteen 
Thank you for your contribution. We are closing this PR since the proposed solution cannot be accepted in the current state. We are working with PO right now to guide you for a proper solution.

Thank you for your patience

slavvka:
@nfourteen Thank you for your attitude but this issue has been already fixed in this PR #15366

PR #12898, opened 2017-12-27
Instant Purchase Unit Tests (Partial Coverage)

pmclain:
### Description
Begins adding some test coverage for Magento_InstantPurchase. Minor refactoring was performed while writing the test cases.

### Contribution checklist


app/code/Magento/InstantPurchase/Controller/Button/PlaceOrder.php
```diff
@@ -104,21 +104,22 @@ public function __construct(
      */
     public function execute()
     {
-        $request = $this->getRequest();
-        if (!$this->doesRequestContainAllKnowParams($request)) {
+        $params = $this->getRequest()->getParams();
```
vkublytskyi: These changes decrease readability of code and introduce code duplication. Testability is not affected anyhow. Controller invokes only two methods of the request: `getParam` and `getParams`. Fetching all parameters into array and use it in procedural style is not helps with testability.

To mock request properly `PHPUnit_Framework_MockObject_Builder_InvocationMocker::willReturnMap` or `PHPUnit_Framework_MockObject_Builder_InvocationMocker::willReturnCallback` may be used.

app/code/Magento/InstantPurchase/Controller/Button/PlaceOrder.php
```diff
@@ -104,21 +104,22 @@ public function __construct(
...
 
-        $paymentTokenPublicHash = (string)$request->getParam('instant_purchase_payment_token');
-        $shippingAddressId = (int)$request->getParam('instant_purchase_shipping_address');
-        $billingAddressId = (int)$request->getParam('instant_purchase_billing_address');
-        $carrierCode = (string)$request->getParam('instant_purchase_carrier');
-        $shippingMethodCode = (string)$request->getParam('instant_purchase_shipping');
-        $productId = (int)$request->getParam('product');
-        $productRequest = $this->getRequestUnknownParams($request);
+        $paymentTokenPublicHash = (string)$params['instant_purchase_payment_token'];
+        $shippingAddressId = (int)$params['instant_purchase_shipping_address'];
```
vkublytskyi: Such refactoring decrease quality of code:
1. 6 rows now contains duplicated logic
2. Instead of reading simply "get named parameter from request" now developer need to remember that `$params` were fetched from a request and interpret ternary operator.

app/code/Magento/InstantPurchase/Controller/Button/PlaceOrder.php
```diff
@@ -104,21 +104,22 @@ public function __construct(
...
+        $billingAddressId = (int)$params['instant_purchase_billing_address'];
+        $carrierCode = (string)isset($params['instant_purchase_carrier']) ? $params['instant_purchase_carrier'] : '';
```
vkublytskyi: This line contains an error (and this is one of the reasons why working with request object are preferable over fetching data into an array and usage of `isset`): type casting will be applied to the condition of the ternary operator, not to the "true" part of result. Se  code example https://3v4l.org/viTNs

app/code/Magento/InstantPurchase/CustomerData/InstantPurchase.php
```diff
@@ -94,7 +94,7 @@ public function getSectionData(): array
         $data = [
             'available' => $instantPurchaseOption->isAvailable()
         ];
-        if (!$instantPurchaseOption->isAvailable()) {
+        if (!$data['available']) {
```
vkublytskyi: This also decrease code readability as now code reader should keep in mind how `$data['available']` was retrieved. Previous code holds logic based on `$instantPurchaseOption` and fill the data (it does not make a decision based on a result, it produce a result from an input).

app/code/Magento/InstantPurchase/Model/InstantPurchaseChooser.php
```diff
@@ -11,7 +11,7 @@
 use Magento\InstantPurchase\Model\ShippingAddressChoose\ShippingAddressChooserInterface;
 use Magento\InstantPurchase\Model\BillingAddressChoose\BillingAddressChooserInterface;
 use Magento\InstantPurchase\Model\ShippingMethodChoose\ShippingMethodChooserInterface;
-use Magento\Store\Model\Store;
+use Magento\Store\Api\Data\StoreInterface;
```
vkublytskyi: Models (Store, Product, Quote) is not used in Instant Purchase by mistake. This was done by intent as models have more methods than API data interfaces. If data interfaces will be used instead of models then at some parts  Instant Purchase will call unknown methods which will work only if model is passed as data interface implementation.

Magento 2 indeed encourage usage of API data interfaces but programming based on contract have more importance. Instant Purchase will not work if someone substitutes data interface preference with own implementation which will not extend corresponding model.
Unfortunately, implementation of  Instant Purchase also relies on the current behavior of Quote module (e.g. save of quote in the middle of order placement at the `\Magento\InstantPurchase\Model\PlaceOrder::placeOrder`).

Thereby, dependency on models (which also marked as `@api`) is actually declaration on what code really depends.

Please revert back dependency on model classes.

app/code/Magento/InstantPurchase/Test/Unit/Block/ButtonTest.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+
+    /**
+     * @param bool $status
+     * @dataProvider isEnabledDataProvider
+     */
+    public function testIsEnabled($status)
+    {
+        $this->configMock->expects($this->once())
```
vkublytskyi: Please don't use `expects` in mocks as this, actually, not test a code but just "freeze" current implementation.

The only valid usage of `excpects` in unit tests are:
- you know that some code has side effects (which is bad practice)
- you fix a performance issue and would like to have a test that heavy method is not invoked more than required to make a job done.

In most cases, method call should return same result for same input arguments and simple method call is not a performance issue (avoid premature optimization). This comes to conclusion that PHPUnit mocks in most cases should mostly map output to input (it's actually not a mock but stub).

There is a great article by Uncle Bob on types of test doubles: https://8thlight.com/blog/uncle-bob/2014/05/14/TheLittleMocker.html.

app/code/Magento/InstantPurchase/Test/Unit/Controller/Button/PlaceOrderTest.php
```diff
@@ -0,0 +1,344 @@
...
+    {
+        $this->requestMock->expects($this->once())
+            ->method('getParams')
+            ->willReturn([]);
+
+        $this->resultMock->expects($this->once())
+            ->method('setData')
+            ->with(['response' => self::GENERIC_ERROR_MESSAGE]);
+        
+        $this->messageManagerMock->expects($this->once())
```
vkublytskyi: Do not test that you call some particular method. With such code you freeze implementation. Instead, test that message manager has a message after  controller execution (this is more relevant to what you are expecting: you want that message is in messages list, you don't care how it appears there). Fake message manager may be used for this purpose.

app/code/Magento/InstantPurchase/Test/Unit/Controller/Button/PlaceOrderTest.php
```diff
@@ -0,0 +1,344 @@
...
+            ->method('addErrorMessage')
+            ->with(self::GENERIC_ERROR_MESSAGE);
+        
+        $this->assertInstanceOf(
```
vkublytskyi: In addition to result type verification, a content of the response should be checked as well. Having JSON result is not enough to be sure that response of controller is correct.

app/code/Magento/InstantPurchase/Test/Unit/Controller/Button/PlaceOrderTest.php
```diff
@@ -0,0 +1,344 @@
...
+            ->method('getIncrementId')
+            ->willReturn($orderIncrementId);
+
+        $successMessage = sprintf('Your order number is: %s.', $orderIncrementId);
+        $this->resultMock->expects($this->once())
+            ->method('setData')
+            ->with(['response' => $successMessage]);
+
+        $this->messageManagerMock->expects($this->once())
+            ->method('addSuccessMessage')
```
vkublytskyi: Probably, we should not verify exact success message. The thing that we really want to verify is that success message *contains* order number.

app/code/Magento/InstantPurchase/Test/Unit/CustomerData/InstantPurchaseTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $this->instantPurchaseMock->expects($this->once())
+            ->method('getOption')
+            ->with($this->storeMock, $this->customerMock)
+            ->willReturn($this->instantPurchaseOptionMock);
+
+        $this->instantPurchaseOptionMock->expects($this->once())
+            ->method('isAvailable')
+            ->willReturn(true);
+
+        $this->assertTrue($this->instantPurchase->getSectionData()['available']);
```
vkublytskyi: The *format* of customer data segment is an API on which depends client-side application. It change would be a backward incompatible. And this is actually a main risk during this class modification. Unit test should verify a *format* of output data.

Instant Purchase module have a good definition what is input of the method and what is output. Try to build most assertions on methods output and avoid use of mock test double (do not assert that some method is invoked). You may use fake objects and stubs to isolate tested class and provide some basic implementation.

app/code/Magento/InstantPurchase/Test/Unit/Block/ButtonTest.php
```diff
@@ -0,0 +1,143 @@
...
+    private $storeMock;
+
+    /**
+     * @var UrlInterface|MockObject
+     */
+    private $urlMock;
+
+    protected function setUp()
+    {
+        $objectManager = new ObjectManager($this);
```
vkublytskyi: Try to avoid usage of object manager in unit tests. Unit tests should be independent of infrastructure as much as possible.

app/code/Magento/InstantPurchase/Test/Unit/Model/InstantPurchaseChooserTest.php
```diff
@@ -0,0 +1,99 @@
...
+        );
+    }
+
+    public function testGetOption()
+    {
+        $this->configMock->expects($this->once())
+            ->method('isModuleEnabled')
+            ->willReturn(true);
+
+        $this->optionFactoryMock->expects($this->once())
```
vkublytskyi: This test does not verify that option object is created and fulfilled correctly.  It just verify that if IP enabled for store it will try to instantiate option. Please add assertions on the method result.

app/code/Magento/InstantPurchase/Test/Unit/Model/InstantPurchaseOptionTest.php
```diff
@@ -0,0 +1,186 @@
...
+            ->getMock();
+    }
+
+    /**
+     * @param bool $available
+     * @dataProvider isAvailableDataProvider
+     */
+    public function testIsAvailable($available)
+    {
+        $option = $this->objectManager->getObject(
```
vkublytskyi: You may use `new` operator in unit test. (You want to verify particular class, no need to manage objects or replace them with interceptors, right?)

app/code/Magento/InstantPurchase/Test/Unit/Model/InstantPurchaseOptionTest.php
```diff
@@ -0,0 +1,186 @@
...
+    public function testGetShippingAddressException()
+    {
+        $option = $this->objectManager->getObject(
+            InstantPurchaseOption::class,
+            [
+                'shippingAddress' => null,
+            ]
+        );
+
+        $this->assertInstanceOf(Address::class, $option->getShippingAddress());
```
vkublytskyi: Dead code. `assertInstanceOf` will be never invoked as exception will be thrown before.
It is enough to have:
```php
/**
 * @expectedException \Magento\Framework\Exception\LocalizedException
 */
public function testGetShippingAddressException()
{
     $option = new InstantPurchaseOption(null, null, null, null);
     $option->getShippingAddress();
}
```

app/code/Magento/InstantPurchase/Test/Unit/Model/InstantPurchaseOptionTest.php
```diff
@@ -0,0 +1,186 @@
...
+                'billingAddress' => $this->addressMock,
+                'shippingMethod' => $this->shippingMethodMock,
+            ]
+        );
+
+        $this->shippingMethodMock->expects($this->once())
+            ->method('getAvailable')
+            ->willReturn($available);
+
+        $this->assertEquals($available, $option->isAvailable());
```
vkublytskyi: This test does not verify correct availability logic (actually, only logic in the class) without addresses and payment token. Suppose data provider will be a good solution to implement such tests.

app/code/Magento/InstantPurchase/Test/Unit/Model/PlaceOrderTest.php
```diff
@@ -0,0 +1,274 @@
...
+                'quoteCreation' => $this->quoteCreateMock,
+                'quoteFilling' => $this->quoteFillingMock,
+                'shippingConfiguration' => $this->shippingConfigurationMock,
+                'paymentConfiguration' => $this->paymentConfigurationMock,
+                'purchase' => $this->purchaseMock,
+            ]
+        );
+    }
+
+    public function testPlaceOrder()
```
vkublytskyi: This class is actually script for invoke other classes in exact order. It has no own logic and totally rely on current behavior of other modules. Unit test will not provide usefull verification here. Valid data and correct work verification may be provided only by integration tests here.
BTW. Fact that implementation have to relay of internal qoute and payment behavior was a reason to not introduce interface but implement order placement with a class.

vkublytskyi (changes requested):
@pmclain Thank you for working on test coverage for Instant Purchase module. I have several high-level remarks according to your implementation:
1. Models are used at some places instead of data interfaces as implementation requires some methods that are defined only in models but not in interfaces and depends on current models behavior. Explicit dependency on modules declare this dependency. It is not good but ok as all used modules marked with an `@api` annotation
2. Instant Purchase uses slightly defined input and output for most of the methods. Unit tests should do assertions on output, not verify invocation of tested class collaborators (use fake and stubs test doubles instead of mocks).
3. Refactoring should increase code readability for humans. If some code is hard to test with PHPunit infrastructure the only possible refactoring is to extract part of logic to another class and cover it with another set of unit tests.  Rewriting code to reduce amount of methods that should be mocked is not a reasonable approach as it complicates implementation of class under test and make code less object oriented.

vkublytskyi:
@pmclain As I mentioned in review Instant Purchase code sometimes depends on the internal behavior of Quote and Payment modules. That was necessary to make it work. Unit tests are not good tool to verify such implementation so you may be also interested in extending integration tests coverage: https://github.com/magento/magento2/tree/2.2-develop/dev/tests/integration/testsuite/Magento/InstantPurchase

app/code/Magento/InstantPurchase/Test/Unit/Model/PlaceOrderTest.php
pmclain: Shouldn't there at least be a test confirming quotes are made inactive when catching exceptions? Failing to do so would replace a customer's current cart with the failed instant purchase cart on the frontend.

vkublytskyi: Yes, fact that quote should be deactivated in case of failure should be verified. But it better to place in integration test as well (simple verification that `setIsActive` is invoked with `false` is not enough, what we really want to test is if order placement failed then after quote object load from repository it will be not active). For simplicity in such case (side effect of the code) we may verify that quote repository `save` method is called and quote object is not active already at this time.

Ideally, we should not create/save quote until order placement but this may lead to failures in the shipping configuration. So as this code is for integration it's better to test with integration test.

pmclain:
@vkublytskyi Thanks for the guidance! I've updated the tests based on the feedback provided. The test for the place order controller has been removed, hopefully I'll find a time block to revisit it soon.

PR #12906, opened 2017-12-28
[Fix] store switcher not showing all stores if the height is bigger than 250px

jalogut:
### Description

Store view switcher in Admin does not display all stores if the height of them is bigger than 250px. Admin cannot select those stores to edit their configuration

Current Bug:
current_bug

PR fix:
fixed

### Manual testing scenarios

1. Create 3 Websites with 3 stores views each
2. Save
3. Click over the "Store View" Admin switcher. 

**Expected**: All stores are displayed
**Actual result**: Not all are displayed. Only the first ones up to 250px height

### Contribution checklist

### DIFF
--- a/app/design/adminhtml/Magento/backend/Magento_Backend/web/css/source/module/main/actions-bar/_store-switcher.less
+++ b/app/design/adminhtml/Magento/backend/Magento_Backend/web/css/source/module/main/actions-bar/_store-switcher.less
@@ -39,7 +39,6 @@
     .dropdown-menu {
         left: 0;
         margin-top: .5em;
-        max-height: 250px;
         overflow-y: auto;
         padding-top: .25em;
 


fooman:
@jalogut I just checked this with 9 store views. I can reproduce that the window only shows some entries. However the dropdown field for me is scrollable:

many-stores

Does that work for you as well?
 
While I agree that this field can probably be made a bit bigger I am not sure if making this completely unlimited (just imagining someone with 100 store views).

avoelkl:
I was wondering about this also quite some time ago and then found out the field is scroll-able. I agree it's not very user friendly @jalogut. Maybe showing the scrollbar (if available/needed) would help?

jalogut:
(facepalm) @fooman thank you very much for the info. 
@avoelkl I agree with that. I created a new PR https://github.com/magento/magento2/pull/12931

PR #12909, opened 2017-12-28
Don't escape params during exportTypes generation

bordeo:
Disable escape params on export grid url generation

### Description
On Ordered Product Report if the period is "day" or "month" export want work because the url generated for export is urlencoded twice.

1. magento/magento2#10992: Ordered Products (Products Sold) Report CSV / XML Export Broken

### Manual testing scenarios
1. Make some orders on different day and months
2. Go to Reports -> Products -> Ordered
3. Chose "Show By:" month or day
4. Press Refresh
5. Press export
6. Now you can see the csv and xml not empty


fooman:
@bordeo thanks for your pull request. It currently sounds like the issue you have encountered is fairly specific to the products sold report. However the proposed change here would impact all exports that use the same widget. Can you confirm that this issue is indeed affecting other exports as well? If not it might be better to apply this fix only for the mentioned report export.

sidolov:
@bordeo , I am closing this PR now due to inactivity.
Thank you for collaboration!

PR #12929, opened 2017-12-29
Issues #10559 - Extend swatch using mixins (M2.2)

srenon:

magento/magento2#10559: Extending swatch functionality using javascript mixins does not work in Safari and MS Edge

### Manual testing scenarios

1. See testing instruction in magento/magento2#10559

### DIFF
--- a/app/code/Magento/Catalog/Pricing/Render/FinalPriceBox.php
+++ b/app/code/Magento/Catalog/Pricing/Render/FinalPriceBox.php
@@ -115,7 +115,7 @@ class FinalPriceBox extends BasePriceBox
     {
         return '<div class="price-box ' . $this->getData('css_classes') . '" ' .
             'data-role="priceBox" ' .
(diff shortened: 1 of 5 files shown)


nicobatty:
Thanks a lot @srenon for your fix!

By the way, it seems like this file have the same issue on the product view form for the price-box mixin support:

https://github.com/magento/magento2/blob/e177f452801bd2e4213fead60556c43464dc0e79/app/code/Magento/Catalog/view/frontend/templates/product/view/form.phtml#L42-L57

I applied the same method you did for the price-box on the listing which seems to work.

For those wondering, I replaced it with this:

```php
<script type="text/x-magento-init">
    {
        "[data-role=priceBox][data-price-box=product-id-<?= $block->escapeHtml($_product->getId()) ?>]": {
            "priceBox": {
                "priceConfig": <?= /* @escapeNotVerified */ $block->getJsonConfig() ?>
            }
        }
    }
</script>
```

PR #12935, opened 2017-12-31
Add Ability To Separate Frontend / Adminhtml in New Relic

mpchadwick:
### Description

Adds a new setting which, when enabled, reports frontend and adminhtml as separate apps to New Relic.

frontend and adminhtml are **in addition to** the "main" app which includes both (for backwards compatibility). The user needs to set "New Relic Application Name" to use this feature as the area is appended to the appname setting, separated by and underscore.

image

image

Useful as slow transactions in admin activity can cause false positives New Relic alerts / skew average response time metrics. Using this feature allows monitoring and alerting on only frontend traffic by configuring alerts for just the frontend New Relic app.

### Manual testing scenarios

1. Pull in changes to a testing environment with New Relic agent installed
2. Ensure "Enable New Relic Intergration" is set to yes
3. Ensure "New Relic Application Name" is set
4. Turn on "Send Adminhtml and Frontend as Separate Apps Setting"
5. Navigate the frontend and backend of Magento
6. Login to New Relic and observe 3 apps ("main" app, "adminhtml" app, and "frontend" app).

### Contribution checklist

### DIFF
--- a/app/code/Magento/NewRelicReporting/Model/NewRelicWrapper.php
+++ b/app/code/Magento/NewRelicReporting/Model/NewRelicWrapper.php
@@ -42,6 +42,19 @@ class NewRelicWrapper
     }
 
     /**
+     * Wrapper for 'newrelic_set_appname'
+     *
+     * @param string $appName
+     * @return void
+     */
+    public function setAppName($appName)
+    {
+        if (extension_loaded('newrelic')) {
+            newrelic_set_appname($appName);
+        }
+    }
+
+    /**
      * Checks whether newrelic-php5 agent is installed
      *
      * @return bool
--- a/app/code/Magento/NewRelicReporting/Model/NewRelicWrapper.php
+++ b/app/code/Magento/NewRelicReporting/Model/NewRelicWrapper.php
@@ -47,7 +47,7 @@ class NewRelicWrapper
      * @param string $appName
      * @return void
      */
-    public function setAppName($appName)
+    public function setAppName(string $appName)
     {
         if (extension_loaded('newrelic')) {
             newrelic_set_appname($appName);
--- a/app/code/Magento/NewRelicReporting/Plugin/StatePlugin.php
+++ b/app/code/Magento/NewRelicReporting/Plugin/StatePlugin.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\NewRelicReporting\Plugin;
+
+use Magento\Framework\App\State;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\NewRelicReporting\Model\Config;
+use Magento\NewRelicReporting\Model\NewRelicWrapper;
+
+class StatePlugin
+{
+    /**
+     * @var Config
+     */
+    private $config;
+
+    /**
+     * @var NewRelicWrapper
+     */
+    private $newRelicWrapper;
+
+    /**
+     * @param Config $config
+     * @param NewRelicWrapper $newRelicWrapper
+     */
+    public function __construct(
+        Config $config,
+        NewRelicWrapper $newRelicWrapper
+    ) {
+        $this->config = $config;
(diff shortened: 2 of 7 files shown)


mpchadwick:
Just to document, based on testing done in response to [this conversation on Twitter](https://twitter.com/maxpchadwick/status/947478645647597570), this will also catch other area codes such as webapi_rest.

image

ihor-sviziev:
I reopened this PR in order to restart build in travis

app/code/Magento/NewRelicReporting/Model/NewRelicWrapper.php
```diff
@@ -41,6 +41,19 @@ public function reportError($exception)
         }
     }
 
+    /**
+     * Wrapper for 'newrelic_set_appname'
+     *
+     * @param string $appName
+     * @return void
+     */
+    public function setAppName($appName)
```
ihor-sviziev: I'd like to define variable type to "string" there, we're not supporting php 5.6 anymore

app/code/Magento/NewRelicReporting/Plugin/StatePlugin.php
```diff
@@ -0,0 +1,83 @@
...
+    public function afterSetAreaCode(State $state, $result)
+    {
+        if (!$this->shouldSetAppName()) {
+            return;
+        }
+
+        try {
+            $this->newRelicWrapper->setAppName($this->appName($state));
+        } catch (LocalizedException $e) {
+            return;
```
ihor-sviziev: 1. Why should we ignore such exception?
2. Probably we should have `return $result` here

app/code/Magento/NewRelicReporting/Plugin/StatePlugin.php
```diff
@@ -0,0 +1,83 @@
...
+     * @param State $subject
+     * @param null $result
+     * @return void
+     *
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function afterSetAreaCode(State $state, $result)
+    {
+        if (!$this->shouldSetAppName()) {
+            return;
```
ihor-sviziev: Probably we should have `return $result` here

app/code/Magento/NewRelicReporting/Plugin/StatePlugin.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        try {
+            $this->newRelicWrapper->setAppName($this->appName($state));
+        } catch (LocalizedException $e) {
+            return;
+        }
+    }
+
+    private function appName(State $state)
```
ihor-sviziev: Would be better to rename it to "setAppName"

ihor-sviziev (changes requested):
In general looks good, but here is some small issues. Can you also cover your changes with tests?

app/code/Magento/NewRelicReporting/Plugin/StatePlugin.php
mpchadwick: This is a getter. Would you prefer `getAppName`?

ihor-sviziev:
Hi @mpchadwick,
Your changes looks good, could you cover them with tests in order to prevent regression in this place?

mpchadwick:
@ihor-sviziev I will add tests. May take a little while. Still trying to get the integration tests to run in my local environment.

ihor-sviziev (changes requested):
@mpchadwick please review static test, they are failing currently

mpchadwick:
@ihor-sviziev fixed

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-1860](https://jira.corp.magento.com/browse/ENGCOM-1860) has been created to process this Pull Request

magento-engcom-team:
Hi @mpchadwick. Thank you for your contribution. 

ihor-sviziev:
Hi @mpchadwick,
I don't have enough time for fixing all the issues that appeared in https://github.com/magento/magento2/pull/15947. Could you create own forwardport with requested changes from https://github.com/magento/magento2/pull/15947?

Issue #5948, opened 2016-08-02
Magento 2 configurable product selection stock issue


dapdx:
Setting up configurable products and using the text swatch option to select shirt sizes. The issue is the out of stock shirt text swatches are still showing up, clickable, and have no styling to show that they are out of stock. I'd like the swatch to still show up, but not be clickable and appear faded out or something so you know they are out of stock. Any help?

I'm not seeing any class for out of stock being passed through that I can style with CSS. 

This worked fine in Magento 1, where on drop downs the out of stock simple products were not selectable, but in Magento 2, both on drop downs and text swatches they still show up and you can click on them, but when added to the cart you get the "product not in stock" error.

veloraven:
@dapdx thank you for your report.
Please provide description according to the template: with steps, actual result and expected result. 
Please also provide the used version. If the problem is actual for a specific tag, please, specify it and be sure that the latest update was used.

dapdx:
Magento 2.1
PHP 7

Steps to Reproduce
1. Create configurable product (in this case example t-shirt)
2. Create linked simple prodcuts (sizes S, Med, Large, XL)
3. Set Size attribute to Text Swatch
4. Set Medium size to out of stock, others are in-stock

Expected Result
1. When Medium size is out of stock, on the front end it should have a class assigned to it so you can style it with CSS to make it not available for purchase or signify that it is out of stock.

Actual Result
1. On front end all (S, M, L, XL) text swatches appear, none have a strike through, or greyed out or missing. There is no "out of stock" class applied to the swatch so you can style it with CSS.

JohnWean:
I also have the same problems and then I have to use an extension from the third part to solve it although this issue can be overcome quite easily in Magento 1 as you have said. 
I have been using this Magento 2 extension - [Configurable product grid table view](http://bsscommerce.com/magento-configurable-product-grid-table-view-for-magento-2.html)  bought from a Magento provider to make out of stock product still visible but unavailable to choose quantity in the box . The quantity box is faded and cannot fill in.

andidhouse:
I can confirm this on a fresh 2.1 system. 

Really bad bug as we have lots of sizes :-/

duhon:
@dapdx Thanks for reporting this issue.
We've created internal ticket MAGETWO-56480  to fix it.

otg-chris:
This bug came up when I updated a site from 2.02 to 2.1. Before the out-of-stock options were not shown. A quick fix to hide them (probably won't work with multiple options)
[...]

Issue #10454, opened 2017-08-07
CE 2.1.7 Attributes - Display Out of Stock Products


stroblee:
### Preconditions

1. Magento CE 2.1.7
2. Apache 2.4.27
3. PHP 7..0.22
4. Varnish 4.1.8

### Steps to reproduce

1. Install Magento 2.1.7 CE Sample Data
2. Admin > Stores > Configuration > Catalog > Inventory > Stock Options > Display Out of Stock Products = Yes.
3. CLI php bin/magento indexer:reindex
4. Admin > Products > Catalog > Breathe-Easy Tank-XS-Puple, Breathe-Easy Tank-XS-White, Breathe-Easy Tank-XS-Orange. Set either Quantity = 0 or Stock Status = Out of Stock.
5. CLI php bin/magento cache:clean

### Expected result

1. Frontend product - Breathe-Easy Tank. The swatch attribute Size XS' should display with class disabled (red strike through) for all Simple products part of Configurable product.

### Actual result

1. Frontend product - Breathe-Easy Tank. The swatch attribute Size XS not displayed at all.

### Other Information
1. If one Simple product from Configurable options, i.e. Breathe-Easy Tank-XS-Puple, Breathe-Easy Tank-XS-White, Breathe-Easy Tank-XS-Orange has either Quantity => 1 or Stock Status = In Stock then the remaining out of stock products will display the swatch attribute 'Size XS' with class disabled (red strike through) for all Simple products part of Configurable product.

dankocherga:
The issue seems to be in a way how Magento retrieves products for showing as a configurable options.

In the [Magento\ConfigurableProduct\Block\Product\View\Type\Configurable:: getAllowProducts](https://github.com/magento/magento2/blob/2.1/app/code/Magento/ConfigurableProduct/Block/Product/View/Type/Configurable.php#L129-L139) method it checks for the `getSkipSaleableCheck` flag which is `false` and only set to `true` during admin checkout.

Since the flag is `false`, it goes down to the [Magento\ConfigurableProduct\Model\Product\Type\Configurable::getSalableUsedProducts](https://github.com/magento/magento2/blob/2.1/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php#L1348-L1360) where stock status is **always** validated.

In other words, "Display Out of Stock Products" configuration does not make any difference in building configurable product options.

torlokken:
Any update on this? Major issue for us.

[...]

PR #12936, opened 2018-01-01
FIX: out-of-stock options for configurable product visible on frontend as sellable

coderimus:
This PR is for configurable product out-of-stock option fix with several improvements for generally used methods. Please, check **Description** and **Code changes explanation** sections for more detail information.

### Description
If one of the configurable product options will be out-of-stock but has status enabled, it will be visible on the frontend (product page with swatches). 
But, when a customer selects it and tries to add to the cart the next message will be shown: "You need to choose options for your item." And item will not be added to the cart. 
The main problem here that the out-of-stock and not salable option is shown as normal one without any notification before adding to the cart and when a customer tries to add it to the cart this generates useless load on the system.

The most relevant issue is https://github.com/magento/magento2/issues/5948 
Despite the fact that this issue was closed, it still can be reproduced on the Magento2 2.2.2. 

### Manual testing scenarios
1. Select configurable product using swatches
2. Choose one random option (simple product) and change it's qty to 0. You will get status "Out-of-stock"
3. Go to the product page, reload it
4. Select option with out-of-stock status and press "Add to cart button"
5. The page will be reloaded and error message will be shown

Accepted criteria: out-of-stock option should be shown as not-selectable and there is no ability to add it to the cart.

### Code changes explanation
1.  `Magento\ConfigurableProduct\Model\Product\Type\Configurable::loadUsedProducts(\Magento\Catalog\Model\Product $product, $cacheKey, $salableOnly = false)` method had 2 issues:

- the `$salableOnly` variable never used since 100.2.0 because of deprecated method `getSalableUsedProducts` and, as a result, the stock status filter was not added to the collection object
- even if the `$salableOnly` filter will be `true` and `$this->salableProcessor->process($collection)` will be executed, it will not get any effect because items were previously loaded in `$collection = $this->getConfiguredUsedProductCollection($product);`   It means that collection will not be updated with the real stock status data. 
Items (`$this->_items`) were loaded before by `$this->getItems()` method in `$collection->addMediaGalleryData()` and `$collection->addTierPriceData()`. 
This is why I have moved `$collection->addMediaGalleryData()` and `$collection->addTierPriceData()` to another method and execute them after all other collection modifications. 

Also, I removed the `if($salableOnly) {...}` condition because there is no need to skip it. With other provided changes product collection will contain all products with correct stock statuses. There will not be skipped products which is equal to current behaviour.

The `$collection->setFlag('has_stock_status_filter',false);` line was added to add stock status filter.

2. `Magento\ConfigurableProduct\Model\Product\Type\Collection\SalableProcessor::process(Collection $collection, $isFilterInStock = true)` has 3 improvements:

- This method now has second param `$isFilterInStock` with default value equals to `true`. It makes it more flexible because now you can just add stock status to collection or filter it by status equals to 1 (in stock)

- `!$collection->hasFlag($stockFlag) || !$collection->getFlag($stockFlag)` this change allow develoeprs to check what value has collcetion flag. This was added because the `hasFlag()' method checks only if current key exists - `array_key_exists($flag, $this->_flags);`

- `$stockStatusResource->addStockDataToCollection($collection, $isFilterInStock);` now uses `$isFilterInStock` value as parametr

3. `Magento\CatalogInventory\Model\ResourceModel\Stock\Status` has next changes

- `addStockStatusToSelect`, `addStockDataToCollection`, `addIsInStockFilterToCollection` now have `$websiteId` with `null` as default value and `$this->getWebsiteId($websiteId)` method for website condition check

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
+++ b/app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
@@ -5,8 +5,8 @@
  */
 namespace Magento\CatalogInventory\Model\ResourceModel\Stock;
 
-use Magento\CatalogInventory\Model\Stock;
 use Magento\CatalogInventory\Api\StockConfigurationInterface;
+use Magento\CatalogInventory\Model\Stock;
 use Magento\Framework\App\ObjectManager;
 
 /**
@@ -46,19 +46,23 @@ class Status extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
      * @param \Magento\Store\Model\WebsiteFactory $websiteFactory
      * @param \Magento\Eav\Model\Config $eavConfig
      * @param string $connectionName
+     * @param \Magento\CatalogInventory\Api\StockConfigurationInterface $stockConfiguration
      */
     public function __construct(
         \Magento\Framework\Model\ResourceModel\Db\Context $context,
         \Magento\Store\Model\StoreManagerInterface $storeManager,
         \Magento\Store\Model\WebsiteFactory $websiteFactory,
         \Magento\Eav\Model\Config $eavConfig,
-        $connectionName = null
+        $connectionName = null,
+        $stockConfiguration = null
     ) {
         parent::__construct($context, $connectionName);
 
         $this->_storeManager = $storeManager;
         $this->_websiteFactory = $websiteFactory;
         $this->eavConfig = $eavConfig;
+        $this->stockConfiguration = $stockConfiguration ?: ObjectManager::getInstance()
+            ->get(StockConfigurationInterface::class);
     }
 
     /**
@@ -204,7 +208,7 @@ class Status extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
      */
     public function addStockStatusToSelect(\Magento\Framework\DB\Select $select, \Magento\Store\Model\Website $website)
     {
-        $websiteId = $this->getStockConfiguration()->getDefaultScopeId();
+        $websiteId = $this->getWebsiteId($website->getId());
         $select->joinLeft(
             ['stock_status' => $this->getMainTable()],
             'e.entity_id = stock_status.product_id AND stock_status.website_id=' . $websiteId,
@@ -221,7 +225,7 @@ class Status extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
      */
     public function addStockDataToCollection($collection, $isFilterInStock)
     {
-        $websiteId = $this->getStockConfiguration()->getDefaultScopeId();
+        $websiteId = $this->getWebsiteId();
         $joinCondition = $this->getConnection()->quoteInto(
             'e.entity_id = stock_status_index.product_id' . ' AND stock_status_index.website_id = ?',
             $websiteId
@@ -255,7 +259,7 @@ class Status extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
      */
     public function addIsInStockFilterToCollection($collection)
     {
-        $websiteId = $this->getStockConfiguration()->getDefaultScopeId();
+        $websiteId = $this->getWebsiteId();
         $joinCondition = $this->getConnection()->quoteInto(
             'e.entity_id = stock_status_index.product_id' . ' AND stock_status_index.website_id = ?',
             $websiteId
@@ -275,6 +279,19 @@ class Status extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
             Stock\Status::STATUS_IN_STOCK
         );
         return $this;
+    }
+
+    /**
+     * @param \Magento\Store\Model\Website $websiteId
+     * @return int
+     */
+    private function getWebsiteId($websiteId = null)
+    {
+        if (null === $websiteId) {
+            $websiteId = $this->stockConfiguration->getDefaultScopeId();
+        }
+
+        return $websiteId;
     }
 
     /**
@@ -334,19 +351,5 @@ class Status extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
             }
         }
         return $statuses;
-    }
-
-    /**
-     * @return StockConfigurationInterface
-     *
-     * @deprecated 100.1.0
-     */
-    private function getStockConfiguration()
-    {
-        if ($this->stockConfiguration === null) {
-            $this->stockConfiguration = \Magento\Framework\App\ObjectManager::getInstance()
-                ->get(\Magento\CatalogInventory\Api\StockConfigurationInterface::class);
-        }
-        return $this->stockConfiguration;
     }
 }
--- a/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php
+++ b/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php
@@ -10,11 +10,11 @@ use Magento\Catalog\Api\Data\ProductAttributeInterface;
 use Magento\Catalog\Api\Data\ProductInterface;
 use Magento\Catalog\Api\Data\ProductInterfaceFactory;
 use Magento\Catalog\Api\ProductRepositoryInterface;
-use Magento\ConfigurableProduct\Model\Product\Type\Collection\SalableProcessor;
 use Magento\Catalog\Model\Config;
+use Magento\Catalog\Model\Product\Gallery\ReadHandler as GalleryReadHandler;
+use Magento\ConfigurableProduct\Model\Product\Type\Collection\SalableProcessor;
 use Magento\Framework\App\ObjectManager;
 use Magento\Framework\EntityManager\MetadataPool;
-use Magento\Catalog\Model\Product\Gallery\ReadHandler as GalleryReadHandler;
 
 /**
  * Configurable product type implementation
@@ -267,7 +267,6 @@ class Configurable extends \Magento\Catalog\Model\Product\Type\AbstractType
             $productRepository,
             $serializer
         );
-
     }
 
     /**
@@ -1277,6 +1276,8 @@ class Configurable extends \Magento\Catalog\Model\Product\Type\AbstractType
      * Load collection on sub-products for specified configurable product
      *
      * Load collection of sub-products, apply result to specified configurable product and store result to cache
+     * Please note $salableOnly parameter is used for backwards compatibility because of deprecated method
+     * getSalableUsedProducts
      * Number of loaded sub-products depends on $salableOnly parameter
      * $salableOnly = true - result array contains only salable sub-products
      * $salableOnly = false - result array contains all sub-products
@@ -1293,7 +1294,7 @@ class Configurable extends \Magento\Catalog\Model\Product\Type\AbstractType
         if (!$product->hasData($dataFieldName)) {
             $usedProducts = $this->readUsedProductsCacheData($cacheKey);
             if ($usedProducts === null) {
-                $collection = $this->getConfiguredUsedProductCollection($product);
+                $collection = $this->getConfiguredUsedProductCollection($product, false);
                 if ($salableOnly) {
                     $collection = $this->salableProcessor->process($collection);
                 }
@@ -1387,13 +1388,18 @@ class Configurable extends \Magento\Catalog\Model\Product\Type\AbstractType
      * Retrieve related products collection with additional configuration
      *
      * @param \Magento\Catalog\Model\Product $product
+     * @param bool $skipStockFilter
      * @return \Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Product\Collection
      */
-    private function getConfiguredUsedProductCollection(\Magento\Catalog\Model\Product $product)
-    {
+    private function getConfiguredUsedProductCollection(
+        \Magento\Catalog\Model\Product $product,
+        $skipStockFilter = true
+    ) {
         $collection = $this->getUsedProductCollection($product);
+        if ($skipStockFilter) {
+            $collection->setFlag('has_stock_status_filter', true);
+        }
         $collection
-            ->setFlag('has_stock_status_filter', true)
             ->addAttributeToSelect($this->getCatalogConfig()->getProductAttributes())
             ->addFilterByRequiredOptions()
             ->setStoreId($product->getStoreId());
--- a/app/code/Magento/ConfigurableProduct/Test/Unit/Model/Product/Type/ConfigurableTest.php
+++ b/app/code/Magento/ConfigurableProduct/Test/Unit/Model/Product/Type/ConfigurableTest.php
@@ -8,20 +8,20 @@ namespace Magento\ConfigurableProduct\Test\Unit\Model\Product\Type;
 use Magento\Catalog\Api\Data\ProductExtensionInterface;
 use Magento\Catalog\Api\Data\ProductInterface;
 use Magento\Catalog\Model\Config;
-use Magento\ConfigurableProduct\Model\Product\Type\Configurable;
-use Magento\Framework\EntityManager\EntityMetadata;
-use Magento\Framework\EntityManager\MetadataPool;
-use Magento\Customer\Model\Session;
-use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Attribute\CollectionFactory;
-use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Attribute\Collection;
 use Magento\Catalog\Model\Product\Configuration\Item\Option\OptionInterface;
-use Magento\Framework\Api\ExtensionAttribute\JoinProcessorInterface;
+use Magento\ConfigurableProduct\Model\Product\Type\Collection\SalableProcessor;
+use Magento\ConfigurableProduct\Model\Product\Type\Configurable;
 use Magento\ConfigurableProduct\Model\Product\Type\Configurable\AttributeFactory;
+use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Attribute\Collection;
+use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Attribute\CollectionFactory;
+use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Product\Collection as ProductCollection;
 use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\ConfigurableFactory;
+use Magento\Customer\Model\Session;
 use Magento\Eav\Model\Entity\Attribute\Frontend\AbstractFrontend;
 use Magento\Eav\Model\Entity\Attribute\Source\AbstractSource;
-use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Product\Collection as ProductCollection;
-use Magento\ConfigurableProduct\Model\Product\Type\Collection\SalableProcessor;
+use Magento\Framework\Api\ExtensionAttribute\JoinProcessorInterface;
+use Magento\Framework\EntityManager\EntityMetadata;
+use Magento\Framework\EntityManager\MetadataPool;
 
 /**
  * @SuppressWarnings(PHPMD.LongVariable)


coderimus:
Hello dear @magento-engcom-team !
Can you please review this PR when you will have a time :) Just worry that it can be lost under the tons of newly pushed changes)
Looking forward to your response,
Alex

dmanners:
Hi @Coderimus thank you for this pull request. We aim to get around this pull request during the next few days and will let you know if we need anything else from your side.

coderimus:
Hi @adrian-martinez-interactiv4 !
Thank you for taking this PR into the review.  Sorry if I interrupt you but if it is possible can you please update me with the possible date when you will check it? :)
Best regards,
Alex

adrian-martinez-interactiv4:
Hi @Coderimus , I'll check it tomorrow, and update you as soon as possible. I apologize for the delay.

adrian-martinez-interactiv4:
Hi again, @Coderimus , **I've trying to reproduce the issue with no luck**, I've tried with cache enabled and disabled, indexes in Update On Save mode, even with config setting Display Out of Stock Products set to Yes and No values, always with the same result: when I update the simple product to remove its stock, it dissapears from product page; when I restore its stock, it shows up again.

I'm using attribute swatches, as marked in the testing scenario. Following is the process I've been doing, with the result, please check if I am missing something:

- Initial status:
captura de pantalla 2018-01-18 a las 19 57 55

captura de pantalla 2018-01-18 a las 19 58 40

captura de pantalla 2018-01-18 a las 19 59 16

- Then, removing the stock for that product:
captura de pantalla 2018-01-18 a las 20 00 48

- Without erasing cache, only reloading product page, option disappears:
captura de pantalla 2018-01-18 a las 20 02 08

I can't reproduce the issue, am I missing something?

coderimus:
@adrian-martinez-interactiv4 thank you for your time spent on this review and such detail description. I think that together we found one more aspect of this problem. This issue can be reproduced **for a product having more than one item for options and options qty should be bigger than 1 and using swatches**. In your example, you have 1 option with 2 products. If you would have 2 options with several products you could reproduce it. 
I tested and reproduced this issue with the latest release 2.2 with sample data. Please, check screenshots and description:

1. Choose a configurable product with 2 options with several products. 
image

3. Choose any simple product, set it's qty to 0 and save. I did this without advanced inventory. Just put 0 into qty input and then press save. You will get status "Out of stock"
image

4. Go to the product page on the frontend and you will see that the option is still available and can be select.
**Screen #1 see black XS:**
image

**Screen #2 check error:**
image

My settings stock settings:
image
Index reindexed and cache flashed (just to be 100% sure)

Please, find time if it is suitable for you and try to reproduce it with my instructions.

Looking forward to your response,
Alex

adrian-martinez-interactiv4:
Hi again @Coderimus , I've been able to reproduce the issue following your steps, applying your changes seems to solve the problem and mark the option as disabled in frontend, I'm starting a deeper review over the proposed changes.

app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
```diff
@@ -217,11 +218,13 @@ public function addStockStatusToSelect(\Magento\Framework\DB\Select $select, \Ma
     /**
      * @param \Magento\Catalog\Model\ResourceModel\Product\Collection $collection
      * @param bool $isFilterInStock
+     * @param \Magento\Store\Model\Website | null $websiteId
      * @return \Magento\Catalog\Model\ResourceModel\Product\Collection $collection
      */
-    public function addStockDataToCollection($collection, $isFilterInStock)
+    public function addStockDataToCollection($collection, $isFilterInStock = true, $websiteId = null)
```
adrian-martinez-interactiv4: There is no need to make already mandatory param $isFilterInStock optional, already existing calls to this methods include explicitly this flag. I think the idea is preset a default value for this field to simplify calls, but by now I would preserve this field as mandatory without default value, to force methods define flag value because this class is not marked as api and default value could change at any time, modifying the behaviour of classes using this method with the default value.

app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
```diff
@@ -202,9 +202,10 @@ public function getProductCollection($lastEntityId = 0, $limit = 1000)
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
      * @return Status
      */
-    public function addStockStatusToSelect(\Magento\Framework\DB\Select $select, \Magento\Store\Model\Website $website)
+    public function addStockStatusToSelect(\Magento\Framework\DB\Select $select, $websiteId = null)
```
adrian-martinez-interactiv4: \Magento\Store\Model\Website $website => $websiteId = null.
- There is no need to make already mandatory param $website optional, see other review comments for further explanation.
- It's preferable to used typed parameters for methods, instead a generic variable. In this case, with this change, you are allowing to pass to this method as second parameter any kind of undesired variables or objects. New private method getWebsiteId also does not check $websiteId param type, so you could end with potential runtime errors when trying to build the query with the unknown-typed $websiteId param.

Please, consider reverting $websiteId param to its previous definition and adapt `$websiteId = $this->getWebsiteId($websiteId);` to `$websiteId = $this->getWebsiteId($website->getId());`, or use \Magento\Store\Model\StoreManager::getWebsite method to get a valid website from a websiteId.

app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
```diff
@@ -277,6 +282,19 @@ public function addIsInStockFilterToCollection($collection)
         return $this;
     }
 
+    /**
+     * @param \Magento\Store\Model\Website $websiteId
+     * @return int
+     */
+    private function getWebsiteId($websiteId)
+    {
+        if ($websiteId === null) {
```
adrian-martinez-interactiv4: While it is not mandatory, please consider putting the operand that cannot be assigned on the left whenever you write a condition, to reduce possible errors.

app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
```diff
@@ -277,6 +282,19 @@ public function addIsInStockFilterToCollection($collection)
...
+            $websiteId = $this->getStockConfiguration()->getDefaultScopeId();
```
adrian-martinez-interactiv4: Avoid using deprecated \Magento\CatalogInventory\Model\ResourceModel\Stock\Status::getStockConfiguration method; please consider injecting $stockConfiguration as optional param in the constructor, and instantiate it in the constructor if needed. Don't hesitate to ask if you need more info about this.

app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php
adrian-martinez-interactiv4: Dependant of other review comments resolution.

app/code/Magento/ConfigurableProduct/Model/Product/Type/Collection/SalableProcessor.php
```diff
@@ -38,19 +38,20 @@ public function __construct(StatusFactory $stockStatusFactory)
...
+    public function process(Collection $collection, $isFilterInStock = true)
     {
         $collection->addAttributeToFilter(
             ProductInterface::STATUS,
             Status::STATUS_ENABLED
         );
 
         $stockFlag = 'has_stock_status_filter';
-        if (!$collection->hasFlag($stockFlag)) {
+        if (!$collection->hasFlag($stockFlag) || !$collection->getFlag($stockFlag)) {
```
adrian-martinez-interactiv4: I see no advantage in adding ` || !$collection->getFlag($stockFlag)`, since the value of this flag is irrelevant as long it is set, because its presence should be only used to mark the collection has already added the stock data to collection, to prevent query errors if \Magento\CatalogInventory\Model\ResourceModel\Stock\Status::addStockDataToCollection is called twice with the same collection before the collection is loaded. Also, this processor logic is very similar to 
\Magento\CatalogInventory\Helper\Stock::addIsInStockFilterToCollection, if we preserved this here, we'd have to adapt also this method.

app/code/Magento/ConfigurableProduct/Model/Product/Type/Collection/SalableProcessor.php
```diff
@@ -38,19 +38,20 @@ public function __construct(StatusFactory $stockStatusFactory)
      * when filter already added and therefore may break the collection.
      *
      * @param Collection $collection
+     * @param bool $isFilterInStock
      * @return Collection
      */
-    public function process(Collection $collection)
+    public function process(Collection $collection, $isFilterInStock = true)
```
adrian-martinez-interactiv4: Please, adapt \Magento\ConfigurableProduct\Test\Unit\Model\Product\Type\Collection\SalableProcessorTest to cover new parameter cases

app/code/Magento/ConfigurableProduct/Model/Product/Type/Collection/SalableProcessor.php
adrian-martinez-interactiv4: `if (!$collection->hasFlag($stockFlag)) {`

app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php
```diff
@@ -1294,9 +1292,11 @@ private function loadUsedProducts(\Magento\Catalog\Model\Product $product, $cach
             $usedProducts = $this->readUsedProductsCacheData($cacheKey);
             if ($usedProducts === null) {
                 $collection = $this->getConfiguredUsedProductCollection($product);
-                if ($salableOnly) {
-                    $collection = $this->salableProcessor->process($collection);
-                }
+                $collection->setFlag('has_stock_status_filter',false);
```
adrian-martinez-interactiv4: Please remove this line: `$collection->setFlag('has_stock_status_filter',false);`

app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php
```diff
@@ -1294,9 +1292,11 @@ private function loadUsedProducts(\Magento\Catalog\Model\Product $product, $cach
...
+                $collection = $this->salableProcessor->process($collection, false);
```
adrian-martinez-interactiv4: `$collection = $this->salableProcessor->process($collection, $salableOnly);`

app/code/Magento/ConfigurableProduct/Test/Unit/Model/Product/Type/ConfigurableTest.php
```diff
@@ -394,6 +394,11 @@ public function testGetUsedProducts()
             ->willReturn('result');
 
         $this->productCollectionFactory->expects($this->any())->method('create')->willReturn($productCollection);
+        $this->salableProcessor
+            ->expects($this->once())
+            ->method('process')
+            ->with($productCollection)
```
adrian-martinez-interactiv4: process() method is usually called with its second parameter as false, but test uses default method second param value, which value is true.
Change `->with($productCollection)` with `->willReturnMap([
                [$productCollection, false, $productCollection]
            ]);`

app/code/Magento/ConfigurableProduct/Test/Unit/Model/Product/Type/ConfigurableTest.php
```diff
@@ -394,6 +394,11 @@ public function testGetUsedProducts()
...
+            ->will($this->returnValue($productCollection));
```
adrian-martinez-interactiv4: `->will($this->returnValue($productCollection))`: Remove this line after applying previous change, as return values are specified in map depending on method call params.

adrian-martinez-interactiv4 (dismissed):
@Coderimus Please, review changes and let me know your thoughts, thank you!

coderimus:
@adrian-martinez-interactiv4 thank you for this super detail and informative review. I will apply all requested changes and will follow them for all my new contributions! I will update PR with them tomorrow.
With great respect,
Alex

coderimus:
@adrian-martinez-interactiv4 sorry for the delay in reply. Had some blockers.
I implemented requested changes you reported in the review but some of them I left because without them explained problem still exists :( The core of this problem is in this method, as I can see: `Magento\ConfigurableProduct\Model\Product\Type\Configurable::loadUsedProducts(\Magento\Catalog\Model\Product $product, $cacheKey, $salableOnly = false)` 

Previously, it was
```php
    private function loadUsedProducts(\Magento\Catalog\Model\Product $product, $cacheKey, $salableOnly = false)
    {
        $dataFieldName = $salableOnly ? $this->usedSalableProducts : $this->_usedProducts;
        if (!$product->hasData($dataFieldName)) {
            $usedProducts = $this->readUsedProductsCacheData($cacheKey);
            if ($usedProducts === null) {
                $collection = $this->getConfiguredUsedProductCollection($product);
                if ($salableOnly) {
                    $collection = $this->salableProcessor->process($collection);
                }
                $usedProducts = array_values($collection->getItems());
                $this->saveUsedProductsCacheData($product, $usedProducts, $cacheKey);
            }
            $product->setData($dataFieldName, $usedProducts);
        }

        return $product->getData($dataFieldName);
    }
```
as you can see the `$collection = $this->salableProcessor->process($collection);` would not be executed because the `$salableOnly = false` was always false due to the previous methods execution results.

I also did this ` $collection->setFlag('has_stock_status_filter',false);` because without this value next code would not be executed:
```php
    public function process(Collection $collection, $isFilterInStock = true)
    {
        $collection->addAttributeToFilter(
            ProductInterface::STATUS,
            Status::STATUS_ENABLED
        );

        $stockFlag = 'has_stock_status_filter';
        if (!$collection->hasFlag($stockFlag) || !$collection->getFlag($stockFlag)) {
            $stockStatusResource = $this->stockStatusFactory->create();
            $stockStatusResource->addStockDataToCollection($collection, $isFilterInStock);
            $collection->setFlag($stockFlag, true);
        }

        return $collection;
    }
```

I used this flag just to add stock status. 

This issue is quite tricky in debugging and as a result difficult to explain all details briefly. Please, if you will have time for detail test, pull my changes and try to debug mentioned methods. Sorry if my description of changes is not so informative.

Thank you so much,
Alex

adrian-martinez-interactiv4:
Hi @Coderimus , I've been reviewing why proposed changes didn't work, and I think I have found the main problem with this issue. The method `\Magento\ConfigurableProduct\Model\Product\Type\Configurable::getConfiguredUsedProductCollection` method, used in `\Magento\ConfigurableProduct\Model\Product\Type\Configurable::loadUsedProducts`,  forced to set flag `has_stock_status_filter` to true, what skipped stock status check. Now, stock status check is performed in frontend by `\Magento\CatalogInventory\Model\AddStockStatusToCollection::beforeLoad` method, that is a frontend plugin. Sorry I have added some commits to your branch, but as you said, this issue is quite tricky and suggest changes and explaining all modifications to make and what changes should be revertes is long, however, feel free to contact me on Slack for a detailed explanation about the proposed changes.

If you are comfortable with changes, you can squash commits and push force your branch so you can get full credit for this PR, as it belongs to you.

adrian-martinez-interactiv4 (approved):

magento-engcom-team:
@Coderimus thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

adrian-martinez-interactiv4:
@Coderimus now that this is accepted and merged, would you consider to backport it to 2.3 if needed? Thank you in advance

coderimus:
@adrian-martinez-interactiv4 thank you! Sure, I will prepare the backport for this issue to 2.3 on Monday next week and will add you as a reviewer! :)

duckchip:
We're having a similar issue that is at least as important to fix but isn't solved by this patch.

In previous comments of @Coderimus:
https://github.com/magento/magento2/pull/12936#issuecomment-358764728
"This issue can be reproduced for a product having more than one item for options and options qty should be bigger than 1 and using swatches. In your example, you have 1 option with 2 products. If you would have 2 options with several products you could reproduce it."

Though when 1 option type is used let's say 'size' with S, M, L, XL. And we make sure that option M has qty = 0 and is out of stock + the Inventory setting to display products out of stock is enabled, that option 'M' disappears from the source code like described in https://github.com/magento/magento2/pull/12936#issuecomment-358748140 by @adrian-martinez-interactiv4 

This can't be intended behaviour. We have to be able to see all the options **AND** make sure that the option that is out of stock either is crossed out or is hidden, but at least outputted in the source code which is not the case with this patch.

scrivvles:
@duckchip, I am also having the same problem with 2.2.5.  If I have a set of configurable products with simple attributes (like size), and some of them are out of stock, then the text swatches disappear.  This happens regardless of my Store > Configuration > Catalog > Inventory being set to "Display Out of Stock Products : Yes"

Is this part of the same issue, or something different?

josh-carter:
@duckchip @scrivvles We're currently having similar issues to you two. 

I've taken a look and found where this stems from, it is indeed related to [the PR for this issue](https://github.com/magento/magento2/pull/13417/files), more specifically the code which gets the collection of products when `getUsedProducts(...)` is called:

```
$collection = $this->getConfiguredUsedProductCollection($product, false);
```

The false parameter results in the following function not setting the `has_stock_status_filter` on that collection:

```
   /**
     * Prepare collection for retrieving sub-products of specified configurable product
     *
     * Retrieve related products collection with additional configuration
     *
     * @param bool $skipStockFilter
     * @return \Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Product\Collection
     */
    private function getConfiguredUsedProductCollection(
        \Magento\Catalog\Model\Product $product,
        $skipStockFilter = true
    ) {
        $collection = $this->getUsedProductCollection($product);

        if ($skipStockFilter) {
            $collection->setFlag('has_stock_status_filter', true);
        }

        $collection
            ->addAttributeToSelect($this->getAttributesForCollection($product))
            ->addFilterByRequiredOptions()
            ->setStoreId($product->getStoreId());

        $collection->addMediaGalleryData();
        $collection->addTierPriceData();

        return $collection;
    }
```

By not setting that flag, whenever the collection is processed via the [Magento\ConfigurableProduct\Model\Product\Type\Collection\SalableProcessor::process()](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/ConfigurableProduct/Model/Product/Type/Collection/SalableProcessor.php#L43) method, out of stock products are then removed. 

This happens because the aforementioned process method [calls the `addStockDataToCollection()` method on the collection with the `isFilterInStock` parameter set as true](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/ConfigurableProduct/Model/Product/Type/Collection/SalableProcessor.php#L53):

```
$stockStatusResource->addStockDataToCollection($collection, true);
```

This results in a [select query on the Collection where products stock status is `in_stock`](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/CatalogInventory/Model/ResourceModel/Stock/Status.php#L245):

```
        if ($isFilterInStock) {
            $collection->getSelect()->where(
                'stock_status_index.stock_status = ?',
                Stock\Status::STATUS_IN_STOCK
            );
        }
```

@scrivvles I too believe that we should be able to set `Store > Configuration > Catalog > Inventory being set to "Display Out of Stock Products : Yes"` and have our out of stock simple swatches and products showing on the frontend of the site as expected.

@dmanners @Coderimus @adrian-martinez-interactiv4 Keen to hear your thoughts. It could pretty easily be fixed with a conditional to check whether the above configuration is set to No, if it is then within the SalableProcessor class pass the `$isFilterInStock` parameter as `true`, if not, don't pass the parameter as it has a default `false` value. This would prevent out of stock products being removed from any used Configurable child Product Collection and would be, in my opinion, rightly controlled via the configuration.

I'd be happy to create a PR for this.

coderimus:
@josh-carter thank you so much for this detailed review.
I just want to add that this issue looks a little bit more complicated and tricky as it could be.

As I understood, the main idea of the `loadUsedProducts` method is **`Load collection on sub-products for specified configurable product`** .

As you can find the `getUsedProducts()` method, which is the main entry point for the private `loadUsedProducts`, calls it without 3rd parameter - [check here, please](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php#L1251)
```
    /**
     * Returns array of sub-products for specified configurable product
     *
     * Result array contains all children for specified configurable product
     *
     * @param  \Magento\Catalog\Model\Product $product
     * @param  array $requiredAttributeIds
     * @return ProductInterface[]
     */
    public function getUsedProducts($product, $requiredAttributeIds = null)
    {
        $metadata = $this->getMetadataPool()->getMetadata(ProductInterface::class);
        $keyParts = [
            __METHOD__,
            $product->getData($metadata->getLinkField()),
            $product->getStoreId(),
            $this->getCustomerSession()->getCustomerGroupId(),
            $requiredAttributeIds
        ];
        $cacheKey = $this->getUsedProductsCacheKey($keyParts);
        return $this->loadUsedProducts($product, $cacheKey);
    }
```
This 3rd parameter is a `$salableOnly` and it has next meaning according to the method description:
```
    /**
     * Load collection on sub-products for specified configurable product
     *
     * Load collection of sub-products, apply result to specified configurable product and store result to cache
     * Please note $salableOnly parameter is used for backwards compatibility because of deprecated method
     * getSalableUsedProducts
     * Number of loaded sub-products depends on $salableOnly parameter
     * $salableOnly = true - result array contains only salable sub-products
     * $salableOnly = false - result array contains all sub-products
     * $cacheKey - allow store result data in different cache records
```
So it means that this param should be always false and we *have to have* all sub products and if we will not have force `false` here `$collection = $this->getConfiguredUsedProductCollection($product, false);` we will have collection filtered by stock status. 

Also, you can see that this method [getSalableUsedProductCollections](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php#L1267) is also deprecated which means that by *calling getUsedProducts* we need to have all subproducts. But with one condition in mind described in the next paragraph.

And the main part here: **the loadUsedProductCollection method is already sensitive to the stock showing setting**. If you have "Display out of stock products" to Yes you will have collection consists of those methods. If setting eq to No only products which are in stock will be displayed.

### **Possible solution**
From my point of view, the problem is in the Swatch module and how it operates with the received collection. Or the described problem is caused be another place in the code after the call to the getUsedProducts method. Need to be investigated.

From my point of view, we are *not able* to have any settings related condition because it will break the original flow.

@dmanners @adrian-martinez-interactiv4 please, when you will have time check this comment. 

With a great respect,
Alex

josh-carter:
@Coderimus I think we need to move away from the `getUsedProducts()` collection for now at least. You're correct when you say that `the described problem is caused be another place in the code after the call to the getUsedProducts method`. 

The problem is from `loadUsedProducts()` where it calls `$collection = $this->getConfiguredUsedProductCollection($product, false);` with that false parameter. It results in `$collection->setFlag('has_stock_status_filter', true);` **not** being set.

So this could still be seen as "not the problem" if i'm making sense, because really this is still fine, the collection doesn't have any `stock_status_filter` on it. But, following on from this, if the Configurable Parent Product ever has `isSalable()` called on it, it will lead to the `SalableProcessor::process()` method being called on the used collection, this can be seen in the [Configurable Product Type Model](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/ConfigurableProduct/Model/Product/Type/Configurable.php#L723):

```
    /**
     * Check is product available for sale
     *
     * @return bool
     */
    public function isSalable($product)
    {
        $salable = parent::isSalable($product);

        if ($salable !== false) {
            $collection = $this->getUsedProductCollection($product);
            $collection->addStoreFilter($this->getStoreFilter($product));
            $collection = $this->salableProcessor->process($collection);
            $salable = 0 !== $collection->getSize();
        }

        return $salable;
    }
``` 

Once the `process()` method is called:

```
    /**
     * Adds filters to the collection to help determine if product is available for sale.
     *
     * This method adds several additional checks for a children products availability.
     * Children products should be enabled and available in stock to be sold.
     * It also adds the specific flag to the collection to prevent the case
     * when filter already added and therefore may break the collection.
     *
     * @param Collection $collection
     * @return Collection
     */
    public function process(Collection $collection)
    {
        $collection->addAttributeToFilter(
            ProductInterface::STATUS,
            Status::STATUS_ENABLED
        );
        $stockFlag = 'has_stock_status_filter';
        if (!$collection->hasFlag($stockFlag)) {
            $stockStatusResource = $this->stockStatusFactory->create();
            $stockStatusResource->addStockDataToCollection($collection, true);
            $collection->setFlag($stockFlag, true);
        }
        return $collection;
    }
```

We see the check for the flag, which isn't set in the `loadUsedProducts()` method as we've clarified with the false parameter it passes when getting the collection. If it isn't set it calls `addStockDataToCollection($collection, true)` which then results in any products with `stock_status` set as `out_of_stock` as can be seen here:

```
        if ($isFilterInStock) {
            $collection->getSelect()->where(
                'stock_status_index.stock_status = ?',
                Stock\Status::STATUS_IN_STOCK
            );
        }
```

**So is there actually a problem? And if so, where?**

Personally i think there is a problem, or maybe a lack of clarity. When i read the configuration for `Display out of Stock Products` and set it to Yes - I read it as i've just made sure any out of stock products in my store will be displayed everywhere possible. Product Lists, Product View, Collections, **Swatches**.

For me I think there are a couple of options:

"Fix" `SalableProcessor` - My fix is in double quotes because some might not see it as a fix, it could be seen as working perfectly already. However, if we changed it to something similar to below:

```
    /**
     * Rewritten so that if show out of stock products is set to yes
     * it wont remove out of stock products
     *
     * @param Collection $collection
     * @return Collection
     */
    public function process(Collection $collection): Collection
    {
        $collection->addAttributeToFilter(
            ProductInterface::STATUS,
            Status::STATUS_ENABLED
        );

        $stockFlag = 'has_stock_status_filter';
        if (!$collection->hasFlag($stockFlag)) {
            $stockStatusResource = $this->stockStatusFactory->create();
            $isFilterInStock = $this->getIsFilterInStock();
            $stockStatusResource->addStockDataToCollection(
                $collection,
                $isFilterInStock
            );
            $collection->setFlag($stockFlag, true);
        }

        return $collection;
    }

    /**
     * Return Configuration value for showing Out Of Stock Products
     *
     * @return bool
     */
    private function getIsFilterInStock(): bool
    {
        $showOutOfStock = $this->stockConfiguration->isShowOutOfStock();
        return $showOutOfStock ? false : true;
    }
```

This is something i've personally implemented yesterday for a client following on from this issue.

Another option could be to add some more clarification to the configuration comment, just something simple saying that out of stock swatches/options wont be displayed.

Maybe theres a possibility that we could add a further configuration option somewhere to define some sort of further out of stock product display.

### Conclusion

I'd say right now its pretty vague as to whether everything is working how it "should". Whether what we're talking about is actually a fix or whether its technically a feature. For now I'll wait to see what the others say.

cc: @dmanners @adrian-martinez-interactiv4

josh-carter:
@dmanners @adrian-martinez-interactiv4 @Coderimus Found this issue today which is related to our convo - https://github.com/magento/magento2/issues/10454

scrivvles:
For what it's worth, I agree with you completely @josh-carter This is a classic example of [POLA ](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)- if everyone interprets something a certain way without any additional prompting or clarification, then that is the way it should work. (Or it should be modified).

I also would love to see your fix if you don't mind sharing it.

josh-carter:
@scrivvles Linked it on that issue i've mentioned, the "fix" is at - https://github.com/interjar/configurable-child-visibility

But to be honest, i aim to get this fixed or at least cleared up in the core too so hopefully that fix extension will become redundant!

PR #12968, opened 2018-01-03
Update AttributeRepository.php

robbertstevens:
This way we get usefull error messages, when saving a attribute.

### Description
When saving an attribute we don't get very usefull error messages. This way we get usefull error messages, when saving a attribute.

### Manual testing scenarios

1. Trigger an error when adding an attribute
2. See usefull error message

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/AttributeRepository.php
+++ b/app/code/Magento/Eav/Model/AttributeRepository.php
@@ -87,7 +87,7 @@ class AttributeRepository implements \Magento\Eav\Api\AttributeRepositoryInterfa
         try {
             $this->eavResource->save($attribute);
         } catch (\Exception $e) {
-            throw new StateException(__('Cannot save attribute'));
+            throw new StateException(__($e->getMessage()));
         }
         return $attribute;
     }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=12968) <br/>All committers have signed the CLA.

avoelkl:
Hi @robbertstevens!
Thanks for the PR!
I am having a look at it, can you just give me some information which errors you triggered to test the PR?
> Trigger an error when adding an attribute`
Thanks!

robbertstevens:
@avoelkl For example, make an attribute with more than 30 characters

orlangur:
Don't we reveal too much information with such approach? Usually some generic message should be displayed to end user and original exception added to log.

> make an attribute with more than 30 characters

Without fix it is showing `'Cannot save attribute'` currently and more details after fix? Wondering if we can do something on a controller level instead of repository.

robbertstevens:
It's only revealing information for admins, users will never see this. If you worry about hackers or something, you have bigger problems if they see these errors.

orlangur:
> It's only revealing information for admins, users will never see this

Users of Admin UI are still users and should not see too much technical details.

robbertstevens:
Sorry, but the errors thrown are not technical details. They are useful messages which are necessary to show. How the hell am I supposed to know a attribute cannot exceed more than 30 characters when the error message is "Cannot save attribute" this can imply multiple reasons. There is not one message to rule them all. We need different messages for different errors. For the sake of simplicity for users. This PR maybe doesn't show the "easy-my-users-are-imbecils"-type of messages but the messages shown are better than what we have now.

Also users who are in the admin are competent enough to comprehend an error message of more than 5 words.

orlangur:
> am I supposed to know a attribute cannot exceed more than 30 characters 

Such kind of message definitely should be shown and was shown before.

However, we cannot just pass underlying exception to end user. There could be some sensitive technical information, like MySQL credentials or other essential data.

sidolov:
Hi @robbertstevens , I'm closing this PR because provided changes may expose sensitive information to the end user and violate Magento security principles. As alternative way you may add exception logging for described case.

PR #13000, opened 2018-01-04
Set `original_inventory_qty` with UiComponent rendered adminhtml Product Edit page

jmonschke:
### Description
Currently, `catalog/view/adminhtml/templates/catalog/product/tab/inventory.phtml` puts an input element, `original_inventory_qty` in its output so that when you have the edit product page open in the admin, and sales are processed either in the admin or on the frontend, if you go to save the product you are editing, and the qty has changed elsewhere, you won't obliterate those changes, instead it will determine the difference between `qty` and `original_inventory_qty` and apply that as a `qty_correction` on the `cataloginventory_stock_item` record.

This pull request puts the `original_inventory_qty` field on the form that the UiComponent system renders in the admin, and adds it to the posted data so that the server-side code that exists properly applies the `qty_correction` to the product's `cataloginventory_stock_item` record.

1. magento/magento2#9345: Admin product save overwrites inventory qty even when it is unchanged

### Manual testing scenarios
1. In Admin, open the Edit Product Screen.
2. Make a note of the QTY field on the Edit Product Screen.
3. On the frontend of the site, place and complete an order for the product you opened in step 1.
4. Make a note of how many of the item you purchased.
5. Back in the admin, where you still have the edit Product screen open, Click Save.
6. When the page refreshes, the QTY shown on the screen should be the Original QTY - Purchased QTY

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogInventory/Api/Data/StockItemInterface.php
+++ b/app/code/Magento/CatalogInventory/Api/Data/StockItemInterface.php
@@ -20,6 +20,7 @@ interface StockItemInterface extends ExtensibleDataInterface
     const PRODUCT_ID = 'product_id';
     const STOCK_ID = 'stock_id';
     const QTY = 'qty';
+    const ORIGINAL_INVENTORY_QTY = 'original_inventory_qty';
     const IS_QTY_DECIMAL = 'is_qty_decimal';
     const SHOW_DEFAULT_NOTIFICATION_MESSAGE = 'show_default_notification_message';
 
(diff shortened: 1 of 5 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13000) <br/>All committers have signed the CLA.

rogyar:
Hello @jmonschke. 
Thank you for your idea. We had an internal discussion regarding the issue. The plan is to implement a more advanced solution for checking all form details and data versioning for preventing the conflicting changes in the upcoming updates.
The solution provided in current PR is good but it affects only QTY field and does it in a different way that is going to be implemented in the future.

Thanks

jmonschke:
@rogyar Thats interesting, because we were assured on the initial issue that Magento was working as expected. I had to raise a Support ticket through Magento Support and there has been no action on Magento's part regarding that ticket other than to email me every month or so saying that there has been no updated. Now there is a solution that fixes the issue that was documented, and you are rejecting it, not because it fixes the issue, using existing code that is present in Magento 2, but instead you want to eventually do it in a different way? It just sounds like you are saying you aren't going to ever make this work.

PR #13001, opened 2018-01-04
Fix fulltext search of attributes with multiselect option values

superdav42:
### Description
Fixes https://github.com/magento/magento2/issues/13972
A bug was introduced in https://github.com/magento/magento2/commit/0ff69c5cbc85af82cd87371c6fe01718a6e463ac which tried optimizing
getting option labels but did not take into account
$valueId could be a comma separated string. This restores
the old behavior and the performance difference is negligible.
Bug only affected 2.2
I'm a little new to unit tests so feel free to give feedback on it, or we could scrap it all together.

### Manual testing scenarios
Create a custom product attribute with the multiselect input type. Create several options. Make the attribute searchable and shown in advanced search. 
Create product with the attribute and select several values.
Searching by label name should be possible in the quick search or advanced search. 

### Contribution checklist

### DIFF
--- a/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/LinkedProductSelectBuilderComposite.php
+++ b/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/LinkedProductSelectBuilderComposite.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\ConfigurableProduct\Model\ResourceModel\Product;
+
+use Magento\Catalog\Model\ResourceModel\Product\LinkedProductSelectBuilderInterface;
+
+/**
+ * Used in Magento\ConfigurableProduct\Pricing\Price\LowestPriceOptionsProvider
+ * to provide queries to select configurable product option with lowest price
+ *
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13001) <br/>All committers have signed the CLA.

superdav42:
@mslabko Wanted to make you aware of this since you've been working on this code.

mslabko:
Hi @superdav42 ,
Thank you for triggering me. 
Could you please merger 2.2-develop - code was changed and it will conflict with your changes. 
It looks like a new code also does not take into account "multi-select attribute issue" - but if we remove caching we can face with performance degradation in case of a lot of searchable attributes. 

What about to handle a case with a comma-separated value id and caching - can we store several values for $this->attributeOptions[$optionKey] for multiselect attribute?

superdav42:
@mslabko 
I see what you mean. I was just using a small catalog before and don't see any benefit with the cache. I tried it with a larger set and had some success using array_intersect_key with the cache. I'll update it.

mslabko:
Yeah, with the customized medium profile (with a bigger amount of products) or with a large profile from performance toolkit benefits is more visible

Floorsave:
Hi there, is there any way that I can patch my 2.2.3 install with this fix as I really need this functionality?
Any assistance is greatly appreciated

PR #13039, opened 2018-01-07
Feature minimum order amount notice issue

neeta-wagento:
During currency switch notice message shows wrong calculation of minimum order amount.

### Description

This is about notice message of min order amount on cart page.When we have set min order amount and on cart page someone change currency, user see same amount and only currency symbol changed. It should recalculate amount also.
Instead of currency convert, we use price helper then it gives us re calculated price amount in notice message.

2. ...

### Manual testing scenarios

Preconditions

Any Product must be created and enabled in frontend with the price greater then 0 or less then 1000
Steps to reproduce

Enable Minimum Order Amount
Login to [Magento Admin Panel]/ Stores -> Configuration -> Sales -> Sales -> Minimum Order Amount and do the following
Set Enable "Yes"
Enter Amount for example 1000
Leave Description Message text blank
Leave Error to Show in Shopping Cart blank
Save the Configurations by clicking on "Save Config" button on top right corner
Now setup Multiple currencies for that follow below steps
Go to section Stores -> Configuration->General->Currency Setup

Now select multiple currencies from Allowed Currencies by pressing ctrl key
select Brazilian Real, Indian Rupee, Japanese Yen, Russian Ruble, US Dollar

Import Currency Rates
Go to Stores -> Under Currency Title click on Currency Rates.
Select any of service and click on import rates it will import live currency rates.
Go to Frontend Add any product in shopping cart add any product 1 qty which has price less then 1000.
Go to cart page "View and Edit Cart" link which you can see on Minicart Popup or by browsing url [Magento base url]/heckout/cart/
You can see the default message on "Minimum order amount is $1,000.00"
Now change the currencies from header upper right corner. You can see Minimum order amount is $1,000.00 in any currency you selected

##  Expected Result 
32069992-def1c89c-ba58-11e7-8e7e-bdece0d6f268
32069993-df303d34-ba58-11e7-8d97-a021f5f5b31f

### Contribution checklist

### DIFF
--- a/app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
+++ b/app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
@@ -37,8 +37,8 @@ class ValidationMessage
     public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Magento\Store\Model\StoreManagerInterface $storeManager,
-        \Magento\Framework\Locale\CurrencyInterface $currency = null,
-        \Magento\Framework\Pricing\Helper\Data $priceHelper
+        \Magento\Framework\Locale\CurrencyInterface $currency,
+        \Magento\Framework\Pricing\Helper\Data $priceHelper = null
     ) {
         $this->scopeConfig = $scopeConfig;
         $this->storeManager = $storeManager;
--- a/app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
+++ b/app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
@@ -19,22 +19,32 @@ class ValidationMessage
 
     /**
      * @var \Magento\Framework\Locale\CurrencyInterface
+     * @deprecated since 101.0.0
      */
     private $currency;
+
+    /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/Quote/Model/Quote/Validator/MinimumOrderAmount/ValidationMessage.php
```diff
@@ -19,22 +19,32 @@ class ValidationMessage
...
      * @param \Magento\Store\Model\StoreManagerInterface $storeManager
      * @param \Magento\Framework\Locale\CurrencyInterface $currency
+     * @param \Magento\Framework\Pricing\Helper\Data $priceHelper
      */
     public function __construct(
         \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
         \Magento\Store\Model\StoreManagerInterface $storeManager,
-        \Magento\Framework\Locale\CurrencyInterface $currency
+        \Magento\Framework\Locale\CurrencyInterface $currency,
+        \Magento\Framework\Pricing\Helper\Data $priceHelper
```
ishakhsuvarov: This argument should be optional to maintain backward compatibility
More info available in [devdocs](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#adding-a-constructor-parameter)

ishakhsuvarov (changes requested):

ishakhsuvarov (approved):

Issue #12892, opened 2017-12-27
Customer account navigation block container doesn't have a "name" or "as" attribute (fix included)


Lez-2:
### Preconditions
Magento 2.2.0

### Steps to reproduce
1. Open the Customer account link e.g. http://domain.com/customer/account while logged in and you can see through any browser debugger that the html surrounding the customer navigation menu is:
```
<div class="block account-nav">
    <div class="title account-nav-title">
        <strong></strong>
    </div>
<div id="account-nav" class="content account-nav-content">
...
</div>
```
The title is always empty, but it is expected.
And you can't add the title because the containing block of the customer account navigation, which is located in .../vendor/magento/module-customer/view/frontend/layout/customer_account.xml, doesn't have a "name" or "as" attribute which is used to reference the block.

### Expected result
1. the customer_account.xml located in .../vendor/magento/module-customer/view/frontend/layout/ should have the block named something like "customer-account-navigation-containter" (with or without a default title argument):
```
<block class="Magento\Framework\View\Element\Template" name="customer-account-navigation-container" template="Magento_Theme::html/collapsible.phtml" before="-">
    <arguments>
        <argument name="block_css" xsi:type="string">account-nav</argument>
	<argument name="title" xsi:type="string" translate="true">Account menu</argument>
    </arguments>
    <block class="Magento\Customer\Block\Account\Navigation" name="customer_account_navigation" before="-">
...
```

### Actual result
1. The current customer_account.xml doesn't have a name or as attribute for the collapsible template container of the menu: 
```
<block class="Magento\Framework\View\Element\Template" template="Magento_Theme::html/collapsible.phtml" before="-">
    <arguments>
        <argument name="block_css" xsi:type="string">account-nav</argument>
    </arguments>
    <block class="Magento\Customer\Block\Account\Navigation" name="customer_account_navigation" before="-">
...
```

devamitbera:
#mm18in

Lez-2:
Seems this issue already has a fix in M 2.2.2

```
<referenceContainer name="sidebar.main">
            <block class="Magento\Framework\View\Element\Template" name="sidebar.main.account_nav" template="Magento_Theme::html/collapsible.phtml" before="-">
                <arguments>
                    <argument name="block_css" xsi:type="string">account-nav</argument>
                </arguments>
                <block class="Magento\Customer\Block\Account\Navigation" name="customer_account_navigation" before="-">
```

ishakhsuvarov:
[...]

PR #13042, opened 2018-01-07
Customer account navigation block container doesn't have  a name 

devamitbera:
- a name or as attribute
Add the suggestion  but it is only showing Label
-seems not good.So,add a condition to phtml.
It makes title only visible when it has childhtml

1. magento/magento2#12892: Issue title

### Manual testing scenarios

1.Please open the Customer account link e.g. http://domain.com/customer/account
2.loggin as Customer  and then go to the Dashboard page
3.Now, you don't see any Section like **<strong></strong>** at the left section of customer account pages.
4.For case,I have added a **conditions to prevent to show the Only title/label* * whenever child does not have any html output By  adding **if ($block->getChildHtml()):**.It  is helpful for UI respective view.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/view/frontend/layout/customer_account.xml
+++ b/app/code/Magento/Customer/view/frontend/layout/customer_account.xml
@@ -12,6 +12,7 @@
             <block class="Magento\Framework\View\Element\Template" name="sidebar.main.account_nav" template="Magento_Theme::html/collapsible.phtml" before="-">
                 <arguments>
                     <argument name="block_css" xsi:type="string">account-nav</argument>
+                    <argument name="block_title" xsi:type="string">Account Sidebar</argument>
                 </arguments>
                 <block class="Magento\Customer\Block\Account\Navigation" name="customer_account_navigation" before="-">
                     <arguments>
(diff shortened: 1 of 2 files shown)


app/code/Magento/Customer/view/frontend/layout/customer_account.xml
```diff
@@ -12,6 +12,7 @@
             <block class="Magento\Framework\View\Element\Template" name="sidebar.main.account_nav" template="Magento_Theme::html/collapsible.phtml" before="-">
                 <arguments>
                     <argument name="block_css" xsi:type="string">account-nav</argument>
+                    <argument name="block_title" xsi:type="string">Account Sidebar</argument>
```
dmanners: can you also include the `translate="true"` attribute here

dmanners (changes requested):

app/code/Magento/Customer/view/frontend/layout/customer_account.xml
devamitbera: sure.I  am doing.

Thanks for feedback.

dmanners (approved):

dmanners:
As discussed at mm18in I think that the original issue was to add the name attribute to the block (`name="sidebar.main.account_nav"`) which has already been done. I am not sure if we should actually add this title by default @okorshenko what would you think here?

okorshenko:
The issue already fixed. Please see the comment. vhttps://github.com/magento/magento2/issues/12892#issuecomment-355815257

@devamitbera thank you for your contribution

Issue #12828, opened 2017-12-20
Uncaught Error: Script error for: trackingCode error on every frontend page


ghost:
### Preconditions
1. Installed Magento version 2.2.2 with sample data.
2. checked in developer and production both mode.
3. Dotdigitalgroup_Email module is enabled downloaded with Magento version 2.2.2.

### Steps to reproduce
1. Open Magento home page.
2. checked browser console and getting error Uncaught Error: Script error for: trackingCode .
![screenshot_90](https://user-images.githubusercontent.com/14589296/34212307-eb18bb14-e5c1-11e7-9451-fbb57f9e25c5.png)
3. Also, this error prevents to load custom UI component grid at frontend as shown in the screenshot-
![screenshot_95](https://user-images.githubusercontent.com/14589296/34212308-ed8131e2-e5c1-11e7-88e1-8fd169aa8492.png)

### Expected result
1. No error should occur.

### Actual result
1. for every frontend page, checked browser console and getting error Uncaught Error: Script error for: trackingCode .
2. Also, this error prevents to load custom UI component grid at frontend.

magento-engcom-team:
@poojawebkul , thank you for your report.
We were not able to reproduce this issue by following the steps you provided. Please provide more detailed steps to reproduce or try to reproduce this issue on a clean installation without any extentions.

ghost:
@magento-engcom-team issue is updated please check.

Yonn-Trimoreau:
uBlock Origin extension is the source of the issue.

It blocks the trackingCode.js file before it has the time to track you down. And that seems to be justified.

Here is the content of `trackingCode.js` : 
```
define(['domReady!'], function () {
    'use strict';

    /**
     * Create script tag
     */
    function createTag() {
        var connector = document.createElement('script'),
            s = document.getElementsByTagName('script')[0];

        connector.type = 'text/javascript';
        connector.src =
            (document.location.protocol === 'https:' ? 'https://' : 'http://') + 't.trackedlink.net/_dmpt.js';
        s.parentNode.insertBefore(connector, s);
    }

    /**
     * Export/return tracking code init
     * @param {Object} trackingCode
     */
    return function (trackingCode) {
        if (trackingCode.isEnabled) {
            createTag();
        }
    };
});
```
And where does that lead to ? ==> A tracking script.. [http://t.trackedlink.net/_dmpt.js](http://t.trackedlink.net/_dmpt.js)

[...]

PR #13061, opened 2018-01-08
Fix for requireJS loading issues (for ad blockers)

Yonn-Trimoreau:
Please read https://github.com/magento/magento2/issues/12828

### Description

When uBlock (or any ad blocker) forbids `trackingCode.js` file from loading, the exception thrown by RequireJS breaks the JS execution flow, causing unexpected and random issues elsewhere on the website.

This fix catches the RequireJS script loading error, displays it in the console as is and returns true, to avoid execution flow to be broken.

1. magento/magento2#12828: Uncaught Error: Script error for: trackingCode error on every frontend page

### Manual testing scenarios

1. Install and enable uBlock Origin [Chrome/Firefox extension] on any stock Magento 2.2.2 installation
2. Open JS console, watch the issue raised (on any page)
3. Pass an order
4. Without my fix checkout would not work, since JS execution flow was broken by RequireJS script loading error

### Contribution checklist

PS: could not provide unit/integration tests since it's raised by a browser extension

### DIFF
--- a/lib/web/mage/apply/main.js
+++ b/lib/web/mage/apply/main.js
@@ -7,7 +7,7 @@ define([
     'underscore',
     'jquery',
     './scripts'
-], function (_, $, processScripts, consoleLogger) {
+], function (_, $, processScripts) {
     'use strict';
 
     var dataAttr = 'data-mage-init',
--- a/lib/web/mage/apply/main.js
+++ b/lib/web/mage/apply/main.js
@@ -6,8 +6,7 @@
 define([
     'underscore',
     'jquery',
-    './scripts',
-    'Magento_Ui/js/lib/logger/console-logger'
+    './scripts'
 ], function (_, $, processScripts, consoleLogger) {
     'use strict';
 
@@ -34,8 +33,8 @@ define([
                 $(el)[component](config);
             }
         }, function (error) {
-            consoleLogger.error(error);
-            
+            console.error(error);
+
             return true;
         });
     }
--- a/lib/web/mage/apply/main.js
+++ b/lib/web/mage/apply/main.js
@@ -32,6 +32,9 @@ define([
             } else if ($(el)[component]) {
                 $(el)[component](config);
             }
+        }, function(error) {
+            console.error(error);
+            return true;
         });
     }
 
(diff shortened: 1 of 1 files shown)


orlangur:
Is it a common practice in Magento code to call console like `console.error(error);` or maybe there is some wrapper so that console is spoiled in dev mode only?

Yonn-Trimoreau:
It seems there's a consoleLogger yes, let me fix this code style issue and integrate with the existing consoleLogger.

omiroshnichenko:
Hi, @Yonn-Trimoreau. I agree with @orlangur that we must use wrapper, but unfortunately it's located in UI module and we can't depend on modules from `lib/web`. So you can revert your code to previous commit and fix code style(you have extra whitespaces on line before return).

orlangur:
@omiroshnichenko, I've quickly grepped following code snippet

```
if (window.console && !this.element.is(this.options.defaultContainer) && $.mage.isDevMode(undefined)) {
                console.warn('This widget is intended to be attached to the body, not below.');
            }
```

should `$.mage.isDevMode` be utilized here or initial implementation is just fine?

Yonn-Trimoreau:
@omiroshnichenko Done
@orlangur I will add it if @omiroshnichenko validates your idea

omiroshnichenko:
@orlangur I also found this code snippet, but I can't found place where `isDevMode` widget initialized. Also, `consoleLogger` doesn't take into account current mode.

Yonn-Trimoreau:
@orlangur @omiroshnichenko Since this bit of code will apply to all required scripts (through `data-mage-init` exclusively), it may be interesting for production debugging purposes if we leave the error displayed in the console even if devMode is disabled.

Maybe using `console.warn(error);` when in production mode could be less "alarming" but still allow debugging easily.

omiroshnichenko:
@Yonn-Trimoreau Currently, we cannot get current mode in JS. So use `console.error`.

Yonn-Trimoreau:
@omiroshnichenko I still would like to add a check on window.console.error, as I could see in `swagger-ui.js` at line 21987 [`if ('console' in window && typeof window.console.warn === 'function') {`]. Then current PR is ok to merge after Travis has ended his job.

spectravp:
This fix does not work; manually applied to 2.2.3, Safari 11.1, Mac High Sierra

PR #13086, opened 2018-01-10
Add failsafe to items.phtml

samgranger:
### Description
We received a warning that $exist is undefined - not entirely sure what the exact scenario was but adding this as a failsafe.

I have made sure that $exist is defined (null) in the 'default' and 'other' case.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/templates/product/list/items.phtml
+++ b/app/code/Magento/Catalog/view/frontend/templates/product/list/items.phtml
@@ -151,7 +151,7 @@ switch ($type = $block->getType()) {
 }
 ?>
 
-<?php if (isset($exist)):?>
+<?php if (isset($exist) && $exist):?>
 
     <?php if ($type == 'related' || $type == 'upsell'): ?>
         <?php if ($type == 'related'): ?>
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/view/frontend/templates/product/list/items.phtml
```diff
@@ -147,7 +147,11 @@ switch ($type = $block->getType()) {
     break;
 
     case 'other':
+        $exist = null;
     break;
+        
+    default:
+        $exist = null;
```
dmanners: Rather than adding the default options here what about adding an `isset` check as part of the if wrapping the content display.

dmanners (changes requested):

app/code/Magento/Catalog/view/frontend/templates/product/list/items.phtml
samgranger: Changed as requested. Your suggestion is a neater solution.

samgranger (commented):
I have removed the extra declarations for $exist and have added an isset check to the if statement as requested.

dmanners:
@samgranger thank you for the quick update. I will process this pr now.

dmanners (approved):

PR #13096, opened 2018-01-10
Magento Backend Login since 2.1.10 / 2.2 does not autocomplete passwords

driskell:
### Description
Since upgrading to 2.1.10 (even 2.2) the Backend login for Magento 2 no longer autocompletes passwords, a useful browser feature for easing login to websites by allowing significantly more complex and secure passwords to be protected under a single user password that is then protected by corporate policies that enforce complexity and rotation. It also eases the risk during the window between leavers being removed from corporate login and being removed from Magento.

It seems previously Magento tried to use autocomplete "off" to disable this feature. However, many browser vendors consider this to be "abusive" as it disables key browser features, so choose to ignore autocomplete "off" on any password fields.

> There is a general consensus within Chrome that there are no security benefits to autocomplete="off".
>
> (Reference: https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c17)

This PR restores the previously functionality that allowed autocompletion of password in Magento backend login but improves it to use the new autocomplete HTML5 values `username` and `current-password`.

### Manual testing scenarios
1. Load Magento backend login, and login, no prompt is given to save password

or

1. If you previously had saved password for Magento login before upgrading, it is no longer autofilled

### Contribution checklist

### DIFF
--- a/app/code/Magento/Backend/view/adminhtml/templates/admin/login.phtml
+++ b/app/code/Magento/Backend/view/adminhtml/templates/admin/login.phtml
@@ -27,7 +27,7 @@
                        value=""
                        data-validate="{required:true}"
                        placeholder="<?= /* @escapeNotVerified */ __('user name') ?>"
-                       autocomplete="off"
+                       autocomplete="username"
                     />
             </div>
         </div>
@@ -43,7 +43,7 @@
                        data-validate="{required:true}"
                        value=""
                        placeholder="<?= /* @escapeNotVerified */ __('password') ?>"
-                       autocomplete="new-password"
+                       autocomplete="current-password"
                     />
             </div>
         </div>


miguelbalparda:
Thanks for your contribution @driskell! @piotrekkaminski I'd like to have your opinion on this one before processing it.

piotrekkaminski:
Unfortunately this still gets flagged in automated PCI scanning reports that our merchants use and they have to explain this to their QSA why this is not a problem. I don't think we can accept it until the scanning vendors get rid of this (i agree with you) incorrect practice of flagging lack of disabling autocomplete.

driskell:
@piotrekkaminski I was thinking about this and I'm happy to look at extending this PR to include an option. The default can be the current behaviour with autocomplete "off". Then there can be an option in Administration to enable autocomplete.

This would then allow both ball parks to configure to their needs, with the default erring on the side of the PCI industry. WDYT?

miguelbalparda:
Sounds like a plan @driskell! We will be treating that as a feature, so maybe it's easier to close this PR and create a new one with all the changes?

miguelbalparda:
I'll close this for now, will revisit if anything changes. Feel free to reopen the PR or create a new one @driskell!

PR #13097, opened 2018-01-10
Fixed typo in event name

dverkade:
There is a typo in a event name, changed "catelog" to "catalog". Checked complete Magento core code and the wrongly spelled "catelog" string is not found in other places, so this is safe to merge.

### DIFF
--- a/app/code/Magento/CatalogSearch/Model/Indexer/Fulltext/Action/DataProvider.php
+++ b/app/code/Magento/CatalogSearch/Model/Indexer/Fulltext/Action/DataProvider.php
@@ -204,6 +204,12 @@ class DataProvider
             /** @var \Magento\Eav\Model\Entity\Attribute[] $attributes */
             $attributes = $productAttributes->getItems();
 
+            /** @deprecated */
+            $this->eventManager->dispatch(
+                'catelogsearch_searchable_attributes_load_after',
+                ['engine' => $this->engine, 'attributes' => $attributes]
+            );
+            
             $this->eventManager->dispatch(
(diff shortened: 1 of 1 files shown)


app/code/Magento/CatalogSearch/Model/Indexer/Fulltext/Action/DataProvider.php
```diff
@@ -205,7 +205,7 @@ public function getSearchableAttributes($backendType = null)
             $attributes = $productAttributes->getItems();
 
             $this->eventManager->dispatch(
-                'catelogsearch_searchable_attributes_load_after',
+                'catalogsearch_searchable_attributes_load_after',
```
ishakhsuvarov: While this change is perfectly valid, 3-rd party extensions may already be relying on this misspelled event name causing their customizations to break.
For additional reference please check [Backward Compatible Development Guide](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html)

ishakhsuvarov (changes requested):

app/code/Magento/CatalogSearch/Model/Indexer/Fulltext/Action/DataProvider.php
dverkade: Hi @ishakhsuvarov,

Thanks for your feedback. I missed that and you are right, it should be backwards compatible. I have changed this in the new commit.

ishakhsuvarov (approved):

PR #13101, opened 2018-01-10
11828 Fix issue with swatch colour block not showing in admin panel once colour selected (PHP7.1.x issue).

chris-pook:
### Description
Using PHP7.1.x the swatch colour block preview in the admin panel attribute will not show the actual colour as the inline style is set to style"b". This is due to PHP no longer allowing strings to be manipulated as arrays.

1. magento/magento2#11828: Visual Swatches not showing swatch color in admin

### Manual testing scenarios
**Preconditions**
Magento CE 2.2
PHP 7.1.0
**Steps to reproduce**
Inside admin, go to Stores->Product and click on an attribute that contains visual swatches.
**Expected result**
Visual Swatches that have a color assigned should show that color in the swatch box.
**Actual result**
Although the color swatch values are being saved, the visual representation of the color in the box is colorless.

### Contribution checklist
 - [ x] All new or changed code is covered with unit/integration tests (if applicable)
 - [ x] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Swatches/Block/Adminhtml/Attribute/Edit/Options/Visual.php
+++ b/app/code/Magento/Swatches/Block/Adminhtml/Attribute/Edit/Options/Visual.php
@@ -84,15 +84,15 @@ class Visual extends AbstractSwatch
      * Parse swatch labels for template
      *
      * @codeCoverageIgnore
-     * @param null $swatchStoreValue
-     * @return string
+     * @param null|array $swatchStoreValue
+     * @return null|array
      */
     protected function reformatSwatchLabels($swatchStoreValue = null)
     {
         if ($swatchStoreValue === null) {
             return;
         }
-        $newSwatch = '';
+        $newSwatch = [];
         foreach ($swatchStoreValue as $key => $value) {
             if ($value[0] == '#') {
                 $newSwatch[$key] = 'background: '.$value;


ihor-sviziev:
@chris-pook could you explain what it was returning in php 7.0? There is string to array conversion, so probably we were getting some unexpected result

chris-pook:
Hi @ihor-sviziev on php7.0 an array is returned containing the inline style strings as set by the function. The problem is that using php7.1 the strings are truncated due to the odd way they are produced using array notation on `$newSwatch` which is defined as a string.

To see the issue introduced in php7.1 you can run the following code:

```
<?php
$str = '';
$str[3] = 'background';
var_dump($str[3]);
```

With php7.0 this will produce:

```
string(10) "background"
```

With php7.1 this will output:

```
string(1) "b"
```

app/code/Magento/Swatches/Block/Adminhtml/Attribute/Edit/Options/Visual.php
```diff
@@ -84,15 +84,15 @@ public function getJsonConfig()
      * Parse swatch labels for template
      *
      * @codeCoverageIgnore
-     * @param null $swatchStoreValue
-     * @return string
+     * @param array $swatchStoreValue
```
orlangur: It can still be `null`. Could you also squash changes into single commit and force push?

orlangur (changes requested):

chris-pook:
@orlangur updated docblock and squashed commits down to 1 (apologies for multiple updates, messed up a local merge).

orlangur (approved):

PR #13109, opened 2018-01-10
Get current order ID on success page

srenon:
### Description

Currently, Magento is using a mixture of observer or Magento\Checkout\Model\Session to return order id on success page which make some implementation difficult because the session method does not support multi-shipping.

This PR will make it more consistent to access order id using ``Magento\Checkout\Model\Session::getLastOrderIds()`` which return an array ([order_id => increment_id]) to support mutlishipping 

See

Session - https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Checkout/Block/Registration.php#L119

Observer - https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/GoogleAnalytics/Observer/SetGoogleAnalyticsOnOrderSuccessPageViewObserver.php#L57

### Manual testing scenarios

1. not needed

### Contribution checklist

### DIFF
--- a/app/code/Magento/Multishipping/Model/Checkout/Type/Multishipping.php
+++ b/app/code/Magento/Multishipping/Model/Checkout/Type/Multishipping.php
@@ -770,7 +770,8 @@ class Multishipping extends \Magento\Framework\DataObject
             }
 
             $this->_session->setOrderIds($orderIds);
-            $this->_checkoutSession->setLastQuoteId($this->getQuote()->getId());
+            $this->_checkoutSession->setLastQuoteId($this->getQuote()->getId())
+                ->setLastOrderIds($orderIds);
 
             $this->getQuote()->setIsActive(false);
             $this->quoteRepository->save($this->getQuote());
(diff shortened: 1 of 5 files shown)


orlangur:
> Manual testing scenarios
> not needed

This is almost never true (for example, in case of changing documentation only).

app/code/Magento/Multishipping/Model/Checkout/Type/Multishipping.php
```diff
@@ -770,7 +770,8 @@ public function createOrders()
             }
 
             $this->_session->setOrderIds($orderIds);
-            $this->_checkoutSession->setLastQuoteId($this->getQuote()->getId());
+            $this->_checkoutSession->setLastQuoteId($this->getQuote()->getId())
+                ->setLastOrderIds($orderIds);
```
orlangur, line 774: Why do we need to set `$orderIds` in two different session objects?

orlangur:
> This PR will make it more consistent to access order id using `Magento\Checkout\Model\Session::getLastOrderIds()`

I don't see obtaining logic among PR changes.

miguelbalparda:
I am closing this PR for now due to lack of activity.
Please check the code review comments and reopen this ticket if you wish to continue.
Thanks!

PR #13132, opened 2018-01-11
Update the Emogrifier dependency to ^2.0.0

oliverklee:
### Description
Emogrifer 2.0.0 does not introduce any backwards-compatibilty-breaking
API changes. This version adds new features, fixes bugs and improves
the resulting HTML. So the update should be reasonably safe.

### Manual testing scenarios
1. have Magento send an HTML email
2. see that the generated email still looks okay

### Contribution checklist

(Should I as a contributor check those, or will the reviewers do this?)

### DIFF
--- a/lib/internal/Magento/Framework/Css/PreProcessor/Adapter/CssInliner.php
+++ b/lib/internal/Magento/Framework/Css/PreProcessor/Adapter/CssInliner.php
@@ -5,6 +5,7 @@
  */
 namespace Magento\Framework\Css\PreProcessor\Adapter;
 
+use Magento\Framework\App\State;
 use Pelago\Emogrifier;
 
 /**
@@ -17,9 +18,13 @@ class CssInliner
      */
     private $emogrifier;
 
-    public function __construct()
+    /**
+     * @param State $appState
+     */
+    public function __construct(State $appState)
     {
         $this->emogrifier = new Emogrifier();
+        $this->emogrifier->setDebug($appState->getMode() === State::MODE_DEVELOPER);
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "composer/composer": "1.4.1",
         "monolog/monolog": "^1.17",
         "oyejorge/less.php": "~1.7.0",
-        "pelago/emogrifier": "1.2.0",
+        "pelago/emogrifier": "^2.0.0",
         "tubalmartin/cssmin": "4.1.0",
         "magento/magento-composer-installer": ">=0.1.11",
         "braintree/braintree_php": "3.25.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "009029e4e58a3802cc9ad92333fad729",
+    "content-hash": "9203f9f6b859a7748f08eb1100f9e97e",
     "packages": [
         {
             "name": "braintree/braintree_php",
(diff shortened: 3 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13132) <br/>All committers have signed the CLA.

oliverklee:
Just signed the CLA.

oliverklee:
Okay, looks like I need to update the email generation error handling. Working on it.

orlangur:
> Emogrifer 2.0.0 does not introduce any backwards-compatibilty-breaking

Is such policy applied for future versions too? I mean, is `^2.0.0` safe enough or maybe we should stick to `~2.0.0` instead.

> (Should I as a contributor check those, or will the reviewers do this?)

Yeah, feel free to check those, probably reviewer will uncheck some if he has another opinion.

oliverklee:
> Is such policy applied for future versions too? I mean, is ^2.0.0 safe enough or maybe we should stick to ~2.0.0 instead.

Yes, `^2.0.0` will be safe enough. Emogrifier follows semantic versioning very strictly.

orlangur:
@oliverklee,

> Emogrifier follows semantic versioning very strictly.

In such case we should make `~2.0.0` constraint and bump to `~2.1.0` and further manually with proper checks. So that we won't break Magento SemVer commitment and, for instance, receive breaking Emogrifier upgrade to 3.0 in a Magento patch release.

oliverklee:
Rebased, updated the tests and added a setter for the Emogrifier debug mode.

oliverklee:
> In such case we should make ~2.0.0 constraint and bump to ~2.1.0 and further manually with proper checks. So that we won't break Magento SemVer commitment and, for instance, receive breaking Emogrifier upgrade to 3.0 in a Magento patch release.

@orlangur I'm not sure I understand your concern correctly. Are you worried that `^2.0.0` would allow (breaking) updates to 3.0.0?

orlangur:
@oliverklee yeah, it can happen with any `composer update` later.

oliverklee:
@orlangur Thanks for the clarification. `^2.0.0` will only allow updates to 2.x.y, not to 3.x.y, and hence should be safe. Please see https://getcomposer.org/doc/articles/versions.md#caret-version-range- for the documentation of the `^` version constraint.

orlangur:
@oliverklee will it allow Emogrifier upgrade to `2.1` in some of the `2.2.x` Magento patch releases?

oliverklee:
@orlangur That's correct. And as Emogrifier follows semantic versioning, 2.1.x will not break anything over 2.0.0 (and will also not require any code changes in Magento).

I've noticed that some dependencies in the `composer.json` are set using `^`, while others are set using `~`. What's the general policy in the Magento project for using `^` vs. `~` for dependencies?

oliverklee:
@orlangur And I'm also not sure whether my PR should go to the 2.2-develop branch, the 2.3-develop branch, or both. (I couldn't find anything about this in the CONTRIBUTING.md file.) Any advice on this? Thanks!

orlangur:
@oliverklee just noticed there are many ZF component using caret constraint, let's leave the one proposed by you until we face with any issue.

> PR should go to the 2.2-develop branch, the 2.3-develop branch, or both

Any acceptable from SemVer perspective contribution must go to `2.2-develop` first, I'll check if this process is properly documented.

app/code/Magento/Email/Model/Template/Filter.php
```diff
@@ -973,6 +973,9 @@ public function applyInlineCss($html)
         );
         $cssToInline = $this->getCssProcessor()->process($cssToInline);
 
+        $debugMode = $this->_appState->getMode() === \Magento\Framework\App\State::MODE_DEVELOPER;
+        $this->cssInliner->setDebugMode($debugMode);
```
orlangur: Why do we need this?

oliverklee:
@orlangur Okay, thanks for the update and the information!

orlangur (changes requested):

app/code/Magento/Email/Model/Template/Filter.php
oliverklee: Emogrifier 1.2 always threw an exception when it encountered something it could not parse (if it didn't directly bail with an XPath error). The filter then either caught and discarded the exception (in production mode) or added an error message (in non-production mode).

Emogrifier 2.0 now has a separate debug mode. In debug mode, it still throws exceptions. In the non-debug mode (the default), it silently ignores anything it cannot parse, and tries to cope the best it can.

So to have both old filter behaviors, we need to enable/disable the Emogrifier debug mode depending on the environment.

orlangur: Got it, thanks for detailed explanation.

lib/internal/Magento/Framework/Css/PreProcessor/Adapter/CssInliner.php
```diff
@@ -54,6 +54,16 @@ public function disableStyleBlocksParsing()
...
     }
 
+    /**
+     * @param bool $debug
+     *
+     * @return void
+     */
+    public function setDebugMode($debug)
+    {
+        $this->emogrifier->setDebug($debug);
```
orlangur: Let's not expose too much methods when we don't really need it.

Please discard changes in `app/code/Magento/Email/Model/Template/Filter.php` and instead of introducing a new method here inject `AppState` and do `setDebug(true)` in constructor in case of developer mode.

orlangur (changes requested):

lib/internal/Magento/Framework/Css/PreProcessor/Adapter/CssInliner.php
oliverklee: On it.

oliverklee: Done and repushed. I'm not very familiar with how Magento handles DI, so I'm keeping my fingers crossed.

magento-engcom-team:
@oliverklee thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

oliverklee:
@orlangur After this has been merged, what's the process to get this in 2.3-develop, too? Should I create a PR for this?

orlangur:
@oliverklee yeah, you can create a PR for `2.3-develop` even at the same time with `2.2-develop` one. Otherwise your changes will not be missed anyway, core team will port everything from `2.2-develop` to `2.3-develop` by themselves.

oliverklee:
@orlangur Done: #13351

oliverklee:
@orlangur I think this PR still has the review status "changes requested" (by you). If this is obsolete (which I believe), could you please update the review status? Thanks! :coffee:

orlangur (approved):

PR #13137, opened 2018-01-11
Remove zend json from json result controller

dmanners:
### Description
This PR is a reworking of https://github.com/magento/magento2/pull/10367
Since Zend_Json should no longer be used update to the `\Magento\Framework\Serialize\Serializer\Json` instead. So that we do not break backwards compatibility we will need to allow for the type `\Magento\Framework\DataObject` to be passed into the method and but we need to convert this into an array first.

1. magento/magento2#9236

### Manual testing scenarios

1. Check that controllers with result json work with setting data as multiple types.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Controller/Result/Json.php
+++ b/lib/internal/Magento/Framework/Controller/Result/Json.php
@@ -52,9 +52,13 @@ class Json extends AbstractResult
      * @param boolean $cycleCheck Optional; whether or not to check for object recursion; off by default
      * @param array $options Additional options used during encoding
      * @return $this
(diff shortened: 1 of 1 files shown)


ishakhsuvarov:
@dmanners Thank you for the continuous effort on the way of removing outdated dependencies.

Taking into the account previous conversation (the one you linked) I am going to spin off all the functional tests to see if anything is broken this time. I'll get back to you with the results.

Meanwhile I would suggest adding/updating tests to make sure all possible cases are supported.

ishakhsuvarov (approved):

PR #13144, opened 2018-01-11
Log file path when image open throws exception

pmclain:
### Description
Log the offending file when opening an image throws an exception. The current exception message does a good job describing the problem, but does not tell you which file caused the issue. The proposed change would alter the log entry from:  
`Unsupported image format.`  
to:  
`Unsupported image format.: /path/to/my/bad/file.png`

none

### Manual testing scenarios
1. Point product `small_image` to a corrupt/unsupported image.
2. Navigate to product list page containing product from step 1. The page should partially render, likely without any products.
3. View error report in `/var/report` to see the path to problem image.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Image/Adapter/Gd2.php
+++ b/lib/internal/Magento/Framework/Image/Adapter/Gd2.php
@@ -62,7 +62,10 @@ class Gd2 extends \Magento\Framework\Image\Adapter\AbstractAdapter
             throw new \OverflowException('Memory limit has been reached.');
         }
         $this->imageDestroy();
-        $this->_imageHandler = call_user_func($this->_getCallback('create'), $this->_fileName);
+        $this->_imageHandler = call_user_func(
+            $this->_getCallback('create', null, sprintf('Unsupported image format. File: %s', $this->_fileName)),
+            $this->_fileName
+        );
     }
 
     /**
--- a/lib/internal/Magento/Framework/Image/Adapter/ImageMagick.php
+++ b/lib/internal/Magento/Framework/Image/Adapter/ImageMagick.php
@@ -77,7 +77,7 @@ class ImageMagick extends \Magento\Framework\Image\Adapter\AbstractAdapter
         try {
             $this->_imageHandler = new \Imagick($this->_fileName);
         } catch (\ImagickException $e) {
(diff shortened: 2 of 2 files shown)


lib/internal/Magento/Framework/Image.php
```diff
@@ -52,7 +52,11 @@ public function open()
             throw new \Exception("File '{$this->_fileName}' does not exists.");
         }
 
-        $this->_adapter->open($this->_fileName);
+        try {
+            $this->_adapter->open($this->_fileName);
+        } catch (\Exception $e) {
+            throw new \Exception("{$e->getMessage()}: {$this->_fileName}");
```
orlangur: Why not add such information directly in `adapter->open()` method?

orlangur (changes requested):

lib/internal/Magento/Framework/Image.php
pmclain: So it doesn't need to be added to both existing adapters and any future adapters.

orlangur: It's a bit odd to catch exception and throw immediately to add information which was passed to adapter method anyway. Please move this addition to adapter(s) as it's clearly his responsibility to throw such exception.

As always, please do commit amend and force push.

pmclain: @orlangur ready for another look.

TomashKhamlai:
Hello, @pmclain! I am testing the changes that come with this PR and I was not able to reproduce the issue. Could you describe how to get the image that cause the error or just post it here in zip archive? I tried TIFF images and images that are corrupted with HEX-editor but with no success.

Here is what I tried:
question

pmclain:
@TomashKhamlai I'm attaching one of the corrupt images that have been giving me trouble. I wasn't able to upload this via the admin panel. I am primarily seeing this when running imports using external URIs for images, and I have been unable to reproduce reliably enough to determine if the download is failing or needing a more reliable validation mechanism. You may need to assign the product image attribute directly in the database or replace an existing image with a corrupt image, like the attached.
[corrupt_image.zip](https://github.com/magento/magento2/files/1635123/corrupt_image.zip)

PR #13173, opened 2018-01-14
Performance: remove count() form the condition section of a loop

coderimus:
### Description
In this PR I removed the `count()` method from condition section for some loops. When this method is used in the loop condition it will be executed every iteration which is not so good for process time (performance).

N/A

### Manual testing scenarios
N/A

### Contribution checklist

### DIFF
--- a/app/code/Magento/Backend/Model/Widget/Grid/Parser.php
+++ b/app/code/Magento/Backend/Model/Widget/Grid/Parser.php
@@ -30,8 +30,8 @@ class Parser
         $expression = trim($expression);
         foreach ($this->_operations as $operation) {
             $splittedExpr = preg_split('/\\' . $operation . '/', $expression, -1, PREG_SPLIT_DELIM_CAPTURE);
-            if (!empty($splittedExpr)) {
-                $count = count($splittedExpr);
+            $count = count($splittedExpr);
+            if ($count > 1) {
                 for ($i = 0; $i < $count; $i++) {
                     $stack = array_merge($stack, $this->parseExpression($splittedExpr[$i]));
                     if ($i > 0) {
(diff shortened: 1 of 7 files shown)


ihor-sviziev (approved):

ishakhsuvarov:
Hi @Coderimus 
Please take a look at the unit test failures.

app/code/Magento/Backend/Model/Widget/Grid/Parser.php
```diff
@@ -30,8 +30,9 @@ public function parseExpression($expression)
         $expression = trim($expression);
         foreach ($this->_operations as $operation) {
             $splittedExpr = preg_split('/\\' . $operation . '/', $expression, -1, PREG_SPLIT_DELIM_CAPTURE);
-            if (count($splittedExpr) > 1) {
```
ishakhsuvarov, line 33: I assume that old check would not pass if `$splittedExpr` is and array with 1 element, while new one would pass.

ishakhsuvarov (changes requested):

app/code/Magento/Backend/Model/Widget/Grid/Parser.php
coderimus: @ishakhsuvarov thank you for your review! You are totally right. Fix uploaded.

coderimus:
@ishakhsuvarov requested changes added. Please, review them when this will be suitable for you :)

ishakhsuvarov:
Thanks @Coderimus 
Will review and proceed with the merge

ishakhsuvarov (approved):

ihor-sviziev (approved):

PR #13178, opened 2018-01-15
[FEATURE] added ability to create default/fixed value nodes during XSD Schema Validation

matthiashofele:
### Description

This feature adds the functionality / ability to use defined default values from xsd schemas in xml config nodes.

This was probably not implemented due to backwards-compatibility to PHP 5.4
The constant `LIBXML_SCHEMA_CREATE` was introduced in Libxml >= 2.6.14 (as of PHP >= 5.5.2)

As of now magento 2 has a php requirement of `7.0.2|7.0.4|~7.0.6|~7.1.0` and it should be safe to use that extended functionality to harness the full power of XSD-Schemas.

Changes applied to ` \Magento\Framework\Config\Dom::validateDomDocument`
Tests supplied in `\Magento\Framework\Config\Test\Unit\DomTest::testValidateWithDefaultValue`

See [DOMDocument::schemaValidate
](http://php.net/manual/en/domdocument.schemavalidate.php)

### Fixed Issues

XSD schemas can define default values for `xs:attribute` nodes, but the xsd-based schema validation does not merge the defined default values into the XML ``\DomDocument``  (the attribute node is not even created when xsd defines a default value but the xml document does not no reference it explicitly)

### DIFF
--- a/lib/internal/Magento/Framework/Config/Dom.php
+++ b/lib/internal/Magento/Framework/Config/Dom.php
@@ -313,7 +313,7 @@ class Dom
         libxml_set_external_entity_loader([self::$urnResolver, 'registerEntityLoader']);
         $errors = [];
         try {
-            $result = $dom->schemaValidate($schema);
+            $result = $dom->schemaValidate($schema, LIBXML_SCHEMA_CREATE);
             if (!$result) {
                 $errors = self::getXmlErrors($errorFormat);
             }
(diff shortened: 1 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13178) <br/>All committers have signed the CLA.

matthiashofele:
Closed
Moved code to 2.3-develop  -> https://github.com/magento/magento2/pull/13184

PR #13184, opened 2018-01-15
[FEATURE] added ability to create default/fixed value nodes during XSD Schema Validation

matthiashofele:
### Description

This feature adds the functionality / ability to use defined default values from xsd schemas in xml config nodes.

This was probably not implemented due to backwards-compatibility to PHP 5.4
The constant `LIBXML_SCHEMA_CREATE` was introduced in Libxml >= 2.6.14 (as of PHP >= 5.5.2)

As of now magento 2 has a php requirement of `7.0.2|7.0.4|~7.0.6|~7.1.0` and it should be safe to use that extended functionality to harness the full power of XSD-Schemas.

Changes applied to ` \Magento\Framework\Config\Dom::validateDomDocument`
Tests supplied in `\Magento\Framework\Config\Test\Unit\DomTest::testValidateWithDefaultValue`

See [DOMDocument::schemaValidate
](http://php.net/manual/en/domdocument.schemavalidate.php)

### Fixed Issues

XSD schemas can define default values for `xs:attribute` nodes, but the xsd-based schema validation does not merge the defined default values into the merged XML ``\DomDocument``

This behaviour is fixed by calling `\DOMDocument->schemaValidate()` with param `LIBXML_SCHEMA_CREATE`


magento-engcom-team:
@matthiasherold thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

matthiashofele:
To be able to use this feature effectively in production mode  the  interface `\Magento\Framework\Config\ValidationStateInterface` which defaults to `\Magento\Framework\App\Arguments\ValidationState` should always return `true` (actually checks for developer mode).

The `ValidationStateInterface` is injected and used within the class `\Magento\Framework\Config\Reader\Filesystem`

This probably will slow down the caching-process a little (done in `\Magento\Framework\Config\Data`).
Once the caching is done the possibilty of being able to define default values for xml definitions far outweights the caching slowdown.

slavvka (approved):

magento-engcom-team:
[ENGCOM-3466](https://jira.corp.magento.com/browse/ENGCOM-3466) has been created to process this Pull Request

Issue #2068, opened 2015-10-09
The Topmenu class looks procedural


adragus-inviqa:
https://github.com/magento/magento2/blob/4985acecc2a9204712531ba25b0c867ef09a4667/app/code/Magento/Theme/Block/Html/Topmenu.php

Could we please use more templates for rendering the markup? You can't easily change the HTML/menu without having to rewrite the whole class. I'm talking especially about the `_getHtml()`.

orlangur:
Nice try on refactoring this class was performed in https://github.com/magento/magento2/pull/1123.

antonkril:
One thing to consider here is performance. If Topmenu is refactored incorrectly (every item is separate template) on some installations this will result in hundreds of template inclusions (more than rest of the page). But if this will be done with one template for whole menu, all logic from class will move to that template, and there will be no noticeable complexity reduction.

okorshenko:
Closing. Considering on performance.

orlangur:
> But if this will be done with one template for whole menu as it is done in PR referenced by @orlangur, customizability will grow, but most logic from class will move to that template, and there will be no noticeable complexity reduction.

This PR did a step in right direction as to me (moving hardcoded view logic from PHP class to PHTML) but I would suggest to move further with this refactoring. The next step would be unnecessary template complexity elimination by splitting one big phtml into multiple smaller.

> One thing to consider here is performance. If Topmenu is refactored to granular rendering (every item is rendered by renderer with template) on some installations this will result in hundreds of template inclusions (more than rest of the page).

Cannot agree with that. Just like for any other renderer there is no need to create many objects and to include templates again and again. So, in optimal implementation there should be only one instance of each renderer and only one inclusion of each distinct template.

Taking into account recursion, we will need to create separate renderer for 0, 1, 2, 3 etc. menu level but no more than that. Each of them may be reused with something like `$renderer->setMenuItem(...); return $renderer->toHtml();`

JamesAnelay:
People just want to easily customise the menu in a template.

As @antonkril said "If this will be done with one template for whole menu as it is done in PR referenced by @orlangur, customizability will grow, but most logic from class will move to that template, and there will be no noticeable complexity reduction." So that's still an improvement lets do that, don't need to over engineer everything if there is a performance issue with going down that route.

PR #13238, opened 2018-01-17
change visibility of private method

brosenberger:
changing visibility of method getCategoryAsArray in order to be able to override this function in a module instead of copying the whole file.

use case: extend menu with attribute (see https://magento.stackexchange.com/questions/147114/magento-2-1-add-custom-category-attribute-to-topmenu-classes/187120) where the collection can be adapted but the resulting array which is available later in the template/block is not accessable

2. ...

### Manual testing scenarios

1. ...
2. ...

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Plugin/Block/Topmenu.php
+++ b/app/code/Magento/Catalog/Plugin/Block/Topmenu.php
@@ -154,7 +154,7 @@ class Topmenu
      * @param bool $isParentActive
      * @return array
      */
-    private function getCategoryAsArray($category, $currentCategory, $isParentActive)
+    protected function getCategoryAsArray($category, $currentCategory, $isParentActive)
     {
         return [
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13238) <br/>All committers have signed the CLA.

orlangur:
Hello @brosenberger, class is implemented this way for a reason.

It is highly discouraged to use inheritance in order to extend core functionality and especially it is not a good idea to customize core plugins (according to backward compatibility policy they can be freely changed or even removed).

Just use an accepted answer.

Another possible solution is to refactor top menu implementation properly to make it more customizable which was raised, for example, in https://github.com/magento/magento2/issues/2068. I believe such PR would have high chances to be merged.

PR #13257, opened 2018-01-18
[FIX]: Recent orders are not filtered per store at the customer account page

coderimus:
### Description
Hello,

In this PR I provide fix for filtering recent orders block where previously orders were not filtered by store. You can have different accounts with the same emails for stores belong to a website and when you logged-in to one of those stores you will see 5 recent orders but it is possible to see orders from all stores belong to a website.

**Main changes:**

1. `Magento\Sales\Block\Order\Recent`

- Move recent orders select to separate method
- Filter selected collection by current store

2. `Magento\Sales\Test\Unit\Block\Order\RecentTest`

- Change test according to the class method changes

3. `app/code/Magento/Sales/view/frontend/templates/order/recent.phtml`

- Change `sizeof()` to `count()` and execute it only once instead of double call

Best regards,
Alex

N/A

### Manual testing scenarios

1. Create 2 stores on one website
2. Create 2 accounts with the same email for previously added stores
3. Place orders per store
4. Log In to previously created account for which you created orders and check `Recent Orders` section. It will show orders for all stores

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Block/Order/Recent.php
+++ b/app/code/Magento/Sales/Block/Order/Recent.php
@@ -10,6 +10,7 @@ use Magento\Sales\Model\ResourceModel\Order\CollectionFactory;
 use Magento\Customer\Model\Session;
 use Magento\Sales\Model\Order\Config;
 use Magento\Store\Model\StoreManagerInterface;
+use Magento\Framework\App\ObjectManager;
 
 /**
  * Sales order history block
@@ -57,15 +58,16 @@ class Recent extends \Magento\Framework\View\Element\Template
         CollectionFactory $orderCollectionFactory,
         Session $customerSession,
         Config $orderConfig,
-        StoreManagerInterface $storeManager,
-        array $data = []
+        array $data = [],
+        StoreManagerInterface $storeManager = null
     ) {
         $this->_orderCollectionFactory = $orderCollectionFactory;
         $this->_customerSession = $customerSession;
         $this->_orderConfig = $orderConfig;
-        $this->storeManager = $storeManager;
-        parent::__construct($context, $data);
         $this->_isScopePrivate = true;
+        $this->storeManager = $storeManager ?: ObjectManager::getInstance()
+            ->get(StoreManagerInterface::class);
+        parent::__construct($context, $data);
     }
 
     /**
--- a/app/code/Magento/Sales/Block/Order/Recent.php
+++ b/app/code/Magento/Sales/Block/Order/Recent.php
@@ -50,8 +50,8 @@ class Recent extends \Magento\Framework\View\Element\Template
      * @param \Magento\Sales\Model\ResourceModel\Order\CollectionFactory $orderCollectionFactory
      * @param \Magento\Customer\Model\Session $customerSession
      * @param \Magento\Sales\Model\Order\Config $orderConfig
-     * @param \Magento\Store\Model\StoreManagerInterface $storeManager
      * @param array $data
+     * @param \Magento\Store\Model\StoreManagerInterface $storeManager
      */
     public function __construct(
         Context $context,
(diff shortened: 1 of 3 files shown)


app/code/Magento/Sales/Block/Order/Recent.php
```diff
@@ -28,23 +39,31 @@ class Recent extends \Magento\Framework\View\Element\Template
...
     public function __construct(
-        \Magento\Framework\View\Element\Template\Context $context,
-        \Magento\Sales\Model\ResourceModel\Order\CollectionFactory $orderCollectionFactory,
-        \Magento\Customer\Model\Session $customerSession,
-        \Magento\Sales\Model\Order\Config $orderConfig,
+        Context $context,
+        CollectionFactory $orderCollectionFactory,
+        Session $customerSession,
+        Config $orderConfig,
+        StoreManagerInterface $storeManager,
```
avoelkl: Hi @Coderimus, did you have a look to retrieve the current store id other than via adding a new dependency with StoreManager?

avoelkl: If we need this new dependency it needs to be backwards compatible, see here: http://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

avoelkl (changes requested):

avoelkl:
Hi @Coderimus,
Thanks for the PR! Please have a look at my comment/review above. Thanks!

app/code/Magento/Sales/Block/Order/Recent.php
coderimus: @avoelkl thank you so much for your response and review! 
I think I can do this without including new dependency using `$this->_storeManager` property in the Block class and leave the unit test with the Mock object of the storeManagerInterface. As I can see they are the same objects.
I will implement this change, test it and deploy to review ASAP.
Thank you,
Alex

avoelkl: Hi @Coderimus, thanks for your reply. Let me know if you need some help!

coderimus:
Hi @avoelkl, 
I provided requested changes. Please review them when it will be suitable for you :)

**Changes explanation**
I decided to implement backward compatibility and adding a new dependency to the class construct just to avoid overhead with Unit tests and do not refactor this block class too much which is also can generate more issues than benefits for this major version of Magento2.

The main reason why I added dependency and didn't use parent class property is that during block class instance initialization (while tests are running) the `_construct()` method is called and, unfortunately, it doesn't have `_storeManager` property from the parent class. In this case, I think it will be good just to save block class as it is. This block can be widely used and if we will remove/replace setOrders() Magento integrators can face problems with templates or other places where they call `$block->getOrders()`

coderimus:
Hi @avoelkl sorry if I interrupt you but when you will have free time, please, review uploaded changes. :) Thank you!

avoelkl (approved):

avoelkl:
Hi @Coderimus!
Thanks for the update! I continue with processing the PR :)

avoelkl:
Btw I saw there is a failing check from Codacy/PR Quality Review. We can ignore it in this case.

coderimus:
Hi @avoelkl,
Sorry if interrupt you but just want to ask you about next: can I prepare the backport of this issue for 2.3 and 2.4 branches? 
Have a good day,
Alex

jhruehl:
Why? You don't expect the orders to be filtered by store view. Store views are mostly used for languages, so why would you want a customer switching the language not see the orders done in another language.

This is not expected generally, but expected by devops, who clearly didn't set up their stores correctly. If you want different products in another store view, use a different website instead... sometimes I doubt people here know what they are doing.

At least now I know how to revert this "bubu".

jhruehl:
@magento-engcom-team Please revert this. It's generally not expected. Or do you expect in let's say Amazon that you don't see your most recent orders, once you switched the language?

This only helps people, who use store views as websites and probably have never touched the website functionality itself. It messes up the stores of devops, who knew what they were doing. So who do you want to help?

jhruehl:
Though you just have to remove the store_id filter from the order collection in the getRecentOrders() function in the 'magento/module-sales/Block/Order/Recent.php' to get the **genereally** expected behaviour egain.

Issue #12524, opened 2017-12-02
Custom address attribute shown on frontend


versdivers:
I have added a custom attribute to the customer address. Now this should only be visible for the admin and not for the customer.

### Preconditions
1.  Magento 2.2.1
2. PHP 7+

### Steps to reproduce
1.  Create a custom attribute programmaticly code 

```
        $eavSetup->removeAttribute('customer_address', 'your_attribute');
        $eavSetup->addAttribute('customer_address', 'your_attribute', array(
            'type' => 'varchar',
            'label' => 'Exact Address ID',
            'input' => 'text',
            'required' => false,
            'unique' => true,
            'system' => 0,
            'global' => ScopedAttributeInterface::SCOPE_WEBSITE,
            'sort_order' => '1'
        ));
```

```
        $your_attribute = $this->_attributeRepository->get('customer_address', 'your_attribute ');
        $setup->getConnection()
            ->insertOnDuplicate(
                $setup->getTable('customer_form_attribute'),
                [
                    ['form_code' => 'adminhtml_customer_address', 'attribute_id' => your_attribute ->getId()]
                ]
            );
```

2.  Fill in the new attribute in a customer
3.  go to checkout

### Expected result
1.  Its only added to adminhtml_customer_address so it should not be picked up by the knockout to display on checkout

### Actual result
1. Field is not shown in customer account
2. Field is shown at checkout 

![image](https://user-images.githubusercontent.com/18028151/33515593-c35051a6-d765-11e7-8f14-2657a0270542.png)

![image](https://user-images.githubusercontent.com/18028151/33515595-ccd32dca-d765-11e7-8e74-cce7032c6cab.png)

3. not shown in checkout 'add new address' (which is good)

And do not even try starting this conversation with 'we can not reproduce your situation. We are here to help you fix your bugs.

Prove that in de DB table it only has adminhtml_customer_address

![image](https://user-images.githubusercontent.com/18028151/33515620-01a894c2-d766-11e7-8c86-0e478887dd59.png)
![image](https://user-images.githubusercontent.com/18028151/33515624-1612883c-d766-11e7-9f70-0a793b4bc5da.png)

magento-engcom-team:
@versdivers, thank you for your report.
We've created internal ticket(s) MAGETWO-84952 to track progress on the issue.

versdivers:
Just a question but am i getting notified here when there is a fix so i can fix it waiting for an update?

shyamranpara:
[...]

PR #13291, opened 2018-01-21
Custom address attribute shown on frontend #12524

shyamranpara:
Custom attribute to the customer address in checkout should only be visible for the admin and not for the customer.

### Description
Custom Field is shown at checkout directly, there was no way to manipulate from form types codes of attribute. it should manageable from used_in_forms values of customer address attributes. I introduced new form type "checkout_saved_address" to be added in used_in_forms value if you want to display particular address attribute in checkout page in existing address list. "customer_register_address" is working in new address form in checkout already but this value should not be used in existing address list because it happens some attributes are required so need to display in new address form in checkout and need to hide in existing address list from checkout page.

### Fixed Issues

1.  [magento/magento2#12524](https://github.com/magento/magento2/issues/12524) : Custom address attribute shown on frontend  

### Manual testing scenarios

 Create new custom attribute of customer address and set this "checkout_saved_address" value in form types like in following example, if this value is set then only it will display in checkout page exist address list. 
......
 $attribute = $customerSetup->getEavConfig()->getAttribute('customer_address', 'suburb')
     ->addData(['used_in_forms' => [
         'adminhtml_customer_address',
         'customer_address_edit',
         'customer_register_address',
         'checkout_saved_address'
     ]]);
 $attribute->save();
.......

image

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/Model/DefaultConfigProvider.php
+++ b/app/code/Magento/Checkout/Model/DefaultConfigProvider.php
@@ -341,12 +341,12 @@ class DefaultConfigProvider implements ConfigProviderInterface
             foreach ($customer->getAddresses() as $key => $address) {
                 $customerData['addresses'][$key]['inline'] = $this->getCustomerAddressInline($address);
                 $custom_attributes = isset($customerData['addresses'][$key]['custom_attributes'])?$customerData['addresses'][$key]['custom_attributes']:'';
-                if($custom_attributes) {
+                if ($custom_attributes) {
                     foreach ($custom_attributes as $indexkey => $item) {
-                        if(isset($item['attribute_code'])) {
+                        if (isset($item['attribute_code'])) {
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13291) <br/>All committers have signed the CLA.

rogyar:
Hello @shyamranpara. 
Thank you for taking care of the issue. 
Unfortunately, we cannot accept the current pull request as it is because of a couple of conceptional issues:
- The customer saved addresses list is not actually a form. Using the form code for this section brings an inconsistency in the meaning of "form"
- All custom attributes are still loaded from the database and filtered only in a scope of the template. In that way we don't take into account the performance aspect of the solution. Moreover, we are adding additional collection requests

As a solution for this issue we consider a way that will avoid loading custom attributes from the database if the attributes should not be shown in the saved addresses selection list.

Issue #13292, opened 2018-01-21
All SampleData should be installed last in bin/magento setup:install


AntonEvers:
### Preconditions

1. Magento 2.1-develop
2. Magento 2.2-develop
3. Magento 2.3-develop

### Steps to reproduce

1. Create a module that saves a custom customer attribute on the event customer_address_save_before
2. Your module dependency is on 'Magento_Customer'

In my case the code is:
```php
<?php

namespace Example\Customer\Observer;

use Magento\Framework\Event\ObserverInterface;

/**
 * Class SyncNeeded
 * @package Example\Customer\Observer
 */
class SyncNeeded implements ObserverInterface
{
    /**
     * @var \Magento\Customer\Model\CustomerFactory
     */
    private $customerFactory;

    /**
     * @var \Magento\Customer\Model\ResourceModel\CustomerFactory
     */
    private $customerResourceFactory;

    /**
     * SyncNeeded constructor.
     * @param \Magento\Customer\Model\CustomerFactory $customerFactory
     * @param \Magento\Customer\Model\ResourceModel\CustomerFactory $customerResourceFactory
     */
    public function __construct(
        \Magento\Customer\Model\CustomerFactory $customerFactory,
        \Magento\Customer\Model\ResourceModel\CustomerFactory $customerResourceFactory
    ) {
        $this->customerFactory = $customerFactory;
        $this->customerResourceFactory = $customerResourceFactory;
    }

    /**
     * @param \Magento\Framework\Event\Observer $observer
     * @throws \Exception
     */
    public function execute(\Magento\Framework\Event\Observer $observer)
    {
        /** @var \Magento\Customer\Model\Address $address */
        $address = $observer->getEvent()->getCustomerAddress();
        /** @var \Magento\Customer\Model\Backend\Customer $customer */
        $customerId = $address->getCustomer()->getId();

        /** @var \Magento\Customer\Model\Customer $customer */
        $customer = $this->customerFactory->create();
        $customerData = $customer->getDataModel();
        $customerData->setId($customerId);
        $customerData->setCustomAttribute('sync_needed', 1);
        $customer->updateData($customerData);
        /** @var \Magento\Customer\Model\ResourceModel\Customer $customerResource */
        $customerResource = $this->customerResourceFactory->create();
        $customerResource->saveAttribute($customer, 'sync_needed');
    }
}
?>
```

### Expected result

1. `bin:magento setup:install` runs without errors

### Actual result

[...]

PR #13294, opened 2018-01-21
Install sample data after everything else

AntonEvers:
Install sample data after everything else while maintaining the original order among SampleData modules

1. magento/magento2#13292: All SampleData should be installed last in bin/magento setup:install
2. ...

### Manual testing scenarios

1. See #13292

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Module/ModuleList/Loader.php
+++ b/lib/internal/Magento/Framework/Module/ModuleList/Loader.php
@@ -151,6 +151,13 @@ class Loader
             }
         }
 
+        foreach ($expanded as $key => $pair) {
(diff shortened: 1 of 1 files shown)


lib/internal/Magento/Framework/Module/ModuleList/Loader.php
```diff
@@ -151,6 +151,13 @@ private function sortBySequence($origList)
             }
         }
 
+        foreach ($expanded as $key => $pair) {
+            if (preg_match('/SampleData$/', $pair['name'])) {
```
orlangur, line 155: Hi @ajpevers, I don't think such kind of magic is a good idea, let's try to figure out a proper implementation approach in initial issue.

PR #13301, opened 2018-01-21
Add @see tag to deprecated property

akiojalehto:
### Description
Add `@see` tag to deprecated property to point out the desired property

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/AttributeManagement.php
+++ b/app/code/Magento/Eav/Model/AttributeManagement.php
@@ -23,6 +23,7 @@ class AttributeManagement implements \Magento\Eav\Api\AttributeManagementInterfa
     /**
      * @var \Magento\Eav\Model\ResourceModel\Entity\Attribute\Collection
      * @deprecated 100.2.0 please use instead \Magento\Eav\Model\ResourceModel\Entity\Attribute\CollectionFactory
+     * @see $attributeCollectionFactory
      */
     protected $attributeCollection;
 


app/code/Magento/Eav/Model/AttributeManagement.php
```diff
@@ -23,6 +23,7 @@ class AttributeManagement implements \Magento\Eav\Api\AttributeManagementInterfa
     /**
      * @var \Magento\Eav\Model\ResourceModel\Entity\Attribute\Collection
      * @deprecated 100.2.0 please use instead \Magento\Eav\Model\ResourceModel\Entity\Attribute\CollectionFactory
+     * @see $attributeCollectionFactory
```
orlangur, line 26: Is such tag already widely used in core code currently? Did you check that such syntax (no "class::") picked up corectly by PHPDoc parser? Ref: https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.see.pkg.html

akiojalehto: It's working great in IDE and it seems to be already used in the core, [e.g. in File](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Product/Option/Type/File.php#L37). And it is pointing out to the current class' variable.

orlangur (approved):

PR #13302, opened 2018-01-21
Introduce resource model pool to replace UniversalFactory

schmengler:
### Description

I was digging a bit for unnecessary object instantiations and one thing I found was the good old `_init` method in collections where the class name of the resource model is passed. resource models are usually shared instances, but collections misuse a "Magento\Framework\Validator\UniversalFactory" (which has a TODO comment that it should be eliminated) to create a new instance every time.

As discussed with @antonkril I introduced a `ResourceModelPool` as replacement.

For BC, the universalFactory argument stays in place, and the resourceModelPool argument has been added at the end with `null` default. The protected _universalFactory property is not used by the core anymore and is now marked as `@deprecated`

### Contribution checklist


antonkril (approved):

lib/internal/Magento/Framework/Model/ResourceModel/ResourceModelPoolInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+interface ResourceModelPoolInterface
+{
+    /**
+     * Return instance for given class name from pool
+     *
+     * @param string $resourceClassName
+     * @return AbstractResource
+     */
+    public function get($resourceClassName): AbstractResource;
```
buskamuza: Please add `string` type hinting, as 2.3 supports PHP 7.0+

buskamuza (changes requested):

schmengler:
@buskamuza right, I updated the PR!

sivaschenko:
@schmengler sorry for slow response, can you please update your branch and resolve the conflicts

schmengler:
Will do

sivaschenko:
@schmengler see my messages in slack as well, there is a rebased branch with this pull request that you can use https://github.com/sivaschenko/magento2/tree/resource-model-pool-rebased (but I didn't want to force push to your branch)

schmengler:
@sivaschenko So I did it now :)

sivaschenko:
Hi @schmengler running `bin/magento setup:di:compile` results in the following error:

    Invalid ExtensionInterface for nonexistent class Interface in file app/code/Magento/TestModuleExtensionAttributes/Api/Data/FakeExtensibleTwoInterface.php

lib/internal/Magento/Framework/Model/ResourceModel/ResourceModelPool.php
```diff
@@ -0,0 +1,36 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Framework\Model\ResourceModel;
```
ihor-sviziev, line 7: Shouldn't we have strict_types=1 there?

PR #13308, opened 2018-01-22
Improve getCustomAttribute() performance

schmengler:
### Description

When traversing a product collection and using `getCustomAttribute()` within the loop, it triggered a database query each time to retrieve the metadata from the `eav_attributes` table. This was due to the custom attributes being cached in each product model separately. 

I moved it into the resource model so that only one query per entity type is necessary.

### Manual testing scenarios

1. On an instance with sample data or more than a handful of products
2. Profile this code snippet:

        $productCollection = $this->productCollectionFactory->create();
        $productCollection->addAttributeToSelect('*');
        foreach ($productCollection as $product) { /** @var ProductInterface $product */
            $product->getCustomAttributes();
        }

More background and a test tool can be found in this blog by my colleague @avstudnitz who found the issue: https://www.integer-net.com/performance-trap-magento-2-getcustomattribute-function/

After the changes in this PR, database calls should been reduced by (n-1) with n = number of products.

### Contribution checklist


schmengler:
The same mechanism was used for categories, I updated the PR to reflect the changes there analogously

AVoskoboinikov:
Hi @schmengler 
I've cloned your PR to my local branch so I am able to use our internal infrastructure to test performance improvement. Will be back with results.

AVoskoboinikov:
references for internal teams:
[L1](https://bamboo.corp.magento.com/browse/MCCE23-L1132-1) bamboo build
[L2](https://bamboo.corp.magento.com/browse/MCCE23-L2140-1) bamboo build
[L3](https://bamboo.corp.magento.com/browse/MCCE23-L3143-1) bamboo build
[L4](https://bamboo.corp.magento.com/browse/MCCE23-L4136-1) bamboo build
[PAT](https://bamboo.corp.magento.com/browse/MCCE23-PAT122-1) bamboo build
[Sample Data](https://bamboo.corp.magento.com/browse/MCCE23-SDT125-1) bamboo build
[SVC](https://bamboo.corp.magento.com/browse/SVC-SVC23184-1) bamboo build

AVoskoboinikov:
I confirm that PR has improvements in next scenarios:
- Server Side EE API (Get Products - response time) - 41.5%
- Server Side EE API (Get Products - created objects count) - 69.5%
- Server Side EE API (Get Products - select requests count) - 23.6%

- Server Side EE Catalog GraphQL (Query multiple products - response time) - 50%
- Server Side EE Catalog GraphQL (Query multiple products - created objects count) - 79.7%

- Server Side EE Catalog GraphQL (Query configurable product - response time) - 4%
- Server Side EE Catalog GraphQL (Query configurable product - created objects count) - 19.9%

- Server Side EE Catalog GraphQL (Query products with full text search and filter - response time) - 4%
- Server Side EE Catalog GraphQL (Query products with full text search and filter - created objects count) - 60.7%
- Server Side EE Catalog GraphQL (Query products with full text search and filter - select requests count) - 14.9%

- Server Side EE Catalog GraphQL (Query products with full text search only - response time) - 42%
- Server Side EE Catalog GraphQL (Query products with full text search only - created objects count) - 76.3%

app/code/Magento/Catalog/Model/Category.php
```diff
@@ -69,23 +70,6 @@ class Category extends \Magento\Catalog\Model\AbstractModel implements
...
-    const KEY_LEVEL = 'level';
-    const KEY_UPDATED_AT = 'updated_at';
-    const KEY_CREATED_AT = 'created_at';
-    const KEY_PATH = 'path';
-    const KEY_AVAILABLE_SORT_BY = 'available_sort_by';
-    const KEY_INCLUDE_IN_MENU = 'include_in_menu';
-    const KEY_PRODUCT_COUNT = 'product_count';
-    const KEY_CHILDREN_DATA = 'children_data';
-    /**#@-*/
-
```
maghamed, line 88: constants removal from class and moving them to interface is backward incompatible

app/code/Magento/Catalog/Model/ResourceModel/Category.php
```diff
@@ -82,6 +85,16 @@ class Category extends AbstractResource
      */
     protected $aggregateCount;
 
+    /**
+     * @var CategoryAttributeRepositoryInterface
+     */
+    protected $metadataService;
```
maghamed: new properties added to classes should be private, protected properties are now allowed

app/code/Magento/Eav/Model/Entity/AbstractEntity.php
```diff
@@ -1958,4 +1959,25 @@ protected function loadAttributesForObject($attributes, $object = null)
...
         }
     }
+
+    /**
+     * Receive a list of EAV attributes using provided metadata service.
+     *
+     * @param MetadataServiceInterface $metadataService
+     * @return string[]
+     */
+    protected function getEavAttributesCodes(MetadataServiceInterface $metadataService)
```
maghamed: Inheritance based API not encouraged in Magento 2

app/code/Magento/Catalog/Model/ResourceModel/Product.php
```diff
@@ -83,6 +85,16 @@ class Product extends AbstractResource
      */
     private $productCategoryLink;
 
+    /**
+     * @var ProductAttributeRepositoryInterface
+     */
+    protected $metadataService;
```
maghamed: introducing of protected properties/methods are not encouraged in M2

maghamed (changes requested):
a lot of BiC being introduced in the scope of performance improvement.
It looks better to introduce dedicated service for `getCustomAttributesCodes` logic. 
And move all the logic there.
This service could be injected into Product model which will proxy call into it.
No need to bring this logic to Product/Category resource models which contain a lot of logic now (violating Single Responsibility), breaking their contract along with that.
The more granular single responsible services we introduce in M2 codebase - the easier it would be maintain this code in future

app/code/Magento/Catalog/Model/Category.php
schmengler: why? `TheClass::THE_CONSTANT` still works

app/code/Magento/Catalog/Model/ResourceModel/Category.php
schmengler: good catch, I'll change it

app/code/Magento/Eav/Model/Entity/AbstractEntity.php
schmengler: I know but I have to work with what is there (which is inheritance based resource models) or introduce huge changes, probably BIC. Do you have an idea for a simple solution?

schmengler: it's basically moved from `\Magento\Framework\Model\AbstractExtensibleModel::getEavAttributesCodes` (inheritance based model API) - so it's not that I am introducing something new here.

A possible non-inheritance solution I see is to to implement the method in the Category and Product resource models seperately in a non-generic way. Then, if we want to get rid of duplicated logic, add a method `getCustomAttributeCodes($dataObjectClassName)` to `\Magento\Framework\Api\MetadataServiceInterface` (BIC: new method in interface). IMHO this would be subject to a different PR

schmengler: disregard this, I've read your top comment

schmengler:
> It looks better to introduce dedicated service for getCustomAttributesCodes logic.
> And move all the logic there.
> This service could be injected into Product model which will proxy call into it.
> No need to bring this logic to Product/Category resource models which contain a lot of  logic now (violating Single Responsibility), breaking their contract along with that.

I'm going to think this through

schmengler:
@maghamed Here's my draft for such a service. I'd like to discuss it before updating the PR:

```php
<?php
/**
 */

namespace Magento\Eav\Model\Entity;


use Magento\Framework\Api\MetadataServiceInterface;

class GetCustomAttributeCodes
{
    /**
     * @var string[]
     */
    private $customAttributesCodes;

    /**
     * Receive a list of custom EAV attributes using provided metadata service. The results are cached per entity type
     *
     * @param MetadataServiceInterface $metadataService Custom attribute metadata service to be used
     * @param string[] $interfaceAttributes Attribute codes that are part of the interface and should not be
     *                                      considered custom
     * @param string|null $entityType Entity type (class name), only needed if metadata service handles different
     *                                entities
     * @return string[]
     */
    public function execute(
        MetadataServiceInterface $metadataService,
        array $interfaceAttributes,
        string $entityType = null
    ): array {
        $cacheKey = get_class($metadataService) . '|' . $entityType;
        if (!isset($this->customAttributesCodes[$cacheKey])) {
            $customAttributesCodes = $this->getEavAttributesCodes($metadataService, $entityType);
            $this->customAttributesCodes[$cacheKey] = array_values(
                array_diff($customAttributesCodes, $interfaceAttributes)
            );
        }
        return $this->customAttributesCodes;
    }

    /**
     * Receive a list of EAV attributes using provided metadata service.
     *
     * @param MetadataServiceInterface $metadataService
     * @param string|null $entityType
     * @return string[]
     */
    private function getEavAttributesCodes(MetadataServiceInterface $metadataService, string $entityType = null)
    {
        $attributeCodes = [];
        $customAttributesMetadata = $metadataService->getCustomAttributesMetadata($entityType);
        if (is_array($customAttributesMetadata)) {
            /** @var $attribute \Magento\Framework\Api\MetadataObjectInterface */
            foreach ($customAttributesMetadata as $attribute) {
                $attributeCodes[] = $attribute->getAttributeCode();
            }
        }
        return $attributeCodes;
    }
}
``` 

This could be injected into the `Product` and `Category` models as you suggested and their `getCustomAttributeCodes()` would be replaced with
```php
        return $this->getCustomAttributesCodes->execute(
            $this->metadataService,
            ProductInterface::ATTRIBUTES
        );
```

What I don't like about it is that the caller (the product/category model) is responsible to pass the non-custom interface attributes. This is something that the custom attribute metadata service should be able to provide, don't you think? Again, adding a method to `\Magento\Framework\Api\MetadataServiceInterface` would be a major BIC, which I have avoided so far.

What do you think?

Also just to be sure: what's the consensus on state in services? Because caching the results is absolutely necessary.

app/code/Magento/Catalog/Model/ResourceModel/Category.php
```diff
@@ -113,8 +128,21 @@ public function __construct(
...
+
+    public function getCustomAttributesCodes()
+    {
+        if ($this->customAttributesCodes === null) {
+            $this->customAttributesCodes = $this->getEavAttributesCodes($this->metadataService);
+            $this->customAttributesCodes = array_values(
+                array_diff($this->customAttributesCodes, CategoryInterface::ATTRIBUTES)
+            );
+        }
+        return $this->customAttributesCodes;
```
jameshalsall: In the case of products we can improve this even further for larger datasets, as it stands here it will work great for a small number of attributes (< 500) but as soon as we start to go upward of that figure the performance will fall through the floor (each attribute loaded results in `Magento\Eav\Model\ResourceModel\Entity\Attribute::_afterLoad` being called which in turn runs a select query to join additional attribute data).

In my current project we have ~2200 custom attributes, but all of our attribute sets are small (< 100 custom attributes in each one), the problem is that this method loads every attribute regardless of the product's attribute set, to fix it I pulled in this PR as a patch and then changed this method to the following:

```php
    public function getCustomAttributesCodes($attributeSetId)
    {
        if (!isset($this->customAttributesCodes[$attributeSetId])) {
            $codes = [];
            $criteria = $this->searchCriteriaBuilder->addFilter('attribute_set_id', $attributeSetId, 'eq');
            $attributes = $this->attributeRepository->getList($criteria->create())->getItems();

            if (is_array($attributes)) {
                /** @var $attribute \Magento\Framework\Api\MetadataObjectInterface */
                foreach ($attributes as $attribute) {
                    $codes[] = $attribute->getAttributeCode();
                }
            }

            $codes = array_values(
                array_diff($codes, ProductInterface::ATTRIBUTES)
            );

            $this->customAttributesCodes[$attributeSetId] = $codes;
        }

        return $this->customAttributesCodes[$attributeSetId];
    }
```

then had the `Magento\Catalog\Model\Product` pass through the product's attribute set ID, this way we only ever load the attributes for the current attribute set

I think the memoization based on attribute set ID in the `$this->customAttributesCodes[$attributeSetId]` is not required because resource model instances are not shared across products, but I might be wrong  so I've left it in there for safety

whilst this would result in a few more queries (1 for each different attribute set to get the attributes relevant for that product) it means that creating a product is not affected by the total number of attributes in Magento, only by the number of attributes in the product's attribute set

schmengler: > because resource model instances are not shared across products, 

They are, this is what my initial approach is based on. Thanks a lot for your addition, it makes sense and I'd add it to the PR!

But first waiting for @maghamed to discuss the final architecture

app/code/Magento/Catalog/Model/Category.php
maghamed: @schmengler 
Possible fatal error if there is another implementation of the interface with same constant name declaration.
But this is more like the theoretical issue.

Currently, I raised this question for voting on the Internal Architectural Council to allow moving constants **preserving names and values** from Implementation to Interface being implemented. And thus fix our Backward Compatibility Guideline http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html saying 
```
Removing or renaming constants
Do not remove or rename constants.
```
And to adapt internal tests which break on such changes.

I will get back to you shortly as soon as get results.

maghamed:
Hey @schmengler ,

Your `GetCustomAttributeCodes` service looks good to me.
Not sure that we need to provide third parameter `string $entityType = null`
Actually, its presence in MetadataInterface also looks excessive as Default implementation not use it either.
```
/**
 * Default implementation of metadata service, which does not return any real attributes.
 */
class DefaultMetadataService implements MetadataServiceInterface
{
    /**
     * {@inheritdoc}
     */
    public function getCustomAttributesMetadata($dataObjectClassName = null)
    {
        return [];
    }
}
```

>What I don't like about it is that the caller (the product/category model) is responsible to pass the non-custom interface attributes. This is something that the custom attribute metadata service should be able to provide, don't you think?

I agree with you that it's the responsibility of Metadata Service to know what attributes belong to Interface and what attributes are Custom ones. But as you mentioned we can't change existing interface because of BC requirements. 

What we can do is to introduce interface for your service like this 
```
interface GetCustomAttributeCodesInterface
{
    /**
      * Only one parameter - Metadata Service. No need to pass interface Attributes
      **/
    public function execute(MetadataServiceInterface $metadataService): array;
}
```

along with that you provide the implementation 
```
class GetCustomAttributeCodes implements GetCustomAttributeCodesInterface
{
    /**
     * Receive a list of custom EAV attributes using provided metadata service. The results are cached per entity type
     *
     * @param MetadataServiceInterface $metadataService Custom attribute metadata service to be used
     *                                entities
     * @return string[]
     */
    public function execute(
        MetadataServiceInterface $metadataService
    ): array {
        $cacheKey = get_class($metadataService) ;
        if (!isset($this->customAttributesCodes[$cacheKey])) {
            $this->customAttributesCodes[$cacheKey]  = $this->getEavAttributesCodes($metadataService);
        }
        return $this->customAttributesCodes[$cacheKey];
    }

    /**
     * Receive a list of EAV attributes using provided metadata service.
     *
     * @param MetadataServiceInterface $metadataService
     * @return string[]
     */
    private function getEavAttributesCodes(MetadataServiceInterface $metadataService)
    {
        $attributeCodes = [];
        $customAttributesMetadata = $metadataService->getCustomAttributesMetadata();
        if (is_array($customAttributesMetadata)) {
            /** @var $attribute \Magento\Framework\Api\MetadataObjectInterface */
            foreach ($customAttributesMetadata as $attribute) {
                $attributeCodes[] = $attribute->getAttributeCode();
            }
        }
        return $attributeCodes;
    }
}
```

And introduce new Composite services for Product and Category.

```
class GetProductCustomAttributeCodes implements GetCustomAttributeCodesInterface
{

    /**
     * @param GetCustomAttributeCodes $baseCustomAttributeCodes
     */
    public function __construct(
        GetCustomAttributeCodes $baseCustomAttributeCodes
    ) {
        $this->baseCustomAttributeCodes = $baseCustomAttributeCodes;
    }


    public function execute(
        MetadataServiceInterface $metadataService
    ): array {
        $customAttributesCodes = $this->baseCustomAttributeCodes->execute($metadataService);
        return array_diff($customAttributesCodes, ProductInterface::ATTRIBUTES);
    }
}

class GetCategoryCustomAttributeCodes implements GetCustomAttributeCodesInterface
{

    /**
     * @param GetCustomAttributeCodes $baseCustomAttributeCodes
     */
    public function __construct(
        GetCustomAttributeCodes $baseCustomAttributeCodes
    ) {
        $this->baseCustomAttributeCodes = $baseCustomAttributeCodes;
    }


    public function execute(
        MetadataServiceInterface $metadataService
    ): array {
        $customAttributesCodes = $this->baseCustomAttributeCodes->execute($metadataService);
        return array_diff($customAttributesCodes, CategoryInterface::ATTRIBUTES);
    }
}
```

Now you make both Product and Category model depend on `GetCustomAttributeCodesInterface` interface.

And with DI Type you change the incoming object.
```
    <type name="Magento\Catalog\Model\Product">
        <arguments>
            <argument name="getCustomAttributeCode" xsi:type="object">GetProductCustomAttributeCodes</argument>
        </arguments>
    </type>
```
Do the same for Category.

So, summarizing all the above.
- we don't make backward incompatible changes
- we introduce a new interface for service which returns CustomAttributeCodes which uses Metadata. On the level of the interface we don't provide awareness from where we take the info about entity attributes, i.e. Lately we can use Metadata service for it.
- Each entity provides own implementation of this service.

Do proposed solution make sense for you?

schmengler:
@maghamed Thank you for the detailed response, your suggestion sounds good!

Regarding the type argument:

> Actually, its presence in MetadataInterface also looks excessive as Default implementation not use it either.

It's used in the implementation for address attributes, although on second look there is a default value and other values are never passed. I'll leave it out then.

schmengler:
@maghamed About the addition by @jameshalsall : I could add an optional `attributeSet`argument to the `execute` method but we would need to pass it along to `Magento\Framework\Api\MetadataServiceInterface::getCustomAttributesMetadata()` (i.e. add an optional argument to the interface method and its implementations). But I'm not sure if it's always a good idea and maybe too much for this PR. It could be done in a second one, and here I'd already add the optional parameter in the new interface to be prepared.

jameshalsall:
If you make it easy for me to apply that fix later (by allowing the parameter in the interface method) then I can send a separate PR after this one is merged 👍 thanks

maghamed (approved):

magento-engcom-team:
Hi @maghamed, thank you for the review.
[MAGETWO-88021](https://jira.corp.magento.com/browse/MAGETWO-88021) has been created to process this Pull Request

app/code/Magento/Catalog/Model/Category.php
```diff
@@ -105,11 +105,6 @@ class Category extends \Magento\Catalog\Model\AbstractModel implements
      */
     protected $_urlRewrite;
 
-    /**
-     * @var ResourceModel\Category
-     */
-    protected $_resource;
-
```
schmengler: while this is not necessary and not relevant for this PR anymore, it helps with static analysis and IDE support by specifying the concrete type of `$_resource`

app/code/Magento/Catalog/Model/Category.php
```diff
@@ -317,16 +313,6 @@ protected function getCustomAttributesCodes()
...
-    /**
-     * @throws \Magento\Framework\Exception\LocalizedException
-     * @return \Magento\Catalog\Model\ResourceModel\Category
-     * @deprecated because resource models should be used directly
-     */
-    protected function _getResource()
-    {
-        return parent::_getResource();
-    }
-
```
schmengler: same as with `$_resource`. It's not super important but a minor improvement that I made while I was at it. Should it be a separate PR?

app/code/Magento/Catalog/Test/Unit/Model/CategoryTest.php
```diff
@@ -424,39 +441,33 @@ public function testReindexFlatDisabled(
...
     {
-        $interfaceAttributeCode = 'name';
-        $customAttributeCode = 'description';
-
-        $this->resource
-            ->method('getCustomAttributesCodes')
-            ->willReturn([$customAttributeCode]);
-        $this->category->setData($interfaceAttributeCode, "sub");
+        $nameAttributeCode = 'name';
+        $descriptionAttributeCode = 'description';
```
schmengler: why did you rename the variables? For the test it's not important that it's name and description, but that one is an interface attribute and the other is custom

app/code/Magento/Catalog/Test/Unit/Model/ProductTest.php
```diff
@@ -1260,40 +1278,31 @@ public function testGetMediaGalleryImagesMerging()
...
     {
-        $interfaceAttributeCode = 'price';
-        $customAttributeCode = 'color';
-
-        $this->resource
-            ->method('getCustomAttributesCodes')
-            ->willReturn([$customAttributeCode]);
-        $this->model->setData($interfaceAttributeCode, 10);
+        $priceCode = 'price';
+        $colorAttributeCode = 'color';
```
schmengler: same as with "name" and "description" above. you probably reverted both to the previous version

app/code/Magento/Catalog/Test/Unit/Model/ProductTest.php
```diff
@@ -1260,40 +1278,31 @@ public function testGetMediaGalleryImagesMerging()
...
+        $this->getCustomAttributeCodes->expects($this->exactly(3))
+            ->method('execute')
+            ->willReturn([$colorAttributeCode]);
+        $this->model->setData($priceCode, 10);
 
         //The color attribute is not set, expect empty custom attribute array
         $this->assertEquals([], $this->model->getCustomAttributes());
 
         //Set the color attribute;
-        $initialCustomAttribueValue = "red";
```
schmengler: the variables for "initial value" and "new value" wer introduced to make connection between fixture and verification explicit and clear

schmengler:
@jameshalsall ooops, too late, sorry!

app/code/Magento/Catalog/Model/Category.php
schmengler: @maghamed  It wouldn't even be a problem: https://3v4l.org/H4mYd :smile:

mslabko:
Hi @jameshalsall ,
Do you plan to create PR with optimization of loading attributes?

jameshalsall:
Yeah I will do this at the weekend.

mslabko:
Hi @jameshalsall ,
We are going to start regression cycle for 2.3.0 and this performance improvement is very important for us because it affects a lot of test cases.
Will you have time to finish this work until 14/05 or we can do it from our side?

jameshalsall:
@mslabko I will take a look at it now and see how much effort is involved

jameshalsall:
> I could add an optional attributeSetargument to the execute method but we would need to pass it along to Magento\Framework\Api\MetadataServiceInterface::getCustomAttributesMetadata() (i.e. add an optional argument to the interface method and its implementations). But I'm not sure if it's always a good idea and maybe too much for this PR. It could be done in a second one, and here I'd already add the optional parameter in the new interface to be prepared.

so it seems the option `$attributeSetId` parameter was not added in this PR, so I will assume it is safe to add it in the new PR - I don't think this is a problem because 2.3 is not yet released.

mslabko:
@jameshalsall , yes, for 2.3 it's not a problem to add second argument for GetCustomAttributeCodesInterface::execute.
But it must be required argument due to we do not need to load all attributes per entity, only for specified attribute set.

Also, I recommend do not use MetadataServiceInterface for retrieving attributes, due to this is not optimal from a performance perspective. Instead of MetadataServiceInterface it's better to use  \Magento\Eav\Model\Config::getEntityAttributes, which already has a caching mechanism - permanent and application cache. 
On 2.2-develop similar approach give us from 56-95% impoovements for all "product-scenarios" on medium profile


So I suggetst refactor GetCustomAttributeCodesInterface as 
```
—  execute(MetadataServiceInterface $metadataService)
++ execute(int $attributeSetId)
```

jameshalsall:
oops, I already added - https://github.com/magento/magento2/pull/15135

PR #13323, opened 2018-01-23
Bugfix/13266 Update the processing of items for the top menu

arnoudhgz:
### Description
The counter in the foreach loop was not being calculated correctly. This resolved in the 'last' css class not being set on the output when one or more parent items are not active.

With this PR the items will be filtered first before they are being processed for the html output.

1. magento/magento2#13266: Topmenu 'last' class not being set if the a parent is inactive

### Manual testing scenarios
1. Do a clean install of Magento 2.2 with sample data
2. Set a level 1 category (e.g. Gear or Gift Cards) inactive
3. Refresh the appropriate cache
4. Inspect the menu and see there is a class last on the **last** active top menu item

### Contribution checklist

### DIFF
--- a/app/code/Magento/Theme/Block/Html/Topmenu.php
+++ b/app/code/Magento/Theme/Block/Html/Topmenu.php
@@ -6,6 +6,7 @@
 namespace Magento\Theme\Block\Html;
 
 use Magento\Framework\Data\Tree\Node;
+use Magento\Framework\Data\Tree\Node\Collection;
 use Magento\Framework\Data\Tree\NodeFactory;
 use Magento\Framework\Data\TreeFactory;
 use Magento\Framework\DataObject\IdentityInterface;
@@ -29,7 +30,7 @@ class Topmenu extends Template implements IdentityInterface
     /**
      * Top menu data tree
      *
-     * @var \Magento\Framework\Data\Tree\Node
+     * @var Node
      */
     protected $_menu;
 
@@ -164,7 +165,7 @@ class Topmenu extends Template implements IdentityInterface
     /**
      * Add sub menu HTML code for current menu item
      *
-     * @param \Magento\Framework\Data\Tree\Node $child
+     * @param Node $child
      * @param string $childLevel
      * @param string $childrenWrapClass
      * @param int $limit
@@ -192,7 +193,7 @@ class Topmenu extends Template implements IdentityInterface
     /**
      * Recursively generates top menu html from data that is specified in $menuTree
      *
-     * @param \Magento\Framework\Data\Tree\Node $menuTree
+     * @param Node $menuTree
      * @param string $childrenWrapClass
      * @param int $limit
      * @param array $colBrakes
@@ -202,7 +203,7 @@ class Topmenu extends Template implements IdentityInterface
      * @SuppressWarnings(PHPMD.NPathComplexity)
      */
     protected function _getHtml(
-        \Magento\Framework\Data\Tree\Node $menuTree,
+        Node $menuTree,
         $childrenWrapClass,
         $limit,
         $colBrakes = []
@@ -213,6 +214,10 @@ class Topmenu extends Template implements IdentityInterface
         $parentLevel = $menuTree->getLevel();
         $childLevel = $parentLevel === null ? 0 : $parentLevel + 1;
 
+        if ($childLevel === 0) {
+            $this->removeOrphans($children);
+        }
+
         $counter = 1;
         $itemPosition = 1;
         $childrenCount = $children->count();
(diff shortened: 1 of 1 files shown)


app/code/Magento/Theme/Block/Html/Topmenu.php
```diff
@@ -213,18 +214,19 @@ protected function _getHtml(
...
 
-        /** @var \Magento\Framework\Data\Tree\Node $child */
+        /** @var Node $child */
         foreach ($children as $child) {
-            if ($childLevel === 0 && $child->getData('is_parent_active') === false) {
-                continue;
-            }
             $child->setLevel($childLevel);
             $child->setIsFirst($counter == 1);
             $child->setIsLast($counter == $childrenCount);
```
orlangur, line 232: Approach looks generally good to me but I don't the idea of separate filtering loop.

I believe instead of removing `if` and introducing filtering we need just to track last item properly.

For example, remove `$childrenCount` variable, declare `$lastChild = null` before loop, do 
```
$child->setIsLast(false);
$lastChild = $child;
```
in loop and
```
if ($lastChild) {
    $lastChild->setIsLast(true);
}
```
after it.

As always, please squash changes into single commit and force push after changes are performed.

orlangur (dismissed):

app/code/Magento/Theme/Block/Html/Topmenu.php
arnoudhgz: @orlangur thank you for your feedback.

Your approach was also the first thing that I wanted to do, but that does not work because the `$html` with all the html for the `$schildren` is generated in the `foreach` loop.  

If you do the `setIsLast()` after the loop the class `last` is not added to the last visible menu item (which was the whole point for this commit ;) )

_I squashed the commits_ :+1:

arnoudhgz:
@orlangur can you please take a look at my response?

orlangur:
@arnoudhgz yeah-yeah, I saw your response and even started to reply, will search for a tab in my browser :) Sorry for delay and thanks for quicks feedback 👍

arnoudhgz:
Removed the label, the code does not need the update as suggested. There should be taken a closer look in the issue and the proposed solution in this commit.

The suggested change by @orlangur will not help to solve the issue, because the HTML is generated inside the foreach loop.

orlangur:
Label should be removed no earlier than it is confirmed by reviewer. Current approach is not acceptable from computational complexity perspective, I was checking the best way to split loop into two, will share exact suggestion soon.

arnoudhgz:
Great! A solution with perfect code that will also solve the bug has my preference too 😉

arnoudhgz:
@orlangur do you have any idea when you can take a look into this?

app/code/Magento/Theme/Block/Html/Topmenu.php
orlangur: 1. Please remove `$itemPosition` as it seems to be unused
1. Track `$lastChild` as requested earlier
1. Split one loop into two: in first we process children, in second generate actual HTML with concatenation 
1. Feel free to introduce new private methods to get rid of `@SuppressWarnings(PHPMD.CyclomaticComplexity)` and `@SuppressWarnings(PHPMD.NPathComplexity)`

orlangur (changes requested):

orlangur:
Hi @arnoudhgz, please check my previous comment :)

arnoudhgz:
@orlangur the splitting into separate private function would be a good thing to do. But I can't agree with the tracking of the `$lastChild`, you will have to overwrite the last item in the array again with this variable. I can't with a good conscience loop twice over the array in your suggested way. 

In my approach of making the array first smaller you have to loop over less items in the second `foreach`. This seems cleaner, more efficient. 

I don't get the thing about the  `computational complexity perspective`, you mentioned earlier, can you make it more specific what I did wrong compared to that perspective?

Before I start working again on making this function cleaner in more separate private functions can you elaborate more about your suggested approach?

Maybe you can make an example in (pseudo) code...

orlangur:
Hi @arnoudhgz, could you please recreate this branch?

During review I noticed that `$children->delete($child);` is `O(1)` in fact thus I was wrong about nonlinear complexity.

The only thing I would ask you to do is rename `removeOrphans` to something like `removeChildrenWithoutActiveParent` - to not introduce a new term "orphan". 

I lost your PR somewhere among hundreds of open browser tabs, sorry about that.

arnoudhgz:
@orlangur yes, no problem. I will add it again.

Should I add the `declare(strict_types=1);` to the PHP file, or is that only for the `2.3-develop` branch?

orlangur:
@arnoudhgz you can but it's mandatory for newly added files only.

PR #13337, opened 2018-01-23
Add missing PHPDocs for methods

akiojalehto:
### Description
Add missing PHPDocs

### Contribution checklist

### DIFF
--- a/app/code/Magento/InstantPurchase/Model/ShippingMethodChoose/DeferredShippingMethodChooserPool.php
+++ b/app/code/Magento/InstantPurchase/Model/ShippingMethodChoose/DeferredShippingMethodChooserPool.php
@@ -15,6 +15,9 @@ class DeferredShippingMethodChooserPool
 {
     private $choosers;
 
+    /**
+     * @param DeferredShippingMethodChooserInterface[] $choosers
+     */
     public function __construct(array $choosers)
     {
         foreach ($choosers as $chooser) {
@@ -28,6 +31,10 @@ class DeferredShippingMethodChooserPool
         $this->choosers = $choosers;
     }
 
+    /**
+     * @param string $type
+     * @return DeferredShippingMethodChooserInterface
+     */
     public function get($type) : DeferredShippingMethodChooserInterface
     {
         if (!isset($this->choosers[$type])) {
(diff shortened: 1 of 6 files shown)


app/code/Magento/InstantPurchase/Model/ShippingMethodChoose/DeferredShippingMethodChooserPool.php
```diff
@@ -28,6 +31,10 @@ public function __construct(array $choosers)
         $this->choosers = $choosers;
     }
 
+    /**
+     * @param $type
```
oliverklee: They type (`mixed`?) is missing.

akiojalehto: Type is now added 5b51e18

app/code/Magento/InstantPurchase/Model/ShippingMethodChoose/DeferredShippingMethodChooserPool.php
```diff
@@ -15,6 +15,9 @@ class DeferredShippingMethodChooserPool
 {
     private $choosers;
 
+    /**
+     * @param array $choosers
```
orlangur: Please be more specific about type, `DeferredShippingMethodChooserInterface[]` can be used here instead of generic array.

app/code/Magento/InstantPurchase/Model/ShippingMethodChoose/DeferredShippingMethodChooserPool.php
```diff
@@ -28,6 +31,10 @@ public function __construct(array $choosers)
         $this->choosers = $choosers;
     }
 
+    /**
+     * @param mixed $type
```
orlangur: `mixed` pseudo-type must be avoided. Please specify exact type(s).

orlangur (dismissed):

app/code/Magento/InstantPurchase/Model/ShippingMethodChoose/DeferredShippingMethodChooserPool.php
akiojalehto: Both fixed in 885f032

orlangur:
@akiojalehto thanks for quick changes! Could you please also squash all changes in a single commit and force push? (generally, it is a good thing to do always unless you want to specifically consolidate some changes in an atomic commit).

akiojalehto:
@orlangur Sure, commits are now squashed.

Issue #5258, opened 2016-06-25
Const PHP_BINARY breaks integration testing in windows since 2.1


martinmonnier:
**Preconditions**
- Install your PHP in the common windows directory e.g. "C:\Program Files (x86)\Zend\ZendServer\bin" 
- Create a simple integration test and run this test.

**Expected result**
- Test execution 

**Actual result**
- The test could not be executed because the php interpreter "C:\Program " could not be started during preparing the integration enviroment.

**Reason**
- In File "\dev\tests\integration\framework\Magento\TestFramework\Application.php" the `$this->_shell->execute()`parameter uses the PHP_BINARY const without any double qoutes.
- The error does not occurs on 2.0 because PHP_BINARY const were not used. Just the simple "php" command that will be executed by PATH variable.

**Solution**
- Escaping PHP_BINARY const, so it will be called safely by $this->_shell->execute(). e.g. with `escapeshellarg()`

martinmonnier:
Stacktrace on test execution:
`Exception: Der Befehl "C:\Program" ist entweder falsch geschrieben oder
konnte nicht gefunden werden. in D:\Projekte\Magento\public_html\mycustomer\shop\vendor\magento\framework\Shell.php:61
Stack trace:
#0 D:\Projekte\Magento\public_html\mycustomer\shop\dev\tests\integration\framework\Magento\TestFramework\Application.php(464): Magento\Framework\Shell->execute('C:\\Program File...', Array)
#1 D:\Projekte\Magento\public_html\mycustomer\shop\dev\tests\integration\framework\bootstrap.php(68): Magento\TestFramework\Application->install()
#2 D:\Projekte\Magento\public_html\mycustomer\shop\vendor\phpunit\phpunit\src\Util\Fileloader.php(93): include_once('D:\\Projekte\\Mag...')
#3 D:\Projekte\Magento\public_html\mycustomer\shop\vendor\phpunit\phpunit\src\Util\Fileloader.php(77): PHPUnit_Util_Fileloader::load('D:\\Projekte\\Mag...')
#4 D:\Projekte\Magento\public_html\mycustomer\shop\vendor\phpunit\phpunit\src\TextUI\Command.php(806): PHPUnit_Util_Fileloader::checkAndLoad('D:\\Projekte\\Mag...')
[...]

PR #13340, opened 2018-01-24
Added support for PHP paths with spaces in Magento CLI

PivitParkour94:
### Description
Found that some Magento CLI commands (`bin/magento deploy:mode:set production` for example) caused another command to be run, which is all good and fine except because it is then executed at a low level `exec` function call, if the PHP_BINARY variable had a space in it, it would cause it to break.

Originally solved it simply escaping all spaces in the final command string, but that's a terrible security principle, not to mention it causes any arguments to not be recognised.

End solution seems to be as simple as wrapping the `\Magento\Deploy\Model\Filesystem::$functionCallPath` in quotes.

Note: If the current state of having M2 running on Windows is that it won't happen that's a real shame, but as far as I know you can have spaces in filepaths on OSX or *nix (you probably wouldn't, but it's at least possible)

Rather than creating an issue, I just solved it with a PR because it is so simple

### Manual testing scenarios
1. Have a space in your php path
2. Run `bin/magento deploy:mode:set production -vvv`
3. See that the command runs as expected

### Contribution checklist
 - [ x ] All new or changed code is covered with unit/integration tests (if applicable)

### DIFF
--- a/app/code/Magento/Deploy/Model/Filesystem.php
+++ b/app/code/Magento/Deploy/Model/Filesystem.php
@@ -115,7 +115,7 @@ class Filesystem
         $this->userCollection = $userCollection;
         $this->locale = $locale;
         $this->functionCallPath =
-            '"' . PHP_BINARY . '"' . ' -f ' . BP . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'magento ';
+	        escapeshellcmd(PHP_BINARY) . ' -f ' . BP . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'magento ';
     }
 
     /**
--- a/app/code/Magento/Deploy/Model/Filesystem.php
+++ b/app/code/Magento/Deploy/Model/Filesystem.php
@@ -115,7 +115,7 @@ class Filesystem
         $this->userCollection = $userCollection;
         $this->locale = $locale;
         $this->functionCallPath =
-	        escapeshellcmd(PHP_BINARY) . ' -f ' . BP . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'magento ';
+            '"' . PHP_BINARY . '"' . ' -f ' . BP . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'magento ';
     }
 
     /**
--- a/app/code/Magento/Deploy/Model/Filesystem.php
+++ b/app/code/Magento/Deploy/Model/Filesystem.php
@@ -115,7 +115,7 @@ class Filesystem
         $this->userCollection = $userCollection;
         $this->locale = $locale;
         $this->functionCallPath =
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13340) <br/>All committers have signed the CLA.

miguelbalparda:
From the [DevDocs](http://devdocs.magento.com/guides/v2.2/install-gde/system-requirements-tech.html#operating-systems-linux-x86-64), MacOs and Windows are not supported. I haven't seen any server with a space in the PHP_BINARY, I'll check internally to see if this is worth adding and report back.
Can you check the failing tests in the meantime? Same with the CLA.

app/code/Magento/Deploy/Model/Filesystem.php
```diff
@@ -115,7 +115,7 @@ public function __construct(
         $this->userCollection = $userCollection;
         $this->locale = $locale;
         $this->functionCallPath =
-            PHP_BINARY . ' -f ' . BP . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'magento ';
+            '"' . PHP_BINARY . '"' . ' -f ' . BP . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'magento ';
```
oliverklee, line 118: This is incomplete and will break if the path has double quotes in it. We'll also need to use `escapeshellcmd`.

oliverklee (changes requested):

orlangur:
> as far as I know you can have spaces in filepaths on OSX or *nix (you probably wouldn't, but it's at least possible)

@PivitParkour94 while obviously it is technically possible, are you aware of ANY [supported OS](http://devdocs.magento.com/guides/v2.2/install-gde/system-requirements-tech.html#operating-systems-linux-x86-64) actually having such spaces by default?

orlangur:
https://github.com/magento/magento2/search?q=php_binary&type=Code&utf8=✓ this is not the only occurrence and I doubt there is any real reason to complicate the code.

@miguelbalparda described issue is a duplicate of https://github.com/magento/magento2/issues/5258#issuecomment-336942122, as I don't see any new argumentation I would insist on rejecting this PR.

app/code/Magento/Deploy/Model/Filesystem.php
PivitParkour94: that's a much better alternative! `escapeshellcmd`

PivitParkour94:
@orlangur This may not be an issue with default setup, but I think skipping it completely would be an oversight, simply because it doesn't fall into the cookie cut PHP installation. 

There are multiple places that the PHP_BINARY constant is used, but the issue is in running `exec` when the path has a space in it. 

As it is such a simple fix, I thought I'd add it into the core and that would be good (granted I haven't added the PHPUnit test fix... that's my bad) rather than having to create an entire module that adds this support, it would be nice to just have it in core. Of course it is your choice to make, 

Thanks for your time

miguelbalparda:
> simply because it doesn't fall into the cookie cut PHP installation

is an understatement. We have supported and unsupported operating systems and this fix, while simple, seems to be for a system we don't support. We'll come back to this if anything changes in the future. Thanks!

PR #13353, opened 2018-01-24
Add the possibility to begin htmlClass with a - or _

VincentMarmiesse:
Hello

### Description
In the `htmlClass` xsd, the pattern should accept `-` or `_` as valid characters. It's valid when I test it with the W3C website.

### Manual testing scenarios
1. Create or update a container in a layout file and set the `htmlClass` to `-myclass`. It should not raise an error in developer mode.

### DIFF
--- a/lib/internal/Magento/Framework/View/Layout/etc/elements.xsd
+++ b/lib/internal/Magento/Framework/View/Layout/etc/elements.xsd
@@ -112,7 +112,7 @@
 
     <xs:simpleType name="htmlClassType">
         <xs:restriction base="xs:string">
-            <xs:pattern value="[a-zA-Z][a-zA-Z\d\-_]*(\s[a-zA-Z][a-zA-Z\d\-_]*)*"/>
+            <xs:pattern value="[a-zA-Z\-_][a-zA-Z\d\-_]*(\s[a-zA-Z\-_][a-zA-Z\d\-_]*)*"/>
         </xs:restriction>
     </xs:simpleType>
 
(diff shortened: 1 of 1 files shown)


VincentMarmiesse:
I added the `:` which can also be used in HTML classes.

orlangur:
> In the htmlClass xsd, the pattern should accept - or _ as valid characters

Such suggestion violates http://devdocs.magento.com/guides/v2.2/coding-standards/code-standard-less.html#standard-classes. Leading underscore is used for helper classes but only added via JavaScript http://devdocs.magento.com/guides/v2.2/coding-standards/code-standard-demarcation.html#you-must-assign-html-helper-classes-in-javascript-to-modify-presentation-layer.

How `:` can be useful in CSS class name?

VincentMarmiesse:
Hi @orlangur,

Well it's valid in CSS3 so I don't see why Magento would not accept it. Do you know the reason?

I'm currently working on a project where front devs are using classes like `product__header` for example.

They also use `:` in classes like `mb:1` for margin bottom 1 for example. It's also valid in CSS3.

orlangur:
> Well it's valid in CSS3 so I don't see why Magento would not accept it. Do you know the reason?

Because Magento has its own set of standards which are obviously stricter than what is technically possible.

> front devs are using classes like `product__header` for example

It is not a good idea to write Magento customizations not following Magento guidelines. Just show them links I posted. By the way, such class name works for me with current regexp: https://regex101.com/r/YVTcX9/1

> use `:` in classes like `mb:1` for margin bottom 1 for example. It's also valid in CSS3.

I'm not a big expert in CSS but for me such classes look non-semantical. Just like some `.color-red` class.

@VincentMarmiesse I'm rejecting these suggested changes for now, however they are possible if someday standards will be changed. Thanks for collaboration!

PR #13355, opened 2018-01-24
#13354 - [Improvement] Add HTML attributes to containers

VincentMarmiesse:
Hello,

### Description
The aim of this PR is to add the possibility to add more HTML attributes to containers.
For example, I would like to create the following tags in a layout file:

```
<header class="header" id="header" data-sticky data-isheader="true">

</header>
```

1. magento/magento2#13354: [Improvement] Add HTML attributes to containers

### Manual testing scenarios
1. Create or update a container and add `htmlAttributes="data-myattribute"` for example
2. Refresh cache, the created HTML tag should have the `data-myattribute` attribute

### Development status
It is currently working but I would like to discuss about attributes which have a value, ie `data-isheader="true"`.
If I want to do this, I have to create my container like this:

```
<container name="header.container" as="header_container" htmlTag="header" htmlClass="header" htmlId="header" htmlAttributes="data-sticky data-isheader='true'">
</container>
```

Notice the simple quote around `true`, it works but is not very beautiful :)

What do you guys think?

- [x] Add tests into `Magento\Framework\View\Test\Unit\Layout\Reader\ContainerTest`

Thanks!

### DIFF
--- a/lib/internal/Magento/Framework/View/Layout.php
+++ b/lib/internal/Magento/Framework/View/Layout.php
@@ -598,9 +598,11 @@ class Layout extends \Magento\Framework\Simplexml\Config implements \Magento\Fra
             $htmlClass = ' class="' . $htmlClass . '"';
         }
 
+        $htmlAttributes = $this->structure->getAttribute($name, Element::CONTAINER_OPT_HTML_ATTRIBUTES);
+
         $htmlTag = $this->structure->getAttribute($name, Element::CONTAINER_OPT_HTML_TAG);
 
-        $html = sprintf('<%1$s%2$s%3$s>%4$s</%1$s>', $htmlTag, $htmlId, $htmlClass, $html);
+        $html = sprintf('<%1$s%2$s%3$s%4$s>%5$s</%1$s>', $htmlTag, $htmlId, $htmlClass, $htmlAttributes, $html);
 
         return $html;
     }
--- a/lib/internal/Magento/Framework/View/Layout/Element.php
+++ b/lib/internal/Magento/Framework/View/Layout/Element.php
@@ -54,6 +54,8 @@ class Element extends \Magento\Framework\Simplexml\Element
 
     const CONTAINER_OPT_HTML_ID = 'htmlId';
 
+    const CONTAINER_OPT_HTML_ATTRIBUTES = 'htmlAttributes';
+
     const CONTAINER_OPT_LABEL = 'label';
 
     /**#@-*/
--- a/lib/internal/Magento/Framework/View/Layout/Reader/Container.php
+++ b/lib/internal/Magento/Framework/View/Layout/Reader/Container.php
@@ -25,6 +25,7 @@ class Container implements Layout\ReaderInterface
     const CONTAINER_OPT_HTML_TAG = 'htmlTag';
     const CONTAINER_OPT_HTML_CLASS = 'htmlClass';
     const CONTAINER_OPT_HTML_ID = 'htmlId';
+    const CONTAINER_OPT_HTML_ATTRIBUTES = 'htmlAttributes';
     const CONTAINER_OPT_LABEL = 'label';
     const CONTAINER_OPT_DISPLAY = 'display';
     /**#@-*/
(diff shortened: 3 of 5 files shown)


orlangur:
> Notice the simple quote around true, it works but is not very beautiful :)

Pretty counterintuitive actually.

Please provide some insight on why do you need data attributes on containers. Why not simply wrap whatever you have with some `<div data-sticky data-isheader="true"` or declare such `<header` in template?

VincentMarmiesse:
Hi @orlangur,
I'm using it when I want to declare a container without having a template. I could do it with a template as you suggest it but I wanted to avoid templates like

```
<div data-sticky data-isheader="true">
   <?php echo $block->getChildHtml('') ?>
</div>
```

That's why containers exist right?

orlangur:
> I could do it with a template as you suggest it but I wanted to avoid templates like

Why? Proposed layout XML syntax looks harder for maintenance than such templates as to me.

VincentMarmiesse:
Hi @orlangur,

Well let's suppose I want to add such attribute on the `main` container. It's defined like this:

```
<container name="main" label="Main Content Container" htmlTag="div" htmlClass="column main"/>
```

and I want this result in HTML:
```
<div class="column main" data-attribute="value">
   ...
</div>
```

There is currently no way to achieve this right?

orlangur:
@VincentMarmiesse, is it possible to implement syntax like

```
<container name="main" label="Main Content Container" 
    htmlTag="div" htmlClass="column main"
    htmlDataSticky=""
    htmlDataIsheader="true"
    />
```

with XSD?

orlangur:

thomas-kl1:
This feature could be really interesting. 
@orlangur could you give a fresh feedback?

orlangur:
@thomas-blackbird please answer my question https://github.com/magento/magento2/pull/13355#issuecomment-400217428 if you know the answer :)

Originally proposed syntax looks a bit messy.

thomas-kl1:
@orlangur it doesn't seems possible to define wildcard for attribute name.
However it's possible to allow any attribute https://www.w3.org/TR/xmlschema-1/#element-anyAttribute
but it's not really sexy..

orlangur:
@thomas-blackbird thanks for investigation, this is what I was talking about. Definitely having serialized attribute values as value for some attribute is not an improved syntax.

Generally, overcomplicating layout syntax is not a desirable way to go, currently available approach described in https://github.com/magento/magento2/pull/13355#issuecomment-361960876 is exactly what would I recommend in such situation. Straightforward & readable.

joshdavenport:
What about an alternative approach such as below:

    <container name="header.container" as="header_container" htmlTag="header" htmlClass="header" htmlId="header">
        <attribute name="data-sticky" />
        <attribute name="data-isheader" value="true" />
    </container>

With this approach you get the added benefit of being open to adding attributes retrospectively to existing containers:

    <referenceContainer name="header.container">
        <attribute name="data-color" value="#000000" /> 
    </referenceConainer>

And even further, you could then add support to remove others attributes so this isn't a one way street:

    <referenceContainer name="header.container">
        <attribute name="data-sticky" remove="true" /> 
    </referenceConainer>

max-atkins:
Magento 2 significantly improved defining HTML through XML but I am constantly limited whenever I need to add data attributes. Take the JS library AOS for example. I want to define my HTML through XML because it's neater, easier to use and (I assume?) faster. It also limits the amount of templates needed on the site and less PHP calls to getChildHtml().

If I want an AOS block with data attributes, I then have to create a template instead of just defining it in XML. The suggestion made by joshdavenport here: https://github.com/magento/magento2/pull/13355#issuecomment-438286063
would be amazingly useful.

thomas-kl1:
@joshdavenport 
I'm thinking of this, it will allows to define XSD validation right:
```xml
<container name="header.container" as="header_container" htmlTag="header" htmlClass="header" htmlId="header">
    <dataSet>
        <attribute name="isheader" value="true" />
    </dataSet>
</container>
```

max-atkins:
@thomas-blackbird It would probably be good for it to output just what the name is set to.
eg. 
`<attribute name="isheader" value="true" />`
 should output 
`<header class="header" id="header" isheader="true">`
 rather than
`<header class="header" id="header" data-isheader="true">`

This would allow for defining any kind of HTML attribute to accommodate JS libraries that don't use data- as a prefix.

thomas-kl1:
@maxxeh1 Whoops, true, I was misunderstanding, my bad..

What about already existing attributes `htmlClass` and `htmlId`? Should they be deprecated or new element <attribute name="`id`|`class`"../> might be restricted?

orlangur:
> This would allow for defining any kind of HTML attribute to accommodate JS libraries that don't use data- as a prefix.

@maxxeh1 is it a common case? Maybe introduce `<dataAttribute` and `<htmlAttribute` then.

> Should they be deprecated or new element <attribute name="id|class"../> might be restricted?

@thomas-blackbird definitely not restricted, deprecation is a good choice while removal is not possible.

max-atkins:
@orlangur I like htmlClass, htmlTag and htmlId as they are since they are now what is familiar in Magento 2 and will be used far more than these data attributes. `<dataAttribute` and `<htmlAttribute` sounds good but maybe don't allow people to add class or id this way?

It's not massively common but I've seen it before; allowing it would just accommodate any way of doing it.

joshdavenport:
> @orlangur I like htmlClass, htmlTag and htmlId as they are since they are now what is familiar in Magento 2 and will be used far more than these data attributes. `<dataAttribute` and `<htmlAttribute` sounds good but maybe don't allow people to add class or id this way?

Agreed @maxxeh1, people are definitely used to those container attributes now and I also think people perceive `class` and `id` having a special place over other attributes even though they're just plain attributes that have special effects so it still makes sense to keep them.

IMO, they should be kept and this should be implemented such that `<htmlAttribute name="id"` and `<htmlAttribute name="class"` don't break `htmlId` and `htmlClass` (and vice versa) and probably such that the value of `htmlId` and `htmlClass` would have precedence over anything added via `<htmlAttribute name="id"` and `<htmlAttribute name="class"`.

thomas-kl1:
> Maybe introduce <dataAttribute and <htmlAttribute then.

@orlangur As I see dataAttribute, it is useless as you can already add them with <htmlAttribute name="data-*"

@joshdavenport @maxxeh1 Deprecate htmlId and htmlClass is right as it's not a good idea to allow many ways to achieve the same goal. It must be clear to know which is the "best practice" to set the id/class on the container.
Anyway there's already a lot of allowed attributes on container, move `htmlId` and `htmlClass` to <htmlAttribute will be more readable.
Here the full list of container's attributes: name, label,  output, as, after, before, htmlTag, htmlClass, htmlId

joshdavenport:
@thomas-blackbird Agreed regarding dataAttribute, seems unnecessary. 

Regarding htmlId and htmlClass, while I do agree with you it'll be a large breaking change to just remove them. Especially as there is nothing like PHPDoc in place to notify the user that this feature is deprecated (i.e. `@deprecated`) and will go soon. With that in mind, how are we supposed to properly let the developers know the feature will get removed soon? Like I say, while I totally agree it would be cleaner to remove them altogether, if I wasn't fortunately party to this conversation and then this suddenly happened out of nowhere, I'd definitely be really annoyed.

So really, my previous suggestion about keeping them comes more from a position of not knowing how quick we should be to deprecate something so often used. Perhaps @orlangur would have something to contribute to the discussion there?

max-atkins:
@thomas-blackbird This makes sense and I agree; it will cause a lot of work in XML but it'll be better in the long run!

PR #13472, opened 2018-02-02
Allow import of empty attributes

PieterCappelle:
Allow import of empty attributes in Magento2 import functionality. See internal number MAGETWO-61593.
1. https://github.com/magento/magento2/issues/7468

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product.php
@@ -909,6 +909,19 @@ class Product extends \Magento\ImportExport\Model\Import\Entity\AbstractEntity
     }
 
     /**
+     * Return empty attribute value constant
+     *
+     * @return string
+     */
+    public function getEmptyAttributeValueConstant()
+    {
+        if (!empty($this->_parameters[Import::FIELD_EMPTY_ATTRIBUTE_VALUE_CONSTANT])) {
+            return $this->_parameters[Import::FIELD_EMPTY_ATTRIBUTE_VALUE_CONSTANT];
+        }
+        return Import::DEFAULT_EMPTY_ATTRIBUTE_VALUE_CONSTANT;
+    }
+
+    /**
      * Retrieve instance of product custom options import entity
      *
      * @return \Magento\CatalogImportExport\Model\Import\Product\Option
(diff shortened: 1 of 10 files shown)


PieterCappelle:
Any progress on this issue? Unit tests are failing but is because of Travis :)

magento-engcom-team:
@PieterCappelle thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

dmanners:
Hi @PieterCappelle for feedback from testing.

There is no human-readable validation message when "custom_options" is set to "_EMPTYVALUE_" even if there was no value before. When you setup the "custom_options" field to "_EMPTYVALUE_" you end up with the notice `Notice: Undefined index: required in /var/www/html/pr/app/code/Magento/CatalogImportExport/Model/Import/Product/Option.php on line 1124`

import-notice-custom-options

There is a validation for name, SKU and price properties on Backend, however for unknown reasons Merchant is able to create products without price using WebAPI.
Why not to disallow unsetting name, SKU and price via Import?

1) Set "_EMPTYVALUE_" for SKU but leave "url_key" to be "simple1".

1. Url key: '%s' was already generated for an item with the SKU: '%s'. You need to specify the unique URL key manually in row(s): 1

import-url-key

2) Set "_EMPTYVALUE_" for SKU and "url_key"
Press "Check Data" and get 'File is valid! To start import process press "Import" button' message.
Press "Import" button and you will see this error:

1. Imported resource (image) could not be downloaded from external resource due to timeout or access permissions in row(s): 1

import-image-error

3) Navigate to Products > Inventory > Catalog and look that there we created a new product

import-product-with-no-name

I think that if we have validation for empty values in the backend we should have the same validation in the import.

dmanners:
I have updated this PR so that it now includes the deletion of custom options via the csv using the empty value option added by @PieterCappelle

magento-engcom-team:
Hi @PieterCappelle. Thank you for your contribution. 

dweerd:
Does this work for all types of attributes? I have applied this patch to a magento 2.2.5 instance, and although this does work for text fields, it doesn't work as expected for boolean fields and dropdown fields.

Empty values for these kinds of fields are saved as NULL, causing them to be displayed as 'No' on the product details page. If I remove the records with value NULL from the `catalog_product_entity_int` table, they are no longer displayed on the product details page, as expected.

As a quick fix, I have extended `Magento\CatalogImportExport\Model\Import\Product::_saveProductAttributes` like so:

```
protected function _saveProductAttributes(array $attributesData)
{
	parent::_saveProductAttributes($attributesData);

	foreach ($attributesData as $tableName => $skuData) {
		$this->_connection->delete($tableName, ['value IS NULL']);
	}

	return $this;
}
````

Issue #9203, opened 2017-04-11
Upgrade components from ZF1 to ZF2


okorshenko:
**Description**
As a long term goal, we would like to eliminate knowledge about 3rd party libraries from Magento code base. Magento code still can use 3rd party libraries, but they must be wrapped by Magento interfaces and classes (adapters) so that 3rd party libraries can be easily substituted by newest versions or alternative implementations.

**_Note_**
_Each component can be implemented separately. It is not a requirement to fix all components in one task/pull request_

**Acceptance Criteria**

1. Magento interfaces are defined and can be used instead of ZF1 classes in Magento code
2. Default implementations for new interfaces are defined and configured as the preferences for new interfaces
3. Client code of the ZF1 components is refactored to use Magento interfaces
4. Zend classes are used as private services in Magento adapters
5. Backward compatibility is preserved
6. ZF1 components are updated to ZF2

**List of the components**

\# | Component Name <br/> (total usages count) | Module Name | Usages Count
------------- | ------------- | ------------- | -------------
1|Zend_Translate (total: 1)|Framework\Translate|1
2|Zend_Service (total: 1)|Framework\CurrencyInterface.php|1
3|Zend_Session (total: 1)|Framework\Session|1
4|Zend_Wildfire (total: 3)|Framework\Profiler|3
5|Zend_Feed (total: 3)|Magento\Rss|3
6|Zend_Controller (total: 6)|Framework\Profiler|6
7|Zend_Oauth (total: 6)|Magento\Integration|2
&nbsp;|&nbsp;|Framework\Oauth|4
8|Zend_Log (total: 7)|Framework\Test|3
&nbsp;|&nbsp;|Framework\Shell.php|4
9|Zend_Json (total: 51)|Framework\DataObject.php|1
&nbsp;|&nbsp;|Framework\Controller|1
&nbsp;|&nbsp;|Framework\Json|2
&nbsp;|&nbsp;|Framework\Module|3
&nbsp;|&nbsp;|Framework\Webapi|3
&nbsp;|&nbsp;|Magento\Weee|6
&nbsp;|&nbsp;|Framework\Data|4
&nbsp;|&nbsp;|Magento\Theme|31

dmanners:
Love this. Probably cannot do all but I am working through the `Zend_Json` stuff slowly

okorshenko:
Closing this issue.
Created separate ticket per component

PR #13492, opened 2018-02-05
[REFACTORING] refactoring the wishlist module by replacing deprecated methods

coderimus:
### Description
The main goal of this PR is to replace deprecated methods for messenger object and use resource models directly instead of using `_getResource()` and `save()` methods.
I used backwards compatible guide for refactoring and adopted tests for new changes.

N/A

### Manual testing scenarios
N/A

### Contribution checklist

### DIFF
--- a/app/code/Magento/Wishlist/Controller/Index/Add.php
+++ b/app/code/Magento/Wishlist/Controller/Index/Add.php
@@ -11,6 +11,7 @@ use Magento\Framework\Data\Form\FormKey\Validator;
 use Magento\Framework\Exception\NotFoundException;
 use Magento\Framework\Exception\NoSuchEntityException;
 use Magento\Framework\Controller\ResultFactory;
+use Magento\Wishlist\Model\ResourceModel\Wishlist;
 
 /**
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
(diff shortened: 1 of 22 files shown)


orlangur:
Hi @Coderimus, could you share some motivation behind such refactoring?

While resource models are a way to go until new ORM is introduced (see https://github.com/magento/magento2/blob/2.2-develop/lib/internal/Magento/Framework/EntityManager/EntityManager.php#L16) I really doubt it makes sense to refactor deprecated load/save/delete usages with resource models as we will have to do another refactoring to involve new persistence layer later.

coderimus:
Hi @orlangur , thank you for your reply! 
I did this because I didn't know about the new ORM and just thought that it will be good to avoid using deprecated methods :) After your reply, I see that there is no sense in my PR because it will not be the final one and we will need to do one more refactoring when the new ORM will be introduced.
Can I decline/close this PR?
Thank you once more for detail answer!

orlangur:
@Coderimus thanks for quick reply. Yeah, closing this PR for now, was just double checking with you if I'm not aware of some ongoing activity like two-phase refactoring.

Generally, the best way to understand if some refactoring fits well in current roadmap is raising your idea in Slack then it can lead to an issue with clearly defined scope reported in GitHub (like ZF1 -> ZF2 upgrade, for example https://github.com/magento/magento2/issues/9203)

PR #13495, opened 2018-02-05
Removing reprecated dependencies from Braintree module

YevSent:
Braintree module has dependencies on deprecated `\Magento\Payment\Model\Method\AbstractMethod` and `\Magento\Vault\Api\Data\PaymentTokenInterfaceFactory`. Proposed changes remove or replace them.

### Description
- Removed dependency of AbstractMethod
- Replaced usage of deprecated Vault factories

### Contribution checklist

### DIFF
--- a/app/code/Magento/Braintree/Gateway/Response/PayPal/VaultDetailsHandler.php
+++ b/app/code/Magento/Braintree/Gateway/Response/PayPal/VaultDetailsHandler.php
@@ -12,8 +12,8 @@ use Magento\Payment\Gateway\Response\HandlerInterface;
 use Magento\Payment\Model\InfoInterface;
 use Magento\Sales\Api\Data\OrderPaymentExtensionInterface;
 use Magento\Sales\Api\Data\OrderPaymentExtensionInterfaceFactory;
+use Magento\Vault\Api\Data\PaymentTokenFactoryInterface;
 use Magento\Vault\Api\Data\PaymentTokenInterface;
-use Magento\Vault\Api\Data\PaymentTokenInterfaceFactory;
 
 /**
  * Vault Details Handler
@@ -21,7 +21,7 @@ use Magento\Vault\Api\Data\PaymentTokenInterfaceFactory;
 class VaultDetailsHandler implements HandlerInterface
 {
     /**
-     * @var PaymentTokenInterfaceFactory
+     * @var PaymentTokenFactoryInterface
      */
     private $paymentTokenFactory;
 
@@ -41,15 +41,13 @@ class VaultDetailsHandler implements HandlerInterface
     private $dateTimeFactory;
 
     /**
-     * Constructor
-     *
-     * @param PaymentTokenInterfaceFactory $paymentTokenFactory
+     * @param PaymentTokenFactoryInterface $paymentTokenFactory
      * @param OrderPaymentExtensionInterfaceFactory $paymentExtensionFactory
      * @param SubjectReader $subjectReader
      * @param DateTimeFactory $dateTimeFactory
      */
     public function __construct(
-        PaymentTokenInterfaceFactory $paymentTokenFactory,
+        PaymentTokenFactoryInterface $paymentTokenFactory,
         OrderPaymentExtensionInterfaceFactory $paymentExtensionFactory,
         SubjectReader $subjectReader,
         DateTimeFactory $dateTimeFactory
@@ -92,7 +90,7 @@ class VaultDetailsHandler implements HandlerInterface
         }
 
         /** @var PaymentTokenInterface $paymentToken */
-        $paymentToken = $this->paymentTokenFactory->create();
+        $paymentToken = $this->paymentTokenFactory->create(PaymentTokenFactoryInterface::TOKEN_TYPE_ACCOUNT);
         $paymentToken->setGatewayToken($token);
         $paymentToken->setExpiresAt($this->getExpirationDate());
         $details = json_encode([
--- a/app/code/Magento/Braintree/Gateway/Response/VaultDetailsHandler.php
+++ b/app/code/Magento/Braintree/Gateway/Response/VaultDetailsHandler.php
@@ -8,12 +8,14 @@ namespace Magento\Braintree\Gateway\Response;
 use Braintree\Transaction;
 use Magento\Braintree\Gateway\Config\Config;
 use Magento\Braintree\Gateway\SubjectReader;
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\Serialize\Serializer\Json;
 use Magento\Payment\Gateway\Response\HandlerInterface;
 use Magento\Payment\Model\InfoInterface;
 use Magento\Sales\Api\Data\OrderPaymentExtensionInterface;
 use Magento\Sales\Api\Data\OrderPaymentExtensionInterfaceFactory;
+use Magento\Vault\Api\Data\PaymentTokenFactoryInterface;
 use Magento\Vault\Api\Data\PaymentTokenInterface;
-use Magento\Vault\Api\Data\PaymentTokenInterfaceFactory;
 
 /**
  * Vault Details Handler
@@ -22,7 +24,7 @@ use Magento\Vault\Api\Data\PaymentTokenInterfaceFactory;
 class VaultDetailsHandler implements HandlerInterface
 {
     /**
-     * @var PaymentTokenInterfaceFactory
+     * @var PaymentTokenFactoryInterface
      */
     protected $paymentTokenFactory;
 
@@ -49,26 +51,26 @@ class VaultDetailsHandler implements HandlerInterface
     /**
      * VaultDetailsHandler constructor.
      *
-     * @param PaymentTokenInterfaceFactory $paymentTokenFactory
+     * @param PaymentTokenFactoryInterface $paymentTokenFactory
      * @param OrderPaymentExtensionInterfaceFactory $paymentExtensionFactory
      * @param Config $config
      * @param SubjectReader $subjectReader
-     * @param \Magento\Framework\Serialize\Serializer\Json|null $serializer
+     * @param Json|null $serializer
      * @throws \RuntimeException
      */
     public function __construct(
-        PaymentTokenInterfaceFactory $paymentTokenFactory,
+        PaymentTokenFactoryInterface $paymentTokenFactory,
         OrderPaymentExtensionInterfaceFactory $paymentExtensionFactory,
         Config $config,
         SubjectReader $subjectReader,
-        \Magento\Framework\Serialize\Serializer\Json $serializer = null
+        Json $serializer = null
     ) {
         $this->paymentTokenFactory = $paymentTokenFactory;
         $this->paymentExtensionFactory = $paymentExtensionFactory;
         $this->config = $config;
         $this->subjectReader = $subjectReader;
-        $this->serializer = $serializer ?: \Magento\Framework\App\ObjectManager::getInstance()
-            ->get(\Magento\Framework\Serialize\Serializer\Json::class);
+        $this->serializer = $serializer ?: ObjectManager::getInstance()
+            ->get(Json::class);
     }
 
     /**
@@ -103,7 +105,7 @@ class VaultDetailsHandler implements HandlerInterface
         }
 
         /** @var PaymentTokenInterface $paymentToken */
-        $paymentToken = $this->paymentTokenFactory->create();
+        $paymentToken = $this->paymentTokenFactory->create(PaymentTokenFactoryInterface::TOKEN_TYPE_CREDIT_CARD);
         $paymentToken->setGatewayToken($token);
         $paymentToken->setExpiresAt($this->getExpirationDate($transaction));
 
--- a/app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
+++ b/app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
@@ -6,7 +6,6 @@
 namespace Magento\Braintree\Model\Adminhtml\Source;
 
 use Magento\Framework\Option\ArrayInterface;
-use Magento\Payment\Model\Method\AbstractMethod;
 
 /**
  * Class PaymentAction
@@ -22,11 +21,11 @@ class PaymentAction implements ArrayInterface
     {
         return [
             [
-                'value' => AbstractMethod::ACTION_AUTHORIZE,
+                'value' => 'authorize',
                 'label' => __('Authorize'),
             ],
             [
-                'value' => AbstractMethod::ACTION_AUTHORIZE_CAPTURE,
+                'value' => 'authorize_capture',
                 'label' => __('Authorize and Capture'),
             ]
         ];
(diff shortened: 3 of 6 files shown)


app/code/Magento/Braintree/Gateway/Response/PayPal/VaultDetailsHandler.php
```diff
@@ -41,15 +41,13 @@ class VaultDetailsHandler implements HandlerInterface
...
-     *
-     * @param PaymentTokenInterfaceFactory $paymentTokenFactory
+     * @param PaymentTokenFactoryInterface $paymentTokenFactory
      * @param OrderPaymentExtensionInterfaceFactory $paymentExtensionFactory
      * @param SubjectReader $subjectReader
      * @param DateTimeFactory $dateTimeFactory
      */
     public function __construct(
-        PaymentTokenInterfaceFactory $paymentTokenFactory,
+        PaymentTokenFactoryInterface $paymentTokenFactory,
```
ihor-sviziev, line 50: This change is not backward compatible. Please revert changes in this line and add new dependency to PaymentTokenFactoryInterface as latest optional parameter that you will use for current class

app/code/Magento/Braintree/Gateway/Response/VaultDetailsHandler.php
```diff
@@ -22,7 +24,7 @@
 class VaultDetailsHandler implements HandlerInterface
 {
     /**
-     * @var PaymentTokenInterfaceFactory
+     * @var PaymentTokenFactoryInterface
```
ihor-sviziev, line 27: Changing type for protected variables is not backward compatible. Please create new private variable and use it in all places that you need

app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
```diff
@@ -22,11 +21,11 @@ public function toOptionArray()
     {
         return [
             [
-                'value' => AbstractMethod::ACTION_AUTHORIZE,
+                'value' => 'authorize',
```
ihor-sviziev, line 24: It's not good idea to stop using constant and use just string. Is there any idea?

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
```diff
@@ -126,14 +113,11 @@ protected function setUp()
         $this->handler = new VaultDetailsHandler(
             $this->paymentTokenFactory,
             $this->paymentExtensionFactory,
-            $this->subjectReader,
+            new SubjectReader(),
```
ihor-sviziev, line 116: Why are you passing there not mock, but real object?

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
```diff
@@ -142,49 +126,38 @@ public function testHandle()
             'object' => $transaction
         ];
 
-        $this->paymentExtension->expects(static::once())
-            ->method('setVaultPaymentToken')
+        $this->paymentExtension->method('setVaultPaymentToken')
```
ihor-sviziev, line 129: You've updated tests to be less strict, I don't think it's a good idea

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
```diff
@@ -230,8 +197,10 @@ private function getTransaction()
         ];
 
         $transaction = Transaction::factory($attributes);
+        $obj = new \stdClass();
+        $obj->transaction = $transaction;
```
ihor-sviziev, line 201: What is the benefit of doing such change?

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
```diff
@@ -8,62 +8,59 @@
...
-     * @var CreditCardTokenFactory|MockObject
+     * @var PaymentTokenFactoryInterface|MockObject
      */
     private $paymentTokenFactory;
 
     /**
      * @var PaymentTokenInterface|MockObject
      */
-    protected $paymentToken;
+    private $paymentToken;
```
ihor-sviziev, line 52: Changing property visibility protected -> private is not backward compatible change. We can't do that

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
```diff
@@ -72,12 +69,11 @@ class VaultDetailsHandlerTest extends \PHPUnit\Framework\TestCase
...
     {
         $this->paymentToken = $this->createMock(PaymentTokenInterface::class);
-        $this->paymentTokenFactory = $this->getMockBuilder(CreditCardTokenFactory::class)
+        $this->paymentTokenFactory = $this->getMockBuilder(PaymentTokenFactoryInterface::class)
             ->setMethods(['create'])
             ->disableOriginalConstructor()
             ->getMock();
-        $this->paymentTokenFactory->expects(self::once())
-            ->method('create')
+        $this->paymentTokenFactory->method('create')
```
ihor-sviziev, line 76: 1. Probably there should be check that argument was "card"
2. You made checks less strict, it's not good idea. Please add self::once() back

ihor-sviziev (changes requested):

app/code/Magento/Braintree/Gateway/Response/PayPal/VaultDetailsHandler.php
YevSent: @ihor-sviziev, it's not a BIC because it's the private code: not marked as @api and class for "end" integration.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
YevSent: BC policy doesn't apply to tests: they are private code, not marked as @api and should not be extended.

app/code/Magento/Braintree/Gateway/Response/VaultDetailsHandler.php
YevSent: This is not BIC change because it's the private code.

app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
YevSent: I don't see any objections of `string` usage.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
YevSent: `SubjectReader` is a simple helper without dependencies and doesn't require mocking. The real implementation allows to simplify unit test and in general, it doesn't matter how a response and subject will be retrieved.
Another case, if the implementation of `SubjectReader` will be changed in future, and `readPayment()` or `readTransaction()` will be broken this test will show it.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
YevSent: How does `expects()` make this test more useful? It's better to add additional assert.
Such `expects()` "freeze" the test methods and follow Tautological TDD.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
YevSent: Real transaction response looks like:
```
Braintree\Successful->transaction instance of Transaction
```
This method creates
```
stdClass->transaction instance of Transaction
```
In this case, it doesn't matter which class will be used as the fake "wrapper" `Successful` or `stdClass`.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
YevSent: 1. Agree.
2. I don't see how it will make the test more useful, it doesn't matter how many times the `create` method will be called, the main purpose of this factory - create the instance of the class.

orlangur:
@joni-jones @ihor-sviziev is such refactoring a good idea for patch release at all? Could suddenly break existing customizations, I would rather move it to 2.3 if it does not fix any bug but was done for the sake of purity.

YevSent:
@orlangur, I don't see any reasons to don't do such refactoring in a patch release.

The refactored response handlers are not points for 3rd party customization - the Payment Provider Gateway infrastructure assumes that each a part of it can be easy "replaced" via DI.xml, so 3rd party developers should use composition instead of inheritance.

orlangur:
> so 3rd party developers should use composition instead of inheritance.

This is not always true in reality and we still do some BC things just for the sake of those who inherits core classes (like not adding parameters even with default values into protected methods).

Just my 2 cents, I'm not aware how often this functionality is customized at all and if it will break things for some 1% merchants not bringing any value for end user for other 99%, it could be moved to a bigger release.

YevSent:
@orlangur, I agree, we are not living in the ideal world. But starting from 2.2.0 release we have the public and private code and if somebody depends, uses, extends the private code he/she should understand that this code can be changed in any way or even removed.

app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
ihor-sviziev: Using constant where it possible allow you to find all places where this constant used just with "find usages" in any IDE. In Magento 1.x we had a lot of such places where were used strings instead of constants and it was really chaos.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
ihor-sviziev: Previously it was checking exactly how much times it will be called, so it was testing business logic. After your changes - it will not fail in case when it will be executed few times (that was not expected before)

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
ihor-sviziev: Yes, you're correct

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
ihor-sviziev: I's better to create instance of Successful class to be sure that we're testing correct behavior

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
ihor-sviziev: Currently it is testing that this method **was executed** only **once**, so there were actually 2 checks, that you removed by your changes. If this setter will not be executed in some case - business logic will be broken and test will NOT fail

ihor-sviziev:
@joni-jones @orlangur I don't like idea to add backward incompatible changes to patch releases when it possible not to do that. You can see on it as "Each backward incompatible change will cost a lot of businesses in the world some money (dev hours).", let's not spend it without need. In this case - most changes could be adjusted to be just backward compatible.
Also here were introduced a lot "not good" changes to tests, that stopping to test business logic, I don't like at all this idea

app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
YevSent: Agree, but in this case these constants from the deprecated class. The better solution to have such constants in the `\Magento\Payment\Model\MethodInterface` but now I can't introduce them - it will be BIC.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
YevSent: I think it's better to add additional `asserts` instead freezing the test code, in that case, it will be more useful than just checking how many times the method was called.

app/code/Magento/Braintree/Test/Unit/Gateway/Response/VaultDetailsHandlerTest.php
YevSent: >  it was testing business logic

Actually, it just checking if the method called but not how business logic works. I think it better to add additional `asserts` to check if payment token will be created with all needed data rather then just check if the method called or not.

YevSent:
I do not insist to merge these changes to patch release, I can re-create the PR to 2.3-develop, it's not a problem.

> Also here were introduced a lot "not good" changes to tests, that stopping to test business logic

@ihor-sviziev, I can't agree with your statement. I've provided the arguments, why `expects()` have been removed but I agree these tests require more refactoring and additional assertions to test business logic.
On my opinion, these `expects()` make tests methods in TTDD manner but don't test the code.

YevSent:
Created PR for 2.3-develop https://github.com/magento/magento2/pull/13527

app/code/Magento/Braintree/Test/Unit/Gateway/Response/PayPal/VaultDetailsHandlerTest.php
ihor-sviziev: @joni-jones looks like you're talking more about integration tests that are testing just that functionality of the class works correct (check outside). 
But unit tests are testing how class behaves in isolation (see what it actually doing), so this actually correct behavior

YevSent: Not only about integration tests. How `expects()` make a tested class isolated? Also, 
there is more than one approach how to write unit tests https://martinfowler.com/bliki/UnitTest.html

app/code/Magento/Braintree/Model/Adminhtml/Source/PaymentAction.php
ihor-sviziev: I checked, these constants are still used in paypal, auth.net payment methods, sales module. For sure these constants were used in much more places if we will look in custom modules, so this class will live marked as "deprecated" really long time.

In order to fix it - we should re-add these constants to new or existing class/interface that is not marked as deprecated.

I think such changes should be discussed with magento architects.

YevSent: @ihor-sviziev I've added them to `MethodInterface` in the PR for 2.3-develop

PR #13544, opened 2018-02-07
[FIX] Add qty type check to the wishlist update action

coderimus:
### Description
In the wishlist page it is possible to set 0 for qty input if you try to add decimal less than 1 but greater than 0 for products having "Qty Uses Decimals" eq to "No".
From my point of view, it is not so logically because:
1. You can add the decimal value to product which should not have this type of qty
2. You can get 0 and will not be able to proceed with cart/checkout until changing qty to the right format.
3. Just unexpectable behaviour :)

N/A

### Manual testing scenarios
1. Add product which has "Qty Uses Decimals"="No" to the wishlist
2. Go to the wishlist page
3. Set qty to 0.5
4. Press "Update Wish List"
5. Check qty input value - it will have 0 value

### Contribution checklist

### DIFF
--- a/app/code/Magento/Wishlist/Controller/Index/Update.php
+++ b/app/code/Magento/Wishlist/Controller/Index/Update.php
@@ -8,6 +8,11 @@ namespace Magento\Wishlist\Controller\Index;
 use Magento\Framework\App\Action;
 use Magento\Framework\Exception\NotFoundException;
 use Magento\Framework\Controller\ResultFactory;
+use Magento\Wishlist\Model\LocaleQuantityProcessor;
+use Magento\Wishlist\Controller\WishlistProviderInterface;
+use Magento\Framework\Data\Form\FormKey\Validator;
+use Magento\CatalogInventory\Api\StockItemRepositoryInterface;
+use Magento\Framework\App\ObjectManager;
 
 class Update extends \Magento\Wishlist\Controller\AbstractIndex
 {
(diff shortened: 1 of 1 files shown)


sidolov:
Fixed unnecessary load stock item if qty is not double.

app/code/Magento/Wishlist/Controller/Index/Update.php
```diff
@@ -88,6 +102,10 @@ public function execute()
                 $qty = null;
                 if (isset($post['qty'][$itemId])) {
                     $qty = $this->quantityProcessor->process($post['qty'][$itemId]);
+                    if (is_double($qty)) {
+                        $stockItem = $this->stockItemRepository->get($itemId);
+                        $qty = !$stockItem->getIsQtyDecimal() ? floor($qty) : $qty;
```
sidolov, line 107: I have a concern about floor method call, i think better check qty and return message for user that qty is not valid. The same behavior used in case when you try to update qty to decimal in the shopping cart. You can find logic regarding this check in: https://github.com/magento/magento2/blob/36bfa583f12ccf79c754920d6b6288e67fcc9961/app/code/Magento/CatalogInventory/Model/StockStateProvider.php#L164

sidolov (changes requested):

sidolov:
@Coderimus , I am closing this PR now due to inactivity.
Thank you for collaboration!

coderimus:
Hi @sidolov sorry for the delay in the response.
I have one question about your previous comment where you suggested qty check. Can you please specify what type of check do you want me to add? I mean stock qty check (availability of a product in stock or just show error/notification message that qty not correct?
Thank you and sorry once more for the delay.

sidolov:
Hi @Coderimus , i mean show error/notification message that qty not correct, customer needs to know what went wrong and how he can fix the problem

coderimus:
@sidolov thank you for clarification. I will add needed changes and reopen this request.

PR #13569, opened 2018-02-08
Correctly save Product Custom Option values

Jeroenll:
### Description
This occurred because the same value object is used for saving all values. After the save `updateStoredData()` is called and is used in `prepareDataForUpdate()`.
https://github.com/magento/magento2/blob/3561b6614eeafb70cbcbf0dc0254359b857af3d4/lib/internal/Magento/Framework/Model/ResourceModel/Db/AbstractDb.php#L718-L733

1. magento/magento2#5067: Custom option values do not save correctly

### Manual testing scenarios
1. Install module [ReachDigital_CustomOption.zip](https://github.com/magento/magento2/files/1707887/ReachDigital_CustomOption.zip)
2. An extra column as added to the Product Custom Options (http://cloud.h-o.nl/pP3N).
3. Uncheck all `Changes Appearance` checkboxes and save product.
4. Only the first checkbox is saved.

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13569) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=13569) before we can accept your contribution.<br/>

app/code/Magento/Catalog/Model/Product/Option/Value.php
```diff
@@ -190,27 +190,29 @@ public function getProduct()
     public function saveValues()
     {
         foreach ($this->getValues() as $value) {
-            $this->isDeleted(false);
-            $this->setData(
+            $optionValue = clone $this;
```
dmanners: Do you know exactly what values need to be set on this object to save correctly? I am thinking rather than cloning here we add a factory that can create the object as needed. We should also consider the cleanup of these objects once we are done with them.

dmanners (changes requested):

app/code/Magento/Catalog/Model/Product/Option/Value.php
Jeroenll: That depends on what type of custom option you added I think. So no, don't think you have a predefined list of values that need to be set unless we have factories for each option type?

dmanners: @JeroenVanLeusden do you think you could update the code to use a factory rather than a clone and see if this covers all the cases?

Jeroenll: Sure do, hopefully have some time by the end of this week.

Jeroenll: @dmanners Code underneath makes saving stop altogether. Am I missing something?

See that a `Magento\Catalog\Model\Product\Option\Value\Repository` is not present, might be an idea to add that or include the saving of custom options in `Magento\Catalog\Model\Product\Option\Repository`?

```PHP
foreach ($this->getValues() as $value) {
    $option = $this->getOption();
    $data = array_merge($value, [
        'option_id' => $option->getId(),
        'store_id' => $option->getStoreId()
    ]);
    $optionValue = $this->customOptionValueFactory->create(['data' => $data]);

    if ((bool) $optionValue->getData('is_delete') === true) {
        if ($optionValue->getId()) {
            $optionValue->deleteValues($optionValue->getId());
            $optionValue->delete();
        }
    } else {
        $optionValue->save();
    }
}
```

dmanners: Hi @JeroenVanLeusden I would be tempted to add the repository for just working with the option values. Though https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Model/Product/Option/Repository.php#L146 does deal with saving of option values.

Jeroenll: @dmanners What are you proposing? Don't know what the preferred way is to get this done.

dmanners: Hi @JeroenVanLeusden sorry for the delay. I would say your suggestion of adding the saving of custom options in `Magento\Catalog\Model\Product\Option\Repository` would be a good option for this.

Jeroenll:
@dmanners Got it working by using a factory in `Magento\Catalog\Model\Product\Option::afterSave()`. Can you share your thoughts?

dmanners:
Hi @JeroenVanLeusden let me take another look at this for you.

magento-engcom-team:
Hi @JeroenVanLeusden. Thank you for your contribution. 

Issue #11657, opened 2017-10-23
app:config:import fails - flag column 'flag_data' is too short ?


peterjaap:
I got this error; `Import failed: Unable to unserialize value, string is corrupted.` when running `php bin/magento app:config:import` to import my `app/etc/config.php`.

It turns out, the importer does a comparison in `\Magento\Config\Model\Config\Importer::import`. To do this comparison, it retrieves a 'flag'; the `system_config_snapshot` flag. This is saved in the database. The problem is; the `flag_data` column in the flag table in Magento 2 is a TEXT, which is too short; it can only hold 65535 characters! It therefore cuts off the serialized data, making it corrupt.

Maybe this is a solution; `ALTER TABLE flag MODIFY flag_data LONGTEXT COMMENT 'Flag Data';` ?

### Preconditions

1. Magento 2.2

### Steps to reproduce

1. Dump a large enough configuration file and import it (message me for an example file, it contains client data but willing to share with Magento employees) -- you should also be able to create a large config.php by inserting large data sets into the `translation` table, since it also exports i18n settings.

### Expected result

1. It should import `app/etc/config.php`

### Actual result

1. Message; `Import failed: Unable to unserialize value, string is corrupted.`

peterjaap:
I circumvented this issue by altering `flag_data` to a LONGTEXT (from a TEXT) and invalidating the hashes so it will force import my config file; `UPDATE flag SET flag_data = '{"system":"","scopes":"","themes":""}' WHERE flag_code = 'config_hash'`. Flush your cache after running this query.

Another workaround for me was to remove 'i18n' and 'system' from `app/etc/config.php`, run `UPDATE flag SET flag_data = '{}' WHERE flag_code = 'system_config_snapshot';` and `php bin/magento app:config:import`.

peterjaap:
Additional thought; `app:config:dump` dumps EVERYTHING. Maybe the CLI command needs some extra features to be able to only export it partially (scopes, system, themes, i18n) ?

I'm thinking of leaving out system and i18n by default.

Especially the i18n part can grow very big when somebody has used the inline translation functionality a lot, thereby hitting the MySQL TEXT character limit quite soon.

peterjaap:
[...]

PR #13579, opened 2018-02-09
Increase the size of the flag_data field

renevogl:
The `flag_data` field in the `flag` table is too small to reimport a config generated by `bin/magento app:config:dump` with more than 64KB size.
Increasing this field should be considered, as exporting the config and reimporting it is the suggested way for pipeline deployment.

### Description
If a `bin/magento app:config:dump` produces a config.php with more than 64KB of size, the `bin/magento app:config:import` or `bin/magento setup:upgrade` will fail because of the `TEXT` field limitation from `flag_data` in the `flag` table.

This PR increases the `flag_data` field to `MEDIUMTEXT` accepting 16MB.

1. magento/magento2#11657: app:config:import fails - flag column 'flag_data' is too short ?

### Manual testing scenarios
1. Run `bin/magento setup:install`
2. Before applying the PR, the `flag_data` field is a `TEXT` field accepting 64KB
3. After applying the PR, the `flag_data` field is a `MEDIUMTEXT` field accepting 16MB

### Contribution checklist

### DIFF
--- a/setup/src/Magento/Setup/Model/Installer.php
+++ b/setup/src/Magento/Setup/Model/Installer.php
@@ -741,7 +741,7 @@ class Installer
             )->addColumn(
                 'flag_data',
                 \Magento\Framework\DB\Ddl\Table::TYPE_TEXT,
-                '64k',
+                '16m',
                 [],
                 'Flag Data'
             )->addColumn(


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13579) <br/>All committers have signed the CLA.

miguelbalparda:
Hi @aschrammel! Thanks for your contribution! 
Shouldn't this be implemented in an update instead of modifying the installer?

aschrammel:
Hi @miguelbalparda!
Normally I also think such changes should be implemented in an update. Is there a possibility to implement updates on the initial setup process that are applied immediately during setup?
I think in this case it is important that the table is set up correctly from beginning on - i.e. after the initial `bin/magento setup:install`.

orlangur:
@aschrammel initial `bin/magento setup:install` applies all data upgrades immediately :)

orlangur (changes requested):
@renevogl please rework this change as data upgrade, do amend commit and force push.

aschrammel:
@orlangur I've now had a look, but where should we place the upgrade script? The Installer itself is in `setup` directory and the flag management isn't a module itself, but contained in the `lib/`.
And I'm not quite sure if the `Config` module is the right place.

miguelbalparda:
@dmanners anything to add here?

dmanners:
Hey @miguelbalparda I think there should be something that could be done in the `setupFlagTable` method so that we cover the cases when there is already a database and we install Magento as well as the case for a clean installation.

I am not really sure if or how the setup module works with upgrade scripts.

miguelbalparda:
Closing due to lack of activity, feel free to reopen if anything changes.

PR #13581, opened 2018-02-09
Prevents to generate an empty string for url rewrite

real34:
… when using default value from store view in admin

### Description

This PR is a fix for invalid `Use Default Value` saved from the admin on product url key, because the value of `$product->getUrlKey()` is `null` instead of `false`

It seems that the same fix (maybe at another abstraction level) might apply to

1. https://github.com/magento/magento2/issues/12982
2. https://github.com/magento/magento2/issues/10424
3. https://github.com/magento/magento2/issues/12159#issuecomment-343528601

### Manual testing scenarios

1. AS An Admin, create 2 store views
2. Create a product
3. Change the url key at a store level and save the product
4. Edit the product to recheck the `Use Default Value` checkbox

What happens (without this patch):

selection_003

What happens (with this patch):

selection_002

### Contribution checklist

### More

If you think it could be useful to fix a more general issue with EAV `Use Default Value`, let me know and I could spend more time on a fix.

### DIFF
--- a/app/code/Magento/CatalogUrlRewrite/Model/ProductUrlPathGenerator.php
+++ b/app/code/Magento/CatalogUrlRewrite/Model/ProductUrlPathGenerator.php
@@ -124,7 +124,8 @@ class ProductUrlPathGenerator
      */
     public function getUrlKey($product)
     {
-        return $product->getUrlKey() === false ? false : $this->prepareProductUrlKey($product);
+        $usesDefaultValue = $product->getUrlKey() === false || $product->getUrlKey() === null;
+        return $usesDefaultValue ? false : $this->prepareProductUrlKey($product);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/CatalogUrlRewrite/Model/ProductUrlPathGenerator.php
```diff
@@ -124,7 +124,8 @@ public function getCanonicalUrlPath($product, $category = null)
      */
     public function getUrlKey($product)
     {
-        return $product->getUrlKey() === false ? false : $this->prepareProductUrlKey($product);
+        $usesDefaultValue = $product->getUrlKey() === false || is_null($product->getUrlKey());
```
rogyar: Please, avoid using `is_null`, use `$var === null` instead. Thanks

rogyar (changes requested):

rogyar:
Hello @real34. Please, review and fix the failed tests. 
Thanks!

real34:
Hi, after investigation in the failed tests and more reading about related issues I think the issue is much more critical and impacts other EAV attributes in multistore mode. Hence the fix for this issue will probably be totally different so I’ll close this PR for now.

Do you know if the MDVA-6475 Enterprise patch will be backported to the OpenSource? (see https://github.com/magento/magento2/issues/10424#issuecomment-343518915)

Thanks for your quick response.

PR #13602, opened 2018-02-10
Add event to admin user edit form

pmclain:
### Description
The event allows addition of form fields by extensions without needing
to override the block class. Structured similar to the event dispatched
in `Magento\Catalog\Block\Adminhtml\Product\Attribute\Edit\Tab\Front`

N/A

### Manual testing scenarios
1. Create module and observe `adminhtml_backend_user_edit_prepare_form` event.
2. Perform task on form object

### Contribution checklist

### DIFF
--- a/app/code/Magento/Backend/Block/System/Account/Edit/Form.php
+++ b/app/code/Magento/Backend/Block/System/Account/Edit/Form.php
@@ -72,8 +72,7 @@ class Form extends \Magento\Backend\Block\Widget\Form\Generic
      */
     protected function _prepareForm()
     {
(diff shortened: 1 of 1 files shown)


orlangur:
I've checked descendants of `\Magento\Backend\Block\Widget\Form\Generic` class and most of them do not contain such event dispatching in `_prepareForm`.

As `\Magento\Backend\Block\Widget\Form\Generic` was deprecated in favor of UI component implementation adding new events is not idea as after it is refactored implementation will have to be adjusted anyway. Class overriding is not that bad here.

PR #13691, opened 2018-02-16
Fix for Issue-13556 - Sorting in Product Listing via Quantity not work

nuzil:
This Pull Request related to issue: https://github.com/magento/magento2/issues/13556

### Problem
If you edit product attribute 'quantity_and_stock_status' and set "Used for Sorting in Product Listing" to "Yes" sort option will be displayed on category products page but sorting will not work. 

### Reason
1. Attribute 'quantity_and_stock_status' has source model, and for sorting magento calls addValueSortToCollection method of source model.
2. However source model of this attribute ("Magento\CatalogInventory\Model\Source\Stock") does not have its own method "addValueSortToCollection" only inherits parent one which just do nothing (return $this).

### Fix
We added to source model method which contain logic to join to collection (which is always product collection) cataloginventory_stock_item table and add order by qty field.

### Manual testing scenarios
Testing scenarios can be found in issue
https://github.com/magento/magento2/issues/13556

### Contribution checklist

### DIFF
--- a/app/code/Magento/CatalogInventory/Model/Source/Stock.php
+++ b/app/code/Magento/CatalogInventory/Model/Source/Stock.php
@@ -43,6 +43,6 @@ class Stock extends AbstractSource
             []
         );
         $collection->getSelect()->order("stock_item_table.qty $dir");
-        return $this
+        return $this;
     }
 }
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13691) <br/>All committers have signed the CLA.

sidolov:
Changes covered with unit test

Issue #12354, opened 2017-11-20
When the Admin is on a different domain, redirect always fails and causes it to redirect to the front page


4DallasC:
When the admin setting 'admin/url/custom' is used to change the domain for the admin, the redirect always fails.

### Preconditions
    /**
     * Check whether URL is internal
     *
     * @param string $url
     * @return bool
     */
    protected function \Magento\Store\App\Response\Redirect::_isUrlInternal($url)
    {
        if (strpos($url, 'http') !== false) {
            $directLinkType = \Magento\Framework\UrlInterface::URL_TYPE_DIRECT_LINK;
            $unsecureBaseUrl = $this->_storeManager->getStore()->getBaseUrl($directLinkType, false);
            $secureBaseUrl = $this->_storeManager->getStore()->getBaseUrl($directLinkType, true);
            return (strpos($url, $unsecureBaseUrl) === 0) || (strpos($url, $secureBaseUrl) === 0);
        }
        return false;
    }

### Steps to reproduce
1) Change the admin to a custom url via the configurations
2) use a page that responds via the redirect

### Expected result
The page should redirect to the referer url.

### Actual result
The above function fails, because it tests the frontend domains against the custom admin domain and always fails.
This failure, then causes the page to redirect to the frontend and not the custom admin url as expected.

magento-engcom-team:
@4DallasC, thank you for your report.
We've created internal ticket(s) MAGETWO-84133 to track progress on the issue.

briscoda:
@okorshenko I would like to work on this one, can it be assigned to me?

KarlDeux:
@briscoda here you go!

briscoda:
@KarlDeux 

Having trouble reproducing this. It seems like this all works the way that it should. 

I went into the admin and went to:

Store -> Configuration -> Advanced -> Admin 

Where I changed the following values.
<img width="1217" alt="settings_one" src="https://user-images.githubusercontent.com/197987/35762429-8197947e-0864-11e8-8b6f-59d69c954aff.png">


The attached screencasts show what I did to try to reproduce this. Let me know if there is something different that I should do.


Custom admin domain and path:
https://www.screencast.com/t/rRuMArVYUi

Custom admin path only:
https://www.screencast.com/t/9D9LWq6CEXz

4DallasC:
Hello Paul,

The custom domain should be:-
https://admin.screencast.com/t/rRuMArVYUi 
[...]

PR #13714, opened 2018-02-18
update the store app response redirect to also check for the backendu…

briscoda:
…rl when determining if the url isInternal

The goal of this was to create an additional check in the isUrlInternal method to take into account if there is a custom admin domain used. Because the default config returns the default store url, if you are using a custom admin domain without explicitly setting the store id to 0 it will return the wrong url for a test of the admin.

### Description

1. Add a call to the backendFrontNameresolver to get the admin url and use it as an additional check
2. In the Magento\Backend\App\Area\FrontNameresolver add the backendUrl code to its own method to get set into a private variable which can be accessed by itself.

1. magento/magento2#12354: When the Admin is on a different domain, redirect always fails and causes it to redirect to the front page

### Manual testing scenarios
1. Set a custom admin path in the admin
2. Create a test method which uses the admin host as a variable.
3. Pass the variable to the \Magento\Store\App\Response\Redirect::_isUrlInternal($url) method and verify that it returns true instead of false

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/App/Response/Redirect.php
+++ b/app/code/Magento/Store/App/Response/Redirect.php
@@ -8,6 +8,7 @@
 namespace Magento\Store\App\Response;
 
 use Magento\Store\Api\StoreResolverInterface;
+use Magento\Backend\App\Area;
 
 class Redirect implements \Magento\Framework\App\Response\RedirectInterface
 {
@@ -47,6 +48,11 @@ class Redirect implements \Magento\Framework\App\Response\RedirectInterface
     protected $_urlBuilder;
 
     /**
+     * @var \Magento\Backend\App\Area\FrontNameResolver
+     */
+    protected $_backendFrontNameResolver;
+
+    /**
      * Constructor
      *
      * @param \Magento\Framework\App\RequestInterface $request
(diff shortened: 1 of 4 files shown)


app/code/Magento/Store/App/Response/Redirect.php
```diff
@@ -63,6 +70,7 @@ public function __construct(
         \Magento\Framework\Encryption\UrlCoder $urlCoder,
         \Magento\Framework\Session\SessionManagerInterface $session,
         \Magento\Framework\Session\SidResolverInterface $sidResolver,
+        \Magento\Backend\App\Area\FrontNameResolver $backendFrontNameResolver,
```
orlangur: Note that `Backend` module must not be referenced in `Store` module.

briscoda: Ah, I just saw this comment now. I was struggling to find a method that would return the actual full admin url is it has been customized. There are some methods in some deprecated helpers.

https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Backend/App/Area/FrontNameResolver.php#L103

Essentially, what needs to be done is to get this backendUrl inside of this Redirect class. 

My further commits are probably not going to cut it if the Backend module cannot be referenced from here.

GuiltyNL:
What is the status on this issue. I see that @orlangur here and there says that you don't want to use the store redirect class in backend modules, but there doesn't seem to be a solution what to use instead?

PR #13718, opened 2018-02-18
Magetwo 84133-2 Update the _isUrlInternal checker to account for a custom admin url

briscoda:
### Description

Use the storeConfigInterface to retrieve the backend url and add it as a check in the _isUrlInternal method

1. magento/magento2#12354: When the Admin is on a different domain, redirect always fails and causes it to redirect to the front page

### Manual testing scenarios

1. Set a custom admin path in the admin
2. Create a test method which uses the admin host as a variable.
3. Pass the variable to the \Magento\Store\App\Response\Redirect::_isUrlInternal($url) method and verify that it returns true instead of false

### Contribution checklist


app/code/Magento/Store/App/Response/Redirect.php
```diff
@@ -7,10 +7,21 @@
...
 namespace Magento\Store\App\Response;
 
+use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Store\Api\StoreResolverInterface;
+use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\Store;
 
 class Redirect implements \Magento\Framework\App\Response\RedirectInterface
 {
+    const XML_PATH_USE_CUSTOM_ADMIN_PATH = 'admin/url/use_custom_path';
```
orlangur, line 17: Again, `Backend` module depends on `Store` and not vise versa, `Store` module must not receive any knowledge about admin part. I cannot say a word about whether proposed implementation is correct or not, just pointing out to a modularity violation.

It may be a good idea to wait for a full code review before doing further changes.

orlangur: @briscoda another small note, there is no need to create new pull requests, just continue work with the same branch, you can force push at any moment also.

briscoda: I appreciate your notes, thank you. There is also a really simple solution to this too, which would be to just call:

$this->_storeManager->getStore(0);

And explicitly pass it 0 which forces it to return the configured admin domain. It's like a one-line thing which could be inserted into the _isUrlInternal function.

Just for curiosity though, in this current implementation, it calls information from the scopeConfig which is in Framework. So although these constants use config paths which have been set for the admin, I'm not sure if they're strictly tied to the Backend module. But I see what you're saying.

briscoda: Alternatively, maybe a plugin that lived in the Backend module that happened after the isUrlInternal function could accomplish this as well. That way references to the admin paths are not contained in this store module but live there instead.

orlangur:
@briscoda fix needs to be applied to `2.2-develop` first.

sidolov:
Hi @briscoda , I am closing this PR now due to inactivity.

LSERRE:
Any update on this Pull Request? It looks like to fix #12354.

PR #13734, opened 2018-02-19
Cart state (selected shipping method for estimation) is lost on coupon apply since 2.2.0

driskell:
### Description
When you apply a coupon on the cart page, the selected shipping option is lost.

I'm aware this might not be the ideal fix. In general though it seems like the way to go. At the moment there are many sectionData objects in the frontend and most of them mirror some data on the server-side, so when private data version changes it is expected for them to be reloaded from the server. However, the checkout-data sectionData contains data that pertains to front-end state and has no mirror on the server. Thus it cannot be reloaded from the server. I figured the best fix is a way to flag a sectionData as being something that needs to persist across version changes, and I noticed there is already a `data_id` special property so added a `data_persistence` property.

Happy to discuss options to get the desired behaviour here.

N/A

### Manual testing scenarios
1. Select a shipping method
2. Apply a coupon code (even an invalid one)
3. The selected shipping method is lost on page load

After fix, at 3, the selected shipping method is maintained as per Magento 2.1.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/checkout-data.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/checkout-data.js
@@ -20,6 +20,7 @@ define([
          * @param {Object} data
          */
         saveData = function (data) {
+            data['data_persistence'] = true;
             storage.set(cacheKey, data);
         },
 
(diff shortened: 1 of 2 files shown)


rogyar:
Hello @driskell. I have just tested the solution and found an issue. On the latest 2.2-develop branch on an attempt to apply a coupon code (an incorrect one) the page is reloaded with shipping methods selection disabled. The previous shipping method selection is preserved but you cannot select another without reloading the page.

screen shot 2018-04-29 at 12 08 40

sidolov:
@driskell , I am closing this PR now due to inactivity.
Thank you for collaboration!

PR #13759, opened 2018-02-20
Add ObserverInterface to the api

ghost:
### Description
Creating an observer that uses ObserverInterface should not trigger a patch level dependency on magento/framework.

N/A

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Event/ObserverInterface.php
+++ b/lib/internal/Magento/Framework/Event/ObserverInterface.php
@@ -10,6 +10,7 @@ namespace Magento\Framework\Event;
 /**
  * Interface \Magento\Framework\Event\ObserverInterface
  *
+ * @api
  */
 interface ObserverInterface
 {


ihor-sviziev (approved):

magento-engcom-team:
[MAGETWO-88146](https://jira.corp.magento.com/browse/MAGETWO-88146) has been created to process this Pull Request

magento-engcom-team:
@fooman thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

lib/internal/Magento/Framework/Event/ObserverInterface.php
```diff
@@ -10,6 +10,7 @@
 /**
  * Interface \Magento\Framework\Event\ObserverInterface
  *
+ * @api
```
orlangur, line 13: It would be good to add a note also that observers creation should be avoided whenever possible in favor of plugins-interceptors.

Vinai: Why? There are plenty of places where an event observer is better suited than a plugin.

orlangur: Mainly because "There can be only one" intention.

The only valid place for event dispatching I heard of are some controllers in MSI project.

All other places in current code which seem better suited by observers is just unrefactored legacy which is doing to much. In code written with plugins in mind there is no need in event dispatching at all.

Vinai: Lol, there is soooo much legacy code that will never be rewritten because it would break so many extensions. Maybe in Magento 3. But for Magento 2 I don't consider that realistic.

PR #13775, opened 2018-02-21
cron_schedule forever increasing in size. Lots of pending jobs never cleared #11002

gnuzealot:
Reported in #11002 the cron_schedule table can lock up while running multiple update statements against the table. This is due to a high cost query, where a left join is included in the update to ensure uniqueness. Written this way the query forces a full table scan and lock as it is updating a row it must also select. As this table naturally grows in size the cost of the query will keep increasing. This was fixed by removing the left join from the update and doing a separate select query to still ensure that unique job codes will run.

### DIFF
--- a/app/code/Magento/Cron/Model/ResourceModel/Schedule.php
+++ b/app/code/Magento/Cron/Model/ResourceModel/Schedule.php
@@ -61,7 +61,7 @@ class Schedule extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
      * @return bool
      * @since 100.2.0
      */
-     public function trySetJobUniqueStatusAtomic($scheduleId, $newStatus, $currentStatus)
+    public function trySetJobUniqueStatusAtomic($scheduleId, $newStatus, $currentStatus)
     {
         $connection = $this->getConnection();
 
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=13775) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=13775) before we can accept your contribution.<br/>

ihor-sviziev:
Hi @gnuzealot could you review tests that are failed in Travis and adjust your code?

gnuzealot:
@ihor-sviziev all checks are passed now

gwharton:
Note, that this fix will reduce the server load when the cron_schedule table grows unbounded, but does not fix the root cause of #11002 which is the table growing in the first place.

sidolov:
Hi @gnuzealot , looks like the described issue was fixed in the scope of https://github.com/magento/magento2/pull/12497. Can you check the behavior on the latest 2.2-develop? 
Thank you!

sidolov:
Hi @gnuzealot , I am closing this PR now due to inactivity.

PR #13796, opened 2018-02-22
Save CMS Block using repository

Jeroenll:
### Description
Update Save action of CMS block to be consistent with CMS Page.

1. None

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cms/Controller/Adminhtml/Block/Save.php
+++ b/app/code/Magento/Cms/Controller/Adminhtml/Block/Save.php
@@ -7,10 +7,12 @@
 namespace Magento\Cms\Controller\Adminhtml\Block;
 
 use Magento\Backend\App\Action\Context;
+use Magento\Cms\Api\BlockRepositoryInterface;
 use Magento\Cms\Model\Block;
+use Magento\Cms\Model\BlockFactory;
 use Magento\Framework\App\Request\DataPersistorInterface;
 use Magento\Framework\Exception\LocalizedException;
-use Magento\TestFramework\Inspection\Exception;
+use Magento\Framework\Registry;
 
 class Save extends \Magento\Cms\Controller\Adminhtml\Block
 {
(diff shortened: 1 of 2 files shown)


app/code/Magento/Cms/Controller/Adminhtml/Block/Save.php
```diff
@@ -55,27 +73,33 @@ public function execute()
...
             /** @var \Magento\Cms\Model\Block $model */
-            $model = $this->_objectManager->create(\Magento\Cms\Model\Block::class)->load($id);
-            if (!$model->getId() && $id) {
-                $this->messageManager->addError(__('This block no longer exists.'));
-                return $resultRedirect->setPath('*/*/');
+            $model = $this->blockFactory->create();
+
+            $id = $this->getRequest()->getParam('block_id');
+            if ($id) {
+                $model->load($id);
```
dmanners: Could we not also load the block from the repository?

dmanners: Something like `$this->blockRepository->getById($id)`

dmanners (changes requested):

app/code/Magento/Cms/Controller/Adminhtml/Block/Save.php
Jeroenll: Sure, give me 2 sec.

Jeroenll:
@dmanners Updated 😄

dmanners:
@JeroenVanLeusden you will also have to cover the unit tests updates as well. Since `app/code/Magento/Cms/Test/Unit/Controller/Adminhtml/Block/SaveTest.php` will now fail as it is written with the model in mind.

Jeroenll:
Totally forgot that one, will update those tonight and ping you when ready.

Jeroenll:
@dmanners Should be good to go now.

PR #13914, opened 2018-03-01
Pass Expected Data Type in backgroundColor Call (2.2)

ghost:
### Description
`\Magento\Framework\Image::backgroundColor` is expecting `null|array` as valid parameter types. However, in `\Magento\Swatches\Helper\Media::setupImageProperties` the string value '#FFF' is passed into this function call. The goal of this fix is to pass along the correct and expected type, which in this case is going to be an array of RGB values to produce the same end result.

This can also be seen on `2.1-develop` (addressed in #13916) and `2.3-develop` (addressed in #13917).

### Contribution checklist


magento-engcom-team:
@northernco thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

sidolov:
Fixed unit test according to changes

PR #13933, opened 2018-03-03
Configurable product price options by store

simpleadm:
Added compatability of `LowestPriceOptionsProvider` class and environmen emulation fucntionality.
Updated `linkedProductMap` property in lowest price options provider to cache options by store and product id.

### Description
Some modules (e.g. Abandoned Cart, Algolia Search etc.) use store emulation functionality (`\Magento\Store\Model\App\Emulation::startEnvironmentEmulation`) to get product info (including configurable product price) for each store. Unfortunately `LowestPriceOptionsProvider` class save linked products collection at the `$linkedProductMap` property based on requested product id only. That is why configurable product price will be the same for other stores after first emulation.

### Contribution checklist

Forwardport for: https://github.com/magento/magento2/pull/13808

### DIFF
--- a/app/code/Magento/ConfigurableProduct/Pricing/Price/LowestPriceOptionsProvider.php
+++ b/app/code/Magento/ConfigurableProduct/Pricing/Price/LowestPriceOptionsProvider.php
@@ -9,6 +9,7 @@ use Magento\Catalog\Api\Data\ProductInterface;
 use Magento\Catalog\Model\ResourceModel\Product\LinkedProductSelectBuilderInterface;
 use Magento\Framework\App\ResourceConnection;
 use Magento\Catalog\Model\ResourceModel\Product\CollectionFactory;
+use Magento\Store\Model\StoreManagerInterface;
 
 /**
  * Retrieve list of products where each product contains lower price than others at least for one possible price type
@@ -31,7 +32,12 @@ class LowestPriceOptionsProvider implements LowestPriceOptionsProviderInterface
     private $collectionFactory;
 
     /**
-     * Key is product id. Value is array of prepared linked products
+     * @var StoreManagerInterface
+     */
+    private $storeManager;
+
+    /**
+     * Key is product id and store id. Value is array of prepared linked products
      *
      * @var array
      */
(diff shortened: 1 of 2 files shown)


app/code/Magento/ConfigurableProduct/Pricing/Price/LowestPriceOptionsProvider.php
```diff
@@ -41,34 +47,38 @@ class LowestPriceOptionsProvider implements LowestPriceOptionsProviderInterface
...
+        $this->storeManager = $storeManager;
     }
 
     /**
      * {@inheritdoc}
      */
     public function getProducts(ProductInterface $product)
     {
-        if (!isset($this->linkedProductMap[$product->getId()])) {
+        $key = $this->storeManager->getStore()->getId() . '-' . $product->getId();
```
kandy, line 69: Price of product in Magento can be modified only on website level, so better use website id here

simpleadm: You are right, prices and tax class id can be modified only on website level, but scope for 'Special Price From Date' and 'Special Price To Date' can be changed from admin panel http://prntscr.com/in1pba . So that attribute can affect results. That's why i decided to use store id.

kandy: I believe that scope for 'Special Price From Date' is wrong. All price related option should be a website or global level.

simpleadm: You are totally right, but we have what we have and that works fine. If we will change for website level - in environmen emulation we will get different results than on real website. That why i propose to keep it without changes. 


I tried to create 3 products: 2 simple, 1 configurable and set 'Special Price From Date'  for store view level. 
Results:
Default store view: http://prntscr.com/ipevz0
Second store view: http://prntscr.com/ipewma

Product settings: 
http://prntscr.com/ipex7e (Second store view)
http://prntscr.com/ipexld (all stores and default store view is without overrides)

simpleadm: On my opinion, store view level caching shouldn't affect performance.

If i remember correctly 'Special Price From Date' attribute doesn't exist in price index, that why it could be in store view level and works correctly.

simpleadm: The last word is yours ;)

kandy: In any case, your solution fixes the problem and improves the system, and my comment is just a suggestion for improvement, you can ignore it as a minor

PR #13945, opened 2018-03-04
ScopeConfigInterface doesn't define a 'clean' method, so this shouldn…

hostep:
…'t get called on all implementations of this interface.

### Description
This is something I discovered by accident while debugging something else, and is probably a minor issue.
Anyway, so the [ScopeConfigInterface](https://github.com/magento/magento2/blob/2.2.3/lib/internal/Magento/Framework/App/Config/ScopeConfigInterface.php) doesn't define a `clean` method, it is however defined in the [Config](https://github.com/magento/magento2/blob/2.2.3/lib/internal/Magento/Framework/App/Config.php) implementation of that interface.

So the `clean` method can't be called on all implementations of the `ScopeConfigInterface` but only on the `Config` implementation.

I found inspiration over here where the same solution is being used:
- https://github.com/magento/magento2/blob/2.2.3/app/code/Magento/Config/Model/Config/Importer.php#L123-L125
- https://github.com/magento/magento2/blob/2.2.3/app/code/Magento/Config/Console/Command/ConfigSet/ProcessorFacade.php#L126-L128

None that I could find

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/Model/StoreManager.php
+++ b/app/code/Magento/Store/Model/StoreManager.php
@@ -5,6 +5,7 @@
  */
 namespace Magento\Store\Model;
 
+use Magento\Framework\App\Config;
 use Magento\Framework\App\ObjectManager;
 use Magento\Store\Api\StoreResolverInterface;
 use Magento\Store\Model\ResourceModel\StoreWebsiteRelation;
@@ -235,7 +236,9 @@ class StoreManager implements
     {
         $this->currentStoreId = null;
         $this->cache->clean(\Zend_Cache::CLEANING_MODE_MATCHING_ANY_TAG, [StoreResolver::CACHE_TAG, Store::CACHE_TAG]);
-        $this->scopeConfig->clean();
+        if ($this->scopeConfig instanceof Config) {
+            $this->scopeConfig->clean();
+        }
         $this->storeRepository->clean();
         $this->websiteRepository->clean();
         $this->groupRepository->clean();


rogyar:
Hello @hostep. Thank you for your contribution. I agree that the `clean` method ideally should be described within a scope of `ScopeConfigInterface`. But it's not.

By bringing your fix we might create an additional issue. Let's imagine a situation that some custom solution replaces the original `ScopeConfigInterface` implementation by its own. And, obviously, the solution expects the `clean` method to be called. In the current state, the mentioned method is called. But after the fix, the `clean` method will be called only in case of the original implementation.

hostep:
Correct, that was my fear as well. But since I found similar "solutions" as the one I proposed, I decided to go for that one.

But I agree that it would be better to define the `clean` method in the interface itself, and then remove all those `instanceof` checks everywhere. Is that the correct solution? If so then I'll try to update my PR later tonight.

hostep:
@rogyar: I'm really confused by how everything is defined currently, so I'm not exactly sure what to do here.

We have an interface `ScopeConfigInterface`, which only defines getter methods, it looks like this interface is only for getting values, not modifying anything. The reason why I bring this up, is because there is also an `MutableScopeConfigInterface`, which inherits from `ScopeConfigInterface`, which defines a setter and looks like this one is being used for modifying state.
So does it make sense to add a `clean` method to an interface which is build to not modify state? (if I understand it correctly that is)

Second reason why I am confused, the `Magento\Framework\App\Config` class uses the `clean` method to cleanup its caches, but should an interface already know this, if an implementation uses caching for its data?

Then, if you still think we need to add the `clean` method to the `ScopeConfigInterface`, will it be considered backwards compatible, because I thought we weren't allowed to add new methods to an interface? Or should I target the `2.3-develop` branch here?

Feedback would be appreciated, it's really hard to determine what's the correct action to take here...

rogyar:
Hello @hostep. As for adding changes to a core interface - we cannot do that since it will cause the backward compatibility issue. 
Let me summarize the situation. Currently, we don't have the `clean()` method in the interface. It might cause issues in case if `ScopeConfigInterface` implementation differs from the original one. However, 3rd party solutions developers can spot the issue in the current version and make the necessary changes in their code. So, the issue we have is only conceptual. 
By making new changes to the interface/implementation from our side we can bring the real logical issue that might affect functionality of the existing stores. So, the chance that we will bring the issue instead of a fix is really high. From this point of view, I would recommend leaving this part of the logic as it is.
Thanks

hostep:
@rogyar: thanks for the feedback!

I have another idea: would it make sense to introduce a new interface which extends from `ScopeConfigInterface`, but with the `clean` method added?

Then let `Magento\Framework\App\Config` implement this new interface.

And then change the constructors of the following 3 classes to use the new interface instead of `ScopeConfigInterface`:
- https://github.com/magento/magento2/blob/2.2.3/app/code/Magento/Config/Model/Config/Importer.php#L89
- https://github.com/magento/magento2/blob/2.2.3/app/code/Magento/Config/Console/Command/ConfigSet/ProcessorFacade.php#L82
- https://github.com/magento/magento2/blob/2.2.3/app/code/Magento/Store/Model/StoreManager.php#L102

Or would that also be considered backwards compatible breaking?


I agree that the chance the current situation triggers an issue is very low, so I can close this PR if it doesn't make much sense to have this fixed.

hostep:
Let's close this one, as it probably isn't that important to get fixed. This is more a job for some Magento core architect to figure out I think.

PR #13956, opened 2018-03-05
Use event object in 'ajax:addToCart' trigger when adding a product to the cart

koenner01:
### Description
Refactored the implementation of the 'ajax:addToCart' trigger that was used in the ajaxSubmit function when adding a product to the cart. The code now leans more closely towards how Magento has already implemented this type of event triggering.

Also by passing an object in the trigger function instead of passing x number of variables, we allow for more generic solutions when trying to observe multiple triggers with a single function.

1. Follow up on https://github.com/magento/magento2/pull/12875

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/web/js/catalog-add-to-cart.js
+++ b/app/code/Magento/Catalog/view/frontend/web/js/catalog-add-to-cart.js
@@ -97,7 +97,11 @@ define([
                 success: function (res) {
                     var eventData, parameters;
 
-                    $(document).trigger('ajax:addToCart', form.data().productSku, form, res);
+                    $(document).trigger('ajax:addToCart', {
+                        'sku': form.data().productSku,
+                        'form': form,
+                        'response': res
+                    });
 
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/view/frontend/web/js/catalog-add-to-cart.js
```diff
@@ -97,7 +97,11 @@ define([
                 success: function (res) {
                     var eventData, parameters;
 
-                    $(document).trigger('ajax:addToCart', form.data().productSku, form, res);
+                    $(document).trigger('ajax:addToCart', {
+                        'sku': form.data().productSku,
+                        'form': form,
+                        'response': res
+                    });
```
elvinristi, line 104: will this be taken account also in `magento/module-google-tag-manager`

- vendor/magento/module-google-tag-manager/view/frontend/web/js/google-tag-manager-cart.js 

where it expects (in M2.2.0 at least) string as 1st argument?

PR #13770, opened 2018-02-21
Remove not-allowed currencies from the currencies dropdown in Setup

r-martins:
### Description
As reported on #13760 some deprecated currencies were being displayed in the setup process.
This PR removes these and other not-allowed currencies from the setup step 4 (customize your store).

### Fixed Issues (if relevant)
1. Fixes #13760 by not displaying old and deprecated Brazilian currencies
2. Remove not-allowed currencies for being displayed

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #13966, opened 2018-03-06
Setup Lists - Make allowedCurrencies property private

ihor-sviziev:
### Description
In https://github.com/magento/magento2/pull/13770 was introduced new property with protected visibility, should be private.
This commit wasn't included into any release, so we still can fix this issue.

N/A

### Manual testing scenarios
N/A

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Setup/Lists.php
+++ b/lib/internal/Magento/Framework/Setup/Lists.php
@@ -26,7 +26,7 @@ class Lists
      *
      * @var array
      */
-    protected $allowedCurrencies;
+    private $allowedCurrencies;
 
     /**
      * @param ConfigInterface $localeConfig


r-martins:
@ihor-sviziev sorry for not seeing your previous comment on #13770.
I'm just wondering why $allowedLocales should be kept protected while $allowedCurrencies, private.
Thanks mate. 

Update: Both variables has protected visibility, and has public methods to get its information. I also checked that there's no other usage outside the class for both properties. Can you confirm that? Perhaps both might be private...

ihor-sviziev:
Hi @r-martins,
Unfortunately such change will be not backward compatible as it was already included as protected in one of releases.

r-martins:
That's true. Thanks for replying.

PR #13990, opened 2018-03-07
Allow using a symlinked sitemap path

cedricziel:
### Description

This will fix an issue with symlinked sitemaps where the symlink is replaced instead of the symlink target.

### Manual testing scenarios

1. share a sitemap.xml through symlinks throughout releases
2. regenerate a sitemap

### DIFF
--- a/app/code/Magento/Sitemap/Model/Sitemap.php
+++ b/app/code/Magento/Sitemap/Model/Sitemap.php
@@ -570,6 +570,11 @@ class Sitemap extends \Magento\Framework\Model\AbstractModel implements \Magento
         }
 
         $path = rtrim($this->getSitemapPath(), '/') . '/' . $fileName;
+
+        if (is_link($path) && readlink($path)) {
+            $path = readlink($path);
+        }
+
         $this->_stream = $this->_directory->openFile($path);
 
         $fileHeader = sprintf($this->_tags[$type][self::OPEN_TAG_KEY], $type);


miguelbalparda:
Can you please format this with the correct template?

cedricziel:
Done, thank you.

orlangur:
@cedricziel please provide more details on what you are trying to fix.

Note that Magento MUST NOT access any files outside of Magento base path.

cedricziel:
I have an instance that uses shared paths. In essence, deployer creates the following  directory layout:

```
- releases/ <- that's where the webroot is
- - {deployment_id}
- - current <- symlink to the most current {deployment_id}
- - - sitemap.xml <- shared sitemap.xml
- shared
- - sitemap.xml
```

Wenn a new version is deployed, deployer creates a new "release" folder, clones my app code (basically magento project) installs dependencies, compiles the container etc. - Configuration may be shared between releases (it will get symlinked from the shared folder to the release folder), and so do i want the sitemap to be shared as well - basically because there's no command to regenerate it. But sitemap.xml, when regenerated from the magento admin doesnt write to the symlink-target, but rather replaces the symlink with a file.

cedricziel:
Well, the file i want it to write is outside of the base path, but the symlinks is not.

Maybe the right way would be to add a command to be able to regenerate the sitemap as well as per CLI.

cedricziel:
This is superseded by #14053 which is a feature request, since it seems plausible to me that magento shouldn't be able to write outside its basedir and the feature is useful anyways.

orlangur:
Thanks @cedricziel, such CLI command truly makes sense to me.

As to me, if you symlinked sitemap within Magento base path it should work just like a regular file, I believe such case was just overlooked during implementation and not intentionally implemented that way.

cedricziel:
Thanks for triaging the idea - if you like it, it's still there, ready to merge ;)

I'm not sure if I can finish a PR that adds such command - maybe reconsider the current PR?

orlangur:
@cedricziel from http://php.net/manual/en/function.readlink.php it looks like this PR tries to access paths even outside of Magento base path.

> when regenerated from the magento admin doesnt write to the symlink-target, but rather replaces the symlink with a file

This is what I believe could be treated as a valid bug and fixed separately. Even if we have a CLI command this looks like a useful fix to me.

cedricziel:
That's exactly what this PR resolves.

The indirect result is that paths outside the base path (the symlink target that is, if the target is outside the base path) can be written to.

miguelbalparda:
I'll close this, will revisit if anything changes. Thanks!

PR #14017, opened 2018-03-08
Fix for Issue #13950 - Cache issue with configurable products related to currency-conversions

nuzil:
### Preconditions.
Magento website with configured multiple currencies, for some of currencies not configured currency symbol.

### Problem.
Two  Magento block classes used in generating  cache key (in method getCacheKeyInfo()) current currency symbol. It leads to problem that these blocks has same cache for different currencies which has no symbol configured.

### Solution.
Use currency code instead currency symbol in cache key generating to get unique cache of these block for different currencies.


orlangur (changes requested):
Hi @nuzil,

`ConfigurableProduct` case was covered by https://github.com/magento/magento2/pull/13894 already, could you please rewrite this PR so that it contains only one commit and only changes related to `CatalogWidget`?

nuzil:
Hi @orlangur 
Changes are applied and pushed

orlangur:
Thanks @nuzil, changes look good to me now.

By "could you please rewrite this PR" I meant squashing into a single commit so that history is cleaner. Please do it and perform a force push into the same branch.

nuzil:
Sorry we have a mess already in our own repository, I made a new PR
https://github.com/magento/magento2/pull/14559

This one can closed then

sidolov:
Closed according to comment above

PR #14057, opened 2018-03-12
Type cast Ids from `String` to `Integer`

yogeshsuhagiya:
The methods `getGroupId` and `getTaxClassId` should set and return the **Integer** value of Id but they are returning **String** values.
And the method `getSharedWebsiteIds` should return integer type array of website Ids but it's returning the string type array of website Ids.
So before setting the value of Ids, I've typecast the value of Id from String to Integer.

1. magento/magento2#14055: Getting group id from customer session returns wrong data type.

### Manual testing scenarios
N/A

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Model/Customer.php
+++ b/app/code/Magento/Customer/Model/Customer.php
@@ -886,7 +886,7 @@ class Customer extends \Magento\Framework\Model\AbstractModel
                 ScopeInterface::SCOPE_STORE,
                 $storeId
             );
-            $this->setData('group_id', $groupId);
+            $this->setData('group_id', (int)$groupId);
         }
         return $this->getData('group_id');
     }
@@ -900,7 +900,7 @@ class Customer extends \Magento\Framework\Model\AbstractModel
     {
         if (!$this->getData('tax_class_id')) {
             $groupTaxClassId = $this->_groupRepository->getById($this->getGroupId())->getTaxClassId();
-            $this->setData('tax_class_id', $groupTaxClassId);
+            $this->setData('tax_class_id', (int)$groupTaxClassId);
         }
         return $this->getData('tax_class_id');
     }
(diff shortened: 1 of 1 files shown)


app/code/Magento/Customer/Model/Customer.php
```diff
@@ -886,7 +886,7 @@ public function getGroupId()
                 ScopeInterface::SCOPE_STORE,
                 $storeId
             );
-            $this->setData('group_id', $groupId);
+            $this->setData('group_id', (int)$groupId);
         }
         return $this->getData('group_id');
```
ihor-sviziev, line 891: Why haven't you changed added casting from string to int here?

yogeshsuhagiya: I'm not getting you, will you please elaborate?

bydrei: @Yogeshks he is talking about the return. The return should be type casted.

yogeshsuhagiya: Hi @ferrazzuk
Thanks for the correction, I'll update my commit accordingly.

ferrazzuk: @Yogeshks no worries do that for the other files you also changed. Thank you for contributing

lbajsarowicz:
I don't like the idea of change ID type to `int` - I still believe that other DBMS drivers (abstraction) are going to be developed for Magento sooner or later, then UUID can be required instead of Auto-Increment `int` value.

IMHO: We should keep `getId()` method for both *Interface* (you forgot) and *Implementation class* return type `string`.

PS: You have to keep in mind that change to *interface* can affect thousands of Magento 2 instances after update.

bydrei:
@lbajsarowicz I disagree with you for these two reasons:
1. If you have a look at the database (even thought the changes are not being made to the database) the type of group_id is of INT type. Same goes to the group tax id class and website id.
2. Keeping it as a string doesn't keep the consistency of data types and also doesn't  match the function comments and even the comments in the interface.

Its makes sense to keep data type's consistency in this case.

yogeshsuhagiya:
Hi @ferrazzuk, 
Are my changes eligible for the merge with `magento/magento2` branch? Will you please let me know.

ihor-sviziev:
Hi,
I don't mind to add those changes. Cc @orlangur 
Any thoughts?

bydrei:
@Yogeshks I don't take those decision. I'm just a contributor like you

app/code/Magento/Customer/Model/Customer.php
bydrei: @Yogeshks can you move the type casting to the return instead?

orlangur:
Closing PR as mentioned issue is not valid.

> This issue is not valid, see https://github.com/magento/magento2/pull/10519#issuecomment-322029527 for more details.

Thanks for collaboration @ferrazzuk and @Yogeshks!

PR #14060, opened 2018-03-12
Disable add to cart button when redirect to cart enabled

ihor-sviziev:
### Description
Disable add to cart button when redirect to cart enabled, it includes product validation.

N/A

### Pre-requisites
1. Install Magento 2.2.3
2. Go to Admin >> Stores >> Configuration >> Sales >> Checkout
3. In "Shopping Cart" section change "After Adding a Product Redirect to Shopping Cart" to "Yes"

### Manual testing scenarios
1. Go to product page on frontend with very slow internet connection (could be emulated in chrome)
2. Click "Add to Cart" button

### Expected result
1. After clicking "Add to cart" button should became to disabled
2. Button color should be changed to "disabled" (grey)
3. Text on "add to cart" is changed to "Adding...", then to "Added"

### Actual result
1. After clicking "Add to cart" button is not became to disabled
2. Button color is not changed to "disabled" (grey)
3. Text on "add to cart" is not changed to "Adding...", then to "Added"

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/templates/product/view/addtocart.phtml
+++ b/app/code/Magento/Catalog/view/frontend/templates/product/view/addtocart.phtml
@@ -40,17 +40,6 @@
     </div>
 </div>
 <?php endif; ?>
-<?php if ($block->isRedirectToCartEnabled()) : ?>
-<script type="text/x-magento-init">
-    {
-        "#product_addtocart_form": {
-            "Magento_Catalog/product/view/validation": {
-                "radioCheckboxClosest": ".nested"
-            }
-        }
-    }
-</script>
-<?php else : ?>
 <script type="text/x-magento-init">
     {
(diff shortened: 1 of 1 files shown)


rogyar:
Hi @ihor-sviziev. The solution works in general but does not work in some particular cases. For example, if we have a simple product with the custom date option (required). Maybe there are some other cases as well.

ihor-sviziev:
Hi @rogyar,
It looks as separate issue. Date custom option adds new validation (looks like overrides) to the same form. Probably when redirect to shopping cart is disabled - we will have this issue too.

https://github.com/magento/magento2/blob/9c14af7a97bcf338662de365d2613c7ca9dbf8c6/app/code/Magento/Catalog/view/frontend/templates/product/view/options/type/date.phtml#L46-L54

PR #14079, opened 2018-03-13
Popout `count` function from `for` loop

yogeshsuhagiya:
Count the size of array outside for loop and passed the result to for loop.

N/A
### Manual testing scenarios
N/A

### Contribution checklist

### DIFF
--- a/setup/src/Magento/Setup/Module/I18n/Parser/Adapter/Html.php
+++ b/setup/src/Magento/Setup/Module/I18n/Parser/Adapter/Html.php
@@ -31,7 +31,8 @@ class Html extends AbstractAdapter
 
         $results = [];
         preg_match_all(Filter::CONSTRUCTION_PATTERN, $data, $results, PREG_SET_ORDER);
-        for ($i = 0; $i < count($results); $i++) {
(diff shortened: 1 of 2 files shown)


miguelbalparda:
Hi @Yogeshks! I checked the code and it seems to be a minor fix to improve readability. Is this correct or I'm missing something else?

orlangur:
Hi @Yogeshks, according to https://stackoverflow.com/questions/4566314/php-what-is-the-complexity-i-e-o1-on-of-the-function-count such optimization seems to be irrelevant in PHP world.

yogeshsuhagiya:
Hi @miguelbalparda,
You're right, it just a minor fix.

yogeshsuhagiya:
Hi @orlangur

I've reviewed your referral link and I agree with you but isn't it best practice to use function outside loops?

orlangur:
@Yogeshks,

>  isn't it best practice to use function outside loops?

No, it is only relevant in case we have some heavy operation, like `O(N)` counting of array elements, some invariant `strpos` etc.

Thanks for quick response and collaboration!

PR #14085, opened 2018-03-13
The autoloader throws an exception on class_exists.

Nyholm:
### Description
The autoloader throws an exception on `class_exists`. That is clearly an unexpected behavior according to the documentation: http://php.net/manual/en/function.class-exists.php

This will cause issues when including third party libraries that is not written for Magento. 

1. https://github.com/magento/magento2/issues/14074: Magento code generation bug, resulting in mails not sending!

### Manual testing scenarios
Run `class_exists('FooFactory')`. It should return `false` but an error is thrown instead. 

### Contribution checklist

-----------

FYI @NickvdMeij

### DIFF
--- a/lib/internal/Magento/Framework/Code/Test/Unit/GeneratorTest.php
+++ b/lib/internal/Magento/Framework/Code/Test/Unit/GeneratorTest.php
@@ -153,6 +153,20 @@ class GeneratorTest extends \PHPUnit\Framework\TestCase
     }
 
     /**
+     * We have to make sure `class_exists` works for third-party classes
+     */
+    public function testGenerateClassWithNonMagentoCode()
+    {
+        $objectManagerMock = $this->createMock(\Magento\Framework\ObjectManagerInterface::class);
+        $entityGeneratorMock = $this->getMockBuilder(\Magento\Framework\Code\Generator\EntityAbstract::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $objectManagerMock->expects($this->once())->method('create')->willReturn($entityGeneratorMock);
+        $this->model->setObjectManager($objectManagerMock);
+        $this->assertFalse(Generator::GENERATION_SUCCESS === $this->model->generateClass('\FooFactory'));
+    }
+
+    /**
      * Build SUT object
      *
      * @param Io $ioObject


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14085) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=14085) before we can accept your contribution.<br/>

orlangur:
We cannot process PR with broken tests. Prior to a full code review please make sure all Travis CI builds are green.

Nyholm:
This PR is not meant to fix anything. It is meant to **show** that the code has a bug. 

Someone with greater knowledge of Magento internals should use this PR/commit/test and add a fix to the problem.

orlangur:
@Nyholm thank you for your endeavours but factories autogeneration is a [crucial](http://devdocs.magento.com/guides/v2.0/extension-dev-guide/factories.html#writing-factories) Magento feature and it cannot be removed just because it causes troubles in some random library.

Instead of hacking Magento core recommended approach in such situation is to implement compatibility layer.

Nyholm:
One do not have to *remove* code just because they it has a bug in it. One could simply try to *improve* in instead. 

Thank you for taking the time to read and answer my posts.

orlangur:
@Nyholm, from another thread

> I'm open for any improvements into current Magento behavior. How you think autoloder can determine whether factory should be generated when `FooFactory` is requested? Some flags to determine whether code is Magento-style or not does not seems to be a viable option.

One suggestion was

> I would suggest not to create classes when they are being autoloaded. The creation of classes should happen when the project is built (ie when the cache is warming up). That would also make it easier to deploy Magento on read-only filesystems.

having a huge Magento expertise I simply don't see what can be changed to keep both scenarios workable.

I don't ask you for a code, if you ever come up with a conceptual idea just share it. For the reference, here is the current implementation: https://github.com/magento/magento2/blob/2.2-develop/lib/internal/Magento/Framework/Code/Generator.php#L91

Vinai:
An autoloader that throws an exception is not [PSR-4](https://www.php-fig.org/psr/psr-4/) compliant:

> 4. Autoloader implementations MUST NOT throw exceptions, MUST NOT raise errors of any level, and SHOULD NOT return a value.

orlangur:
Thanks @Vinai for pointing that out :)

@Vinai @kandy what do you propose to do with this PR? It does not include any solution, to me it makes more sense to reopen a corresponding issue or report a new one describing what is not so good in current core implementation.

sidolov:
Hi @Nyholm , @orlangur , I'm closing this PR because it does not contain the solution for the described issue. The new issue was created with all needed information https://github.com/magento/magento2/issues/20773. Feel free to add additional info if I missed something. 
Thank you!

PR #14088, opened 2018-03-13
Dispatch events in AccountManagement::checkPasswordStrength()

tdgroot:
Dispatch events in AccountManagement::checkPasswordStrength() to improve extensibility.

### Manual testing scenarios

1. Create an observer that hooks in on `customer_check_password_strength_before` or `customer_check_password_strength_after`.
2. Create a customer, the observer should be called and the password provided as well.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Model/PasswordStrengthChecker.php
+++ b/app/code/Magento/Customer/Model/PasswordStrengthChecker.php
@@ -104,7 +104,7 @@ class PasswordStrengthChecker implements PasswordStrengthCheckerInterface
      *
      * @return int
      */
-    protected function getMinPasswordLength(): int
+    private function getMinPasswordLength(): int
     {
         return (int)$this->scopeConfig->getValue(self::XML_PATH_MINIMUM_PASSWORD_LENGTH);
     }
--- a/app/code/Magento/Customer/Model/PasswordStrengthChecker.php
+++ b/app/code/Magento/Customer/Model/PasswordStrengthChecker.php
@@ -1,0 +1,143 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Customer\Model;
+
+use Magento\Customer\Model\PasswordStrengthChecker\ResultInterface;
+use Magento\Customer\Model\PasswordStrengthChecker\ResultInterfaceFactory;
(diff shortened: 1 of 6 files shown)


orlangur:
Hi @tdgroot, it is not recommended to add any new events in general, plugins-interceptors should be implemented whenever possible instead of event observers.

If current code does not allow to write plugin-interceptor in some case, it should be refactored so that intercepting is possible.

Please provide real-world customization scenario you trying to cover with these new events.

tdgroot:
@orlangur I considered making the method public, but that would mean it actually has to be added to the interface. 

The way I see it, there are 3 options:
- Make the method `checkPasswordStrength()` public.
- Create and additionally call a new (public) method to allow for extension possibilities.
- Create a password validator class to keep the password validation logic in that single class. This also allows for creation of new public methods and new extension possibilities.  

What's your suggestion in this case? 

p.s. as for the real-world customization scenario, I have written a pwned password validator for Magento 2. Unfortunately I had to rewrite the AccountManagement class to create an extension point to add my validation. The module: [magento2-module-pwned-validator](https://github.com/tdgroot/magento2-module-pwned-validator)

orlangur:
@tdgroot oh, nice, I didn't expect to see an existing implementation 👍 

To me it seems the best to move `checkPasswordStrength` method logic into some new `PasswordStrengthChecker` class.

tdgroot:
@orlangur I created a little implementation of the logic inside classes. Please let me know if this is going to work for Magento. If it is, I can start writing tests for the logic.

app/code/Magento/Customer/Model/PasswordStrengthChecker.php
```diff
@@ -0,0 +1,143 @@
...
+
+        return $result;
+    }
+
+    /**
+     * Retrieve minimum password length
+     *
+     * @return int
+     */
+    protected function getMinPasswordLength(): int
```
Jeroenll: Can we mark this method private or remove entirely since it's only used in `check()`?

tdgroot: Spot on! I'll make it private.

tdgroot:
@orlangur what do you think?

orlangur:
@tdgroot will review shortly, no need to fix tests yet - let me check implementation first.

orlangur (changes requested):

orlangur:
Thanks for collaboration!

Issue #10540, opened 2017-08-15
Unable to save multi-select attributes with grouped/configurable products


apsummers:
This issue appears to be similar to #4822, but is more like what @daniel-ifrim describes in the last comment. When updating product attributes on a Grouped Product, attributes using a multi-select field do not get saved correctly, and the values are ignored.

Simply saving the product does not trigger `ArrayBackend::beforeSave` at all. "Save and Duplicate" will trigger `ArrayBackend::beforeSave`, though the data returned by `$object->getData($attributeCode);` only contains the current selection. `$object`, which is passed to `ArrayBackend::beforeSave` _does_ show the correct selections.

Multi-select fields do save correctly on Simple Products, and the other field types (dropdown, text) seem to save correctly on Grouped Products.

### Preconditions

1. Magento 2.1.7
2. PHP 7.0.15
3. MariaDB 10.1

### Steps to reproduce

1. Create a Grouped Product with a multi-select attribute
2. Select one or more new values for that attribute
3. Save

### Expected result

1. The attribute is updated

### Actual result

1. The original attribute value is still there

apsummers:
I've been looking into this more, and I've got some more info to hopefully help narrow this down:

* The issue doesn't exist when the Grouped Product is composed of two Simple Products
* The issue _does_ exist when the Grouped Product contains multiple Configurable Products
* The Configurable Products in the Grouped Product that does exhibit this behavior are using different attribute sets

magento-engcom-team:
@apsummers  Thank you for report. We were not able to reproduce this issue by following the steps you provided. If you'd like to update it, please reopen the issue.

A-1A:
I Found Solution after R&D On Magento database.
Login In your phpmyadmin
Go to your website database 
Open "eav_attribute" table
Find Your multiselect "attribute_id"
Edit Row
If "backend_model" Column blank then paste this **Magento\Eav\Model\Entity\Attribute\Backend\ArrayBackend**
And Also Please check this
https://magento.stackexchange.com/questions/224621/warning-array-filter-expects-parameter-1-to-be-array/230444#230444

PR #14125, opened 2018-03-15
Fix mutliselect attributes save for bundle products (and other cases)

cherednichenkoa:
Fix multi select attributes saving, for product with bundle options
https://github.com/magento/magento2/issues/10540

### Description

To make long story short - 
 product model instance Magento\Catalog\Model\Product, has shared instance inside Magento\Catalog\Model\ResourceModel\Product - if we save a bundle product (possible another cases) thru admin panel, and load simple product childs inside with different attribute set - we will have an issues with saving of multiselect attributes that do not not exist in child product attribute set 
because of attribute list Magento\Catalog\Model\ResourceModel\Product _attributesByCode here will be from last loaded product 

1. magento/magento2#10540
2. magento/magento2#13520

### Preconditions
Magento 2.2.3 EE
PHP 7.0.15
MariaDB 10.1

### Steps to reproduce
Create a Bundle Product with a multi-select attribute (multiselect attribute in bundle product attribute set, that does not exist in child product attribute set)
Select one or more new values for that attribute
Add to the bundle childs,  that have different attribute set from parent 
Save
### Expected result
The multiselect attribute is updated
### Actual result
The original attribute multiselect value is still there

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/etc/di.xml
+++ b/app/code/Magento/Catalog/etc/di.xml
@@ -1073,4 +1073,9 @@
             <argument name="productRepository" xsi:type="object">Magento\Catalog\Api\ProductRepositoryInterface\Proxy</argument>
         </arguments>
     </type>
+    <type name="Magento\Catalog\Model\Product">
+        <arguments>
+            <argument name="resource" xsi:type="object" shared="false">Magento\Catalog\Model\ResourceModel\Product</argument>
+        </arguments>
+    </type>
 </config>


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14125) <br/>All committers have signed the CLA.

app/code/Magento/Catalog/etc/di.xml
```diff
@@ -1073,4 +1073,9 @@
             <argument name="productRepository" xsi:type="object">Magento\Catalog\Api\ProductRepositoryInterface\Proxy</argument>
         </arguments>
     </type>
+    <type name="Magento\Catalog\Model\Product">
+        <arguments>
+            <argument name="resource" xsi:type="object" shared="false">Magento\Catalog\Model\ResourceModel\Product</argument>
```
orlangur, line 1078: As product model is quite popular in Magento code such change leads to creating a lot of additional resource models when they are not really necessary.

Please rework as a more isolated fix, most likely as only multiselect attribute type for bundle options is affected it will be there, not in generic product model.

You actually managed to reproduce
> Fixed Issues (if relevant)
magento/magento2#10540
magento/magento2#13520

correct?

orlangur (changes requested):

cherednichenkoa:
@orlangur , makes sense, however same issue can appear with Grouped Products, Products links (cross sells, related etc) as well . Not sure that memory savings and initialization time(construct is not so heavy) is a good trade off for such case  
Right now Product ResourceModel is working like a container for pre-loaded attribute models . And those attributes can be shared between different products.

I have checked Resource __construct - it looks like it is not so heavy right up to Magento\Catalog\Model\ResourceModel\AbstractResource

Regarding  your proposition - I will double check additional "isolated solution".

app/code/Magento/Catalog/etc/di.xml
cherednichenkoa: Correct

PR #14190, opened 2018-03-20
[Forwardport] Modify Report processor to return 500

### LINKED ISSUE #11512: Incorrect use of 503 status code

AlexWorking:
### Original Pull Request 
 https://github.com/magento/magento2/pull/11513
The report processor is currently returning a HTTP 503 status
code; generally used for temporarily failures to connect to a service
such as when that service is in maintenance mode, when an upstream
proxy is unavailable etc.

This commit modifies the report HTTP code to be a 500. The author
believes this to be a better reflection that the error is miscellaneous
in nature, and that action is required in order to change it (i.e.
it is not a temporary condition)

Fixes #11512

### DIFF
--- a/pub/errors/processor.php
+++ b/pub/errors/processor.php
@@ -215,7 +215,7 @@ class Processor
     public function processReport()
     {
         $this->pageTitle = 'There has been an error processing your request';
-        $this->_response->setHttpResponseCode(503);
+        $this->_response->setHttpResponseCode(500);
 
         $this->showErrorMsg = false;
         $this->showSentMsg  = false;


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14190) <br/>All committers have signed the CLA.

magento-engcom-team:
@AlexWorking thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

orlangur (approved):

magento-engcom-team:
[ENGCOM-976](https://jira.corp.magento.com/browse/ENGCOM-976) has been created to process this Pull Request

PR #14265, opened 2018-03-22
Fix duplicate URL key error output

peterjaap:
### Description
Message was;
`Url key: '%s' was already generated for an item with the SKU: '%s'. You need to specify the unique URL key manually in row(s): 1124`

Fixed message;
`3: Url key: 'venotrain-impuls' was already generated for an item with the SKU: 'Venotrain Impuls'. You need to specify the unique URL key manually in row(s): 1124`

### DIFF
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product.php
@@ -2678,8 +2678,7 @@ class Product extends \Magento\ImportExport\Model\Import\Entity\AbstractEntity
 
             $source->next();
         }
-        $this->
-            ();
+        $this->checkUrlKeyDuplicates();
         $this->getOptionEntity()->validateAmbiguousData();
         return parent::_saveValidatedBunches();
     }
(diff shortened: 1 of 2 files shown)


dmanners:
Hey @peterjaap thanks for this PR. Good catch.

Could you update the integration test to match this change also. Think it is failing as it expects the older message https://travis-ci.org/magento/magento2/jobs/356903288#L1010.

Feel free to reach out to me if you need any help.

Thanks man.

peterjaap:
@dmanners the error says `Failed asserting that 1 matches expected 2.`. So it seems it still receives one error in the 'old' style and one in the 'new' style ?

Also, I'm not sure where to inject the SKU and URL key to have it run the test successfully since it seems to work with an error code instead of the actual error?

dmanners:
Let me see what I can find out for you @peterjaap

dmanners:
Ok @peterjaap so after some step debugging with xdebug I am a little confused. During the old code the system only gets into the line https://github.com/magento/magento2/pull/14265/files#diff-01295aded280e8fb45bea4001c5247abL2709 once in the test case. This is the same as the new code but for some reason another error message is being added. In this case I think it would be fine to update the `validateUrlKeysDataProvider` method to only check for 1 error message in this case is fine.

peterjaap:
Updated the test.

ishakhsuvarov:
@peterjaap @dmanners Please create branch for the PR in your own fork. We can not accept submissions pushed directly to `magento/magento2`

okorshenko:
Please create the branch in your personal fork

PR #14273, opened 2018-03-22
Remove jQuery UI from form-mini.js

jonathanKingston:
### Description
Removal of jQuery UI from the form search as it's a dependency that isn't needed.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Search/view/frontend/web/form-mini.js
+++ b/app/code/Magento/Search/view/frontend/web/form-mini.js
@@ -9,11 +9,11 @@
 define([
     'jquery',
     'underscore',
+    'mage/getKeyNameFromEvent',
     'mage/template',
     'matchMedia',
-    'jquery/ui',
     'mage/translate'
-], function ($, _, mageTemplate, mediaCheck) {
+], function ($, _, getKeyNameFromEvent, mageTemplate, mediaCheck) {
     'use strict';
 
     /**
@@ -202,29 +202,27 @@ define([
          * @return {Boolean} Default return type for any unhandled keys
          */
         _onKeyDown: function (e) {
-            var keyCode = e.keyCode || e.which;
-
-            switch (keyCode) {
-                case $.ui.keyCode.HOME:
+            switch (getKeyNameFromEvent(e)) {
+                case "Home":
                     this._getFirstVisibleElement().addClass(this.options.selectClass);
                     this.responseList.selected = this._getFirstVisibleElement();
                     break;
 
-                case $.ui.keyCode.END:
+                case "End":
                     this._getLastElement().addClass(this.options.selectClass);
                     this.responseList.selected = this._getLastElement();
                     break;
 
-                case $.ui.keyCode.ESCAPE:
+                case "Escape":
                     this._resetResponseList(true);
                     this.autoComplete.hide();
                     break;
 
-                case $.ui.keyCode.ENTER:
+                case "Enter":
                     this.searchForm.trigger('submit');
                     break;
 
-                case $.ui.keyCode.DOWN:
+                case "ArrowDown":
                     if (this.responseList.indexList) {
                         if (!this.responseList.selected) {  //eslint-disable-line max-depth
                             this._getFirstVisibleElement().addClass(this.options.selectClass);
@@ -242,7 +240,7 @@ define([
                     }
                     break;
 
-                case $.ui.keyCode.UP:
+                case "ArrowUp":
                     if (this.responseList.indexList !== null) {
                         if (!this._getFirstVisibleElement().hasClass(this.options.selectClass)) {
                             this.responseList.selected = this.responseList.selected
--- a/lib/web/mage/getKeyNameFromEvent.js
+++ b/lib/web/mage/getKeyNameFromEvent.js
@@ -1,0 +1,35 @@
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+define([
+], function () {
+
+    /**
+     * @param {Event} e The key event
+     */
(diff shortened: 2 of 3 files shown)


ihor-sviziev:
Hi @jonathanKingston,

Could you clarify what is benefit of getting rid of jquery ui specially in this place? AFAIK jquery UI is using in a lot of places, so removing it just from one place - do not changing anything. Not sure if it's reasonable to accept such pull request. Actually previous implementation was much better because there were using constants

jonathanKingston:
@ihor-sviziev the idea is to burn down the implementation areas of jQuery UI. Given the jQuery UI version is old and insecure. The jQuery UI version is also modified making it hard to upgrade.
As this is probably the most common usage, I started here so I could answer this question.

> Actually previous implementation was much better because there were using constants

Seems fair, I mean it might make sense to have a lightweight jQuery plugin which just handles keydown events etc, however I didn't spot many other areas that would use it.

jonathanKingston:
Minus TinyMCE and some other libraries it looks like it could be removed in another PR in the front-end. I could expand this commit to do that? The only thing I am uncertain of is the use of `mage/calendar`

VladimirZaets (commented):
Currently, we can't use `event.key` without the fallback to the old approach because we supporting IE11 and Edge browsers and there aren't fully support `event.key`. 
https://caniuse.com/#feat=keyboardevent-key
Also, if you want you can read more about it https://medium.com/@uistephen/keyboardevent-key-for-cross-browser-key-press-check-61dbad0a067a

jonathanKingston:
> Also, if you want you can read more about it https://medium.com/@uistephen/keyboardevent-key-for-cross-browser-key-press-check-61dbad0a067a

Damn sounds like MDN docs are wrong on the support for this sorry.

Anyway I still think this would be easier to copy out the jQuery UI behaviour for keypresses into a smaller lib as the only other potential jQuery UI on the front end I have found is the calendar meaning in a follow up we could remove it as a FE dependency. Does that sound like an ok solution @VladimirZaets?

orlangur:
Hi @jonathanKingston,

> the idea is to burn down the implementation areas of jQuery UI

While me personally would love to get rid of jQuery/jQuery UI completely (in favor of Angular, but it's not gonna happen :D), how is it even possible considering http://devdocs.magento.com/guides/v2.2/javascript-dev-guide/widgets/jquery-widgets-about.html?

So, to reduce jQuery UI usages some common strategy should be agreed first which then should be concretized in particular action points.

> Given the jQuery UI version is old and insecure

Are there any attack vectors you know of actually affecting Magento?

> The jQuery UI version is also modified making it hard to upgrade

Having backward compatibility in mind, this seems the only viable approach though.

jonathanKingston:
> Are there any attack vectors you know of actually affecting Magento?

Not at the moment, however having very old versions is never good. I wasn't raising this as a vuln or anything, just that there might be attack vectors. I also haven't tried to look into the exploits in UI and match them to Magento.

> Having backward compatibility in mind, this seems the only viable approach though.

I think at this point if the front end can have the code removed then this will help update or removal. I'm also not 100% on how you handle backwards compatibility either.

> how is it even possible considering http://devdocs.magento.com/guides/v2.2/javascript-dev-guide/widgets/jquery-widgets-about.html?

Only the following appear to use UI though:
- https://github.com/magento/magento2/blob/2.0/lib/web/mage/calendar.js
- https://github.com/magento/magento2/blob/2.0/lib/web/mage/collapsible.js
- https://github.com/magento/magento2/blob/2.0/lib/web/mage/dropdown.js
- https://github.com/magento/magento2/blob/2.0/app/code/Magento/Search/view/frontend/web/form-mini.js
- https://github.com/magento/magento2/blob/2.0/lib/web/mage/tabs.js

Also "Magento out of the box does not contain jQuery UI styles", so it really does seem like these small number of references that could be burned down in a few PRs.

> So, to reduce jQuery UI usages some common strategy should be agreed first which then should be concretized in particular action points.

Yeah we have @dmanners at the office today, he said we should have a conversation about architectural changes with the relevant parties :)

orlangur:
> Only the following appear to use UI though:

My main concern is http://devdocs.magento.com/guides/v2.0/coding-standards/code-standard-jquery-widgets.html, there are almost 300 occurrences of `$.widget` in Magento Open Source. But probably something like "it might make sense to have a lightweight jQuery plugin" is possible for widgets too.

> we should have a conversation about architectural changes with the relevant parties :)

Great!

jonathanKingston:
> there are almost 300 occurrences of $.widget in Magento Open Source. But probably something like "it might make sense to have a lightweight jQuery plugin" is possible for widgets too.

Yeah I think we can migrate that to something simpler too.

@orlangur I updated the PR to include a function from an external file, I moved another usage in a different file also.

orlangur:
@VladimirZaets please take a look on this one.

VladimirZaets:
@jonathanKingston 

> Anyway I still think this would be easier to copy out the jQuery UI behaviour for keypresses into a smaller lib as the only other potential jQuery UI on the front end I have found is the calendar meaning in a follow up we could remove it as a FE dependency. Does that sound like an ok solution @VladimirZaets?

It's make sense. But we already have small [library](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Ui/view/base/web/js/lib/key-codes.js) that  transform keycodes to understandable type.
 
So, we can use [this](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Ui/view/base/web/js/lib/key-codes.js) library in current case, but we shouldnt create new one library with the similiar behaviour

jonathanKingston:
I'm not working on this any longer, I'm just going to close my PRs. Feel free to adopt the code after a fix-up etc.

PR #14275, opened 2018-03-22
Removal of legacy TinyMCE3 from core.

jonathanKingston:
### Description
Removal of TinyMCE3 as it appears not to be used from the core any more. The existence of this code makes it hard to audit especially with it's own library versions that are also in core.

### Fixed Issues
1. Removal of legacy code

### Manual testing scenarios
1. Ensure that TinyMCE works as expected

### Contribution checklist


orlangur:
Hi @jonathanKingston, this module is not legacy, it is not even released yet, please check module description: https://github.com/magento/magento2/tree/2.3-develop/app/code/Magento/Tinymce3

jonathanKingston:
@orlangur would we be able to move this to an external dependency or perhaps a path that makes it clear that it's backwards compat.

Being in `app` or `lib` by default makes auditing the code harder as there is so much of it. If users could install an external dependency to get this behaviour instead that would be better.

PR #14311, opened 2018-03-24
Fix type casting for stock item qty increment

aleron75:
This PR is intended to fix the `getQtyIncrements()` method that didn't respect implemented interface according to which a float or a boolean value has to be returned.

Wrong typecasting to `int` doesn't allow to use decimal stock qty increments.

### DIFF
--- a/app/code/Magento/CatalogInventory/Model/Stock/Item.php
+++ b/app/code/Magento/CatalogInventory/Model/Stock/Item.php
@@ -392,7 +392,7 @@ class Item extends AbstractExtensibleModel implements StockItemInterface
     /**
      * Retrieve Quantity Increments
      *
-     * @return int|false
+     * @return float|false
      */
     public function getQtyIncrements()
     {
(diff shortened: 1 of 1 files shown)


app/code/Magento/CatalogInventory/Model/Stock/Item.php
```diff
@@ -401,7 +401,7 @@ public function getQtyIncrements()
                 if ($this->getUseConfigQtyIncrements()) {
                     $this->qtyIncrements = $this->stockConfiguration->getQtyIncrements($this->getStoreId());
                 } else {
-                    $this->qtyIncrements = (int) $this->getData(static::QTY_INCREMENTS);
+                    $this->qtyIncrements = (float) $this->getData(static::QTY_INCREMENTS);
```
orlangur, line 404: Are fractional Qty increments actually supported?

aleron75: >Are fractional Qty increments actually supported?

uhm, at configuration level I see usage of float value: 

https://github.com/magento/magento2/blob/068a97e8ae1615f580e492829bea688496ea4487/app/code/Magento/CatalogInventory/Model/Configuration.php#L278

but you are right, they don't seem to be allowed; the problem is that in the scope of MSI project we are considering them as float, maybe I should discuss with MSI team a bit more about this

okorshenko:
Please create the branch in your personal fork

PR #14324, opened 2018-03-24
Fixed returning null on _getEavWebsiteTable results in describe table error

MikeParkin:
Returning null within _getEavWebsiteTable is intended to mean that the functionality is not required. Unfortunately there are a few places where SQL is generated using the eav website table and does not take into account that it can be null.

### Description

Wrapping the existing queries with conditional checks to see if the table name is null or not.  

1. magento/magento2#6043: Returning null on _getEavWebsiteTable results in describe table error.

### Manual testing scenarios

1. Create a new eav entity without attribute_website table including attribute models (or to make it easier and quicker just return null on the customer one).
2. Retrieve the collection of attributes from your custom entity

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/ResourceModel/Attribute.php
+++ b/app/code/Magento/Eav/Model/ResourceModel/Attribute.php
@@ -62,7 +62,7 @@ abstract class Attribute extends \Magento\Eav\Model\ResourceModel\Entity\Attribu
     {
         $select = parent::_getLoadSelect($field, $value, $object);
         $websiteId = (int)$object->getWebsite()->getId();
-        if ($websiteId) {
+        if ($websiteId && $this->_getEavWebsiteTable()) {
             $connection = $this->getConnection();
             $columns = [];
             $scopeTable = $this->_getEavWebsiteTable();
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14324) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=14324) before we can accept your contribution.<br/>

rogyar:
Hello @MikeParkin. Thank you for your contribution. Please, consider fixing the failing static tests. To reduce the cyclomatic complexity, first of all, you need to reduce the amount of the nested if -> else constructions.

MikeParkin:
@rogyar No problem, will do that later tonight. Regarding cyclomatic complexity, I have changed very little in the code (just added some necessary IF commands). I believe the coding style is single return, rather than early exit? If so, are you wanting the code simplifying (seems bad to do that as part of a small bug fix).

rogyar:
Yep, there are only a few IF statements added. However, every new IF statement that causes deeper conditions nesting increases the cyclomatic complexity. According to the Magento standards, it should be less than 10 within a scope of one method. So, the simplest way is to use the return statement in case if the condition is not met instead of wrapping all logic into another condition, i.e.

```
// app/code/Magento/Eav/Model/ResourceModel/Attribute/Collection.php

if (!$this->_getEavWebsiteTable()) {
    return $this;
}
```
or so.

The more complex way is to move the logic into a separate private method

rogyar:
Hello @MikeParkin. Are you going to proceed with fixing the cyclomatic complexity issue?

rogyar:
Closing this PR due to inactivity. Please, feel free to reopen it if you decide to proceed with your work. Thank you.

Issue #4136, opened 2016-04-13
Widget condition with unexpected character not preventing from saving


devhn:
## Preconditions
1. Magento 2.1

## Steps to reproduce
1. Install Magento from `community` branch.
2. Create a new widget via admin (Content -->Widgets)
3. Widget type: Catalog Products List
4. Layout update: Specific category
5. Widget options (conditions): Set product as new from date. Entered value: {TODAY}
## Expected result
1. It should display an error about unexpected character and not allow to save the widget
## Actual result
1. Upon saving the widget it display and error on frontend (attached)
2. When trying to edit the widget it displays an error on backend (attached)

This happens when using {TODAY} as a time value and not actual formatted date.
Is the {TODAY} parameter obsolete? And will there be an option to delete the widget from the administration in the future as currently there is no selection options with actions within the Widgets section?

[errors.zip](https://github.com/magento/magento2/files/216853/errors.zip)

piotrekkaminski:
internal issue MAGETWO-53440

magento-engcom-team:
@devhn, thank you for your report.
We've created internal ticket(s) MAGETWO-53440 to track progress on the issue.

vasilii-b:
I'm gonna have a look on this.

magento-engcom-team:
Hi @devhn. Thank you for your report. 
The issue has been fixed in magento/magento2#14485 by @vasilii-b in 2.2-develop branch
Related commit(s):
 - [cc1e013ff822aacccf9c8bf9c6c320f5a9f1f606](https://github.com/magento/magento2/commit/cc1e013ff822aacccf9c8bf9c6c320f5a9f1f606)
 - [ff4b005c34c541951d31691515f1e5c33c75798e](https://github.com/magento/magento2/commit/ff4b005c34c541951d31691515f1e5c33c75798e)
 - [6c87bb36a666e17ba7f993b436cea1763699be02](https://github.com/magento/magento2/commit/6c87bb36a666e17ba7f993b436cea1763699be02)
 - [1d1d061ad4eb8c664d9abf9aaa84e7b620a8dd9a](https://github.com/magento/magento2/commit/1d1d061ad4eb8c664d9abf9aaa84e7b620a8dd9a)

The fix will be available with the upcoming 2.2.8 release.

vasilii-b:
@gelanivishal skip forward port for this one

magento-engcom-team:
Hi @devhn. Thank you for your report. 
[...]

PR #14330, opened 2018-03-24
[Forwardport] Fix for Issue #4136, MAGETWO-53440

vasilii-b:
### Description
Full description in #4136.
✔️ __Added calendar initialization for Conditional Rules when a rule is created for the 1st time__

For the condition type date the datepicker was not initialized when creating new conditions.

This PR aims to solve the issue and do not allow the user to write anything to the input.

Issue described in #4136

### Manual testing scenarios
Described in #4136

### Contribution checklist

### Original PR 
https://github.com/magento/magento2/pull/14485

### DIFF
--- a/app/code/Magento/Rule/view/adminhtml/web/rules.js
+++ b/app/code/Magento/Rule/view/adminhtml/web/rules.js
@@ -220,6 +220,8 @@ define([
 
             var elem = Element.down(elemContainer, 'input.input-text');
 
+            jQuery(elem).trigger('contentUpdated');
+
             if (elem) {
                 elem.focus();
 


orlangur (changes requested):
>  Added calendar initialization for Conditional Rules when a rule is created for the 1st time
> For the condition type date the datepicker was not initialized when creating new conditions.

Please change implementation in such a way that calendar is successfully initialized via `data-mage-init` (so that whenever you add it to DOM it is auto-wired).

vasilii-b:
@orlangur I'm not sure if that's possible. Tough you may suggest me where (what module, file) does the parsing of the `data-mage-init` so I can investigate how to solve this in a better way.
As far as I know the `data-mage-init` is processed once the DOM is loaded and does not listen to any other DOM events. Please correct me if I'm wrong.

Thank you!

vasilii-b:
@orlangur nevermind, I figured it out. Please check the latest commit. Thank you!

app/code/Magento/Rule/view/adminhtml/web/rules.js
```diff
@@ -220,6 +221,8 @@ define([
 
             var elem = Element.down(elemContainer, 'input.input-text');
 
+            jQuery(elem).trigger('contentUpdated');
```
orlangur, line 223: Yeah, exactly 👍 

Looks like calendar dependency is not needed anymore, isn't it? Please squash changes into a single commit also.

orlangur (changes requested):

app/code/Magento/Rule/view/adminhtml/web/rules.js
vasilii-b: Indeed. I've removed it. Thank you!

orlangur: Parameter too, please: `+], function (jQuery, calendar) {`

 If you need any assistance with squashing, just tell me.

vasilii-b:
@orlangur changes added

orlangur:
@vasilii-b great! Changes look good to me now, there is one thing I didn't mention before - this PR is targeting `2.3-develop` while we should deliver every contribution into `2.2-develop` first.

Could you please apply similar changes to `2.2-develop` in separate PR and I'll process them both instantly then.

vasilii-b:
@orlangur please see separated request https://github.com/magento/magento2/pull/14485
Thank you a lot for help!

vasilii-b:
Hi @orlangur, can you please update me why this is on `on hold` ?

sidolov:
Hi @vasilii-b , I'm closing this PR because original PR was closed due to inactivity. 

PR #14333, opened 2018-03-24
Registry deprecation

vkublytskyi:
`Magento\Framework\Registry` abused in Magento 2 and used to pass variables through the system. Such approach leads to an implicit temporal coupling which causes bugs and complicates system evolving. To explicitly communicate that Registry usage should be counted as bad practice this `Magento\Framework\Registry` will be deprecated since 2.3.0 release.

### DIFF
--- a/lib/internal/Magento/Framework/Registry.php
+++ b/lib/internal/Magento/Framework/Registry.php
@@ -8,7 +8,11 @@ namespace Magento\Framework;
 /**
  * Registry model. Used to manage values in registry
  *
+ * Registry usage as a shared service introduces temporal, hard to detect coupling into system.
+ * It's usage should be avoid. Use service classes or data providers instead.
+ *
  * @api
+ * @deprecated
  */
 class Registry
 {
@@ -24,6 +28,8 @@ class Registry
      *
      * @param string $key
      * @return mixed
+     *
+     * @deprecated
      */
     public function registry($key)
     {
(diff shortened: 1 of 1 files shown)


lib/internal/Magento/Framework/Registry.php
```diff
@@ -8,7 +8,11 @@
 /**
  * Registry model. Used to manage values in registry
  *
+ * Registry usage as a shared service introduces temporal, hard to detect coupling into system.
+ * It's usage should be avoid. Use service classes or data providers instead.
```
orlangur, line 12: Some good example of service class and data provider wouldn't hurt here I believe.

orlangur (changes requested):

lib/internal/Magento/Framework/Registry.php
vkublytskyi: Service classes are any class that has no internal state and return data. If phrasing is not clear I will be thankful if you suggest more correct.

If there are would be many questions how to avoid registry usage I will describe that in Magento DevBlog post. But in general, that should be pretty obvious for Magento developers.

orlangur: Just any implementation of service class and data provider which looks good in your eyes.

Ok, probably wider explanation makes sense in release notes where this will be actually deprecated and not in code comments.

orlangur (approved):

PR #14356, opened 2018-03-25
Use `LEFT JOIN` When Checking for `null` Attributes

maxbucknell:
### Description

It is not possible to filter an EAV collection by an attribute being `null`.

* magento/magento2#14355: Impossible to Filter Collection by `null` Attribute Value

Also addresses:

* magento/magento2#14312: GraphQL `ProductFilterInput` Requires Field to be a String

### Manual testing scenarios

As addressed by the integration test, load a product collection and try to `addAttributeToFilter` and select an attribute to be `null`. This is a correct usage of the API, and the `where` clause is generated correctly, as `... WHERE ((at_special_price.value IS NULL))`.

However, without this diff, the join is an `INNER JOIN`, which automatically excludes `null` values.

### Discussion

I have chosen to add a little complexity here, by checking for a null comparison and only then selecting a `LEFT JOIN`. It would work fine to use a `LEFT JOIN` for this all the time, but I suspect that `INNER JOIN` was chosen for a reason, perhaps performance. I decided it was best to respect this decision and change as little as possible.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Eav/Model/Entity/Collection/AbstractCollection.php
+++ b/app/code/Magento/Eav/Model/Entity/Collection/AbstractCollection.php
@@ -1498,6 +1498,10 @@ abstract class AbstractCollection extends AbstractDb implements SourceProviderIn
                 $condition
             );
         } else {
+            if (isset($condition['null'])) {
+                $joinType = 'left';
+            }
+
             $this->_addAttributeJoin($attribute, $joinType);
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14356) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @maxbucknell. Thank you for your contribution. 

PR #14371, opened 2018-03-26
comment error reporting on app/bootstrap.php

moamenmashael:
https://github.com/magento/magento2/issues/14370

error_reporting hardcoded to E_ALL

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -7,7 +7,7 @@
 /**
  * Environment initialization
  */
-error_reporting(E_ALL);
+#error_reporting(E_ALL);
 #ini_set('display_errors', 1);
 
 /* PHP version validation */


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14371) <br/>All committers have signed the CLA.

orlangur:
Hi @moamenmashael, please explain why commenting out display in next line

```
#ini_set('display_errors', 1);
```

is not enough.

moamenmashael:
This creates several issues, one of them comes from third-party payment extension, where API should return a clean result, but there was notice of undefined variable, leads to loss of sales for our client.

orlangur:
> The `error_reporting()` function won't be effective if your `display_errors` directive in `php.ini` is set to "Off", regardless of level reporting you set.

From one of the comments on php.net. Could you provide more details how a notice with `display_errors` turned off could break some flow?

xtremeperf:
For the sake of clarity...

error_reporting
This is used to set the run-time error reporting level.  It tells PHP which types (levels) of errors it can recognize and act upon.  Has nothing to do with outputting these errors to screen, log file, etc.

display_errors
This determines whether errors should be printed to the screen as part of the output or if they should be hidden from the user.

I recommend checking out PHP's documentation for error reporting here:
[http://php.net/manual/en/errorfunc.configuration.php#ini.error-reporting](http://php.net/manual/en/errorfunc.configuration.php#ini.error-reporting)

It appears to me that initializing error reporting level to E_ALL during application bootstrapping is completely appropriate in this case, and still allows customization at the system level.

orlangur:
@xtremeperf exactly 😉

orlangur:
Closing due to inactivity. Feel free to reach me out anytime later if you wish to continue work on this pull request, share your thoughts and it may be reopened.

PR #14386, opened 2018-03-27
Add checkPasswordStrength() to API

Jeroenll:
### Description
Add `checkPasswordStrength()` to API so we can extend it without making a preference. With that, I also deprecated `getMinPasswordLength()` since it isn't used anywhere else in the code.

An alternative is to move the strength check to a `PasswordStrengthValidator`. That way no update to the API is needed. Please let me know what is prefered.

1. None.

### Manual testing scenarios
1. None.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Api/AccountManagementInterface.php
+++ b/app/code/Magento/Customer/Api/AccountManagementInterface.php
@@ -153,6 +153,15 @@ interface AccountManagementInterface
     public function resetPassword($email, $resetToken, $newPassword);
 
     /**
+     * Make sure that password complies with minimum security requirements.
+     *
+     * @param string $password
+     * @return void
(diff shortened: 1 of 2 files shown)


orlangur:
Hi @JeroenVanLeusden,

> An alternative is to move the strength check to a PasswordStrengthValidator

We went this way in https://github.com/magento/magento2/pull/14088 already, I will consider marking `PasswordStrengthCheckerInterface` as `@api`. Would be nice if you look into this PR really quickly to check if it covers your case.

> Please let me know what is prefered.

Generally, it is not a good idea to bloat interfaces especially when there is some distinct responsibility which can be moved into separate class. Usually interface is extended only when something was overlooked when it was introduced.

Thanks for collaboration!

PR #14411, opened 2018-03-28
Returning relative paths for validation.xml files

DanCarlyon:
### Description
There is a bug in the `Magento\Framework\Validator` package which means that certain actions, for example saving a customer, will intermittently fail on multi-web server Magento environments which use different web roots per-web server but a single, shared cache instance. The bug is caused by Magento saving *absolute* paths to certain configuration files in the cache, as absolute paths will not be consistent between servers which have different web roots

With this change we now save the relative paths to Validation.xml configuration files. We recently faced an issue on a multi server setup where the paths to Magento were different.

|Server|Magento Root Path|
|------|-------------------------|
|Server 1|`/var/www/vhosts/server1/httpdocs/`|
|Server 2|`/var/www/vhosts/server2/httpdocs/`|
|Server 3|`/var/www/vhosts/server3/httpdocs/`|

By saving the paths relative to the Magento root in the cache we can prepend the Magento root path to the url to return the absolute urls without changing any functionality, the only thing we are changing is what is saved in the cache.

N/A

### Manual testing scenarios
Multi-server required with different web roots with a shared cache instance.
1. Log into the admin
2. Save a customer, this is when this method get's called
3. Saving should still work correctly without any exception being thrown

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Validator/Factory.php
+++ b/lib/internal/Magento/Framework/Validator/Factory.php
@@ -11,11 +11,8 @@
 namespace Magento\Framework\Validator;
 
 use Magento\Framework\Cache\FrontendInterface;
-use Magento\Framework\Module\Dir\Reader;
-use Magento\Framework\ObjectManagerInterface;
 use Magento\Framework\App\Filesystem\DirectoryList;
 use Magento\Framework\Config\FileIteratorFactory;
-use Magento\Framework\Phrase;
 
 class Factory
 {
@@ -23,7 +20,7 @@ class Factory
     const CACHE_KEY = __CLASS__;
 
     /**
-     * @var ObjectManagerInterface
+     * @var \Magento\Framework\ObjectManagerInterface
      */
     protected $_objectManager;
 
@@ -40,7 +37,7 @@ class Factory
     private $isDefaultTranslatorInitialized = false;
 
     /**
-     * @var Reader
+     * @var \Magento\Framework\Module\Dir\Reader
      */
     private $moduleReader;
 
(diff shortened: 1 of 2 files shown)


david-fuehr:
In our multi node setup the same error occurred and for example prevented a successful checkout and customer registration.

As the method `\Magento\Framwork\Validator\Factory::_initializeConfigList()` determines which files should be loaded and only stores the absolute file paths (not the file contents), I would suggest to just skip caching rather than trying to sanitize file paths.

```PHP
    /**
     * Init cached list of validation files
     */
    protected function _initializeConfigList()
    {
        if (!$this->_configFiles) {
            $this->_configFiles = $this->moduleReader->getConfigurationFiles('validation.xml');
        }
    }
```

@DanCarlyon what do you think?

DanCarlyon:
@david-fuehr Not sure if it's wise to bypass the caching 🤔 It's there for a reason. The patch in this PR works (Our production site has had it live now since March and no longer has the problem) I just need to sort the unit tests out.

david-fuehr:
I think both versions do the job - but allow me some thoughts on this fix:

I looked through your changes and think that the solution works good when the file names are stored in cache. It should save some execution time assembling the absolute path rather than searching the files in the module directories.
When you first determine the files and iterate over them to modify the path, you introduce some string operations and also read the file content (calling the method `\Magento\Framework\Config\FileIterator::current` on each element). The result is not cached and thus read again when the file list is iterated the next time, so this operation seems more expensive.

Also, in each case - cached or not, the file contents are read and merged in every request (see method `\Magento\Framework\Validator\Config::_merge` - inherited from parent). So for me the advantage of caching the file names and not the merged content is not clear.
When I look through the core files, the method `\Magento\Framework\Module\Dir\Reader::getConfigurationFiles` is called several times. I found no second occurrence, where the file paths but not the content is cached. In many cases it seems, as though the content is not cached at all.

ishakhsuvarov:
Hi @DanCarlyon 
Unfortunately, we can not accept fixes to the 2.1-develop branch which are not yet part of 2.2 and 2.3
We are closing this PR now, please feel free to submit relevant fix to the newer versions first.

Thank you

PR #14422, opened 2018-03-28
Fix product url model to not include category in url if 'Use Categories Path for Product URLs' config is not enabled

tkotosz:
Fix product url model to not include category in url if 'Use Categories Path for Product URLs' config is not enabled

### Description
If the "Use Categories Path for Product URLs" config is not enabled the Product Url model still generates a product url which includes category id. So the fix for the issue to respect this config setting.

Note that the Product Collection already uses this config setting in (magento/module-catalog/Model/ResourceModel/Product/Collection.php -> addUrlRewrite method) which means it is working properly for collections already.

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Url.php
+++ b/app/code/Magento/Catalog/Model/Product/Url.php
@@ -5,6 +5,12 @@
  */
 namespace Magento\Catalog\Model\Product;
 
+use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\Framework\Filter\FilterManager;
+use Magento\Framework\Session\SidResolverInterface;
+use Magento\Framework\UrlFactory;
+use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\StoreManagerInterface;
 use Magento\UrlRewrite\Model\UrlFinderInterface;
 use Magento\UrlRewrite\Service\V1\Data\UrlRewrite;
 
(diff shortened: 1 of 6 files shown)


tkotosz:
Note: need to finish integration tests

orlangur:

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14422) <br/>All committers have signed the CLA.

tkotosz:
@orlangur hi, the fix is done, just need to complete the integration tests because they are failing. Could you please let me know where can I find a documentation about how to run the integration tests in local? Then I will try to find the time to finish it.

orlangur:
@tkotosz here you are: https://devdocs.magento.com/guides/v2.2/test/integration/integration_test_execution.html

Please squash all changes and force push after builds are green.

tkotosz:
@orlangur thanks, I have squashed the commits and fixed the tests. Hopefully the build will green now. :crossed_fingers:

tkotosz:
@orlangur I think the current build failure is not related to my PR. Could you please have a look?

app/code/Magento/Catalog/Model/Product/Url.php
```diff
@@ -19,46 +25,53 @@ class Url extends \Magento\Framework\DataObject
...
     /**
-     * @param \Magento\Framework\UrlFactory $urlFactory
-     * @param \Magento\Store\Model\StoreManagerInterface $storeManager
-     * @param \Magento\Framework\Filter\FilterManager $filter
-     * @param \Magento\Framework\Session\SidResolverInterface $sidResolver
-     * @param UrlFinderInterface $urlFinder
-     * @param array $data
+     * @var ScopeConfigInterface
+     */
+    protected $scopeConfig;
```
dmanners: Since this is a new variable I would recommend that we make this private. For Magento 2 we recommend that new class variables are added as private.

app/code/Magento/Catalog/Model/Product/Url.php
```diff
@@ -19,46 +25,53 @@ class Url extends \Magento\Framework\DataObject
...
-        \Magento\Framework\UrlFactory $urlFactory,
-        \Magento\Store\Model\StoreManagerInterface $storeManager,
-        \Magento\Framework\Filter\FilterManager $filter,
-        \Magento\Framework\Session\SidResolverInterface $sidResolver,
+        UrlFactory $urlFactory,
+        StoreManagerInterface $storeManager,
+        FilterManager $filter,
+        SidResolverInterface $sidResolver,
         UrlFinderInterface $urlFinder,
+        ScopeConfigInterface $scopeConfig,
```
dmanners: Can you please add this as an optional dependancy and add it after the `array $data = []`. By making this change we will be able to keep your code as a backwards compatible change. See https://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#adding-a-constructor-parameter for more information.

app/code/Magento/Catalog/Test/Unit/Model/Product/UrlTest.php
```diff
@@ -212,6 +230,18 @@ public function getUrlDataProvider()
                 ['_direct' => '/product/url/path', '_query' => []],
                 null,
                 null,
+                false,
+            ], [
```
dmanners, line 232: Thanks for adding a test case to cover the category being set as well as updating the other cases to work with the false flag.

dmanners (changes requested):
Thanks for this PR. If you could update the code to have a private variable and optional constructor injection then I will be happy to process this PR.

tkotosz:
@dmanners I just applied the suggested changes. I will squash the commits if the build is the still green. :)

dmanners:
Thanks for the changes @tkotosz I will start to process this now. If I need anything else I will drop you a message here.

app/code/Magento/Catalog/Model/Product/Url.php
```diff
@@ -19,54 +25,63 @@ class Url extends \Magento\Framework\DataObject
...
-        \Magento\Framework\Filter\FilterManager $filter,
-        \Magento\Framework\Session\SidResolverInterface $sidResolver,
+        UrlFactory $urlFactory,
+        StoreManagerInterface $storeManager,
+        FilterManager $filter,
+        SidResolverInterface $sidResolver,
         UrlFinderInterface $urlFinder,
-        array $data = []
+        array $data = [],
+        ScopeConfigInterface $scopeConfig = null
```
dmanners: For this code we are getting the following when running our static tests.

```
/var/www/html/app/code/Magento/Catalog/Model/Product/Url.php:23	The class Url has a coupling between objects value of 13. Consider to reduce the number of dependencies under 13.
```

Can you think of a way of refactoring these dependencies and class so that it is not such a high number?

dmanners (changes requested):

app/code/Magento/Catalog/Model/Product/Url.php
tkotosz: yep, I can add a different class as dependency instead of the ScopeConfigInterface and move the logic of getting the "product url use category" flag value to that class... I think that might reduce the number of coupled classes... will have a look as soon as I have time

sivaschenko:
Hi @tkotosz thank for the collaboration, are you going to continue progress on this pull request?

tkotosz:
Hi @sivaschenko , Thanks for the reminder I will get back to this PR this week, I forgot that there is 1 last thing to fix in this.

sivaschenko:
Hi @tkotosz I am closing this pull request due to inactivity, feel free to reopen if you'd like to continue progress on it.

tkotosz:
oh it seems since the original PR some other version of the fix was applied already (using the scopeconfig in the same way) so I will close this one

PR #14445, opened 2018-03-29
[FIX] Simplify ternary operators for catalog module

coderimus:
### Description
Hello,
Worked on the internal issue I have found that there is possible to simplify some ternary operators for catalog module. I would be happy to do this for all modules in future by phpcs-fixer automatically but as I can see it doesn't have such ability.

N/A

### Manual testing scenarios
N/A

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/CategoryRepository.php
+++ b/app/code/Magento/Catalog/Model/CategoryRepository.php
@@ -129,7 +129,7 @@ class CategoryRepository implements \Magento\Catalog\Api\CategoryRepositoryInter
      */
     public function get($categoryId, $storeId = null)
     {
-        $cacheKey = null !== $storeId ? $storeId : 'all';
+        $cacheKey = $storeId ?? 'all';
         if (!isset($this->instances[$categoryId][$cacheKey])) {
             /** @var Category $category */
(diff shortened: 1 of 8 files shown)


app/code/Magento/Catalog/Model/CategoryRepository.php
```diff
@@ -129,7 +129,7 @@ public function save(\Magento\Catalog\Api\Data\CategoryInterface $category)
      */
     public function get($categoryId, $storeId = null)
     {
-        $cacheKey = null !== $storeId ? $storeId : 'all';
+        $cacheKey = null !== $storeId ?: 'all';
```
adrian-martinez-interactiv4: This change breaks expected functionality, try:
```
$cacheKey = $storeId ?? 'all';
```

coderimus:
Hi @adrian-martinez-interactiv4 thank you for your detail review. I implemented requested fix and squashed commits to one. Please, review it.

orlangur:
>  I would be happy to do this for all modules in future by `php-cs-fixer` automatically but as I can see it doesn't have such ability.

Just do it :wink: There is no reason to waste time on performing such changes manually.

Issue #14424, opened 2018-03-28
Mistake with loading an order in Invoice model


springimport:
### Preconditions
I try to get an extension attributes from Order using the Invoice model.

### Steps to reproduce
Use [getOrder()](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/Model/Order/Invoice.php#L239) in Ivoice model.

### Expected result
Invoice getOrder() uses orderRepository ([Shipment](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/Model/Order/Shipment.php#L204) as an example).

### Actual result
Invoice getOrder() uses orderFactory.

PHP 7.1. Magento 2.2.3.

orlangur:
@springimport there is no mistake here, just don't mix legacy models with new approach sticked to repositories.

You are not supposed to get order with extension attributes like `$invoice->getOrder()`, such approach violates Law of Demeter. Simply obtain order model via order repository.

springimport:
@orlangur Ok, what should I to do?
```php
$this->orderRepository->get($invoice->getOrderid());
```

orlangur:
@springimport yep, code like this should do the trick.

PR #14450, opened 2018-03-29
Fix mistake with loading an order in Invoice model

TomashKhamlai:
Should allow you to get an extension attributes from Order using the Invoice model.

### Description
In this pull request was changed the way of getting information about the Invoice.

### Fixed Issues
1. magento/magento2#14424: Mistake with loading an order in Invoice model

### Manual testing scenarios
1. Check that now you can get extension attributes using WebAPI client
2. Check that Magento Swagger also returns extension attributes without error.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Invoice.php
+++ b/app/code/Magento/Sales/Model/Order/Invoice.php
@@ -96,9 +96,9 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
     protected $_invoiceConfig;
 
     /**
-     * @var \Magento\Sales\Model\OrderFactory
+     * @var \Magento\Sales\Api\OrderRepositoryInterface
      */
-    protected $_orderFactory;
+    protected $orderRepository;
 
     /**
      * @var \Magento\Framework\Math\CalculatorFactory
@@ -126,7 +126,7 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
      * @param \Magento\Framework\Api\ExtensionAttributesFactory $extensionFactory
      * @param AttributeValueFactory $customAttributeFactory
      * @param Invoice\Config $invoiceConfig
-     * @param \Magento\Sales\Model\OrderFactory $orderFactory
+     * @param \Magento\Sales\Api\OrderRepositoryInterface $orderRepository
      * @param \Magento\Framework\Math\CalculatorFactory $calculatorFactory
      * @param \Magento\Sales\Model\ResourceModel\Order\Invoice\Item\CollectionFactory $invoiceItemCollectionFactory
      * @param Invoice\CommentFactory $invoiceCommentFactory
@@ -142,7 +142,7 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
         \Magento\Framework\Api\ExtensionAttributesFactory $extensionFactory,
         AttributeValueFactory $customAttributeFactory,
         \Magento\Sales\Model\Order\Invoice\Config $invoiceConfig,
-        \Magento\Sales\Model\OrderFactory $orderFactory,
+        \Magento\Sales\Api\OrderRepositoryInterface $orderRepository,
         \Magento\Framework\Math\CalculatorFactory $calculatorFactory,
         \Magento\Sales\Model\ResourceModel\Order\Invoice\Item\CollectionFactory $invoiceItemCollectionFactory,
         \Magento\Sales\Model\Order\Invoice\CommentFactory $invoiceCommentFactory,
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14450) <br/>All committers have signed the CLA.

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -236,7 +236,7 @@ public function setOrder(\Magento\Sales\Model\Order $order)
     public function getOrder()
     {
         if (!$this->_order instanceof \Magento\Sales\Model\Order) {
```
ishakhsuvarov, line 238: Since `OrderRepository` may return any class implementing `OrderInterface` this check may be false-negative in some scenarios on customized system.

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -142,7 +142,7 @@ public function __construct(
         \Magento\Framework\Api\ExtensionAttributesFactory $extensionFactory,
         AttributeValueFactory $customAttributeFactory,
         \Magento\Sales\Model\Order\Invoice\Config $invoiceConfig,
-        \Magento\Sales\Model\OrderFactory $orderFactory,
```
ishakhsuvarov, line 145: It's not allowed to change constructor signature, especially for classes marked with `@api` annotation.
More information is available here: http://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/index.html

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -236,7 +236,7 @@ public function setOrder(\Magento\Sales\Model\Order $order)
     public function getOrder()
     {
         if (!$this->_order instanceof \Magento\Sales\Model\Order) {
-            $this->_order = $this->_orderFactory->create()->load($this->getOrderId());
+            $this->_order = $this->orderRepository->get($this->getOrderId());
```
ishakhsuvarov, line 239: Doc block type hint of `$_order` property was not updated accordingly.

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -96,9 +96,9 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
     protected $_invoiceConfig;
 
     /**
-     * @var \Magento\Sales\Model\OrderFactory
+     * @var \Magento\Sales\Api\OrderRepositoryInterface
      */
-    protected $_orderFactory;
```
ishakhsuvarov, line 101: Removing `$_orderFactory` may break clients already relying on this property. It's better to deprecate it.

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -236,7 +236,7 @@ public function setOrder(\Magento\Sales\Model\Order $order)
     public function getOrder()
     {
         if (!$this->_order instanceof \Magento\Sales\Model\Order) {
-            $this->_order = $this->_orderFactory->create()->load($this->getOrderId());
+            $this->_order = $this->orderRepository->get($this->getOrderId());
         }
         return $this->_order->setHistoryEntityName($this->entityType);
```
ishakhsuvarov, line 241: * Method doc block still states that return type is `Magento\Sales\Model\Order` which may not be true in some cases. It's best to change it to `OrderInterface` now. 
* Method clients must be checked to avoid unexpected consequences.
* `setOrder` method of this class is now inconsistent with `getOrder`. It's best to check possibility of aligning signatures and behaviors.

app/code/Magento/Sales/Test/Unit/Model/Order/InvoiceTest.php
```diff
@@ -74,6 +75,11 @@ protected function setUp()
                 ]
             )
             ->getMock();
+        $this->orderRepository = $this->getMockBuilder(OrderRepository::class)
```
ishakhsuvarov, line 78: It's best to mock `OrderRepositoryInterface` here

app/code/Magento/Sales/Test/Unit/Model/Order/InvoiceTest.php
```diff
@@ -74,6 +75,11 @@ protected function setUp()
...
+            ->disableOriginalConstructor()
+            ->setMethods(['get'])
```
ishakhsuvarov, line 80: What is the reason for `setMethods` here?

app/code/Magento/Sales/Test/Unit/Model/Order/InvoiceTest.php
```diff
@@ -156,17 +160,13 @@ public function testGetOrderLoadedById()
     {
         $orderId = 100000041;
         $this->modelWithoutOrder->setOrderId($orderId);
-        $this->order->expects($this->once())
-            ->method('load')
-            ->with($orderId)
-            ->willReturnSelf();
+        $this->orderRepository->expects($this->once())
```
ishakhsuvarov, line 163: It may be a good idea to check both cases: 
* in-object cache is used
* new entity is created

ishakhsuvarov:
I would also suggest another testing scenario: verify that none of the currently existing properties in web api response is missing/changed after the change.

sidolov:
@TomashKhamlai , I am closing this PR now due to inactivity.
Thank you for collaboration!

PR #14452, opened 2018-03-29
Remove else statements from \Magento\Framework\Session\SessionManager

adrian-martinez-interactiv4:
### Description
Remove else statements from \Magento\Framework\Session\SessionManager

None, improvement

### Manual testing scenarios
None, improvement

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/Session/SessionManager.php
+++ b/lib/internal/Magento/Framework/Session/SessionManager.php
@@ -317,11 +317,8 @@ class SessionManager implements SessionManagerInterface
      */
     public function destroy(array $options = null)
     {
-        if (null === $options) {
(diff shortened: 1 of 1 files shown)


lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -504,11 +501,7 @@ public function regenerateId()
             return $this;
         }
 
-        if ($this->isSessionExists()) {
-            session_regenerate_id(true);
-        } else {
-            session_start();
-        }
+        $this->isSessionExists() ? session_regenerate_id(true) : session_start();
```
orlangur, line 504: Ternary operator is not a good idea when there is no return value.

By the way, in terms of code calisthenics ternary operator is still else :)

PR #14484, opened 2018-04-01
magneto/magento2#12362 Do not remove session instantly

swnsma:
Give configurable time delay to capture session id.
Remove old sessions only after new session id cookie was received.

### Description
Current pull request contains fix for the issue #12362.
Problem related to specific case of session work in PHP.
In case of call function 'session_regenerate_id(true)' there is required to be sure that renewed cookie will be successfully delivered to the browser.
Current PR provide functionality to keep old session till customer will not receive cookie.

### Fixed Issues

1. magento/magento2#12362: Concurrent (quick reload) requests on checkout cause cart to empty - related to session_regenerate_id
2.  magento/magento2#13427: [2.1.11] Add to cart, try to checkout, cart is empty but mini-cart has items.
3. magento/magento2#4301: Hit fast twice F5 on checout page, customer loggs out automatically

#### Pre-Conditions
1. Magento 2.2.0
2. PHP7
#### Steps to reproduce

1. Add product to cart
2. Navigate to checkout
3. Reload the page as quick as possible (F5 multiple times)
#### Expected result

1. Reload the checkout as usual
2. Cart contains session items
#### Actual result

1. Redirects to empty cart

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/etc/config.xml
+++ b/app/code/Magento/Store/etc/config.xml
@@ -84,7 +84,7 @@
                 <use_http_x_forwarded_for>0</use_http_x_forwarded_for>
                 <use_http_user_agent>0</use_http_user_agent>
                 <use_frontend_sid>1</use_frontend_sid>
-                <old_session_access_delta>300</old_session_access_delta>
+                <old_session_access_delta>30</old_session_access_delta>
             </session>
             <browser_capabilities>
                 <cookies>1</cookies>
--- a/app/code/Magento/Store/etc/config.xml
+++ b/app/code/Magento/Store/etc/config.xml
@@ -84,6 +84,7 @@
                 <use_http_x_forwarded_for>0</use_http_x_forwarded_for>
                 <use_http_user_agent>0</use_http_user_agent>
                 <use_frontend_sid>1</use_frontend_sid>
+                <old_session_access_delta>300</old_session_access_delta>
             </session>
             <browser_capabilities>
                 <cookies>1</cookies>
--- a/lib/internal/Magento/Framework/Session/SessionManager.php
+++ b/lib/internal/Magento/Framework/Session/SessionManager.php
@@ -15,6 +15,12 @@ use Magento\Framework\Session\Config\ConfigInterface;
  */
 class SessionManager implements SessionManagerInterface
 {
+    const SESSION_OLD_TIMESTAMP = 'destroyed';
+
+    const NEW_SESSION_ID = 'new_session_id';
+
+    const OLD_SESSION_ID = 'old_session_id';
+
     /**
      * Default options when a call destroy()
      *
@@ -186,8 +192,7 @@ class SessionManager implements SessionManagerInterface
             $sid = $this->sidResolver->getSid($this);
             // potential custom logic for session id (ex. switching between hosts)
             $this->setSessionId($sid);
-            session_start();
-            $this->validator->validate($this);
+            $this->sessionStart();
             $this->renewCookie($sid);
 
             register_shutdown_function([$this, 'writeClose']);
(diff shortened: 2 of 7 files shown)


lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -200,6 +205,33 @@ public function start()
         return $this;
     }
 
+    /**
+     * Start session.
+     *
+     * @throws \Magento\Framework\Exception\SessionException
+     */
+    private function sessionStart()
```
larsroettig: sessionStart accesses the super-global variable $_SESSION.
https://app.codacy.com/app/shyamranpara/magento2/pullRequest?prid=1494337

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -517,6 +549,37 @@ public function regenerateId()
         return $this;
     }
 
+    /**
+     * Regenerate session id, with saving relation between two sessions.
+     */
+    protected function regenerateSessionId()
```
larsroettig: regenerateSessionId accesses the super-global variable $_SESSION.

Superglobal variables in PHP should not be accessed directly. Instead of writing:

`$name = $_POST['foo'];`
You should encapsulate the call in an single class. Most PHP frameworks already wrap Superglobals into a request object so you can use that.


https://app.codacy.com/app/shyamranpara/magento2/pullRequest?prid=1494337

app/code/Magento/Store/etc/config.xml
```diff
@@ -84,6 +84,7 @@
                 <use_http_x_forwarded_for>0</use_http_x_forwarded_for>
                 <use_http_user_agent>0</use_http_user_agent>
                 <use_frontend_sid>1</use_frontend_sid>
+                <old_session_access_delta>300</old_session_access_delta>
```
orlangur: I don't think putting such a sensitive configuration option into Admin UI is a good idea. 300 seconds is a huge delay, I believe we should pick up some reasonable value.

Ideally solution must not involve any timeout at all allowing just to create new session seamlessly.

larsroettig (changes requested):
Hi @swnsma,
Thank you for contribution. Currently, I cannot accept your **Pull request** for merge. First, the codacy build is **red** see review comments. The second problem is the static test see https://travis-ci.org/magento/magento2/jobs/360827154

Best regards,

Lars

orlangur (changes requested):
Apply changes against `2.2-develop` branch first so that this PR can be processed.

app/code/Magento/Store/etc/config.xml
swnsma: @orlangur, thank you for the review.
I agree that such configuration should not be placed in Admin UI.

About timeout, I have added it to the solution as additional security frontier in session recovery.
Old session will be removed as soon as response with new session id will be received.
And this timeout is for that cases when new session was generated but response with new session id will never be received.

Unfortunately, one of PHP session flaw is that you can not be sure that cookie with session id was delivered to browser or no.

About timeout, I agree that 5m is a huge delay.  I have decreased the default number to 30s.

lib/internal/Magento/Framework/Session/SessionManager.php
swnsma: @larsroettig, thank you for the review.
The problem has been fixed.

lib/internal/Magento/Framework/Session/SessionManager.php
swnsma: @larsroettig, thank you for the review.
The problem has been fixed.

swnsma:
Pull request into 2.2-develop has been created: #14487

PR #14527, opened 2018-04-04
Added a better posibility to add an outputHandler

quisse:
### Description
Added a method which does nothing but can be used to create a plugin for

1. magento/magento2#12371: Adding a handler to the Magento\Catalog\Helper\Output Class

### Manual testing scenarios

1. before: https://gist.github.com/quisse/b78ee3d121a4c9b8dd141334c8fa27a8
2. after: https://gist.github.com/quisse/10e2a5a90f8dc55f82f972efe138e3ad

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cms/Model/PageRepository.php
+++ b/app/code/Magento/Cms/Model/PageRepository.php
@@ -110,7 +110,7 @@ class PageRepository implements PageRepositoryInterface
      */
     public function save(\Magento\Cms\Api\Data\PageInterface $page)
     {
-        if (empty($page->getStoreId())) {
+        if ($page->getStoreId() === null) {
             $storeId = $this->storeManager->getStore()->getId();
             $page->setStoreId($storeId);
         }


sidolov:
Hi @quisse , I agree with you that current solution hasn't valid point for extending. Adding new empty method that called from the constructor is not a good idea, the mentioned helper is legacy code and doesn't match our [Technical guidelines](https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html) and we don't want to make this code more complicated and unobvious. Current implementation needs more complex solution and refactoring and for now your [current](https://gist.github.com/quisse/b78ee3d121a4c9b8dd141334c8fa27a8) implementation the best way to extend `\Magento\Catalog\Helper\Output`

quisse:
Hi @sidolov , i was expecting this isn't the best solution but hey, it's a start 😀. Can you please point me in a direction about how to refactor or shall i close this pull request & remove the non issue label from [magento/magento2#12371](https://github.com/magento/magento2/issues/12371) so this can be picked up by somebody who knows how to handle?

quisse:
Closed this because I've moved this to other branch. I still want to create a new pr with proper refactoring @sidolov .

quisse:
@sidolov

PR #14607, opened 2018-04-09
Remove foreach and break statements

thomas-kl1:
Retrieves the first element key of the options.

### Description

I was reading some piece of code when I see te "foreach" and "break" statements. I know that the "break" should be avoided, so here my suggestion to retrieve the key of the first element of the array.

Is there a goal to do it via a foreach and break?

### DIFF
--- a/app/code/Magento/Rule/Model/Condition/AbstractCondition.php
+++ b/app/code/Magento/Rule/Model/Condition/AbstractCondition.php
@@ -83,18 +83,14 @@ abstract class AbstractCondition extends \Magento\Framework\DataObject implement
         $this->loadAttributeOptions()->loadOperatorOptions()->loadValueOptions();
 
         $options = $this->getAttributeOptions();
-        if ($options) {
-            foreach (array_keys($options) as $attr) {
-                $this->setAttribute($attr);
-                break;
-            }
+        if (!empty($options)) {
+            $attrs = array_keys($options);
+            $this->setAttribute($attrs[0]);
         }
         $options = $this->getOperatorOptions();
-        if ($options) {
-            foreach (array_keys($options) as $operator) {
-                $this->setOperator($operator);
-                break;
-            }
+        if (!empty($options)) {
+            $operators = array_keys($options);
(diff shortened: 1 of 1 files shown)


app/code/Magento/Rule/Model/Condition/AbstractCondition.php
```diff
@@ -83,18 +83,14 @@ public function __construct(Context $context, array $data = [])
         $this->loadAttributeOptions()->loadOperatorOptions()->loadValueOptions();
 
         $options = $this->getAttributeOptions();
-        if ($options) {
-            foreach (array_keys($options) as $attr) {
-                $this->setAttribute($attr);
-                break;
-            }
+        if (!empty($options)) {
```
orlangur: Change back to `if ($options)`, no need for a redundant syntax.

app/code/Magento/Rule/Model/Condition/AbstractCondition.php
```diff
@@ -83,18 +83,14 @@ public function __construct(Context $context, array $data = [])
...
+            $attrs = array_keys($options);
+            $this->setAttribute($attrs[0]);
```
orlangur: More effective way to obtain key of the first element: https://stackoverflow.com/a/1028677/8640867

app/code/Magento/Rule/Model/Condition/AbstractCondition.php
```diff
@@ -520,9 +516,10 @@ public function getTypeElementHtml()
     public function getAttributeElement()
     {
         if (null === $this->getAttribute()) {
-            foreach (array_keys($this->getAttributeOption()) as $option) {
-                $this->setAttribute($option);
-                break;
+            $options = $this->getAttributeOptions();
+            if (!empty($options)) {
```
orlangur: No need to change logic: please remove this condition.

orlangur (changes requested):
Please apply described changes in all places and prepare as PR against `2.2-develop` branch first.

PR #14704, opened 2018-04-14
Catalog price rule save is too slow, ignore on demand reindex

chrom:
**Preconditions**
Have a large enough data set
Set indexing mode to Reindex on schedule
**Steps to reproduce**
Go to admin and create new catalog price rule that has condition based on category
Hit save
**Expected result**
New catalog price rule should be saved within reasonable time
**Actual result**
Request times out
**Details**
There are actually more than one thing wrong here:

1. Function after save of Magento\CatalogRule\Model\Rule walks through collection of products and triggers the reindex, regardless of indexing setting. It may do nothing, but the fact that it does go over the data set makes this action really slow. Plus, depending on third party integrations, it may trigger unwanted sync at this point. And I know that what third party extensions do is not under your control, just pointing out that something is triggered and that is not supposed to happen at this time.
3. They way it fetches the affected products is wrong. Function [getMatchingProductIds](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/CatalogRule/Model/Rule.php#L289-L321) will not scope the collection by category. And if that is your only condition, you will get whole catalog, even though the rule will affect only two products in that category. With the large enough set, code will end up checking potentially hundreds of thousands of products instead of just one category.

Combining these two, you get additional actions triggered where not necessary, on a data set that is too large in the first place hence the transaction timing out. I have observed this on just 140k products in the catalog.

### DIFF
--- a/app/code/Magento/CatalogRule/Model/Rule.php
+++ b/app/code/Magento/CatalogRule/Model/Rule.php
@@ -8,8 +8,10 @@ namespace Magento\CatalogRule\Model;
 use Magento\Catalog\Model\Product;
 use Magento\Catalog\Model\ProductFactory;
 use Magento\Catalog\Model\ResourceModel\Product\CollectionFactory;
+use Magento\CatalogRule\Api\Data\RuleExtensionInterface;
 use Magento\CatalogRule\Api\Data\RuleInterface;
 use Magento\CatalogRule\Helper\Data;
+use Magento\CatalogRule\Model\Data\Condition\Converter;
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14704) <br/>All committers have signed the CLA.

magento-engcom-team:
@chrom thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

magento-engcom-team:
Hi @chrom. Thank you for your contribution. 

Issue #14788, opened 2018-04-20
Impossible to change configurable product attributes position on PDP


rocketweb:
Configurable Product Attributes on product details page are not sorted according to configurable product attribute option position.
It's impossible to change configurable product attributes position. 

### Preconditions

Magento 2.2-develop

### Steps to reproduce

1. Create attributes used for creating configurable product:
First attribute:  'size' with sort_order 100
Second attribute: 'color' with sort_order 200
2. Create configurable product based on this attributes.
3. Now on PDP page would be shown size attribute at the top, color attribute at the bottom under size. 
4. Try to change attributes position. It's required to show color first, size second. 
Via upgrade script try to change attributes position: 
$eavSetup->updateAttribute(Product::ENTITY, 'color','position', 10, 200);
$eavSetup->updateAttribute(Product::ENTITY, 'size','position', 11, 210);

### Expected result

Configurable Product Attributes on product details page should be sorted according to configurable product attribute option position.

### Actual result

Configurable Product Attributes on product details page  sorted by attribute id.

ghost:
@rocketweb, thank you for your report.
We've acknowledged the issue and added to our backlog.

AleksLi:
@rocketweb  I found the reason for that behavior. I've investigated the issue and found that you were changed the wrong arguments.
`$eavSetup->updateAttribute(Product::ENTITY, 'color','position', 10, 200);`
The position you were trying to change is the position of the attribute in [category layer navigation](https://monosnap.com/file/oDfblxX2DK6UNqryKJJt3MQ4B0MZnA) 

The real one you are looking for is [this one](http://joxi.ru/52a5RYetElXRE2). Then go to the [step 2](https://monosnap.com/file/Sgtm3GWJdpJ0H6YpN1VvxUmdrTLHwV). Those changes will be written to `catalog_product_super_attribute` table and affect the position on Product Page.

**@magento-engcom-team  I guess we can close this one.**

AleksLi:
@rocketweb Reach out to me if something is still unclear. 
CC::@magento-engcom-team

magento-engcom-team:
Hi @AleksLi. Thank you for working on this issue. 
Looks like this issue is already verified and confirmed. But if you want to validate it one more time, please, go though the following instruction:
- [ ] 1. Add/Edit `Component: XXXXX` label(s) to the ticket, indicating the components it may be related to.

[...]

PR #14810, opened 2018-04-23
 Impossible to change configurable product attributes position on PDP #14788

ghost:
Configurable Product Attributes on product details page are not sorted according to configurable product attribute option position.
It's impossible to change configurable product attributes position. 

### Preconditions

Magento 2.2-develop

### Steps to reproduce

1. Create attributes used for creating configurable product:
First attribute:  'size' with sort_order 100
Second attribute: 'color' with sort_order 200
2. Create configurable product based on this attributes.
3. Now on PDP page would be shown size attribute at the top, color attribute at the bottom under size. 
4. Try to change attributes position. It's required to show color first, size second. 
Via upgrade script try to change attributes position: 
$eavSetup->updateAttribute(Product::ENTITY, 'color','position', 10, 200);
$eavSetup->updateAttribute(Product::ENTITY, 'size','position', 11, 210);

### Expected result

Configurable Product Attributes on product details page should be sorted according to configurable product attribute option position.

### Actual result

Configurable Product Attributes on product details page  sorted by attribute id.

### DIFF
--- a/app/code/Magento/ConfigurableProduct/Model/Config/Source/ProductAttributesSort.php
+++ b/app/code/Magento/ConfigurableProduct/Model/Config/Source/ProductAttributesSort.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * Copyright © 2013-2017 Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\ConfigurableProduct\Model\Config\Source;
+
+class ProductAttributesSort implements \Magento\Framework\Option\ArrayInterface
+{
+    /**
+     * Options array
+     *
+     * @var array
+     */
+    protected $options;
+
+    const PER_PRODUCT = 0;
+
+    const GLOBAL_ATTRIBUTE_SETTING = 1;
+
+    /**
+     * @return array
+     */
+    public function toOptionArray()
+    {
+        if (!$this->options) {
+            $this->options = [
+                ['label' => __('Position Per Product'), 'value' => self::PER_PRODUCT],
+                ['label' => __('Global Attribute Setting'), 'value' => self::GLOBAL_ATTRIBUTE_SETTING],
+            ];
+        }
+        return $this->options;
+    }
+}
--- a/app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
+++ b/app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
@@ -8,6 +8,8 @@ namespace Magento\ConfigurableProduct\Model;
 
 use Magento\Catalog\Model\Product;
 use Magento\ConfigurableProduct\Model\Product\Type\Configurable\Attribute;
+use Magento\ConfigurableProduct\Model\Config\Source\ProductAttributesSort;
+use Magento\Framework\App\Config\ScopeConfigInterface;
 
 /**
  * Class ConfigurableAttributeData
@@ -16,6 +18,22 @@ use Magento\ConfigurableProduct\Model\Product\Type\Configurable\Attribute;
  */
 class ConfigurableAttributeData
 {
+    const XML_PATH_CONF_PRODUCT_ATTRIBUTES_SORT = 'catalog/frontend/product_attributes_sort_by';
+
+    /**
+     * @var ScopeConfigInterface
+     */
+    protected $scopeConfig;
+
+    /**
+     * @param ScopeConfigInterface $scopeConfig
+     */
+    public function __construct(
+        ScopeConfigInterface $scopeConfig
+    ) {
+        $this->scopeConfig = $scopeConfig;
+    }
+
     /**
      * Get product attributes
      *
@@ -37,7 +55,7 @@ class ConfigurableAttributeData
                     'code' => $productAttribute->getAttributeCode(),
                     'label' => $productAttribute->getStoreLabel($product->getStoreId()),
                     'options' => $attributeOptionsData,
-                    'position' => $productAttribute->getPosition(),
+                    'position' => $this->getPosition($attribute),
                 ];
                 $defaultValues[$attributeId] = $this->getAttributeConfigValue($attributeId, $product);
             }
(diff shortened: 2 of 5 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14810) <br/>All committers have signed the CLA.

magento-engcom-team:
@rocketweb thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

VladimirZaets:
Hi @rocketweb , please, sign CLA, otherwise, we can't process your pull request.

VladimirZaets:
hi @rocketweb 
In the testing process, we find the scenario that is failed. Please, can you look and fix it.

**Problems**: The order of attributes that was set in admin, is incorrect on PDP.
**Steps to reproduce:**
Create attributes used for creating configurable product:
Attribute: 'size' with Catalog Input Type for Store Owner=Text Swatch
Attribute: 'color' with Type for Store Owner=Dropdown
Create configurable product based on this attributes.
Set attribute order: first attribute 'size' and second attribute 'color'.
Save product.
Go to Product detailed page on frontend
**Expected result:**
Attribute order same as was set in admin.
**Actual result:**
Configurable Product Attributes on product details page sorted by attribute id.

rocketweb:
Hello,

According to testing process scenario:

In current Magento implementation attributes sort order saved per product. Yes you could change attributes sort order while creating product in Admin as mentioned in scenario. But it's impossible to change attributes sort order for existing products without updating thousands of products or re-save all simple products for each configurable. 
Does it really make sense in such case if it's impossible to change saved values?
When products exported to Magento or when we need to change attributes sort order for all products. For example, if required  colors always show first and sizes always show second. That is the **main purpose of this fix** and the real client's case.

From the other hand we have field Position used to sort attributes in Layered Navigation. This field also could be used to change attributes sort order on PDP page (as was done in provided fix). Globally for all products. In this case sort order could be easily changed via Admin panel for attribute.

If case with keeping attributes sort order per product so important the solution could be:
Add configuration setting for PDP page under Stores -> Configuration -> Catalog -> Storefront
Sort Products attributes by position = Global attribute setting / Per product
and use sorting provider according to this setting.

"Global attribute setting" -  Use Global settings from product attribute position field. Used when we need global unified behavior. For example, colors always first and sizes always second. No matter how was moved attributes in Admin when configurable product created in Admin

"Per product" -  Use sort order from product super attribute position (current behavior)

VladimirZaets:
hi @rocketweb, I agree that it's can be more common practice, but unfortunately, this is backward incompatible changes. This fix can effect attributes sort order on our customers. 
Your fix is needful, but you should do it with backward compatible way.

rocketweb:
Hi @VladimirZaets, 
so what is decision according to provided solution in previous comment using configuration?
Stores -> Configuration -> Catalog -> Storefront 
Sort Products attributes by position = Global attribute setting / Per product

We could prepare fix where config option added and by default selected current type of sorting (Per product). This would be backward compatible for existing customers.  But if customer needs sorting other way - globally for all attributes he just switch config option to Sort Products attributes by position = Global attribute setting.

VladimirZaets:
Hi @rocketweb.
Yes, the solution with using configuration will be great. Thanks

app/code/Magento/ConfigurableProduct/Model/Config/Source/ProductAttributesSort.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Magento\ConfigurableProduct\Model\Config\Source;
+
+class ProductAttributesSort implements \Magento\Framework\Option\ArrayInterface
+{
+    /**
+     * Options array
+     *
+     * @var array
+     */
+    protected $options;
```
VladimirZaets, line 15: Thank for changes.  Please change the protected property to private. According [Magento technical guideline](https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html)  we can't introduce new protected properties

app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
```diff
@@ -8,6 +8,8 @@
```
VladimirZaets, line 8: Please don't change file mode permission. `100644 → 100755`

app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
```diff
@@ -16,6 +18,22 @@
  */
 class ConfigurableAttributeData
 {
+    const XML_PATH_CONF_PRODUCT_ATTRIBUTES_SORT = 'catalog/frontend/product_attributes_sort_by';
+
+    /**
+     * @var ScopeConfigInterface
+     */
+    protected $scopeConfig;
```
VladimirZaets, line 26: Please change the protected property to private. According [Magento technical guideline](https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html)  we can't introduce new protected properties

app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
```diff
@@ -16,6 +18,22 @@
...
+
+    /**
+     * @param ScopeConfigInterface $scopeConfig
+     */
+    public function __construct(
```
VladimirZaets, line 31: Due to [Magento backward-compatible guide](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/), we can't add newly required params to the constructor.

Please use `$this->newDependency = $newDependency ?: \Magento\Framework\App\ObjectManager::getInstance()->get(\New\Dependency\Interface::class);
` pattern to add new dependency

app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
```diff
@@ -48,6 +66,35 @@ public function getAttributesData(Product $product, array $options = [])
...
     }
 
+    /**
+     * Get configurable product attribute position
+     *
+     * @param $attribute
+     *
+     * @return mixed
+     */
+    protected function getPosition($attribute)
```
VladimirZaets, line 76: Please change the protected property to private. According [Magento technical guideline](https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html)  we can't introduce new protected properties

app/code/Magento/ConfigurableProduct/Model/ConfigurableAttributeData.php
```diff
@@ -48,6 +66,35 @@ public function getAttributesData(Product $product, array $options = [])
...
+
+        return $attribute->getPosition();
+    }
+
+    /**
+     * Get configuration value for configurable products attributes sorting settings
+     *
+     * @return int
+     */
+    protected function getConfigurableProductAttributesSortConfig()
```
VladimirZaets, line 90: Please change the protected property to private. According [Magento technical guideline](https://devdocs.magento.com/guides/v2.2/coding-standards/technical-guidelines.html)  we can't introduce new protected properties

VladimirZaets:
Hi @rocketweb, I am closing this PR now due to inactivity.
Thank you for collaboration

PR #14960, opened 2018-05-03
Remove direct use of Object Manager

AnshuMishra17:
Remove direct use of Object Manager by adding dependency in constructor.

### Description
Remove direct use of object manager by adding dependency in constructor

### Manual testing scenarios
1. Go to Content -> CMS Pages
2. Delete CMS pages from edit page and massaction as well.

### DIFF
--- a/app/code/Magento/Cms/Controller/Adminhtml/Page/Delete.php
+++ b/app/code/Magento/Cms/Controller/Adminhtml/Page/Delete.php
@@ -6,6 +6,8 @@
  */
 namespace Magento\Cms\Controller\Adminhtml\Page;
 
+use Magento\Backend\App\Action\Context;
+
 class Delete extends \Magento\Backend\App\Action
 {
     /**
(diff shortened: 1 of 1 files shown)


app/code/Magento/Cms/Controller/Adminhtml/Page/Delete.php
```diff
@@ -15,6 +17,23 @@ class Delete extends \Magento\Backend\App\Action
      */
     const ADMIN_RESOURCE = 'Magento_Cms::page_delete';
 
+    /**
+     * @var \Magento\Cms\Model\Page
+     */
+    private $_cmsPage;
```
dmanners, line 23: Since this property is private we can rename it so that it does not include the underscore `$cmsPage` would be fine in this case.

app/code/Magento/Cms/Controller/Adminhtml/Page/Delete.php
```diff
@@ -30,7 +49,7 @@ public function execute()
             $title = "";
             try {
                 // init model and delete
-                $model = $this->_objectManager->create(\Magento\Cms\Model\Page::class);
+                $model = $this->_cmsPage;
```
dmanners, line 52: Could we not directly use the `$this->cmsPage` here instead of setting against the `$model` variable.

dmanners (changes requested):
Thank you for your pull request. I have added a few comments on naming and also usage of variables. Could you take a look into updating these.

dmanners:
@AnshuMishra17 also PRs should be created from your fork rather than directly on the Magento 2 repository. Could you recreate this PR from a branch on your fork? If you need with this please let me know.

AnshuMishra17:
@dmanners It would be great if you can share the steps to create pull request from fork repo, I will create a new pull request with the suggested changes.

PR #14973, opened 2018-05-03
Fix unstable session manager

elioermini:
### Description

Fix of the unstable behaviour of the session management with high volume of concurrent sessions.

### Fixed Issues
1. magento/magento2#12362: Concurrent (quick reload) requests on checkout cause cart to empty - related to session_regenerate_id

### Manual testing scenarios

1. Reach the checkout page with products in the cart
2. Press CMD+R or ALT+F5 multiple times to refresh the page, after the fix the user is not being logged out and redirected to the empty cart page.

We already deployed the fix on our high traffic website and the issue is fixed.

### DIFF
--- a/lib/internal/Magento/Framework/Session/SessionManager.php
+++ b/lib/internal/Magento/Framework/Session/SessionManager.php
@@ -183,10 +183,21 @@ class SessionManager implements SessionManagerInterface
             // Need to apply the config options so they can be ready by session_start
             $this->initIniOptions();
             $this->registerSaveHandler();
+            if (isset($_SESSION['new_session_id'])) {
+                // Not fully expired yet. Could be lost cookie by unstable network.
+                session_commit();
+                session_id($_SESSION['new_session_id']);
+            }
             $sid = $this->sidResolver->getSid($this);
             // potential custom logic for session id (ex. switching between hosts)
             $this->setSessionId($sid);
             session_start();
+            if (isset($_SESSION['destroyed'])) {
+                if ($_SESSION['destroyed'] < time() - 300) {
+                    $this->destroy(['clear_storage' => true]);
+
+                }
+            }
             $this->validator->validate($this);
             $this->renewCookie($sid);
 
@@ -501,7 +512,31 @@ class SessionManager implements SessionManagerInterface
             return $this;
         }
 
-        $this->isSessionExists() ? session_regenerate_id(true) : session_start();
+        if ($this->isSessionExists()) {
+            //regenerate the session
+            session_regenerate_id();
+            $new_session_id = session_id();
+
+            $_SESSION['new_session_id'] = $new_session_id;
+
+            // Set destroy timestamp
+            $_SESSION['destroyed'] = time();
+
+            // Write and close current session;
+            session_commit();
+            $oldSession = $_SESSION;   //called after destroy - see destroy!
+            // Start session with new session ID
+            session_id($new_session_id);
+            ini_set('session.use_strict_mode', 0);
+            session_start();
+            ini_set('session.use_strict_mode', 1);
+            $_SESSION = $oldSession;
+            // New session does not need them
+            unset($_SESSION['destroyed']);
+            unset($_SESSION['new_session_id']);
+        } else {
+            session_start();
+        }
         $this->storage->init(isset($_SESSION) ? $_SESSION : []);
 
         if ($this->sessionConfig->getUseCookies()) {
--- a/lib/internal/Magento/Framework/Session/SessionManager.php
+++ b/lib/internal/Magento/Framework/Session/SessionManager.php
@@ -195,7 +195,6 @@ class SessionManager implements SessionManagerInterface
             if (isset($_SESSION['destroyed'])) {
                 if ($_SESSION['destroyed'] < time() - 300) {
                     $this->destroy(['clear_storage' => true]);
-
                 }
             }
             $this->validator->validate($this);
--- a/lib/internal/Magento/Framework/Session/SessionManager.php
+++ b/lib/internal/Magento/Framework/Session/SessionManager.php
@@ -514,18 +514,20 @@ class SessionManager implements SessionManagerInterface
         if ($this->isSessionExists()) {
             //regenerate the session
             session_regenerate_id();
-            $new_session_id = session_id();
+            $newSessionId = session_id();
 
-            $_SESSION['new_session_id'] = $new_session_id;
+            $_SESSION['new_session_id'] = $newSessionId;
 
             // Set destroy timestamp
             $_SESSION['destroyed'] = time();
 
             // Write and close current session;
             session_commit();
-            $oldSession = $_SESSION;   //called after destroy - see destroy!
+            
+            //called after destroy()
+            $oldSession = $_SESSION;
             // Start session with new session ID
-            session_id($new_session_id);
+            session_id($newSessionId);
             ini_set('session.use_strict_mode', 0);
             session_start();
             ini_set('session.use_strict_mode', 1);
--- a/lib/internal/Magento/Framework/Session/SessionManager.php
+++ b/lib/internal/Magento/Framework/Session/SessionManager.php
@@ -5,9 +5,6 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
-
-// @codingStandardsIgnoreFile
-
 namespace Magento\Framework\Session;
 
 use Magento\Framework\Session\Config\ConfigInterface;
@@ -18,6 +15,11 @@ use Magento\Framework\Session\Config\ConfigInterface;
  */
 class SessionManager implements SessionManagerInterface
 {
+   /**
+    * Session destroyed threshold in seconds
+    */
+    const SESSION_DESTROYED_THRESHOLD = 300;
+    
     /**
      * Default options when a call destroy()
      *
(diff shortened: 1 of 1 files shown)


miguelbalparda:
There are a couple of fixes referenced in #12362, some also for 2.2. Have you seen those?

elioermini:
Hi @miguelbalparda not yet, I'll have a deeper look next week. Thanks.

VladimirZaets:
Hi @elioermini , thank you for collaboration.

magento-engcom-team:
@elioermini thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

acetronaut:
Hi @elioermini & @magento-engcom-team

@elioermini Did you make any changes / enhancements in this pull request on the code that i wrote here? https://github.com/magento/magento2/issues/12362

There have been reports i can see of a possible issue with http to https, although i think people should ideally have sitewide SSL, i think it will need some testing to ensure there is no issue here.

@magento-engcom-team - I wrote this code back in november. I have also got this deployed on several high turnover sites - both CE and EE versions and this seems to be running for us with no issue. Clients are no longer complaining - and we have done some extensive hotjar sessions where we can see this issue no longer present for customers, as well as it increasing conversion / reducing dropouts, especially on mobile

One thing i am interested to know though - is we are nearly 6 months after this issue. I can see this issue present on some EE websites online now, i am curious to why there seems to have been lack of acknowlegement on the scale of the issue, and also why the issue seems to have been added to your milestone and then subsequently removed.

Further to this

public function regenerateId() was updated in 2.1.9 and most likely the reason this isnt apparent on all M2 instances

https://github.com/magento/magento2/blob/2.1.8/lib/internal/Magento/Framework/Session/SessionManager.php

https://github.com/magento/magento2/blob/2.1.9/lib/internal/Magento/Framework/Session/SessionManager.php

So a revert to the previous code may resolve the issue - however possibly re-introduce a different bug this was changing anyway. Could you let me know and give me a test case please for what issue you were trying to solve with this update which introduced this issue and why you chose not to roll it back.

The change you made in 2.1.9 was probably for a reason, and if you have a example of a problem the change you made was solving - i would advise that this pull request also needs retesting against this.

Also have noticed some changes a while ago and attempts to make changes, as well as closing some other tickets off in favour of a "Magento Solution" so i am curious whether you can give the magento take on this issue and what has been looked into?

It took me 2 solid (and very long) days to write and test this and i am sure you have some very clever people in there that you could assign this to who should be able to look at this in full and also in respect to scenarios that i wont have tested (for example http to https as all of our clients are now 100% SSL).

As soon as i became aware of this issue it became a showstopper / top priority for us due to its nature of affecting sales. I was hoping to see some progress from your team by now on the issue. Maybe there has been, and if there has could you please make this public. 

I hope you decide to engage a little more on the issue with some technical / developer feedback from the changes i made at least. im sure if you had have done this already a lot of clever people in the community could have helped chip in to solve this issue with you

I look forward to your reply @magento-engcom-team

aeu:
Subscribed

aeu:
Subscribed

elioermini:
Hi @acetronaut yes that's basically the code you added in the comment plus this line to renew the cookie https://github.com/elioermini/magento2/blob/6ee737bb43adcd9b40c170170c5ea8e5d5ad6731/lib/internal/Magento/Framework/Session/SessionManager.php#L201

Thanks a lot for your contribution, it appears to be close to what is being suggested on php.net http://php.net/manual/en/function.session-regenerate-id.php. 
It's working very well on our side.
I'm not sure how to solve the issue of the CI at the moment. Looks like it doesn't like the `php_ini` directive. It would be nice a hint from the core team.

piotrekkaminski:
Recent internal PR https://github.com/magento/magento2ce/pull/2496 resolves the issues for https only by removing regenerate but this PR should resolve it for mixed mode (http/https) when regenerate is still security best practice.

ishakhsuvarov:
Hi @elioermini @acetronaut @piotrekkaminski 

Similar to provided in this Pull Request fix has been evaluated by the core team when investigating this issue. Unfortunately, in some scenarios, specifically when using Redis for session storage it may cause significant performance degradation on some scenarios. @elioermini Have you tried this commit with Redis session storage? 

Current solution, which is already available in `2.2-develop` branch https://github.com/magento/magento2/commit/a6f146cd9302fbbb6305df9613c7ce3eea7a05ae is to eliminate session regeneration for stores using full `https` mode.

We are going to check the specific solution provided in this Pull Request with possible redis issues in mind and let you know the result.

Thank you for collaboration.

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -5,6 +5,9 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+
+// @codingStandardsIgnoreFile
```
ishakhsuvarov: It's not recommended to suppress warnings in all new and updated code. For this specific case I would suggest narrowing down the ignore section to the specific lines, indicated by the test.

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -183,10 +186,20 @@ public function start()
...
+                // Not fully expired yet. Could be lost cookie by unstable network.
+                session_commit();
+                session_id($_SESSION['new_session_id']);
+            }
             $sid = $this->sidResolver->getSid($this);
             // potential custom logic for session id (ex. switching between hosts)
             $this->setSessionId($sid);
             session_start();
+            if (isset($_SESSION['destroyed'])) {
+                if ($_SESSION['destroyed'] < time() - 300) {
```
ishakhsuvarov: Hard coded `300` seems to be a session lifetime. I would suggest replacing it with the configuration value.

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -501,7 +514,32 @@ public function regenerateId()
...
+            // Set destroy timestamp
+            $_SESSION['destroyed'] = time();
+
+            // Write and close current session;
+            session_commit();
+            // Called after destroy()
+            $oldSession = $_SESSION;
+            // Start session with new session ID
+            session_id($newSessionId);
+            ini_set('session.use_strict_mode', 0);
```
ishakhsuvarov: Do you have any examples when manipulating strict mode on the fly is required? I've tested solution briefly with session strict mode enabled and these lines commented and have not seen any issue with this so far.

ishakhsuvarov (changes requested):

lib/internal/Magento/Framework/Session/SessionManager.php
elioermini: Hi @ishakhsuvarov should I move it as const of `\Magento\Framework\Session\SessionManagerInterface` ?

lib/internal/Magento/Framework/Session/SessionManager.php
elioermini: It's because is how is shown in the example 2 here http://php.net/manual/en/function.session-regenerate-id.php. Actually now that I think about the line 532 and 533 should be swapped. However you are right, seems not useful: even with strict_mode enabled I'm able to set the session_id from $newSessionId. Maybe @acetronaut has further thoughts about.

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -15,6 +15,11 @@
  */
 class SessionManager implements SessionManagerInterface
 {
+   /**
+    * Session destroyed threshold in seconds
+    */
+    const SESSION_DESTROYED_THRESHOLD = 300;
```
sidolov: Please, remove constant and use session lifetime value from the system configuration

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -501,7 +516,29 @@ public function regenerateId()
             return $this;
         }
 
-        $this->isSessionExists() ? session_regenerate_id(true) : session_start();
+        // @codingStandardsIgnoreStart
```
sidolov: Why was this code fragment suppressed?

lib/internal/Magento/Framework/Session/SessionManager.php
```diff
@@ -183,10 +188,20 @@ public function start()
...
+                // Not fully expired yet. Could be lost cookie by unstable network.
+                session_commit();
+                session_id($_SESSION['new_session_id']);
+            }
             $sid = $this->sidResolver->getSid($this);
             // potential custom logic for session id (ex. switching between hosts)
             $this->setSessionId($sid);
             session_start();
+            if (isset($_SESSION['destroyed'])) {
+                if ($_SESSION['destroyed'] < time() - self::SESSION_DESTROYED_THRESHOLD) {
```
sidolov: Combine this expression with previous if statement

sidolov (changes requested):

lib/internal/Magento/Framework/Session/SessionManager.php
elioermini: It has been replaced with the if / else block till line 540

aeu:
@acetronaut That was a really good find on the change to regenerateId(), and I agree with you, that this is most likely the change that broke it.

sidolov:
@elioermini , any updates?

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=14973) <br/>All committers have signed the CLA.

sidolov:
Hi @elioermini  , I fixed minor issues. Please, sign CLA, otherwise, we can't process your pull request

elioermini:
Sorry the delay I got side-tracked, can someone help me making the hardcoded values configurable? Thanks

sidolov:
@elioermini , I think we are done with all the requested changes. The solution looks good to me

sidolov (approved):

magento-engcom-team:
[ENGCOM-1440](https://jira.corp.magento.com/browse/ENGCOM-1440) has been created to process this Pull Request

magento-engcom-team:
Hi @elioermini. Thank you for your contribution. 

PR #15100, opened 2018-05-09
[Backport] Fixed issue products grid operations in admin cart price rule edit page for user which has no access to CatalogRule module

sergiy-v:
Original PR: #14886.

## Preconditions
Magento 2.2 have multiple admin user roles with access restrictions to some resources. Role has access to cart price rules, but it restricted to catalog rules rules.

## Problem
When admin user with role mentioned above tried in edit/new cart price rule, in conditions settings add “SKU” condition, and then select products from loaded product grid – all grid navigation, sorting and filtering operations will lead to “403” redirect.  Reason
When magento loaded products grid for selecting product sku for cart price rules it use controller action which is inherited from similar one form CatalogRule module, this controller render its block Magento\CatalogRule\Block\Adminhtml\Promo\Widget\Chooser\Sku (which generate product grid HTML and JavaScript). This block method “getGridUrl” generates urls with route 'catalog_rule/*/chooser', so all requests for updating products grid data will lead to CatalogRule module controller. But current user has no access to CatalogRule routes so it get 403 response.  

## Solution
Modify method Magento\CatalogRule\Block\Adminhtml\Promo\Widget\Chooser\Sku::getGridUrl() method. Change route parameter which it pass in “getUrl” method from “catalog_rule//chooser” to “/*/chooser”.

### DIFF
--- a/app/code/Magento/CatalogRule/Block/Adminhtml/Promo/Widget/Chooser/Sku.php
+++ b/app/code/Magento/CatalogRule/Block/Adminhtml/Promo/Widget/Chooser/Sku.php
@@ -207,7 +207,7 @@ class Sku extends \Magento\Backend\Block\Widget\Grid\Extended
     public function getGridUrl()
     {
         return $this->getUrl(
-            'catalog_rule/*/chooser',
+            '*/*/chooser',
             ['_current' => true, 'current_grid_id' => $this->getId(), 'collapse' => null]
         );
     }


VladimirZaets:
Hi @sergiy-v.
Current PR doesn't fixed problem.

**Steps to reproduce**:
Go To Magento Admin
Create New User Role and grant to it permission for Shopping Cart rules, but NOT for Catalog Price Rules
Create New user and assign to created role
Login to Admin with new user
Go to Shopping Cart rules
Go to Conditions Section and create sub-selection of products with "SKU is:" and try to choose products

**Expected result:**
Product grid is shown

**Actual result:**
Redirecting to admin/sales_rule/ page with 404 Error

sidolov:
@sergiy-v , I am closing this PR now due to inactivity.
Thank you for collaboration!

PR #14342, opened 2018-03-24
Add ability to set custom product option with negative price

mszydlo:
### Description
Enabled ability to set negative prices for custom product options.

### Fixed Issues
1. magento/magento2#7333: Unable to set negative custom option fixed price in admin view

### Manual testing scenarios

1. Create product.
2. Add custom option and set negative value.
3. Save product. For fixed price absolute value of option's price cannot exceed product's special/regular price and if it does - 'Invalid option value' message is shown. For percentage the value of 100 cannot be exceeded.
4. If negative value is in proper range product saves correctly and option is displayed in storefront with proper + or - sign

PR #13393, opened 2018-01-26
Allow negative price for custom option: fixes #7333 in GitHub (internal MAGETWO-60573)

luckyduck:
### Description

The price for custom options should be allowed to be negative. The contained changes do just that. In addition, a cosmetic problem is solved. Once options are allowed to be negative, available options (in a drop-down) add a plus sign as prefix to the price, resulting in '+-1,50 €" for a negative option price (-1,50 in this example). This is fixed as well.

### Fixed Issues (if relevant)

1. magento/magento2#7333: Unable to set negative custom option fixed price in admin view

### Manual testing scenarios

1. Add a custom option to any product.
2. Enter a negative price (for example -3) for a custom option, see #7333 for details.
3. Save the product
4. Visit the product url 
5. Take a look at the Drop-Down (select element). You'll find the option text like "+-3 €" (or dollar)

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

PR #15179, opened 2018-05-12
Fix #7333 Unable to set negative custom option fixed price in admin view.

### LINKED ISSUE #7333: Unable to set negative custom option fixed price in admin view.
In versions 2.1.1 and 2.1.2, setting custom options for a product with fix price settings will not allow a negative value. I'm not sure if this is a simple ui problem or if there is a validator issue. Either way, this was always an option in Magento 1.x in which you can set discounts on an item via custom option. Uploading values via CSV works correctly, however making admin view changes for given product will fail validation via negative values.

### Preconditions

1. Magento 2.1.2
2. PHP 7

### Steps to reproduce

1. Freshly install Magento.
2. Create product with a custom option.
3. Attempt to place a negative value in the custom option price input.

### Expected result

1. Item validates and clients using the custom option allows a price reduction.

### Actual result

1. Cannot save item. Validator fails.

capture

dverkade:
Fixes the ability to set a negative custom option price in the backend. 

### Description
In Magento 1 it was possible to set a negative price change to a custom option. This PR fixes issue #7333 and makes it possible to set a negative price in the backend of Magento. This PR fixes the backend so that the negative price change can actually be entered in the Magento admin. There is a separate PR for the frontend changes which is PR #14975 

There also seem to be two other PR's for this issue which are stale and are not moving along: 
#13393
#14342

### Fixed Issues
1. magento/magento2#7333: Unable to set negative custom option fixed price in admin view.

### Manual testing scenarios
1. Create a new simple product in the backend and try to create new custom options where the price change has a negative value (for instance, "-10").

### DIFF
--- a/app/code/Magento/Catalog/Model/Product/Option/Validator/DefaultValidator.php
+++ b/app/code/Magento/Catalog/Model/Product/Option/Validator/DefaultValidator.php
@@ -132,7 +132,7 @@ class DefaultValidator extends \Magento\Framework\Validator\AbstractValidator
      */
     protected function validateOptionValue(Option $option)
     {
-        return $this->isInRange($option->getPriceType(), $this->priceTypes) && !$this->isNegative($option->getPrice());
+        return $this->isInRange($option->getPriceType(), $this->priceTypes);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dverkade:
Let me check the failing tests.

VladimirZaets:
Hi, @dverkade, thank you for collaboration. If you need some help I will glad to help you.

PR #15216, opened 2018-05-15
Make possible to set NULL values using config:set command

jalogut:
### Description
Currently is not possible to set `null` values using `config:set` command. That is needed, if we want to overwrite `core_config_data` values using `config.php` and `env.php` files.

This PR makes that possible by replacing "NULL" param value with `null` before the value is saved into the config files.

### Manual testing scenarios

1. `bin/magento config:set <path> NULL`
2. I would expect this value set to `null`. However, the current status is that "NULL" string is saved instead.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Config/Console/Command/ConfigSetCommand.php
+++ b/app/code/Magento/Config/Console/Command/ConfigSetCommand.php
@@ -173,9 +173,14 @@ class ConfigSetCommand extends Command
                     $lockTargetPath = ConfigFilePool::APP_CONFIG;
                 }
 
+                $value = $input->getArgument(static::ARG_VALUE);
+                if ($value === 'NULL') {
+                    $value = null;
+                }
+
                 return $this->processorFacadeFactory->create()->processWithLockTarget(
                     $input->getArgument(static::ARG_PATH),
-                    $input->getArgument(static::ARG_VALUE),
+                    $value,
                     $input->getOption(static::OPTION_SCOPE),
                     $input->getOption(static::OPTION_SCOPE_CODE),
                     $lock,


hostep:
What if you want to use the actual word "NULL", that won't be possible anymore? (I realize almost nobody will want to do this, but still ...)

In the n98/magerun2 project, they fixed it with an extra flag `--no-null`, which change the interpretation when you provide `NULL`, either the value `null` or the word "NULL", maybe that's an idea which you can use over here as well?
See: https://github.com/netz98/n98-magerun2/blob/fb75eb2/src/N98/Magento/Command/Config/Store/SetCommand.php#L77-L86

jalogut:
Yes, I know about the flag on n98. It’s just that I do not think anybody will need “NULL” as a string. I also think that it should be avoided, so I did not want to provide this option. If you think that it is needed, I wouldn’t have any problem to add it.

hostep:
I also believe almost nobody will need "NULL" as a string, but still, one day, someone will probably need it for some obscure reason.
Let's wait for someone from the Magento team to give some feedback here to see if the extra flag is needed or not.

orlangur (changes requested):
I don't like that we should use capsed `NULL` here and that string behavior is changed.

Can we have a separate command like `bin/magento config:unset <path>` or `bin/magento config:set-null <path>` maybe?

orlangur:

jalogut:
Hi @orlangur

I’d rather go for the option `—no-null` as @hostep mentioned. Adding a especific command just for null values doesn't seem like the right solution for me. What do you think? 

If you use config propagation, you need to assign `NULL` value to unset config settings. With config propagation files, you configure values but not different actions depending on the value.

joshuaadickerson:
unset seems best. Makes sure nobody will mistake it.

orlangur:
Raised in Slack.

orlangur:
Hi @jalogut, among maintainers the preferred version is `bin/magento config:unset <path>`, as this is a new feature please create a PR against `2.3-develop`.

PR #15251, opened 2018-05-16
[BUGFIX] [issue-13720] Allow more then 2 products with the same posit…

lewisvoncken:
…ion in the Admin for Related Products

### Description

Only 2 related products are showing in backend

1. magento/magento2#13720: Only 2 related products are showing in backend

### Steps to reproduce

1. Update products related products using csv and update via magento default import.
2. Flush the cache storage and reindexing .
3. Check the frontend and backend

### Expected result

1. Updated related products should come in the frontend of product page and backend of product edit page

### Actual result

1. The update related products are showing in the frontend .But only 2 related products are showing in the backend 

### Apply changes
Now the expected result will be the outcome.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/ProductLink/CollectionProvider.php
+++ b/app/code/Magento/Catalog/Model/ProductLink/CollectionProvider.php
@@ -54,13 +54,14 @@ class CollectionProvider
         }
 
         foreach ($output as $item) {
-            $itemPosition = $item['position'];
-            if (!isset($sorterItems[$itemPosition])) {
-                $sorterItems[$itemPosition] = $item;
-            } else {
-                $newPosition = $itemPosition + 1;
-                $sorterItems[$newPosition] = $item;
+            $itemPosition = (int)$item['position'];
+            while (true) {
+                if (!isset($sorterItems[$itemPosition])) {
+                    break;
+                }
+                $itemPosition += 1;
             }
+            $sorterItems[$itemPosition] = $item;
         }
         ksort($sorterItems);
         return $sorterItems;


VladimirZaets:
Hi @lewisvoncken, thank you for collaboration.

I think to use `while(true)` it's not a better solution. Maybe you can find another solution.
Anyway, I think to use the recursive function in the current case looks better. Are you agree?

Can you please cover your changes by tests?

robbertstevens:
@VladimirZaets i agree the solutions isn't the best, but atleast it works now.

VladimirZaets:
@robbertstevens can you refactor this code uses recursive function instead of `while(true)`

robbertstevens:
@VladimirZaets no i cant, and what do you mean you think this is not a better solution? Now it doesnt even work. With his solutions atleast the feature works. 

It's like going to the garage and you need some air in your tires, but the electric pump is broken so you need to do in manually, and you say well what ever, ill drive with empty tires

VladimirZaets:
@robbertstevens we try to not introduce new not reliable code. This code is not reliable, it can introduce infinite loop and other bugs. I provided information how to refactor this code and  this fix should be covered by tests.

Thank you for collaboration

orlangur:
@lewisvoncken I don't think there is any use for recursion here and I don't think a loop is infinite sometimes, however, old implementation was `O(n)` and your one is `O(n*n)` which needs to be fixed.

n0kit31:
How come this bug is closed? Issue still happens in 2.2.5!

hostep:
I've opened a new PR using a different way for fixing this, which will hopefully get accepted by the Magento devs: https://github.com/magento/magento2/pull/17885

Issue #5326, opened 2016-06-28
Can't Use a Symlinked Module During Development


astorm:
Steps to reproduce
1. Create a simple module in `app/code` using the instructions below, or using the [attached zip file](https://github.com/magento/magento2/files/339488/Pulsestorm_Symlinkbug.zip) (or via your own means, adjusting the URLs mentioned below)
2. Confirm module is created and installed by accessing URL `http://example.magento.com/pulsestorm_symlinkbug`
3. Navigate to `Stores -> Configuration -> Developer -> Template` and set "Allow Symlinks" to `1`
4. Move the `app/code/Pulsestorm/Symlinkbug` folder to _a place outside the Magento root_
5. Create a symlink at `app/code/Pulsestorm/Symlinkbug` which points to _a place outside the Magento root_ (ex. `~/Documents/magento-modules/Symlinkbug`)
6. Access the URL at `http://example.magento.com/pulsestorm_symlinkbug` again
## Expected result: 
1. Page loads identically both times
## Actual result
1. The controller action endpoint responds, but without rendering the `phtml` template
## Context

This bug impacts developers who use symlinks in their deployment setups, and in their development workflows.  Magento added the "Allow Symlinks" setting explicitly to enable these workflows.  The problem appears to be the code that validates the templates in 

```
#File: vendor/magento/framework/View/Element/Template/File/Validator.php
public function isValid($filename)
{
    $filename = str_replace('\\', '/', $filename);
    if (!isset($this->_templatesValidationResults[$filename])) {
        $this->_templatesValidationResults[$filename] =
            ($this->isPathInDirectories($filename, $this->_compiledDir)
                || $this->isPathInDirectories($filename, $this->moduleDirs)
                || $this->isPathInDirectories($filename, $this->_themesDir)
                || $this->_isAllowSymlinks)
            && $this->getRootDirectory()->isFile($this->getRootDirectory()->getRelativePath($filename));
    }

    return $this->_templatesValidationResults[$filename];
}
```

The `$this->getRootDirectory()->isFile($this->getRootDirectory()->getRelativePath($filename))` check fails -- and confusingly based on the names, seems to be a check that explicitly runs counter to the `$this->_isAllowSymlinks` check. 

How to handle this is a policy decision that's best left to the framework/platform owner, but removing the ability to symlink components in Magento 2 runs counter to an early stated goal of Magento 2 (formalizing components so they could run anywhere), and regardless of current goals is hostile to developer experience. 
## Module Creation Instructions

[...]

PR #15252, opened 2018-05-16
Fix to have possibility use symlinked modules outside magento root dir

likemusic:
### Description

During magento extention development we need to test module in different magento versions. In this case its very useful to make symlink to developing module in many magento hosts and take module source in one place (with on opened and configured IDE).

Currently if symlinked module source placed outside magento root directory then we get exception like this:

```
Exception #0 (Magento\Framework\Exception\ValidatorException): Invalid template file: '/home/httpd/htdocs/users/ivashchenko/puq220ee/app/code/Aitoc/ProductUnitsAndQuantities/view/adminhtml/templates/system/config/use_quantities_validation.phtml' in module: 'Magento_Backend' block's name: 'js_schedule_block2'
```

1. magento/magento2#5326: Can't Use a Symlinked Module During Development

### Manual testing scenarios

1. Place extension that have custom templates outside magento root directory.
2. Make symlink to `app/code/{Vendor}/{ExtentionName}`  directory.
3. Use extension.

### Contribution checklist

### DIFF
--- a/lib/internal/Magento/Framework/View/Element/Template/File/Validator.php
+++ b/lib/internal/Magento/Framework/View/Element/Template/File/Validator.php
@@ -34,6 +34,11 @@ class Validator
     protected $_filesystem;
 
     /**
+     * @var \Magento\Framework\Filesystem\Io\File
+     */
+    protected $_file;
+
+    /**
      * Allow symlinks flag
      *
      * @var bool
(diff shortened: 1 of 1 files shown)


orlangur:
Same as https://github.com/magento/magento2/issues/13375#issuecomment-431146467 - paths outside of Magento base path should not be accessible for security reasons.

PR #15279, opened 2018-05-17
[fix] typo in method name _isCategoryMov[e]able

mhauri:
### Description
Method name `\Magento\Catalog\Block\Adminhtml\Category::_isCategoryMoveable` contained typo.

Renamed it to `_isCategoryMovable`.

### Contribution checklist


app/code/Magento/Catalog/Block/Adminhtml/Category/Tree.php
```diff
@@ -393,7 +393,7 @@ public function buildNodeName($node)
      * @param Node|array $node
      * @return bool
      */
-    protected function _isCategoryMoveable($node)
+    protected function _isCategoryMovable($node)
```
orlangur, line 396: Protected methods cannot be removed, please check https://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/index.html#php:
> Public and protected method removal 
> Mark the method with the `@deprecated` tag instead of removing it.

orlangur (changes requested):

app/code/Magento/Catalog/Block/Adminhtml/Category/Tree.php
mhauri: Does this mean I should add a deprecated tag to the method and make a Forwardport PR for 2.3 with renaming the method? Or how does that work @orlangur ?

Jeroenll: Mark the old method as deprecated and proxy that call to the new method.
```PHP
protected function _isCategoryMoveable($node)
{
    $this-> _isCategoryMovable($node)
}

protected function _isCategoryMovable($node)
{
    // Logic
}
```

mhauri:
I will create a new PR for this!

app/code/Magento/Catalog/Block/Adminhtml/Category/Tree.php
orlangur: @mhauri the same for 2.3, deprecated methods will be removed at some point but not so quickly. General rule is "if something is deprecated in `2.X.0`, it can be removed in `2.X+1.0`" if I remember correctly.

PR #15287, opened 2018-05-17
[Backport] Handle empty or incorrect lines in a language CSV

VitaliyBoyko:
Original PR: https://github.com/magento/magento2/pull/12304

**Description:**
Reading language packs will break if the CSV file contains incorrect lines. For example an accidental empty line at the end of the file will cause an undefined index error.


DanielRuf (approved):

hostep:
The original PR seems to have one extra newline in the csv file. Or is it just github's rendering that is confusing?

DanielRuf:
> The original PR seems to have one extra newline in the csv file. Or is it just github's rendering that is confusing?

This is the line ending.

hostep:
Yes, I know, but I mean this:

original:
original

new: 
new

Unless I'm blind, there is an extra line in the original PR, right?

DanielRuf:
@hostep you are right, probably different EOL settings or just forgotten? Should be probably added here too as files should end with an empty line normally (this is alo what VSCode automatically does).

hostep:
I actually think there are two empty lines, since github doesn't render the last empty line, hence why they use that symbol for indicating there is no empty line at the end. So in this case, there is no symbol and we see an empty line, so I assume there are 2 empty lines.

@VitaliyBoyko: it's probably best if you can exactly copy that file from the original PR including the exact same number of empty lines, thanks!

magento-engcom-team:
Hi @VitaliyBoyko. Thank you for your contribution. 

PR #15288, opened 2018-05-17
[Backport] Fix typo in single product layout hanlde

VitaliyBoyko:
Original PR: #12786

### Description
The constant SINGLE_PRODUCT_LAYOUT_HANLDE contains a typo. It should be SINGLE_PRODUCT_LAYOUT_HANDLE.

### DIFF
--- a/app/code/Magento/Widget/Model/Widget/Instance.php
+++ b/app/code/Magento/Widget/Model/Widget/Instance.php
@@ -30,6 +30,11 @@ class Instance extends \Magento\Framework\Model\AbstractModel
 
     const PRODUCT_LAYOUT_HANDLE = 'catalog_product_view';
 
+    /**
+     * @deprecated
+     */
+    const SINGLE_PRODUCT_LAYOUT_HANLDE = 'catalog_product_view_id_{{ID}}';
+
     const SINGLE_PRODUCT_LAYOUT_HANDLE = 'catalog_product_view_id_{{ID}}';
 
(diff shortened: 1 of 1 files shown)


DanielRuf (approved):

app/code/Magento/Widget/Model/Widget/Instance.php
```diff
@@ -30,7 +30,7 @@ class Instance extends \Magento\Framework\Model\AbstractModel
 
     const PRODUCT_LAYOUT_HANDLE = 'catalog_product_view';
 
-    const SINGLE_PRODUCT_LAYOUT_HANLDE = 'catalog_product_view_id_{{ID}}';
+    const SINGLE_PRODUCT_LAYOUT_HANDLE = 'catalog_product_view_id_{{ID}}';
```
sidolov: Constant removing is backward incompatible, please, fix issue according to [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.1/contributor-guide/backward-compatible-development/index.html)

sidolov (changes requested):

app/code/Magento/Widget/Model/Widget/Instance.php
VitaliyBoyko: Hi @sidolov 
Do you mean to mark the constant with typo deprecated and not delete it? Then just declare correct one.

sidolov: Yes, you are right

PR #15304, opened 2018-05-18
Fix for issue-6578

bartoszherba:
Fix for #6578 
    - add LinkTokenManager class responsible for modification of RpToken on customer entity
    - add LinkTokenManager unit tests
    - udpdate relevant unit/integrations tests for AccountManagement class
    - mark AccountManagement::changeResetPasswordLinkToken() method as deprecated and delegate responsibility to token manager

### Description
On password reset request we should care only about preserving relevant data. Validation of whole model with all its relation is not our interest here.
The main problem is that the customer knows to much and some of its data as tokens should be a separate class. We can't do that easily. Therefore I created token manager class to take care of setting new token on the customer without using repository which leads to extended and unnecessary logic execution.

### Fixed Issues
Now customer's address data are irrelevant with regards of password reset request.

### Manual testing scenarios
1. Register a customer and log in
2. Create an address with empty phone number
3. Log out
4. Go to login view and try to reset your password

As a result you should be able to receive email with further instructions as expected.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Api/LinkTokenManagerInterface.php
+++ b/app/code/Magento/Customer/Api/LinkTokenManagerInterface.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * @package  Magento\Customer
+ * @author Bartosz Herba <bherba@divante.pl>
+ * @copyright 2017 Divante Sp. z o.o.
+ * @license See LICENSE_DIVANTE.txt for license details.
+ */
+
+namespace Magento\Customer\Api;
+
+use Magento\Customer\Api\Data\CustomerInterface;
+use Magento\Framework\Exception\InputException;
+
+/**
+ * Interface LinkTokenManagerInterface
+ */
+interface LinkTokenManagerInterface
+{
+    /**
+     * Change and store new reset password link token
+     *
+     * @param CustomerInterface $customerData
+     * @param string $passwordLinkToken
+     *
+     * @throws InputException
+     *
+     * @return bool
+     */
+    public function changeToken(CustomerInterface $customerData, string $passwordLinkToken): bool;
+}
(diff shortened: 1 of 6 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15304) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=15304) before we can accept your contribution.<br/>

orlangur:
> Now in case that for some reasons, like after data migration from M1, telephone field is empty

Phone field is required in both M1 and M2. If migration produces invalid data, it should be fixed instead of hacking core.

Making phone field not required, on the other hand, seems a huge backward-incompatible change for all assuming it was required.

> the real, underlying exception message is hidden, in our case, it was: "Phone" is a required value. "Phone" length must be equal or greater than 1 characters.. as a fix, full field validation must be ignored during password resets.

Please rework a fix in such a way that changing phone is_required flag is not needed.

bartoszherba:
@orlangur 
I changed the approach to the issue. Please, evaluate it again.

orlangur (changes requested):

orlangur:
Thanks for collaboration!

jordipujol-mcneel:
The issue is still present in 2.3
It happens with any required customer field.
The solution proposed in https://github.com/magento/magento2/pull/15304/commits/e6050641e00b77b96a5fe3de03edbb9644f7dbb5 does not work because the exception fires on this line:
$this->customerResource->save($customer);
when trying to save token.

Regards,
JOINSO

PR #15538, opened 2018-05-26
Add the Symfony VarDumper component to allow the use of `dump()`

jacquesbh:
Allow the use of `dump()` in our code.

### Description

It's always cool as a developer of being able to debug some variables.
In M1 we used `Zend_Debug::dump()` and in M2 I wanted to use the `dump()` function from the Symfony VarDumper Component, because it's great.

So I added the dependency into the Developer module as a dev dependency. That way it's possible to not install it with the `--no-dev` option in composer.

var-dump

### Manual testing scenarios

Just try `dump("foo");` or `dump($object);` in your code.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Developer/composer.json
+++ b/app/code/Magento/Developer/composer.json
@@ -8,7 +8,8 @@
         "php": "~7.1.3||~7.2.0",
         "magento/framework": "*",
         "magento/module-config": "*",
-        "magento/module-store": "*"
+        "magento/module-store": "*",
+        "symfony/var-dumper": "^4.0"
     },
     "type": "magento2-module",
     "license": [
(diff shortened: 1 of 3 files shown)


app/code/Magento/Developer/composer.json
```diff
@@ -10,6 +10,9 @@
         "magento/module-config": "*",
         "magento/module-store": "*"
     },
+    "require-dev": {
```
vkublytskyi: This is won't have any effect as `require-dev` section used only for root composer package  so it will be ignored for git based installations and for composer based installations

jacquesbh: Hum. Good point. I'll fix that.

jacquesbh:
@vkublytskyi I made the change.

The addition of the PHP 7.2 Polyfill by the component is breaking the tests in PHP 7.2. So, I have to take a look at it before merging obviously.

orlangur (changes requested):
Changes must be applied to `2.2-develop` first.

Simply add it to `require-dev` section of root `composer.json`.

orlangur:

PR #15240, opened 2018-05-16
Remove unused namespace from ui export

yuriyDne:
### Description
The Export model have some unused namespaces: `Magento\Framework\Filesystem\Directory\WriteInterface.`
Just removing it and update description

### Manual testing scenarios
1. For example, go to orders list;
2. Export list of order by XML or CSV;
3. All should export as expected.

### Contribution checklist
- [x] Pull request has a meaningful description of its purpose
- [x] All commits are accompanied by meaningful commit messages
- [x]  All new or changed code is covered with unit/integration tests (if applicable)
- [x]   All automated tests passed successfully (all builds on Travis CI are green)

PR #15611, opened 2018-05-30
[Forwardport] Remove unused namespace from ui export

dmytro-ch:
### Original Pull Request 
 https://github.com/magento/magento2/pull/15240

### Description
The Export model have some unused namespaces: `Magento\Framework\Filesystem\Directory\WriteInterface.`
Just removing it and update description

### Manual testing scenarios
1. For example, go to orders list;
2. Export list of order by XML or CSV;
3. All should export as expected.

### Contribution checklist
- [x]  All new or changed code is covered with unit/integration tests (if applicable)
- [x]   All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Ui/Model/Export/ConvertToCsv.php
+++ b/app/code/Magento/Ui/Model/Export/ConvertToCsv.php
@@ -8,7 +8,6 @@ namespace Magento\Ui\Model\Export;
 use Magento\Framework\App\Filesystem\DirectoryList;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Filesystem;
-use Magento\Framework\Filesystem\Directory\WriteInterface;
 use Magento\Ui\Component\MassAction\Filter;
 
 /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/Ui/Model/Export/ConvertToCsv.php
```diff
@@ -17,7 +16,7 @@
 class ConvertToCsv
 {
     /**
-     * @var WriteInterface
+     * @var DirectoryList
```
rogyar, line 19: Seems like there's some mistake. In the constructor we init `$directory` property as a result of $filesystem->getDirectoryWrite() method, which creates an instance of Magento\Framework\Filesystem\Directory\Write using the corresponding factory. So, basically, it returns `WriteInterface` but not `DirectoryList`. The same situation is for the `ConvertToXml` class.
Please, correct me if I'm wrong.

rogyar (changes requested):

dmytro-ch:
@rogyar thank you for notice. You are right. We should close this pull request.

PR #15623, opened 2018-05-31
Fix translations

cherednichenkoa:
Fix unsupported translate tags in mini cart content 
### Description

Minicart template has several usages of `<tranlsate />` tags , however according to this 
Config (that has inside defined preg expressions for translator template parser)
```
            <argument name="patterns" xsi:type="array">
                <item name="i18n_translation" xsi:type="string"><![CDATA[~i18n\:\s*(["'])(.*?)(?<!\\)\1~]]></item>
                <item name="translate_wrapping" xsi:type="string"><![CDATA[~translate\=("')([^\'].*?)\'\"~]]></item>
                <item name="mage_translation_widget" xsi:type="string"><![CDATA[~(?:\$|jQuery)\.mage\.__\((?s)[^'"]*?(['"])(.+?)(?<!\\)\1(?s).*?\)~]]></item>
                <item name="mage_translation_static" xsi:type="string"><![CDATA[~\$t\((?s)[^'"]*?(["'])(.+?)\1(?s).*?\)~]]></item>
            </argument>
```
Magento supports such definition of translate attribute` <span translate="'Text to translate'">` but not the tag <translate>,
this tag will be translated without problem on frontend, if such translation has additiona properl definitions 
When magento initialize translations for FE, and put it in Magento\Translation\Model\Js\Config::DICTIONARY_FILE_NAME file , it parses templates to get proper set of translations, in case template has <tranlsate /> definition, it will not match preg's, so it will not be translated on frontend

1. "Item in cart" translation does not work 
Magento has around 20 definitions of` <template />` tag, so it does not make sense to add additional preg to create additional load for template parser, only for 20 templates

### Manual testing scenarios

1. Create additional locale 
2. Add own label + translation in locale csv (for exaple "My custom text")
3. And this text to KO template via "`<translate />`" tag 
4. This translate will not not work 

### Contribution checklist

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/template/minicart/content.html
+++ b/app/code/Magento/Checkout/view/frontend/web/template/minicart/content.html
@@ -30,8 +30,12 @@
             <span class="count" if="maxItemsToDisplay < getCartLineItemsCount()" text="maxItemsToDisplay"/>
             <translate args="'of'" if="maxItemsToDisplay < getCartLineItemsCount()"/>
             <span class="count" text="getCartLineItemsCount()"/>
-            <translate args="'Item in Cart'" if="getCartLineItemsCount() === 1"/>
-            <translate args="'Items in Cart'" if="getCartLineItemsCount() > 1"/>
+                <!-- ko if: (getCartLineItemsCount() === 1) -->
+                    <span translate="'Item in Cart'"/>
+                <!--/ko-->
+                <!-- ko if: (getCartLineItemsCount() > 1) -->
+                    <span translate="'Items in Cart'"/>
+                <!--/ko-->
         </div>
 
         <each args="getRegion('subtotalContainer')" render=""/>


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15623) <br/>All committers have signed the CLA.

VladimirZaets:
@cherednichenkoa Yes, JS will not translate phrases that don't exist in CSV file, but after command `bin/magento i18n:collect-phrases` all phrases that should be translated will be added to CSV file. I'm not sure that your fix can affect it, I'm investigating it now.

cherednichenkoa:
@VladimirZaets , I am debugging it now

cherednichenkoa:
@VladimirZaets I quick went thru debugger , and here is what I have found 

Take a look at this file - this cli command uses it to parse html templates (it is our case)
https://github.com/magento/magento2/blob/2.2-develop/setup/src/Magento/Setup/Module/I18n/Parser/Adapter/Html.php

This is preg match used to parse .html templates for translation, as you can see, it does not cover all cases that magento uses on frontend translates 
` 
 /**
     * Covers
     * <span></span>
     * <th class="col col-method" data-bind="i18n: 'Select Method'"></th>
     */
    const HTML_FILTER = "/i18n:\s?'(?<value>[^'\\\\]*(?:\\\\.[^'\\\\]*)*)'/i";

`
You can try to parse content of file I have modified with this preg match, it will not find anything 


Here is list of patters that magento uses to build js-tranlations.json for frontend , as you may see it has more types of translates it covers 
`    
        <argument name="patterns" xsi:type="array">
                <item name="i18n_translation" xsi:type="string"><![CDATA[~i18n\:\s*(["'])(.*?)(?<!\\)\1~]]></item>
                <item name="translate_wrapping" xsi:type="string"><![CDATA[~translate\=("')([^\'].*?)\'\"~]]></item>
                <item name="mage_translation_widget" xsi:type="string"><![CDATA[~(?:\$|jQuery)\.mage\.__\((?s)[^'"]*?(['"])(.+?)(?<!\\)\1(?s).*?\)~]]></item>
                <item name="mage_translation_static" xsi:type="string"><![CDATA[~\$t\((?s)[^'"]*?(["'])(.+?)\1(?s).*?\)~]]></item>
            </argument>
`


I think it is problem with different flow for this command and Magento js-translations.json logic,we can add more pattens thru di.xml to this parser, what do you think ?

I will double check it later, that what I have found now, hope it will help you

VladimirZaets:
@cherednichenkoa I think they have a different responsibility, before adding new patterns we should fully understand of use cases. 
Anyway, it exceeds the scope of the current task.

cherednichenkoa:
@VladimirZaets okay, you are trying to test my change with this console command, I have checked why it works so, what are the our further steps ? Not sure that I am clear what is wrong for now

VladimirZaets:
@cherednichenkoa I will merge your PR. The fix looks fine. Also, I will investigate the reason why CLI command works incorrectly. 
Will be great if you try to investigate it and will create new one PR with the solution for this issue.

cherednichenkoa:
@VladimirZaets , ok got it, I will check it for sure

magento-engcom-team:
Hi @cherednichenkoa. Thank you for your contribution. 

PR #15687, opened 2018-06-02
Fixes for #15393

simonjanguapa:
Issue fix for: https://github.com/magento/magento2/issues/15393
1. Fixed the check for allowed HTML tags OR wysiwyg for textarea attribute types.
2. Fix for options so that they can be rendered as HTML as well.

### DIFF
--- a/dev/tests/integration/testsuite/Magento/Catalog/Model/ProductGettersTest.php
+++ b/dev/tests/integration/testsuite/Magento/Catalog/Model/ProductGettersTest.php
@@ -217,7 +217,7 @@ class ProductGettersTest extends \PHPUnit\Framework\TestCase
         $expected = [
             'Option 2',
             'Option 3',
-            'Option 4 &quot;!@#$%^&amp;*'
+            'Option 4 "!@#$%^&*'
         ];
         self::assertEquals(
             $expected,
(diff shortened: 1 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15687) <br/>All committers have signed the CLA.

dev/tests/integration/testsuite/Magento/Catalog/Model/ProductGettersTest.php
```diff
@@ -217,7 +217,7 @@ public function testGetAttributeTextArray()
         $expected = [
             'Option 2',
             'Option 3',
-            'Option 4 &quot;!@#$%^&amp;*'
```
ishakhsuvarov, line 220: It would be a good idea to cover more cases with the test. Possibly combinations of plain/html setting for attribue

ishakhsuvarov:
Hi @simonjanguapa Thank you for contribution.
We are experiencing troubles with the verification of the initial case, which has been fixed, could you please guide us on correct steps?

simonjanguapa:
- create a dropdown / multiselect attribute
- make sure one or more of the values contain html tags (< strong >value< /strong > for instance, without spaces ofcourse)
- make sure "allow HTML on frontend" and "use in product listing" are set to "Yes"
- add this new attribute to an attribute set
- create a product based on this attribute set and select at least one of the values containing HTML tags
- look up the product on the frontend. You'll see < strong >value< /strong > without the fix, but something like **value** with this fix.

kedarjoyner:
@ishakhsuvarov It looks like this PR has 3 successful checks. Any idea of when this might get pulled into the project? My team is also looking for this change. Thank you for your help!

magento-engcom-team:
Hi @simonjanguapa. Thank you for your contribution. 

PR #15721, opened 2018-06-03
[Forwardport] Fix Magento_ImportExport not supporting unicode characters in column names

tdgroot:
### Original Pull Request 
 https://github.com/magento/magento2/pull/15197
### Description
Column names like `vitamin_a_µg` were being marked invalid.

### Manual testing scenarios

1. Import a catalog product csv with a column like `vitamin_a_µg`

### Contribution checklist

### DIFF
--- a/app/code/Magento/ImportExport/Model/Import/AbstractEntity.php
+++ b/app/code/Magento/ImportExport/Model/Import/AbstractEntity.php
@@ -831,7 +831,7 @@ abstract class AbstractEntity
                 if (!$this->isAttributeParticular($columnName)) {
                     if (trim($columnName) == '') {
                         $emptyHeaderColumns[] = $columnNumber;
-                    } elseif (!preg_match('/^[a-z][a-z0-9_]*$/', $columnName)) {
+                    } elseif (!preg_match('/^[a-z][\w]*$/u', $columnName)) {
                         $invalidColumns[] = $columnName;
                     } elseif ($this->needColumnCheck && !in_array($columnName, $this->getValidColumnNames())) {
                         $invalidAttributes[] = $columnName;
--- a/app/code/Magento/ImportExport/Model/Import/Entity/AbstractEntity.php
+++ b/app/code/Magento/ImportExport/Model/Import/Entity/AbstractEntity.php
@@ -803,7 +803,7 @@ abstract class AbstractEntity
                     if (!$this->isAttributeParticular($columnName)) {
                         if (trim($columnName) == '') {
                             $emptyHeaderColumns[] = $columnNumber;
-                        } elseif (!preg_match('/^[a-z][a-z0-9_]*$/', $columnName)) {
+                        } elseif (!preg_match('/^[a-z][\w]*$/u', $columnName)) {
                             $invalidColumns[] = $columnName;
                         } elseif ($this->needColumnCheck && !in_array($columnName, $this->getValidColumnNames())) {
                             $invalidAttributes[] = $columnName;


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15721) <br/>All committers have signed the CLA.

app/code/Magento/ImportExport/Model/Import/AbstractEntity.php
```diff
@@ -831,7 +831,7 @@ public function validateData()
                 if (!$this->isAttributeParticular($columnName)) {
                     if (trim($columnName) == '') {
                         $emptyHeaderColumns[] = $columnNumber;
-                    } elseif (!preg_match('/^[a-z][a-z0-9_]*$/', $columnName)) {
+                    } elseif (!preg_match('/^[a-z][\w]*$/u', $columnName)) {
```
orlangur, line 834: Why first character is supposed to be a Latin letter?

orlangur (changes requested):

app/code/Magento/ImportExport/Model/Import/AbstractEntity.php
tdgroot: I'd say that a regex like `'/^[\w]+$/u'` would work perfectly fine as well, but I wanted to keep old behavior for whatever reason it was designed this way. Should I change it to `'/^[\w]+$/u'`?

orlangur: Yes. So that name like `µg_vitamin_a_µg` is also supported. Not sure it can start from digit or underscore though, could you check attribute validation when you create/update it via Admin UI please?

tdgroot: Sure thing. I'll get back to you.

tdgroot: The attribute code validation when saving happens right here: [Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php#L161](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Catalog/Controller/Adminhtml/Product/Attribute/Save.php#L161). The regex seems to be `^[a-z\x{600}-\x{6FF}][a-z\x{600}-\x{6FF}_0-9]{0,30}$`. 

It starts to look like we need a validation class for attribute codes in general, so we can unify this stuff. But that's a story for another time. 

@orlangur What's your suggestion for now?

orlangur: @tdgroot ok, so it's not "Magento_ImportExport not supporting unicode characters in column names" actually but such attribute code are generally forbidden. I've raised an internal discussion to understand the correct approach.

tdgroot: @orlangur sure, I'd like to add something to it. I created the attribute in my example using the Magento API methods. To me it seems that validation of the attribute code shouldn't happen in a controller, but rather in the EAV model/repository class.

orlangur:
Checking with @VladimirZaets and @sidolov on this.

sivaschenko:
@orlangur , @tdgroot as there were no updates on attribute code format I will accept this pull request as a port from 2.2 to syncranize the validation between 2.2 and 2.3. 

Please open a nwe pull request to fix the first letter attribute code validation in case there will be any result of technical discussion. Also, it might be useful to reach out to architeture team opening a pull request to https://github.com/magento/architecture with the proposal.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3439](https://jira.corp.magento.com/browse/ENGCOM-3439) has been created to process this Pull Request

orlangur:
@sivaschenko please do not change PR status without communication with others, especially when it is `on hold`.

This PR must be rejected according to discussion in maintainers channel and original one must be reverted for both `2.2=develop` and `2.1-develop`.

Issue #13153, opened 2018-01-12
Credit memo for orders with zero TotalPaid


klakovsky:
### Preconditions

1. Magento 2.3-develop
2. Create order with zero Amount (all Grand Total, Total Paid, Total Refunded, Total Due, Subtotal, Shipping & Handling, Tax are zero ) in status = "Pending" and 1 SKU qty = 1.

### Steps to reproduce
1. First Set tax calculation setting according to this screenshot
![57360176-14df4180-71ac-11e9-874b-6dcdd2a1f1d1](https://user-images.githubusercontent.com/50043226/57431130-642a7e00-724f-11e9-8cba-cfe0f1e314c0.png)
2. then Create a Cart price rule.
![Screenshot (5)](https://user-images.githubusercontent.com/50043226/57431290-f03ca580-724f-11e9-9c99-b7015d0d1eeb.png)
=> Total discount: $104.00
3. Tax percentage: 7% (all product prices on our store include tax, and our flat rate shipping fee includes tax as well)
4. Create a bundle product according to this screenshot
![2019-05-09-11-36-i-13153-2-3-develop instances magento-community engineering](https://user-images.githubusercontent.com/50043226/57431453-7f49bd80-7250-11e9-81fd-814b40720bbf.png)
5. Go to frontend, Add bundle product into the cart 
![Screenshot (6)](https://user-images.githubusercontent.com/50043226/57431702-3e05dd80-7251-11e9-87d6-6967d3f24375.png)
6. Go to checkout
7. Continue as guest or as authenticated user and place order
8. Create Invoice and now Order is status = "Processing"
9. Now create Credit memo and refund and return product
### Expected result
1. Credit memo's grand total should be zero.
![Screenshot (8)](https://user-images.githubusercontent.com/50043226/57432276-05ff9a00-7253-11e9-869e-6b66db1531e1.png)

### Actual result
1. I'm seeing a negative dollar value in the credit memo
![Screenshot (7)](https://user-images.githubusercontent.com/50043226/57432092-7bb73600-7252-11e9-8538-4e00524397c8.png)

-----------------------------------------------------------------------------------------------------------------------------------------

magento-engcom-team:
@klakovskiy, thank you for your report.
We've acknowledged the issue and added to our backlog.

viral-wagento:
@klakovskiy 

I think, Once the invoice/shipment is generated, you can not cancel order, only option available is to refund that order, if customer do not want to proceed for ordered items.

[...]

PR #15742, opened 2018-06-04
Changes for : enable creditmemo feature for zero order total.

viral-wagento:
### Description

When customer place an order of zero ordertotal using Zero Subtotal Checkout Payment method and Admin create an invoice - order goes into Processing. After we can't cancel order. 

At such scenario, requirement comes to create creditmemo and return ordered item back in stock.
Magento doesn't allow to create a creditmemo for zero order total. So creating credit memo is required to return ordered items back in stock.

1. magento/magento2#13153 : Credit memo for orders with zero TotalPaid

### Manual testing scenarios

1. Place an order of Zero amount using Zero Subtotal Checkout Payment method.
2. Create an invoice for that order.
3. As per Magento behaviour, we can't see Credit Memo Option for Zero Total Order.
4. But Now, you can see 'Credit Memo' option, create credit memo for that order and return item back to stock.
5. We can check same steps for Order of amount greater than zero, It will work proper for them too.

### Actual Result
For Zero Order Total, we can't see Credit Memo Option Available, and so we are unable to generate it.
actual-result

### Expected Result
One should be able to generate Credit Memo even for Zero OrderTotal. 
expected-result

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15742) <br/>All committers have signed the CLA.

Issue #15245, opened 2018-05-16
2.2.4: Wrong home page loaded in multi store setup 


jokeputs:
### Preconditions
1. Magento 2.2.4
2. Stores > Configuration > General > Web > Url Options > Add Store Code to Urls: Yes
3. Add two websites with each a store view, for example:
a. An international website with a store view with store code 'international'. This is your default website.
b. A website for Belgium with a store view with store code 'nl_be'.
4. All websites use the same base url, for example: http://demo.shop.com/
5. Create two CMS pages:
a. A page that will be the home page for the international store view. 
b. A page that will be the home page for the nl_be store view.  
6. Configure the home pages: Stores > Configuration > General > Web > Default Pages > CMS Home Page:
a. Use the international home page as your default configuration
b. Set the nl_be home page for the Belgian website

### Steps to reproduce
1. Visit the nl_be store view: http://demo.shop.com/nl_be/

### Expected result
1. The home page for nl_be will be loaded.

### Actual result
1. The home page for international is loaded.

### Possible cause
I've already done some debugging and this commit seems to cause the issue: c18e36b804fb3c86caa39d9976ce5db7497001a7. 

Because `Magento\Framework\Session\Config\ConfigInterface` has been added as a constructor argument to `Magento\Framework\App\Response\Http`, `Magento\Framework\Session\Config` will be created during the `$bootstrap->createApplication()` part of `pub/index.php`:
````
$bootstrap = \Magento\Framework\App\Bootstrap::create(BP, $params);
/** @var \Magento\Framework\App\Http $app */
$app = $bootstrap->createApplication(\Magento\Framework\App\Http::class);
$bootstrap->run($app);
````

Next `Magento\Framework\Session\Config` in its own constructor will call `Magento\Framework\App\Config\ScopeConfigInterface` to get a configuration value. But the current store has not been determined yet at this point. This will only happen in the `$bootstrap->run()` part of `pub/index.php`. 

Magento will fallback to the default store and this will cause the `$resolvedScopeCodes` array in `Magento\Framework\App\Config\ScopeCodeResolver` to have a wrong value for key `null`. Which in turn will cause the wrong homepage being loaded for the nl_be store view.

I'm willing to make a pull request for this but I don't see a good solution.

callumstar:
Hi @jokeputs,

Did you find a temporary workaround for this issue? I'm having a similar issue where the Locale options aren't being set for multi-stores due to this also (see #15205).

jokeputs:
[...]

Issue #15205, opened 2018-05-14
Upgraded to Magento 2.2.4 from Magento 2.1.9 - Locale and Store Configuration issues 'Store View' Locale not being used on frontend, 'Default Config' Locale being used instead  


callumstar:
Magento multi-store installations aren't using the store view specific values from the Store Configuration if they have been set to be different to the global 'Default Configuration'.

### Preconditions
1. Magento 2.2.4
2. Multi-storefront installation
3. PHP 7.0.28

### Steps to reproduce
1. Set global locale in Store Configuration in the admin (e.g. 'English (United Kingdom)')
2. On one of the individual storefronts in the Magento installation, set locale to be different value to the global (e.g. 'French')
3. Save configuration and then review frontend.

### Expected result
1. Locale should be using 'fr-FR' and the <html> lang attribute should be set to be 'fr'.
2. Translations should be using the 'French' locale

### Actual result
1. The frontend shows all storefronts in the global locale ('English') instead of using the storefront specific value.
2. This means that all different language storefronts use 'English' and not their respective languages.

callumstar:
Further to the above, we have done some further research and no other language apart from en_GB gets compiled in the 'pub/static/' folder regardless of what language/locale we have set.

LSERRE:
I got the same issue in 2.2.4, Default locale configuration is used for all stores view. This issue is really big.

andy17612:
I tested others language, like Chinese , Japanese it can showing the language. *But something error on frontend , my theme some file is losed. On english frontend is working.I already changed file 777 . Ran update,compile or static-content:deploy , something are not working still.

callumstar:
Hi @magento-engcom-team / @magento-team,

Is there any reponse on this at all? It seems to be a pretty fundamental issue for users with multi-store websites and one that doesn't look to have been tested overly well by yourselves. It looks like the "Default Configuration' is used for all store views when loaded regardless if you have chosen to 'Use Default Values' or not.

I'm seeing similar issues now appear in the Issue Tracker:
- #15245

hostep:
@callumstar: I tried to reproduce this on a cleanly installed 2.2.4 but couldn't reproduce (sorry, didn't test upgrading from 2.1.9), I performed following steps:
- Installed Magento 2.2.4 with composer
- Installed a French language pack using composer: https://github.com/Imaginaerum/magento2-language-fr-fr
- Ran `bin/magento setup:upgrade`
[...]

Issue #15873, opened 2018-06-05
2.2.4 EE Unable to change themes of stores


silvanaguilar:
On Magento 2.2.4 EE I'm not able to apply different themes on different stores.

**Preconditions**
1. Win 10, Mac Sierra or Ubuntu 18.04 (tested on all of them)
2. Magento 2.2.4
3. Docker 18.03.1, PHP 7.1.13

**Steps to reproduce**
1. Go to the administrator
2. Create one more store under Stores/All Stores in order to have the "Default Store View" and "Second Store View".
3. Go to Content/Design/Configuration and apply the Magento Blank theme to the first store and Luma to the second one
4. Now, go to Store/Configuration -> General/Web -> Url Options and select **YES** on the Add Store Code to Urls option (so at the end of the domain the code of the store will be added)
5. Go to the frontend and visit both stores. 
ie: my first store url is **_magento2.test/default_** and my second url is **_magento2.test/second_**

**Expected result**
6. Default Store View should display Blank theme
7. Second Store View should display Luma Theme

**Actual result**
8. Both stores are displaying Blank theme

![2a-stores](https://user-images.githubusercontent.com/2109774/41003033-f8e17b20-68eb-11e8-8839-636ffde301b3.JPG)
![2b-website-code](https://user-images.githubusercontent.com/2109774/41003034-f905b1de-68eb-11e8-98f5-97656d42e8d4.JPG)
![3-design-conf](https://user-images.githubusercontent.com/2109774/41003036-f93138a4-68eb-11e8-916d-54386b16210b.JPG)
![4-store-code-to-urls](https://user-images.githubusercontent.com/2109774/41003037-f9564478-68eb-11e8-9987-5ec45e11b829.JPG)
![8a-second-website](https://user-images.githubusercontent.com/2109774/41003038-f97f6b82-68eb-11e8-886f-288df4c5aaaa.png)
![8b-default-website](https://user-images.githubusercontent.com/2109774/41003039-f9a40564-68eb-11e8-8689-1ad2745da97f.png)

DanielRuf:
Did you clear all Magento caches and deploy the static content?

silvanaguilar:
Yes, I absolutely did both

hostep:
@silvanaguilar: this is a know bug in 2.2.4, it was reported before, if you use multiple websites, the configuration fallback no longer works properly. See these other tickets: https://github.com/magento/magento2/issues/15205, https://github.com/magento/magento2/issues/15245

[...]

PR #15929, opened 2018-06-07
Postpone instantiation of session config by using a proxy

fmarangi:
### Description
The session config is initialized too early, on initialisation of the HTTP response object. The init of this config object triggers the load of store-specific configuration, which then caches the resolved config path. This happens *before* the pathinfo is processed, which defines the current store in multi-storeview environment. By using a proxy the instantiation is postponed.

This problem was introduced in c18e36b804fb3c86caa39d9976ce5db7497001a7.

1. magento/magento2#15205
2. magento/magento2#15245

### Manual testing scenarios
1. Magento 2.2.4 
2. Multiple storeviews, e.g. for different languages (store `en` for `en_US` and `de` for `de_DE`)
3. Config `web/url/use_store` set to `1`
4. Navigate to `/de`

### Contribution checklist

### DIFF
--- a/app/code/Magento/MediaStorage/Model/File/Storage/Response.php
+++ b/app/code/Magento/MediaStorage/Model/File/Storage/Response.php
@@ -7,7 +7,6 @@ namespace Magento\MediaStorage\Model\File\Storage;
 
 use Magento\Framework\App\Response\Http;
 use Magento\Framework\App\Request\Http as HttpRequest;
-use Magento\Framework\Session\Config\ConfigInterface;
 use Magento\Framework\Stdlib\Cookie\CookieMetadataFactory;
 use Magento\Framework\Stdlib\CookieManagerInterface;
 
@@ -35,7 +34,6 @@ class Response extends Http implements
      * @param CookieMetadataFactory $cookieMetadataFactory
      * @param \Magento\Framework\App\Http\Context $context
      * @param \Magento\Framework\Stdlib\DateTime $dateTime
-     * @param ConfigInterface $sessionConfig
      * @param \Magento\Framework\File\Transfer\Adapter\Http $transferAdapter
      */
     public function __construct(
@@ -44,10 +42,9 @@ class Response extends Http implements
         CookieMetadataFactory $cookieMetadataFactory,
         \Magento\Framework\App\Http\Context $context,
         \Magento\Framework\Stdlib\DateTime $dateTime,
-        ConfigInterface $sessionConfig,
         \Magento\Framework\File\Transfer\Adapter\Http $transferAdapter
     ) {
-        parent::__construct($request, $cookieManager, $cookieMetadataFactory, $context, $dateTime, $sessionConfig);
+        parent::__construct($request, $cookieManager, $cookieMetadataFactory, $context, $dateTime);
         $this->_transferAdapter = $transferAdapter;
     }
 
--- a/app/code/Magento/MediaStorage/Model/File/Storage/Response.php
+++ b/app/code/Magento/MediaStorage/Model/File/Storage/Response.php
@@ -7,6 +7,7 @@ namespace Magento\MediaStorage\Model\File\Storage;
 
 use Magento\Framework\App\Response\Http;
 use Magento\Framework\App\Request\Http as HttpRequest;
+use Magento\Framework\Session\Config\ConfigInterface;
 use Magento\Framework\Stdlib\Cookie\CookieMetadataFactory;
 use Magento\Framework\Stdlib\CookieManagerInterface;
 
@@ -34,6 +35,7 @@ class Response extends Http implements
      * @param CookieMetadataFactory $cookieMetadataFactory
      * @param \Magento\Framework\App\Http\Context $context
      * @param \Magento\Framework\Stdlib\DateTime $dateTime
+     * @param ConfigInterface $sessionConfig
      * @param \Magento\Framework\File\Transfer\Adapter\Http $transferAdapter
      */
     public function __construct(
@@ -42,9 +44,10 @@ class Response extends Http implements
         CookieMetadataFactory $cookieMetadataFactory,
         \Magento\Framework\App\Http\Context $context,
         \Magento\Framework\Stdlib\DateTime $dateTime,
+        ConfigInterface $sessionConfig,
         \Magento\Framework\File\Transfer\Adapter\Http $transferAdapter
     ) {
-        parent::__construct($request, $cookieManager, $cookieMetadataFactory, $context, $dateTime);
+        parent::__construct($request, $cookieManager, $cookieMetadataFactory, $context, $dateTime, $sessionConfig);
         $this->_transferAdapter = $transferAdapter;
     }
 
--- a/app/etc/di.xml
+++ b/app/etc/di.xml
@@ -233,6 +233,11 @@
             <argument name="pathInfoProcessor" xsi:type="object">Magento\Backend\App\Request\PathInfoProcessor\Proxy</argument>
         </arguments>
     </type>
+    <type name="Magento\Framework\App\Response\Http">
+        <arguments>
+            <argument name="sessionConfig" xsi:type="object">Magento\Framework\Session\Config\ConfigInterface\Proxy</argument>
+        </arguments>
+    </type>
     <preference for="Magento\Framework\Session\SaveHandlerInterface" type="Magento\Framework\Session\SaveHandler" />
     <type name="Magento\Framework\Session\SaveHandlerFactory">
         <arguments>
--- a/lib/internal/Magento/Framework/App/Response/Http.php
+++ b/lib/internal/Magento/Framework/App/Response/Http.php
@@ -71,14 +71,14 @@ class Http extends \Magento\Framework\HTTP\PhpEnvironment\Response
         CookieMetadataFactory $cookieMetadataFactory,
         Context $context,
         DateTime $dateTime,
-        ConfigInterface $sessionConfig = null
+        ConfigInterface $sessionConfig
     ) {
         $this->request = $request;
         $this->cookieManager = $cookieManager;
         $this->cookieMetadataFactory = $cookieMetadataFactory;
         $this->context = $context;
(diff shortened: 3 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15929) <br/>All committers have signed the CLA.

hostep:
Nice one @fmarangi !

Unfortunately applying your fix appears to break loading certain assets.
For example the response of the request to `https://example.com/static/version1528396674/frontend/Magento/blank/en_US/requirejs-config.js` contains:
```
Fatal error: Uncaught ArgumentCountError: Too few arguments to function Magento\Framework\App\Response\Http::__construct(), 5 passed in vendor/magento/module-media-storage/Model/File/Storage/Response.php on line 47 and exactly 6 expected in vendor/magento/framework/App/Response/Http.php on line 68

Call Stack
#   Time    Memory  Function    Location
1   0.0001  419920  {main}( )   .../static.php:0
2   0.0553  3669672 Magento\Framework\App\Bootstrap->createApplication( )   .../static.php:12
3   0.0553  3669672 Magento\Framework\App\ObjectManager->create( )  .../Bootstrap.php:232
4   0.0553  3669672 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->create( )    .../ObjectManager.php:56
5   0.0557  3681008 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->_resolveArguments( ) .../Developer.php:59
6   0.0558  3682288 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->resolveArgumentsInRuntime( ) .../Developer.php:34
7   0.0560  3684264 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->resolveArgument( )   .../AbstractFactory.php:230
8   0.0560  3684264 Magento\Framework\App\ObjectManager->create( )  .../AbstractFactory.php:146
9   0.0560  3684264 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->create( )    .../ObjectManager.php:56
10  0.0662  4106832 Magento\Framework\ObjectManager\Factory\Dynamic\Developer->createObject( )  .../Developer.php:66
11  0.0662  4107216 Magento\MediaStorage\Model\File\Storage\Response\Interceptor->__construct( )    .../AbstractFactory.php:111
12  0.0662  4107216 Magento\MediaStorage\Model\File\Storage\Response\Interceptor->__construct( )    .../Interceptor.php:14
13  0.0662  4107216 Magento\MediaStorage\Model\File\Storage\Response\Interceptor->__construct( )    .../Response.php:47
```

Which makes sense, since the call to the parent constructor of the class [Magento\MediaStorage\Model\File\Storage\Response](https://github.com/magento/magento2/blob/2.2.4/app/code/Magento/MediaStorage/Model/File/Storage/Response.php#L47) doesn't have a sixth argument which it passes along.

Can you take a further look at this? Thanks!

hostep:
I tested a bit more, looks like if I only apply your changes to the `app/etc/di.xml` file and don't apply the changes to the `lib/internal/Magento/Framework/App/Response/Http.php` file. It fixes everything.

I tested with the 3 scenario's:
- #15205: different locales configured works again
- #15245: different homepages configured works again
- #15873: different themes configured works again  (can you also add this one to the 'fixed issues' list?)

Feel free to update your PR, unless you have a particular reason why you want to change that other class?

fmarangi:
@hostep hi, thanks a lot for testing my changes! I updated my PR with a possible fix for the Media Storage, would you mind having a look at it?

hostep:
Thanks, I won't have time to test this today, maybe tomorrow.

Just out of curiosity, what's the reason for changing the constructor of the `Magento\Framework\App\Response\Http` class? When I tested it yesterday, those changes weren't necessary to fix the problem. So I'm just curious why you think this is needed?

Thanks!

fmarangi:
It is not really needed, but I find it quite it quite ugly to use the ObjectManager in a constructor when you can define everything via DI and have more control, I thought I'd tidy that up too. It's definitely going to be refactored at some stage, since it's a bit against M2 principles.

The reason why it was done like that is probably because of the error you found, instead of changing the subclass the OM was used directly as a workaround / quick win.

hostep:
Ah ok, I suppose it was done for backwards compatible reasons, see this document: https://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/, the section 'Adding a constructor parameter'.

So I think it would be best to only keep the changes to the `di.xml` file, and remove all the other changes, otherwise I'm afraid this PR will be denied, but I'm not 100% sure. The Magento Community Maintainers can probably tell you more about this (@ihor-sviziev for example?)

fmarangi:
Ok, I see, I'll reset my changes then, thanks for looking into that, learned something new.

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-1924](https://jira.corp.magento.com/browse/ENGCOM-1924) has been created to process this Pull Request

magento-engcom-team:
@fmarangi thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

hostep:
@magento-engcom-team, @ihor-sviziev: I see you've added this to the 2.2.6 milestone, but I think it would be better to add this fix to 2.2.5 if it gets accepted.

Reason being: 2.2.5 is a security release, so you want as much people possible upgrading to that version. If big major bugs were introduced in 2.2.4 (like the one which is fixed by this PR), a fix for such major bugs should get included in 2.2.5, otherwise there will be people not wanting to upgrade to 2.2.5, which is not what you want since it's a security release.

I would also like to propose a new way for releasing new versions (using high version examples, just because):

- 2.2.105 (bugfixes, new features, performance improvements, new bugs)
- 2.2.106 (no new features, no performance improvements, only fixes for major bugs released in previous release, so this one is a more stable version for adding security fixes on top)
- 2.2.107 (security release, on top of a stable version without any major bugs)

Would something like this make more sense? Because bringing out security releases on top of a version with major bugs, is not the right way to go I think.

Just my 2 cents :)

VladimirZaets:
Hi @fmarangi, thanks for collaboration. The PR introduces about 15% performance degradation. Unfortunately, we can't accept this PR. The result of performance tests in the attachment.

screen shot 2018-06-15 at 11 23 41

hostep:
@VladimirZaets: please tell me Magento is working on a proper solution for fixing #15205, #15245 & #15873 then. This is really a big priority which needs to be fixed before 2.2.5 is released. Thanks!

fmarangi:
@VladimirZaets I really cannot imagine how this fix (using a Proxy for the session config) would cause the figures you posted.

My guess is that the bug introduced in c18e36b had the side effect of causing a considerable performance improvement since store-specific and website-specific configurations were never loaded, thus saving some time and memory. Anyway it is a serious issue in my opinion, since language configs, payment configs, themes etc are not loaded correctly in a multiple storeview environment.

Could you please tell me how the performance tests are executed? Did repeated tests have the same results? I'd like to look into that and help provide a solution.

ishakhsuvarov:
Hi @fmarangi 

Performance tests are executed using jMeter and multiple variations of performance scenarios, you can see one in `setup/performance-toolkit/benchmark.jmx`. By default it is at least 60 loops on each scenario with disabled full page cache. You may learn more on the scenarios here: https://github.com/magento/magento2/tree/2.2/setup/performance-toolkit

By default, we use Magento installation with a generated "small profile" of data, which contains around 800 products and small amounts of other entities. You may find more info on generation here: https://devdocs.magento.com/guides/v2.2/config-guide/cli/config-cli-subcommands-perf-data.html

The Pull Request in question was measured by performance CICD job not 1 or 2 times, but multiple, all with the same result, to eliminate possible errors or deviations. From our experience, degradations of 10-15% are not caused by deviations but real issues. 

@hostep I am now checking internal backlog to figure out status on the issue. I'll get back with the details, if I find any.

ishakhsuvarov:
@fmarangi 
> My guess is that the bug introduced in c18e36b had the side effect of causing a considerable performance improvement since store-specific and website-specific configurations were never loaded, thus saving some time and memory.

I'll investigate performance numbers history for `2.2.x` to find out if this is the case.

ishakhsuvarov:
@fmarangi @hostep Reopening for additional investigation

magento-engcom-team:
Hi @fmarangi. Thank you for your contribution. 

hostep:
@magento-engcom-team or @ishakhsuvarov : no chance you can include this fix in Magento 2.2.5?

The [Release Notes for Magento 2.2.4](https://devdocs.magento.com/guides/v2.2/release-notes/ReleaseNotes2.2.4CE.html) already include that it's not recommended to upgrade to that version because of the bugs this PR fixes:
> Known issue
> Customers have reported the following behavior after upgrading to Magento 2.2.4 in deployments that span multiple websites:

> Magento multi-store installations do not use the store view-specific values from the store configuration settings if these settings have different values than the global default configuration settings. Instead, Magento uses the default configuration for all store views. GitHub-15205, GitHub-15245

> We do not recommend upgrading to Magento 2.2.4 if you deploy across multiple websites. Note that this problem is not triggered if you have only a single website with multiple stores or store views.

So if 2.2.5 is a security release, you want people upgrading to it and not repeat that same message that it's not recommended to upgrade to that version I think, that would look very weird.

sidolov:
Hi @hostep , fix **will be available in 2.2.5 release**. The comment above was autogenerated due to delivery to 2.2-develop branch, this changes were delivered to 2.2.5 release line in the separate internal pull request.

hostep:
@sidolov: excellent, thanks for the update!

Also big thanks to @fmarangi for providing the fix! :)

stkams:
A similar issue occurs in 2.2.5 when you switch "Add Store Code to Urls" to "Yes" after browsing first with the option set to "No". A cookie "store" was set which should be completely ignored when "Add Store Code to Urls" is Yes

ihor-sviziev:
Hi @stkams,

Could you create new issue with steps to reproduce and maybe some additional details?

PR #15981, opened 2018-06-10
Fix issue #14112

riccardougolini:
- set null OptionId
- set null all ProductLink
- updated unit test

to prevent overwrite of duplicated bundle product in table `catalog_product_bundle_selection`

### DIFF
--- a/app/code/Magento/Bundle/Model/Product/CopyConstructor/Bundle.php
+++ b/app/code/Magento/Bundle/Model/Product/CopyConstructor/Bundle.php
@@ -6,20 +6,20 @@
 namespace Magento\Bundle\Model\Product\CopyConstructor;
 
 use Magento\Catalog\Model\Product;
+use Magento\Catalog\Model\Product\CopyConstructorInterface;
 use Magento\Catalog\Model\Product\Type;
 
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=15981) <br/>All committers have signed the CLA.

app/code/Magento/Bundle/Model/Product/CopyConstructor/Bundle.php
```diff
@@ -6,28 +6,38 @@
...
 {
     /**
      * Duplicating bundle options and selections
-     *
      * @param Product $product
      * @param Product $duplicate
      * @return void
      */
-    public function build(Product $product, Product $duplicate)
+    public function build(Product $product, Product $duplicate): void
```
sivaschenko, line 20: Please remove the strict :) "void" return type declaration as it can result in a warning on instances where the class is extended

sivaschenko (changes requested):
Thank you for the contribution. Please see the code review notes.

sidolov:
Hi @riccardougolini , I am closing this PR now due to inactivity.

PR #16004, opened 2018-06-11
[Forwardport] Fixed the prepare items method for configurable products

### LINKED ISSUE #1335: Fix Magento\Sales\Service\V1\ShipOrderTest::testConfigurableShipOrder
### Preconditions

1. Magento 2 with MSI.

### Steps to reproduce

1. Run test Magento\Sales\Service\V1\ShipOrderTest::testConfigurableShipOrder.

### Expected result

1. Test run finished without any errors.
2. Unskip test Magento\Sales\Service\V1\ShipOrderTest::testConfigurableShipOrder

### Actual result

1. Test run fails.

### Note 
Fail is caused by erroneous fixture Magento/Sales/_files/order_configurable_product.php, which creates incorrect Order:
image

tephereth:
### Description

Trying to fix the issue of [MSI #1335](https://github.com/magento-engcom/msi/issues/1335) we discovered a problem related to the deleted order items in the `prepareItems()` method inside the `ShipmentFactory` class.
I've noticed that the range of deleted order items processed by the mentioned method was too wide because it's supposed to delete only virtual and bundle items with products not shipped separately.
The previous condition checked wether the product was virtual or had a parentId and not separately shipped, but this check deleted also simple products related to configurables introducing an issue with shipments for configurable products.
My proposed change would check if the product is virtual or bundle (not shipped seperately) to delete from processed items.
This issue doesn't have any consequence in 2.2 and 2.3 because the method `getAllItems()` of entity shipment is never used inside the core modules; but it will affect the MSI classes related to the `SourceDeductionProcessor` where this method is used to cycle the shipment items.

Fixed: [MSI #1335](https://github.com/magento-engcom/msi/issues/1335) with PR [MSI#1394](https://github.com/magento-engcom/msi/pull/1394)

### Contribution checklist

Port for https://github.com/magento/magento2/pull/16396

### DIFF
--- a/app/code/Magento/Sales/Model/Order/ShipmentFactory.php
+++ b/app/code/Magento/Sales/Model/Order/ShipmentFactory.php
@@ -105,7 +105,7 @@ class ShipmentFactory
             /** @var \Magento\Sales\Model\Order\Shipment\Item $item */
             $item = $this->converter->itemToShipmentItem($orderItem);
             if ($orderItem->getIsVirtual() ||
-                ($orderItem->getProductType() === \Magento\Catalog\Model\Product\Type::TYPE_BUNDLE
+                ($orderItem->getProductType() !== \Magento\Catalog\Model\Product\Type::TYPE_SIMPLE
                     && !$orderItem->isShipSeparately())) {
                 $item->isDeleted(true);
             }
--- a/app/code/Magento/Sales/Model/Order/ShipmentFactory.php
+++ b/app/code/Magento/Sales/Model/Order/ShipmentFactory.php
@@ -105,7 +105,8 @@ class ShipmentFactory
             /** @var \Magento\Sales\Model\Order\Shipment\Item $item */
             $item = $this->converter->itemToShipmentItem($orderItem);
             if ($orderItem->getIsVirtual() ||
-                ($orderItem->getProductType() === \Magento\Catalog\Model\Product\Type::TYPE_BUNDLE && !$orderItem->isShipSeparately())) {
+                ($orderItem->getProductType() === \Magento\Catalog\Model\Product\Type::TYPE_BUNDLE
+                    && !$orderItem->isShipSeparately())) {
                 $item->isDeleted(true);
             }
 
--- a/app/code/Magento/Sales/Model/Order/ShipmentFactory.php
+++ b/app/code/Magento/Sales/Model/Order/ShipmentFactory.php
@@ -104,7 +104,8 @@ class ShipmentFactory
 
             /** @var \Magento\Sales\Model\Order\Shipment\Item $item */
             $item = $this->converter->itemToShipmentItem($orderItem);
-            if ($orderItem->getIsVirtual() || ($orderItem->getParentItemId() && !$orderItem->isShipSeparately())) {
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=16004) <br/>All committers have signed the CLA.

app/code/Magento/Sales/Model/Order/ShipmentFactory.php
```diff
@@ -104,7 +104,9 @@ protected function prepareItems(
 
             /** @var \Magento\Sales\Model\Order\Shipment\Item $item */
             $item = $this->converter->itemToShipmentItem($orderItem);
-            if ($orderItem->getIsVirtual() || ($orderItem->getParentItemId() && !$orderItem->isShipSeparately())) {
+            if ($orderItem->getIsVirtual() ||
+                ($orderItem->getProductType() === \Magento\Catalog\Model\Product\Type::TYPE_BUNDLE
```
orlangur: This is incorrect module dependency. Although `bundle` constant is placed in `Catalog` module for historical reasons, `Sales` module must not be aware of bundle product type business logic.

If it's not too hard, please move the `bundle_selection_attributes` part to `Bundle` module as well.

Please cover described case with integration test.

orlangur (changes requested):

orlangur:
@tephereth also, please create PR targeting `2.2-develop` first.

tephereth:
Uploaded a better solution not involving bundle. As said previously this issue affect only MSI, there are no consequences in 2.3 or 2.2. The fix is verifiable via MSI as described here: https://github.com/magento-engcom/msi/pull/1394
After I'll do a backport even on 2.2

orlangur:
Hi @tephereth, we need to start from `2.2-develop` anyway, no matter whether is affects MSI or anything else. Then it can be forwardported to `2.3-develop` shortly.

Why do we need to remove `$orderItem->getParentItemId()` check?

tephereth:
Hi @orlangur, actually the above check marks simple products as `deleted`, then when in MSI stock system the shipment order is listing the products with the method `getAllItems`, those simple products are not cycled and removed from the process. The process itself crashes in the `SourceDeduction` when configurable products are in the shipment and so the related test `ShipOrderTest::testConfigurableShipOrder`.
According to @maghamed this seems to be a bug introduced by a previous workaround implemented few months ago and inside the shipment order u must have simple products but not complex products (that are just containers and don't have a specified quantity).
This PR fix that problem.
About the 2.2 no problem, I'll send the same PR for the `2.2-develop` right today.

orlangur:
@tephereth,

>  this seems to be a bug introduced by a previous workaround implemented few months ago and inside the shipment order u must have simple products but not complex products

ok, is it just blame of the same rows showing?

tephereth:
Created backport for `2.2-develop`:
https://github.com/magento/magento2/pull/16396

orlangur (approved):

magento-engcom-team:
[ENGCOM-2266](https://jira.corp.magento.com/browse/ENGCOM-2266) has been created to process this Pull Request

PR #16032, opened 2018-06-12
[Backport] Avoid calling function repetitively in loop conditionals

chirag-wagento:
### Original Pull Request 
 https://github.com/magento/magento2/pull/16027
Avoid calling function repetitively in loop conditionals.

### Description
Avoid calling function repetitively in loop conditionals for better performance.

### Fixed Issues
N/A


orlangur:
See https://github.com/magento/magento2/pull/16027#issuecomment-396613822

PR #16075, opened 2018-06-12
Wishlist Rest API implementation

Tickles-BV:
### Description

This pull request is a small start at adding an api part for the wishlist functionality in magento 2.
Because it adds new interfaces I have tried to only implement the smallest possible set of changes so this can be discussed and implemented further.

I have added a wishlist repository, which allows one to retrieve a wishlist for a customer, and to add products to a customer's wishlist. I've also added an interface for the wishlist model itself, also for use in the rest api. The wishlist model interface currently only interfaces 'getName', because I'm unsure which properties should be added to it.

I have a few questions about implementation details.

1. How do I know which attributes should be made available in data api? For instance, should you get a list of wishlist products when retrieving a wishlist by customer? Or should you simply get a wishlist id, which can then be used to retrieve the items via another API call. This seems like a cleaner seperation of logic, but I'd like confirmation.
2. I've added two methods to the repository: `getWishlistForCustomer` and `addWishlistForCustomer`, should these be in the repository? Or should these be extracted to a WishlistManagement class.
3. I've simply made the api route 'V1/wishlist' are there conventions for this? If so, where can I find these?
4. In the wishlist model, method 'getSharingCode' is added via @property docblock tag, can I add these to api response without adding to new const and wrapping in own method?

 Sorry for the amount of questions but I'm trying to wrap my head around a lot of magento 2 internals implementing this.

1. magento/community-features#56: API: Wishlist

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=16075) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @Tickles-BV. Thank you for your contribution 

Tickles-BV:
I've updated the WishlistInterface and now it encapsulates all the properties available in the wishlist model, wishlist model now properly implements this and all the data is returned in api calls. 

This answers both question 1 and 4 on my end, the others I'd still like to have someone look at.

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Magento\Wishlist\Api;
+
+/**
+ * Interface WishlistRepositoryInterface
+ * @api
+ * @package Magento\Wishlist\Api
+ */
+interface WishlistRepositoryInterface
```
sidolov, line 17: 1. The repository should have methods for CRUD operations: get, getList, save and delete. 
2. Wishlist Repository should manipulate Wishlist entity only 
3. Use strict types for methods

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Magento\Wishlist\Api\data;
```
sidolov: First letter in namespace should be uppercase - `Magento\Wishlist\Api\Data`

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    const NAME = 'name';
+
+    const SHARING_CODE = 'sharing_code';
+
+    const CUSTOMER_ID = 'customer_id';
+
+    const SHARED = 'shared';
+
+    const UPDATED_AT = 'updated_at';
```
sidolov: This field seems to be redundant. Do you have the specific case?

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Magento\Wishlist\Api\data;
+
+interface WishlistInterface
+{
+    /**#@+
+     * Constants defined for keys of  data array
+     */
+    const NAME = 'name';
```
sidolov: Wishlist name is immutable in the current implementation, why we need this field in api?

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Magento\Wishlist\Api\data;
+
+interface WishlistInterface
```
sidolov: 1. Some setters for fields are missing 
2. Use strict types for methods 
3. Setter and getter for extension attributes are missing

app/code/Magento/Wishlist/Model/Wishlist.php
```diff
@@ -486,17 +490,15 @@ public function addNewItem($product, $buyRequest = null, $forciblySetQty = false
      */
     public function setCustomerId($customerId)
     {
-        return $this->setData($this->_getResource()->getCustomerIdFieldName(), $customerId);
+        return $this->setData(self::CUSTOMER_ID, $customerId);
```
sidolov: This change may break 3d party extension that uses this extension point for customer_id field

app/code/Magento/Wishlist/Model/WishlistRepository.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public function addWishlistForCustomer($customerId, $productId)
+    {
+        /** @var Wishlist $wishlist */
+        $wishlist = $this->wishlistFactory->create()->loadByCustomerId($customerId, true);
+
+        $product = $this->productRepository->getById($productId);
+
+        try {
+            $item = $wishlist->addNewItem($product);
```
sidolov: Wishlist item is the different entity and managing should be in a separate repository (for example WishlistItemRepository)

app/code/Magento/Wishlist/Model/WishlistRepository.php
```diff
@@ -0,0 +1,66 @@
...
+            throw new NoSuchEntityException(__('No wishlist for customer.'));
+        }
+
+        return $wishlist;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function addWishlistForCustomer($customerId, $productId)
```
sidolov: Please, try to use product SKU instead of product id.

app/code/Magento/Wishlist/etc/di.xml
```diff
@@ -8,6 +8,8 @@
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
     <preference for="Magento\Wishlist\Model\AuthenticationStateInterface" type="Magento\Wishlist\Model\AuthenticationState" />
     <preference for="Magento\Wishlist\Controller\WishlistProviderInterface" type="Magento\Wishlist\Controller\WishlistProvider" />
+    <preference for="Magento\Wishlist\Api\WishlistRepositoryInterface" type="Magento\Wishlist\Model\WishlistRepository" />
```
sidolov, line 11: Preference for WishlistInterface is missing

app/code/Magento/Wishlist/etc/webapi.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" ?>
+<routes xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Webapi:etc/webapi.xsd">
+    <route method="GET" url="/V1/wishlist/:customerId">
```
sidolov: Please, use wishlist id

app/code/Magento/Wishlist/etc/webapi.xml
```diff
@@ -0,0 +1,15 @@
...
+        <service class="Magento\Wishlist\Api\WishlistRepositoryInterface" method="getWishlistForCustomer"/>
+        <resources>
+            <resource ref="anonymous"/>
```
sidolov: Please, use separate resources for customer and admin with corresponded access rights. 
Anonymous shouldn't have access to wishlist at all (the only customers may have wishlist)

sidolov (changes requested):

ishakhsuvarov:
Hi @Tickles-BV 
Thank you for this Pull Request. Designing a new API always requires a lot of effort.

I'll try to answer your questions briefly:
> For instance, should you get a list of wishlist products when retrieving a wishlist by customer?

It would be better to handle this in a separate service, getting wishlist contents.

> I've added two methods to the repository should these be in the repository?

Concept of repositories used by Magento only allows to CRUD methods to be present in the `RepositoryInterface`. Methods in question may be moved to a `MagementInterface` or even to a separate commands.

> I've simply made the api route 'V1/wishlist' are there conventions for this? If so, where can I find these?

I'm not aware of any specific convention, so it would be best to make it consistent with other modules and use common sense.

> In the wishlist model, method 'getSharingCode' is added via @property docblock tag, can I add these to api response without adding to new const and wrapping in own method?

I'm pretty sure it would not work with the WebAPI framework. Additionally, it's always the best way to have everything declared explicitly. Wishlist module code seems to be quite "legacy" so it's not a great example.

Please take a look at the code review we have done. It's not 100% complete at the moment, but it doesn't make sense to dig too deep in the code which is subject to change, we could work on it iteratively. 

Thank you for collaboration

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
Tickles-BV: Should not all fields used by getter and setter methods be specified in api constants?

app/code/Magento/Wishlist/etc/webapi.xml
Tickles-BV: Who should have access? And how do I know who implements the rights? I'm looking at other files for examples but it is confusing. For instance, webapi.xml for lots of modules just use 'Magento_Catalog::products' but I can't find where this is defined. Do you have place with example I can look at?

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
Tickles-BV: Should api not have method to retrieve name? Or should I only remove NAME constant and keep getName method and remove setName method? If so, why remove constant?

app/code/Magento/Wishlist/Model/Wishlist.php
Tickles-BV: This is the way it is implemented in new code such as BookmarkInterface, what is preferred solution? Removing field from constants and just keeping getId and setId in interface?

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
Tickles-BV: Resolved

app/code/Magento/Wishlist/Api/data/WishlistInterface.php
Tickles-BV: I've added setter for customerId, with 2, do you mean declare('strict_types') or that I should use typehints

app/code/Magento/Wishlist/etc/di.xml
Tickles-BV: Added

Tickles-BV:
Hi @ishakhsuvarov, @sidolov 

I've created WishlistRepository that implements getById, getList, save, delete and deleteById.
Old WishlistRepository methods have been moved to WishlistManagement, but that is probably still wrong.

I've added acl record for Wishlist and added 'me' routes in webapi.xml

Also WishlistRepository is now covered by integration test WishlistRepositoryTest

Please answer my questions about api constants and which getters and setters should be in the interface.

Also should I write WishlistItemRepository in this pull request or is it out of scope?

Thanks.

ishakhsuvarov:
@Tickles-BV We are going to review it shortly and get back to you with the updates. 

`WishlistItem` interfaces look like they would perfectly fit into the scope of this PR.

Tickles-BV:
One more question,

"Setter and getter for extension attributes are missing"

Does this mean I should make Wishlist model extend AbstractExtensibleModel and implement WishlistExtensionInterface and concrete class?

app/code/Magento/Wishlist/Api/Data/WishlistInterface.php
```diff
@@ -0,0 +1,100 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Wishlist\Api\Data;
+
+interface WishlistInterface
```
ishakhsuvarov: Please add support for extension attributes to this interface

app/code/Magento/Wishlist/Api/Data/WishlistInterface.php
```diff
@@ -0,0 +1,100 @@
...
+
+    const NAME = 'name';
+
+    const SHARING_CODE = 'sharing_code';
+
+    const CUSTOMER_ID = 'customer_id';
+
+    const SHARED = 'shared';
+
+    const UPDATED_AT = 'updated_at';
```
ishakhsuvarov: There is not need to expose `updated_at` via Wishlist data object. Please consider removing it together with getter and setter.
Possibly it's still valid to maintain this field in `WishlistItem`

app/code/Magento/Wishlist/Api/Data/WishlistInterface.php
ishakhsuvarov: Please add type hinting to all method arguments and return types.

app/code/Magento/Wishlist/Api/WishlistManagementInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Wishlist\Api;
+
+interface WishlistManagementInterface
```
ishakhsuvarov: Please add type hints to all methods for arguments and return types

ishakhsuvarov: Existing wishlist feature supports more actions, e. g. "share wishlist".
There should be investigated and added as well

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function save(\Magento\Wishlist\Api\Data\WishlistInterface $wishlist): WishlistInterface;
+
+    /**
+     * Delete wishlist by passing object
+     *
+     * @param \Magento\Wishlist\Api\Data\WishlistInterface $wishlist
+     * @return bool
+     */
+    public function delete(\Magento\Wishlist\Api\Data\WishlistInterface $wishlist);
```
ishakhsuvarov: It is pretty safe to assume that this method would only proxy to `deleteById`. I would suggest removing it.

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+interface WishlistRepositoryInterface
+{
+    /**
+     * Get Wishlist by id
+     * @param int $id
+     * @return \Magento\Wishlist\Api\Data\WishlistInterface
+     * @throws \Magento\Framework\Exception\NoSuchEntityException
+     */
+    public function getById($id): \Magento\Wishlist\Api\Data\WishlistInterface;
```
ishakhsuvarov, line 33: It would make sense to rename this method to `get`

app/code/Magento/Wishlist/Api/Data/WishlistInterface.php
ishakhsuvarov: I'd suggest adding wishlist items to this interface as well.

app/code/Magento/Wishlist/Api/WishlistManagementInterface.php
ishakhsuvarov: I would suggest to split this interface into multiple separate interfaces, pretty much interface per method, as currently, a class which implements this interface (especially the updated version with more features) would have too much responsibilities.

app/code/Magento/Wishlist/Api/WishlistManagementInterface.php
```diff
@@ -0,0 +1,27 @@
...
+     */
+    public function getWishlistByCustomerId($customerId);
+
+    /**
+     * @param int $customerId
+     * @param string $sku
+     * @throws \Magento\Framework\Exception\StateException
+     * @return int
+     */
+    public function addWishlistItemByCustomerId($customerId, $sku);
```
ishakhsuvarov, line 20: Please consider, that wishlist items also support custom options, which can not be resolved via sku only. Possibly it's best to introduce `WishlistItem` and use it here.

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
```diff
@@ -0,0 +1,59 @@
...
+    public function getList(\Magento\Framework\Api\SearchCriteriaInterface $searchCriteria);
+
+    /**
+     * Save wishlist object
+     *
+     * @param Data\WishlistInterface $wishlist
+     * @return \Magento\Wishlist\Api\Data\WishlistInterface
+     * @throws \Magento\Framework\Exception\CouldNotSaveException
+     */
+    public function save(\Magento\Wishlist\Api\Data\WishlistInterface $wishlist): WishlistInterface;
```
ishakhsuvarov: To make this method more CQS-friendly it may be updated to return only id of the created item, instead of the whole item. Returning void may be a good idea as well, but would greatly limit use cases, as there is no predefined if for wishlist.

app/code/Magento/Wishlist/Model/Wishlist.php
```diff
@@ -3,32 +3,30 @@
...
- * @method string getUpdatedAt()
- * @method \Magento\Wishlist\Model\Wishlist setUpdatedAt(string $value)
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  * @SuppressWarnings(PHPMD.TooManyFields)
  *
  * @api
  * @since 100.0.2
  */
-class Wishlist extends \Magento\Framework\Model\AbstractModel implements \Magento\Framework\DataObject\IdentityInterface
+class Wishlist extends AbstractModel implements IdentityInterface, WishlistInterface
```
ishakhsuvarov: This class seems to be a legacy model, and trying to implement `WhishlistInterface` on top of it may be one of the less-optimal solutions. 

I would suggest implementing a separate data transfer object to implement the interface.

ishakhsuvarov (changes requested):

ishakhsuvarov:
Hi @Tickles-BV 
Thank you for the updates. This PR looks much more complete now.

There is still some items missing, main one is the `WishlistItem` and repository for it. Looking more closely at wishlist items it appears that apart from sku it can also bear custom options and qty of the item. It's best to check existing features and make sure interface would cover all of the requirements.

Some features, for example related to sharing, are not covered with this API as well, however I suppose that sharing a wishlist is the main idea of creating it :) Again, it's best to research what's already available, and at least cover all of it with the APIs.

Please let me know if you have any questions regarding the review. You may also contact me and the team via community engineering slack workspace, to speed up communication.

ishakhsuvarov:
@Tickles-BV Adding answers to questions above.
> Old WishlistRepository methods have been moved to WishlistManagement, but that is probably still wrong.

It's a step in a right direction 👍 

> Also WishlistRepository is now covered by integration test WishlistRepositoryTest

While it's pretty similar, It's suggested to cover web api's with api-functional tests.

> Does this mean I should make Wishlist model extend AbstractExtensibleModel and implement WishlistExtensionInterface and concrete class?

Interface should extend `\Magento\Framework\Api\ExtensibleDataInterface` while model may inherit `\Magento\Framework\Api\AbstractExtensibleObject`. Extension interfaces are generated automatically, as long as you define getter and setter for extension attributes with the correct method signature.

> Please answer my questions about api constants and which getters and setters should be in the interface.

It's best to design the API which exposes only meaningful fields, basically following YAGNI principle. In the example of this code review, I'm not sure if `updated_at` would be ever used and worth supporting in the future.

> Also should I write WishlistItemRepository in this pull request or is it out of scope?

After further looking into it – it's definitely yes.

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
deefco: Changed to returning id

Tickles-BV: I've changed this method to return just the id

app/code/Magento/Wishlist/Model/Wishlist.php
Tickles-BV: I've created a new dto in Model\Data\Wishlist which now implements the interface, wishlist now has public function getDataModel which copies properties to new implementation and returns. Repository has been changed to directly return new dto instead of legacy one. All of this has been based on implementation in Customer files Model\Customer, Model\Data\Customer and CustomerRepository.

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
Tickles-BV: Deleted

app/code/Magento/Wishlist/Api/WishlistRepositoryInterface.php
Tickles-BV: Renamed

app/code/Magento/Wishlist/Api/Data/WishlistInterface.php
Tickles-BV: Removed updated_at property

app/code/Magento/Wishlist/Api/Data/WishlistInterface.php
Tickles-BV: I've added function called getItems to wishlist interface (based on naming scheme I could find in other api interfaces) which returns array of wishlist item interfaces

Tickles-BV: I've also made wishlist dto extensible as requested

Tickles-BV:
I've added wishlist item interface now, along with repository. Repository just implements crud operations same as wishlist repository. I'm going to figure out what other things should be exposed to service contracts.

For example: Sharing wishlist via email appears to be implemented directly in the Shared controller. Should this be extracted to service contract?

Tickles-BV:
I've added repository for wishlist item and wishlist item option. Those both now have extended interfaces, item is totally new, option has been changed to new one. Not sure if this effects backwards compatibilty? Option model has had getters and setters changed to adhere to new interface which has type hints, this is definitely b/c breaking, I'd like to hear suggestions on this. Another new DTO perhaps? :)

I'd like opinions on the way the new Wishlist DTO has been implemented. @ishakhsuvarov Said to implement seperate concrete DTO, this has been done by adding getDataModel proxy call that sets data from legacy model onto new model and returns that instead. Implementation is based on what has been done in Customer model (pretty much copied) though someone said this is deprecated way of doing things, it is the only example implementation I could find.

I've added 2 spi classes, AddItemToWishlistInterface and AddOptionToWishlistItemInterface, both based on new MSI model where class has 1 function called 'execute'. I'd like some guidance on figuring out what functionality to extract to SPI layer.

Finally I have a question about DTO implementation. Current Wishlist Model is "clean" implementation (only getters and setters), say I want to have getItems functionality that gets all wishlist items. Should I implement loading these items in plugin on after_load event? Or afterSave()? Or something else entirely? Same goes for saving wishlist, all Items should be saved to keep consistency. Should saving be implemented by plugin? Or afterSave method, or in Repository using SaveHandler? (This is implementation used in Quote Module)

I know these are a lot of questions but I'd like to have a basic idea of where we're going with this pull request because the amount of work keeps growing and I'd really not like to redo any of it because of miscommunication.

deefco:
@gelanivishal could you please stop changing / doing commits in my pull request? If you require changes to be made please leave them as a comment here so I can discuss this with you.

gelanivishal:
@Tickles-BV please revet back my changes.

misha-kotov:
@sidolov @ishakhsuvarov can you guys provide guidance on the above questions by @Tickles-BV ? Let me know if you need me from use case/product perspective.

misha-kotov:
I wanted to include a collection of use cases that we need to keep in mind when developing the APIs. We can use these as a sanity check to ensure we have everything in place.
cc: @sidolov @ishakhsuvarov  

Shopper Wishlist Use cases: 
- Wishlist level: 
   - Create wishlist
   - Delete wishlist (can delete all wishlists except id=1)
   - Update wishlist (name, public/private)
   - Share wishlist (list of emails, message)
- Wishlist Item level: 
   - Add wishlist item 
   - Remove wishlist item
   - Update item (configuration, qty, comment)

Use cases we don't need dedicated endpoints for: 
- Move items to wishlist: can be accomplished with `Add wishlist item`, `Remove wishlist item`
- Copy items to wishlist: can be accomplished with `Add wishlist item`

In Admin, today we only allow the admin to delete and reconfigure products on customer wishlists. I'm not sure if these are the only ones by design or due to incompletion.

Spaggel:

magento-engcom-team:
Hi @Spaggel. Thank you for your request. I'm working on Magento instance for you

orlangur (changes requested):
@Tickles-BV please prepare a new pull request against `2.3-develop` as new features should be only added there: https://devdocs.magento.com/guides/v2.2/contributor-guide/contributing.html#rules

PR #16204, opened 2018-06-17
Improvement the validation of the eav entity attribute

eduard13:
### Description
This additional validation, is checking for allowed characters within an eav attribute code, which allows only letters, number, underscores and the first character should be a letter.

### Contribution checklist


magento-engcom-team:

app/code/Magento/Eav/Model/Entity/Attribute.php
```diff
@@ -253,6 +253,16 @@ public function beforeSave()
             );
         }
 
+        /**
+         * Check attribute_code for allowed characters
+         */
+        if (isset($this->_data['attribute_code'])
+            && !preg_match('/^[a-z]+[a-z0-9_]+$/', $this->_data['attribute_code'])) {
```
adeptofvoltron: that regex, also force us to have attribute_code with at least 2 chars. Maybe it is ok, I just want to make you aware of that.

orlangur: Model is not right place for validation, there is some validation already present in controller, isn't it?

orlangur (changes requested):

app/code/Magento/Eav/Model/Entity/Attribute.php
eduard13: @orlangur I suggest to have this validation also when we need to create some eav attributes programmatically. As for the controllers, there is a frontend validation in the backend while an attribute is created.

orlangur: @eduard13 frontend validation is not enough. I'm pretty sure this is validated on PHP side also. Validation should not be added to model itself, when you create attribute programmatically - you should validate attribute code prior to passing it into model.

eduard13: @orlangur I agree with you that we should check the attribute code before saving it, but sometimes you don't know that some chars aren't allowed, like . (dots), and you successfully save the attribute, but later you aren't able to properly work with the attribute (because the getters and setters aren't well generated).

By the way, if we'll take a look closer in this method, there are other several validation rules, so it is needed to refactor this method and move all the validations to a separate file?

orlangur: @eduard13 didn't check the code out of your changeset :) I need some time to understand the whole picture, please be patient.

eduard13: Hi @orlangur any suggestions here?

orlangur (changes requested):
@eduard13 just discovered that there is already validation for attribute code length here, so, checking for allowed characters should be fine.

Please introduce an `AttributeCodeValidator` class and reuse it here and in `\Magento\Catalog\Model\Product\Attribute\Repository::generateCode` and in `\Magento\Catalog\Model\Product\Attribute\Repository::validateCode`.

eduard13:
@orlangur, after digging deeper, I've noticed that when the attributes are created using Setup services, the Attribute model isn't used for saving it, instead of that it is inserted directly in DB. So I've moved the validation in the validating method and reverted the previously made changes.

Moreover, I don't see the reason to use a different validation message, than it is used elsewhere.

orlangur:
Copy-pasting is not a good idea.

Please introduce an `AttributeCodeValidator` class and reuse it here, in `\Magento\Catalog\Model\Product\Attribute\Repository::generateCode` and in `\Magento\Catalog\Model\Product\Attribute\Repository::validateCode`.

sidolov:
Hi @eduard13 , will you continue progress with PR?

orlangur:
@sidolov we need a reply from @VladimirZaets first in order to understand a proper validation rule.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3442](https://jira.corp.magento.com/browse/ENGCOM-3442) has been created to process this Pull Request

sivaschenko:
Hi @eduard13, it's preferable to merge changes to 2.3 before merging to 2.2 on the current stage. Please [port](https://devdocs.magento.com/guides/v2.2/contributor-guide/contributing.html#port) the pull request to 2.3-develop branch

eduard13:
Hi @sivaschenko, thank you. I agree with that, as this PR was made when the mainline was 2.2. Will do it asap.

orlangur (changes requested):
@eduard13 review NOT passed, please pay attention to https://github.com/magento/magento2/pull/16204#issuecomment-403433664. Sergey is right about delivering everything via `2.3-develop` first nowadays, please create such a PR containing necessary changes.

@sivaschenko please DO NOT touch labels when you don't know what are they intended for.

sivaschenko:
@orlangur , @eduard13 , @VladimirZaets Gentlemen, do you have any updates on this PR?

eduard13:
@sivaschenko we can close this one, I will open a new one for 2.3 with the mentioned reviews fixed.
Thanks everyone.

PR #16571, opened 2018-07-05
Simplify code product alert email

arnoudhgz:
The send method was complex and had some duplication in code. This is improved by using private methods for small parts of this method and making the if/else statements more logical (else isn't even used anymore).

As result the PHPMD suppress warning annotations could be removed.

For better string comparison the currently defined types (price and stock) are set as constants.

### Contribution checklist

### DIFF
--- a/app/code/Magento/ProductAlert/Model/Email.php
+++ b/app/code/Magento/ProductAlert/Model/Email.php
@@ -1,9 +1,35 @@
 <?php
+
 /**
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
 namespace Magento\ProductAlert\Model;
+
+use Magento\Catalog\Model\Product;
+use Magento\Customer\Api\CustomerRepositoryInterface;
+use Magento\Customer\Api\Data\CustomerInterface;
+use Magento\Customer\Helper\View;
+use Magento\Framework\App\Area;
+use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\Framework\Data\Collection\AbstractDb;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Exception\MailException;
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Framework\Mail\Template\TransportBuilder;
+use Magento\Framework\Model\AbstractModel;
+use Magento\Framework\Model\Context;
+use Magento\Framework\Model\ResourceModel\AbstractResource;
+use Magento\Framework\Registry;
+use Magento\ProductAlert\Block\Email\AbstractEmail;
+use Magento\ProductAlert\Block\Email\Price;
+use Magento\ProductAlert\Block\Email\Stock;
+use Magento\ProductAlert\Helper\Data;
+use Magento\Store\Api\Data\StoreInterface;
+use Magento\Store\Model\App\Emulation;
+use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\StoreManagerInterface;
+use Magento\Store\Model\Website;
 
 /**
  * ProductAlert Email processor
@@ -14,7 +40,7 @@ namespace Magento\ProductAlert\Model;
  * @api
  * @since 100.0.2
  */
-class Email extends \Magento\Framework\Model\AbstractModel
+class Email extends AbstractModel
 {
     const XML_PATH_EMAIL_PRICE_TEMPLATE = 'catalog/productalert/email_price_template';
 
@@ -32,14 +58,14 @@ class Email extends \Magento\Framework\Model\AbstractModel
     /**
      * Website Model
      *
-     * @var \Magento\Store\Model\Website
+     * @var Website
      */
     protected $_website;
 
     /**
      * Customer model
      *
-     * @var \Magento\Customer\Api\Data\CustomerInterface
+     * @var CustomerInterface
      */
     protected $_customer;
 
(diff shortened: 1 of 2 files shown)


magento-engcom-team:
Hi @arnoudhgz. Thank you for your contribution 

app/code/Magento/ProductAlert/Model/Email.php
```diff
@@ -286,110 +319,136 @@ protected function _getStockBlock()
...
+     *
+     * @return string
+     */
+    private function getTemplateConfigPath()
+    {
+        if ($this->_type === self::TYPE_PRICE) {
+            return self::XML_PATH_EMAIL_PRICE_TEMPLATE;
+        }
+
+        return self::XML_PATH_EMAIL_STOCK_TEMPLATE;
```
orlangur: Please replace all such conditionals with ternary operator.

arnoudhgz: @orlangur thank you for your feedback, I changed it...

VladimirZaets:
Hi @arnoudhgz, thank you for collaboration. Please enable `strict type` mode for this uses `declare(strict_types=1)` declaration.

app/code/Magento/ProductAlert/Model/Email.php
```diff
@@ -286,110 +319,128 @@ protected function _getStockBlock()
...
     }
+
+    /**
+     * Retrieve the store for the email
+     *
+     * @param int $customerStoreId
+     *
+     * @return StoreInterface|Store
+     */
+    private function getStore($customerStoreId)
```
VladimirZaets: Please use `type hints` for the method arguments.

app/code/Magento/ProductAlert/Model/Email.php
```diff
@@ -286,110 +319,128 @@ protected function _getStockBlock()
...
+            ? $this->_priceProducts
+            : $this->_stockProducts;
+    }
+
+    /**
+     * Retrieve template config path based on type
+     *
+     * @return string
+     */
+    private function getTemplateConfigPath()
```
VladimirZaets: Please use `type hints` for the method arguments.

app/code/Magento/ProductAlert/Model/Email.php
```diff
@@ -286,110 +319,128 @@ protected function _getStockBlock()
...
+            ? $this->_getPriceBlock()
+            : $this->_getStockBlock();
+    }
+
+    /**
+     * Retrieve the products for the email based on type
+     *
+     * @return array
+     */
+    private function getProducts()
```
VladimirZaets: Please use `type hints` for the method arguments.

app/code/Magento/ProductAlert/Model/Email.php
```diff
@@ -286,110 +319,128 @@ protected function _getStockBlock()
...
+            ? $this->_storeManager->getStore($customerStoreId)
+            : $this->_website->getDefaultStore();
+    }
+
+    /**
+     * Retrieve the block for the email based on type
+     *
+     * @return Price|Stock
+     */
+    private function getBlock()
```
VladimirZaets: Please use `type hints` for the method arguments.

arnoudhgz:
@VladimirZaets  I have added the strict types and included type hinting where possible.

I saw some existing protected methods like `setWebsiteId` where I could not define a strict type, because the input can be also string in dependend code....

arnoudhgz:
@VladimirZaets I see that with the strict types give now errors in the unit tests. I will first fix this before you have to check again.

arnoudhgz:
@VladimirZaets  I will fix the errors in the unit test first before you have to check again.

VladimirZaets:
Hi @arnoudhgz, maybe you need any help?

arnoudhgz:
@VladimirZaets  Yes,  I would like that. 
I am struggling with the strict type for the `setCustomerData` method. First I had `CustomerInterface` but that did not workout, `DataObject` worked for the unit tests, but now it fails on the integration test.

VladimirZaets:
@arnoudhgz as I see, the `$customer` should be `CustomerInterface` type. I searched usages of this method but I can't find any place where was passed param with other type. Can you please explain what the problem when you add `CustomerInterface` as a type?

arnoudhgz:
@VladimirZaets  I had this first as `CustomerInterface` but than the unit tests said that a `DataObject` was given. The `CustomerRepositoryInterface` has a return type of `CustomerInterface` but it actually returns a `DataObject`. 

That is why I changed it to `DataObject`.

_When using `CustomerInterface`:_
image

VladimirZaets:
@arnaucoll I think the problem is in the unit test that returns DataObject. What test is failed?

arnoudhgz:
@VladimirZaets  please see my updated comment with screenshot ^^

VladimirZaets:
@arnoudhgz The problem is in the test. 
In [this](https://github.com/magento/magento2/blob/6b7a9b3d3529ea9518fab12d0852afb3c73dc649/app/code/Magento/ProductAlert/Test/Unit/Model/ObserverTest.php#L299) line uses `DataObject` instead of `CustomerInterface` to create the customer mock. 

Can you please fix it?

arnoudhgz:
@VladimirZaets it seems solved now. There is only the timeout in Travis.

VladimirZaets:
@arnoudhgz Thanks for fixes

VladimirZaets:
Hi @arnoudhgz, can you please make refactoring and use private properties instead of constants.

arnoudhgz:
@VladimirZaets I can, but it seems wrong, I think the class `Magento\ProductAlert\Model\Observer` should use `EMAIL::TYPE_PRICE` abd `EMAIL::TYPE_STOCK` instead of hardcoded strings in `$email->setType('price');` and `$email->setType('stock');`.

Do you think I should maybe refactor this Observer also to make the code better (remove complexities)?

arnoudhgz:
@VladimirZaets are you really sure I should use private properties? These values are constants imho, they can't be changed with setters or inside other functions. 

According to the defnition of constants on the PHP website:
> A constant is an identifier (name) for a simple value. As the name suggests, that value cannot change during the execution of the script

VladimirZaets:
Hi @arnoudhgz. I discussed this question and we take decisions what we should do in the similar cases. 
We should use constants if value uses in the few classes. If their value uses in the scope of the class we should use private static declaration.

arnoudhgz:
@VladimirZaets thank you for the explanation, after I changed it to private static variables some test failed in Travis. It was about having to many variables on the class (the maximum is 15). So I changed the code to compare on strings (price or stock).

VladimirZaets:
Hi @arnoudhgz. I discussed this issue with my teammates and we take the decision that we can't add strict type and type checking to the already existing files due to [backward-compatible reason](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/). It's my fail, sorry. Can you please revert the changes related to strict types. Thanks

arnoudhgz:
@VladimirZaets  I will change it back.... :+1:

arnoudhgz:
@VladimirZaets I removed the strict types checking + the return types and parameter typehinting for existing functions.

arnoudhgz:
@VladimirZaets  do I still need to update the code? Could you provide some feedback?

VladimirZaets:
Hi @arnoudhgz. Sorry for the delay, we took your PR for delivery

magento-engcom-team:
Hi @arnoudhgz. Thank you for your contribution. 

PR #16800, opened 2018-07-14
[2.2-dev] Move functions.php into Framework

ghost:
See discussion and comments in #16705 - this is a pull request for 2.2

### Contribution checklist

### DIFF
--- a/app/functions.php
+++ b/app/functions.php
@@ -6,17 +6,22 @@
 
 /**
  * Create value-object \Magento\Framework\Phrase
- *
+ * @deprecated The global function __() is now loaded via Magento Framework, the below require is only
+ *             for backwards compatibility reasons and this file will be removed in a future version
+ * @see        Magento\Framework\Phrase\__.php
+ * @SuppressWarnings(PHPMD.ShortMethodName)
  * @return \Magento\Framework\Phrase
  */
-function __()
-{
-    $argc = func_get_args();
+if (!function_exists('__')) {
+    function __()
+    {
+        $argc = func_get_args();
 
-    $text = array_shift($argc);
-    if (!empty($argc) && is_array($argc[0])) {
-        $argc = $argc[0];
+        $text = array_shift($argc);
+        if (!empty($argc) && is_array($argc[0])) {
+            $argc = $argc[0];
+        }
+
+        return new \Magento\Framework\Phrase($text, $argc);
     }
-
-    return new \Magento\Framework\Phrase($text, $argc);
 }
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -31,8 +31,6 @@ require_once __DIR__ . '/autoload.php';
 // Sets default autoload mappings, may be overridden in Bootstrap::create
 \Magento\Framework\App\Bootstrap::populateAutoloader(BP, []);
 
-require_once BP . '/app/functions.php';
-
 /* Custom umask value may be provided in optional mage_umask file in root */
 $umaskFile = BP . '/magento_umask';
 $mask = file_exists($umaskFile) ? octdec(file_get_contents($umaskFile)) : 002;
--- a/dev/tests/unit/framework/bootstrap.php
+++ b/dev/tests/unit/framework/bootstrap.php
@@ -11,7 +11,6 @@ if (!defined('TESTS_TEMP_DIR')) {
 }
 
 require_once __DIR__ . '/autoload.php';
-require BP . '/app/functions.php';
 
 setCustomErrorHandler();
 
--- a/lib/internal/Magento/Framework/Phrase/__.php
+++ b/lib/internal/Magento/Framework/Phrase/__.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+/**
+ * Create value-object \Magento\Framework\Phrase
+ * @SuppressWarnings(PHPMD.ShortMethodName)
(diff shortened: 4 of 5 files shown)


magento-engcom-team:
Hi @fooman. Thank you for your contribution 

fooman:
Updated the PR to fix the noted short name violation of __.php  (squashed into 1 commit)

lib/internal/Magento/Framework/Phrase/__function.php
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+/**
+ * Create value-object \Magento\Framework\Phrase
+ *
+ * @return \Magento\Framework\Phrase
+ */
+function __()
```
ihor-sviziev: Wasn't idea to put this function in Magento/Framework/Phrase/ - `\Magento\Framework\Phrase` namespace?

fooman: I don't think that was proposed so far. Wouldn't namespacing the function mean all existing templates would need to be adjusted?

ihor-sviziev: Let's wait @orlangur for review

fooman:
@orlangur any thing else to do here before it can be merged?

fooman:
@orlangur or @ihor-sviziev anything else needed here?

orlangur:
@fooman sorry for delay, started review almost immediately but didn't finish it. Will do now.

lib/internal/Magento/Framework/Phrase/__function.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
orlangur: Which rule was violated by `__.php` name?

lib/internal/Magento/Framework/Phrase/registration.php
```diff
@@ -0,0 +1,4 @@
+<?php
+if (!function_exists('__')) {
+    require '__function.php';
```
orlangur: What is the use for such registration file?

lib/internal/Magento/Framework/composer.json
```diff
@@ -45,7 +45,8 @@
             "Magento\\Framework\\": ""
         },
         "files": [
-            "registration.php"
+            "registration.php",
+            "Phrase/__function.php"
```
orlangur: I meant to just require `__.php` in registration of framework.

app/functions.php
```diff
@@ -1,22 +1,15 @@
...
- */
-function __()
-{
-    $argc = func_get_args();
-
-    $text = array_shift($argc);
-    if (!empty($argc) && is_array($argc[0])) {
-        $argc = $argc[0];
+$vendorDir = require VENDOR_PATH;
+if (!function_exists('__')) {
```
orlangur: Why not simply move such logic into `__.php`? It's better to mark `app/functions.php` as deprecated without any changes.

orlangur (changes requested):

lib/internal/Magento/Framework/Phrase/__function.php
fooman: It was actually the function name itself `__()` which triggered the violation. As this was previously not occurring I assumed it was related to the file name. It turned out that wasn't the case and that __() was always in violation but making the changes brought this up.

If you have a strong preference I can change the filename __function.php back to __.php.

lib/internal/Magento/Framework/Phrase/registration.php
fooman: The main idea of splitting this out is the following:

The standard way of installing Magento 2 should be through Composer. In this case you will get the __function.php included directly through the Composer autoloader - no further requires needed. Also this registration file would not be loaded.

Only if you clone from Github `app/etc/NonComposerComponentRegistration.php` will kick in and find this registration file which in turn will require __function.php.

fooman: @orlangur we could mark this file deprecated as well as it should get removed at the same time as app/functions.php

lib/internal/Magento/Framework/composer.json
fooman: To me the final goal should be to use Composer to supply autoloaded files. Using registration.php is an intermediate step which we could do without. It also keeps the purpose of the registration file.

lib/internal/Magento/Framework/Phrase/__function.php
orlangur: Please do so.

lib/internal/Magento/Framework/Phrase/registration.php
orlangur: Such file is a bit confusing, please just include `__.php` in `registration.php` of Magento Framework.

lib/internal/Magento/Framework/composer.json
orlangur: > Using registration.php is an intermediate

There is no need to do something differently until we get rid of them.

So, instead of including `__.php` separately just do it within `registration.php`.

fooman:
@orlangur okay changed this to how you suggested so that we can get this merged

orlangur (approved):
@fooman great! Even simpler than I thought it will be 👍

magento-engcom-team:
[ENGCOM-2915](https://jira.corp.magento.com/browse/ENGCOM-2915) has been created to process this Pull Request

magento-engcom-team:
Hi @fooman. Thank you for your contribution. 

PR #17004, opened 2018-07-22
Fix for the checkout-breaking bug, related to the state/province field. Issue: 14052

ivancukns:
### Description

The reason I made this update is because the variable $allowedRegions is always an empty array in cases when a country which doesn't have a selection of regions for choosing is selected by users.

This elseif is checking if the region_id is in_array, but the array is empty in these cases and will always trigger the error. Also, I believe that regions collection does not exist for countries for which users have to type the name of their region, so this elseif only has sense for countries with region collections, otherwise it is always triggering the error.

1. magento/magento2#14052: M2.2.3 Checkout - Once integer is stored for State/Province field, it can not be changed to alphanumeric

### Manual testing scenarios

1. While on the checkout page, select the US as the shipping country and select a state.
2. Change the shipping country to the United Kingdom again.
3. It should be possible to complete the checkout proccess, while this is not possible without this change.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Model/Address/AbstractAddress.php
+++ b/app/code/Magento/Customer/Model/Address/AbstractAddress.php
@@ -648,7 +648,7 @@ class AbstractAddress extends AbstractExtensibleModel implements AddressModelInt
                     //If country actually has regions and requires you to
                     //select one then it must be selected.
                     $errors[] = __('%fieldName is a required field.', ['fieldName' => 'regionId']);
-                } elseif ($regionId && !in_array($regionId, $allowedRegions, true)) {
+                } elseif ($regionId && !in_array($regionId, $allowedRegions, true) && !empty($allowedRegions)) {
                     //If a region is selected then checking if it exists.
                     $errors[] = __(
                         'Invalid value of "%value" provided for the %fieldName field.',


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=17004) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @ivancukns. Thank you for your contribution 

VladimirZaets:
Hi @ivancukns. Thanks for the collaboration. As I understand, if `$allowedRegions` is always empty that means that the region isn't required? 

If I am right, we can add the check to `$isRequired` property in the condition

ivancukns:
Hi @VladimirZaets, $allowedRegions is always empty when you choose a country which does not have a selection of regions offered to users to pick one during the checkout, and instead of a select element, they have a text input field and have to type their region name, like for example the UK. 

I am not sure is it possible to make the region field mandatory if its a text input field, but if yes, then the $isRequired check wouldnt work for this issue.

sidolov:
Hi @ivancukns , looks like described issue already fixed by Magento core team in commit: 5b87891f5c58a8f8026c05dbc468852f8860afbc
I'm closing this PR as not relevant. 
Thank you for the contribution!

PR #17104, opened 2018-07-24
[Backport] Magento UI - Cleanup of undefined mixins parameters and usage of "leaking" variables scope

mage2pratik:
### Original Pull Request 
 https://github.com/magento/magento2/pull/11371
# Backstory
It's not a secret that I don't like LESS, but because of SASS Blank project, I often have to deep dive into Magento UI LESS library code.
Unfortunately, 9 of 10 times I got a punch of weirdness of language "features" mixed with the inconsistency of the implementation right in the face. 
That's why I decided to do something with it and clean up most annoying places.

It may be considered as heavily opinionated changes, but all of this code is 100% safe and didn't change the output CSS, the goal is to make working on this code less painful, especially for porting into languages that care more about scope and execution order, but it also should help people working directly on LESS, because all of the changes can be considered as a good practice in any programing language.

# Technical reason of changes
In LESS variables have kinda weird scope rules, which I'm considering as a leaking scope and the source of confusion i.e. functions (mixins) invoked inside function declaration inherits all variables of the parent function.

An example in a JS-like code to show weirdness of the LESS scoping:
```js
function parent (param) {
  child();
}

function child() {
  return param;
}

parent('red'); // it will return 'red', but should throw an error of undefined variable
```
It's not a common practice in other languages used in FE development and it not clear for developers, so will be nice to avoid it.

The goal is to pass all necessary values to mixin only through parameters or using global variables as default values of parameters.

Same example, but with this new rules applied:

```js
const sample = 'red';

function parent (param = sample) {
  child(param);
}

function child(param) {
  return param;
}

parent(); // it will return 'red'
parent('blue'); // it will return 'blue'
```

# Description of changes
I removed all fallbacks to variables not existing in the global scope, defined all variables used inside mixins as parameters and added all missing parameters to the places where mixins are invoked.
Also mixin that was used just once, was moved and simplified, to reduce usage of weird LESS scoping, where variables defined in mixin are accessible in the place where mixin is invoked.

### DIFF
--- a/lib/web/css/source/lib/_icons.less
+++ b/lib/web/css/source/lib/_icons.less
@@ -285,6 +285,10 @@
     .lib-icon-text-hide();
 }
 
+._lib-icon-image-url(@_path) when not (@_path = false) {
+    .lib-css(background-image, url(@_path));
+}
+
 ._lib-icon-image(
     @_icon-image,
     @_icon-image-width,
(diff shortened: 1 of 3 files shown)


VladimirZaets:
Hi @mage2pratik, thanks for collaboration. The changes in PR are not applicable to Magento 2.1 version and can effect existing functionality.

PR #17387, opened 2018-08-06
increase length of sequence_table column in sales_sequence_meta table

jhruehl:
### Description

Having a too small length for the sequence_table column can lead to integrity constraint errors on setup:upgrade, because longer table names will be cut off (at least the store id suffix of those sales sequence table names).
In our case it was the PostfinanceCW extension, which lead to the tables "sequence_postfinancecw_transaction_0", "sequence_postfinancecw_transaction_1",... and so on (for each store id one table), which are 36 characters long.
So the names for the admin store view (id 0) got cut off in the sales_sequence_meta table and as a consequence lead to an integrity constrained, when the recurring script in the Magento_SalesSequence module tries to add the meta data for the sequence_postfinancecw_transaction_1 table.

So a character length of only 32 is definitely not long enough. I guess 50 character should suffice though. ¯\_(ツ)_/¯

### DIFF
--- a/app/code/Magento/SalesSequence/etc/db_schema.xml
+++ b/app/code/Magento/SalesSequence/etc/db_schema.xml
@@ -41,7 +41,7 @@
         <column xsi:type="varchar" name="entity_type" nullable="false" length="32" comment="Prefix"/>
         <column xsi:type="smallint" name="store_id" padding="5" unsigned="true" nullable="false" identity="false"
                 comment="Store Id"/>
-        <column xsi:type="varchar" name="sequence_table" nullable="false" length="32" comment="table for sequence"/>
+        <column xsi:type="varchar" name="sequence_table" nullable="false" length="64" comment="table for sequence"/>
         <constraint xsi:type="primary" referenceId="PRIMARY">
             <column name="meta_id"/>
         </constraint>


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=17387) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @jhruehl. Thank you for your contribution 

orlangur (changes requested):
Why those table names are not passed through `getTableName` performing a truncation + some MD5 addition?

jhruehl:
@orlangur Just change the field to text or increase the length to 256. The number of rows in this table will probably never come near 200 in whatever reasonable setup. And even with 1000 rows you'll have to severly misconfigure your database for that to have a perfomance impact in any way.

Furthermore hashing (or adding a hash) may seem to be a solution, but it's not good for readability.

But if you still object, feel free to think of a solution, which includes 32 characters, where the important information is still readable and which doesn't lead to problems or has to be overhauled in the long run.

orlangur:
@jhruehl I mean there is already a mechanism to not have too long table names, just wondering why it wasn't used here.

sidolov:
Hi @jhruehl , I am closing this PR now due to inactivity.

jhruehl:
@sidolov Actually I've done my part and offered a solution, which doesn't hurt and just increases the length of a column in a table, which will never reach a size where this could have an impact.

And because I guess no one on your side has the time or is willing to do react to orlangur's question, I'm definitely not interested (not my job), I guess it's better to initially accept my pull request than letting the issue be lost in nothingness over time and having people raising questions every now and then.

To summarize my opinion in more clear word: "It's just an increase of a stupid column lenght value, which type should have been text anyway." Hier wird aus einer Mücke ein Elefant gemacht.

heldchen:
ran into this today. sequence name is silently truncated by myslq, thus getting 
```
  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'postfinancecw_transaction-0' for key 'SALES_SEQUENCE_META_ENTITY_TYPE_STORE_ID', query was: INSERT INTO `sales_sequence_meta` (`entity_type`, `store_id`, `sequen
  ce_table`) VALUES (?, ?, ?)
```
during a `setup:db-schema:upgrade` as sequence table is truncated to 32 chars. this pull request should have been accepted.

orlangur:
@jhruehl it was determined in issue you reported that correct value is 64 characters.

Could you please change this value and amend your commit?

jhruehl:
@orlangur I've changed it to 64, ran into a merge conflict and corrected it by comparing the given version of the db_schema.xml file in the current 2.3.0-develop branch. Hope this suffices.

orlangur:
@jhruehl oh, now there are four commits :) Can you please squash into single one and force push? I can assist with that if needed.

orlangur:
@jhruehl oh, now there are four commits :) Can you please squash into single one and force push? I can assist with that if needed.

VladimirZaets (dismissed):

magento-engcom-team:
[ENGCOM-3757](https://jira.corp.magento.com/browse/ENGCOM-3757) has been created to process this Pull Request

orlangur:
@jhruehl could you please update branch as requested?

@VladimirZaets it is not a good idea to update a number in few consecutive commits.

p-bystritsky:
@orlangur done.

orlangur (approved):

magento-engcom-team:
Hi @jhruehl. Thank you for your contribution. 

PR #17660, opened 2018-08-17
Update getChildren in Magento\Catalog\Model\Category to have $asArray boolean parameter

cruxicheiros:
### Description
Give getChildren a boolean parameter, $asArray, to allow it to return a result as an array instead of as a string.

This makes it follow the same pattern as getAllChildren, which fulfils a similar purpose in the code. It also makes it possible to directly get and use an array of child IDs instead of needing to explode the returned string of child IDs (which is imploded within getChildren).

### Manual testing scenarios
1. Add a category to the database.
2. Add two subcategories.
3. Using CategoryRepository.get(), create a Category with the data of the category created in step 1.
4. Verify that the result of calling the Category's getChildren method with no parameters is a string
5. Verify that the result of calling the Category's getChildren method with $asArray = true is an array

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Category.php
+++ b/app/code/Magento/Catalog/Model/Category.php
@@ -796,11 +796,17 @@ class Category extends \Magento\Catalog\Model\AbstractModel implements
      * @param boolean $recursive
      * @param boolean $isActive
      * @param boolean $sortByPosition
-     * @return string
+     * @param boolean $asArray return result as array instead of comma-separated list of IDs
+     * @return array|string
      */
-    public function getChildren($recursive = false, $isActive = true, $sortByPosition = false)
+    public function getChildren($recursive = false, $isActive = true, $sortByPosition = false, $asArray = false)
     {
-        return implode(',', $this->getResource()->getChildren($this, $recursive, $isActive, $sortByPosition));
+        $children = $this->getResource()->getChildren($this, $recursive, $isActive, $sortByPosition);
+        if ($asArray) {
+            return $children;
+        } else {
+            return implode(',', $children);
+        }
     }
 
     /**


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=17660) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @cruxicheiros. Thank you for your contribution 

app/code/Magento/Catalog/Model/Category.php
```diff
@@ -796,11 +796,17 @@ public function getAllChildren($asArray = false)
      * @param boolean $recursive
      * @param boolean $isActive
      * @param boolean $sortByPosition
-     * @return string
+     * @param boolean $asArray return result as array instead of comma-separated list of IDs
+     * @return array|string
      */
-    public function getChildren($recursive = false, $isActive = true, $sortByPosition = false)
+    public function getChildren($recursive = false, $isActive = true, $sortByPosition = false, $asArray = false)
```
sidolov, line 802: We are not allowed to add new parameters to public methods according to [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#php)

sidolov (changes requested):

app/code/Magento/Catalog/Model/Category.php
cruxicheiros: This shouldn't create a compatibility issue because it defaults to `false` meaning the default behaviour is still the same.

orlangur: @cruxicheiros it does not matter, guide is clear about that. it is such strict to prevent a situation when additional parameters were added in custom code.

cruxicheiros: There is a method directly above this, [getAllChildren](https://github.com/magento/magento2/blob/48b3816ad6b3235a68976679200184d9e63120b0/app/code/Magento/Catalog/Model/Category.php#L783), which already follows the proposed pattern. So, I still hold that this change improves consistency.

However, there are alternate ways to resolve it. Would it be better as a seperate method - `getChildrenAsArray`? Or could it go into a breaking release?

sidolov: @cruxicheiros yes, separate method is the best way to resolve it

sidolov:
Hi @cruxicheiros , I am closing this PR now due to inactivity.

PR #17965, opened 2018-09-07
Fixes for add to cart button in translation csv

priti2jcommerce:
### Description

The "Add to Cart" Button text is not translating correctly after the product added to the cart.When product page load, the translation for the "Add to Cart" is working as expected and button not getting translated into "Add to Basket".

1. #3423

### Manual testing scenarios

1. On Product page goes through the process of being added to the cart as expected, when add to cart process is completed the "Add to Cart" button shows Add To Basket from the translation and this work in all version of 2.1.x also supporting all browsers.

### Contribution checklist

### DIFF
--- a/app/i18n/Magento/en_US/en_US.csv
+++ b/app/i18n/Magento/en_US/en_US.csv
@@ -1,0 +1,3 @@
+"Add to Cart","Add To Basket"
+"Adding","Adding Basket"
+"Added","Added Basket"
--- a/app/i18n/Magento/en_US/en_US.csv
+++ b/app/i18n/Magento/en_US/en_US.csv
@@ -1,3 +1,3 @@
 "Add to Cart","Add To Basket"
-"Adding","Adding Basket"
+"Adding...","Adding Basket"
 "Added","Added Basket"


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=17965) <br/>All committers have signed the CLA.

rogyar:
Hi @priti2jcommerce. Thank you for your contribution. The mentioned issue https://github.com/magento/magento2/issues/3423 is not about missing translation. The problem is the add to cart button is not translated correctly for some scenarios even if the translation exists. 
Also, the "basket" term is not used within a scope of Magento terminology. The term described in the issue is about a custom translation.

priti2jcommerce:
Hi rogyar,
I referred the following to create language package
https://devdocs.magento.com/guides/v2.3/config-guide/cli/config-cli-subcommands-i18n.html

I have implemented magento versions (2.1.8 and 2.2.5)  with custom theme and found that its working, 
And I have done with the following process for custom theme:

creating the language package in custom theme :
add directory app/i18n/vendor/package , add all of the following files:
composer.json
en_US.csv
registration.php 
language.xml 

And run the following commands:
php bin/magento setup:upgrade
php bin/magento setup:static-content:deploy

Please check the above mentioned steps with custom theme and let me know if further details needed.

rogyar:
Hi @priti2jcommerce. Please, refer to the following comment which describes the actual problem
https://github.com/magento/magento2/issues/3423#issuecomment-302946418

The problem is when a custom translation is added to the different states of the "Add to cart" button the system still does not translate the button title.

What has been done in the current PR: additional translations were added to the i18 module. Should the added translations change the incorrect system behavior for custom language packs somehow?  

According to your changes, we should be able to do the same in a custom language pack and have the correct translations. Which means, the original issue is not reproducible and it's enough to have proper translations in a custom language pack. Please, correct me if I'm wrong.

priti2jcommerce:
Hi @rogyar,

I have checked the comment here 
#3423 (comment) 

yes, you are right! we should be able to do the same in a custom language pack and have the correct translations.

priti2jcommerce:
I have used like this with custom theme.
[app.zip](https://github.com/magento/magento2/files/2382398/app.zip)

ishakhsuvarov:
Hi @priti2jcommerce 

Unfortunately, we can not accept different wording from the one already used system-wide. We are closing this PR now. Additionally, it's best to submit new fixes to the 2.2-develop or 2.3-develop branches, as it's challenging to keep releases aligned when initial fix is delivered to 2.1.

Please consider the comments above by @rogyar and use 2.2-develop as a target if you wish to continue. 

Thank you

Issue #4694, opened 2016-05-25
Magento 2.1 - Can't update custom attribute value for customer (account edit form)


tanghia:
Magento 2.0.2 -> 2.0.6 update value of custom attributes on account edit page successfully, but magento 2.1 can't update (no error message or javascript error).
### Steps to reproduce
1. Create account with custom attributes successfully
2. Login
3. Edit account information 
### Expected result
1. Value of custom attributes changed
### Actual result
1. New values is not update.

Magento also hardcode some attribute_set_id in: CustomerMetadataInterface and AddressMetadataInterface (1 and 2). Magento 2.1 also change js and some template files (account page) which is not compatible with extensions and theme of old versions.

ravichandrarv:
i have created custom attributes for customer and product......now i want to make them read only  for some user and for some user make it editable.......

wert2all:
Thanks for reporting. Internal ticket MAGETWO-56126 was created.

antboiko:
Hi @tanghia , please let me inform you, this issue has been already fixed in a mainline and will be also available in one of the future patches.

I'm closing this ticket. Please feel free to reopen or create a new one.

Best,
Anton.

PR #17968, opened 2018-09-07
Fix Customer custom attributes lost after save

Thundar:
### Description

The customer model has an attribute `attribute_set_id` that links to its attribute set. It is used while saving, as there is a check to see if its attributes are in the set or not, using this link.
We want to avoid a `null` `attribute_set_id`.

### Fixed Issues 

1. magento/magento2#12479: Saving Customer Model directly causes loss of data

### Manual testing scenarios

1. Disable EAV cache.
2. Create a custom attribute for Customer entity.
3. Update a customer in the admin and add a value for this new attribute.
4. Programmatically load the customer, make some change (e.g. firstname) and then save directly using the model.
5. Reload the customer: the custom value is not lost.

No regression manual test:
6. Programmatically load the customer, add a different not-yet-defined attribute to customer (e.g. `setNotYetDefined(1)`) and save directly using the model.
7. Reload the customer: the custom value is still not lost, while `not_yet_defined` is lost.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Model/Address.php
+++ b/app/code/Magento/Customer/Model/Address.php
@@ -154,9 +154,6 @@ class Address extends \Magento\Customer\Model\Address\AbstractAddress
         // Need to explicitly set this due to discrepancy in the keys between model and data object
         $this->setIsDefaultBilling($address->isDefaultBilling());
         $this->setIsDefaultShipping($address->isDefaultShipping());
-        if (!$this->getAttributeSetId()) {
-            $this->setAttributeSetId(AddressMetadataInterface::ATTRIBUTE_SET_ID_ADDRESS);
-        }
         $customAttributes = $address->getCustomAttributes();
         if ($customAttributes !== null) {
             foreach ($customAttributes as $attribute) {
--- a/app/code/Magento/Customer/Model/Customer.php
+++ b/app/code/Magento/Customer/Model/Customer.php
@@ -358,13 +358,6 @@ class Customer extends \Magento\Framework\Model\AbstractModel
             $this->setId($customerId);
         }
 
-        // Need to use attribute set or future updates can cause data loss
-        if (!$this->getAttributeSetId()) {
-            $this->setAttributeSetId(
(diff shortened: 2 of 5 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=17968) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @Thundar. Thank you for your contribution 

Thundar:
Maybe related evolution: https://github.com/magento/magento2/issues/4677
Surely related: https://github.com/magento/magento2/issues/4694
By the comments, it should be already fixed on 2.3, but I cannot check it.

slavvka:
Hey @Thundar. Thank you for the contribution. I have a question though. As far as I can see `attribute_set_id` should be pre-populated on customer instance creation (and exists during whole customer object life cycle). That gives us required variable set and a possibility to extend. I am not sure that having the code pre-population run every time you save customer is a good idea. Please consider refactoring it to pre-populate this variable in the factory, for example. Moreover the address model also may have the similar issue, that is also would be good to have fixed. And in that case the original code that you copied from `updateData` would be also not needed. Does it make sense?

Thundar:
Hello @slavvka , thank you for your answer and the hints. They makes sense.
The nuisance is that the same code is both in `CustomerRepository::save` function and in `Customer::updateData`. I see that having a third replication would be worst.

The Address model benefits of a custom `getAttributeSetId` method (since MAGETWO-69111) that make it always not null: I added the same to the customer. This looks me the cleaner way (and it is easy to navigate). I think this is the best way to grant extendibility.
I also removed unused code.

slavvka:
Hey @Thundar. You are right. Thank you for the contribution. Would you mind to port the PR to 2.3 as well? You could do it now or after this PR is merged with the special tool.

slavvka (approved):

magento-engcom-team:
[ENGCOM-3082](https://jira.corp.magento.com/browse/ENGCOM-3082) has been created to process this Pull Request

magento-engcom-team:
Hi @Thundar. Thank you for your contribution. 

PR #18364, opened 2018-10-03
Avoid static methods in \Magento\Catalog\Model\Product\Visibility

DmitryFurs:
### Description
This pr will avoid [static methods](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Product/Visibility.php#L90) in `\Magento\Catalog\Model\Product\Visibility`. 

It looks like this code is simply [ported without refactoring]( https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Catalog/Model/Product/Visibility.php#L136)

In m2 these methods must be non-static public methods. Thus, we can replace this model via DI or create a plugin for it. But for now, we can't create a plugin for `getAllOption` method and put new options to all the places where magento uses this source model.

Even though this class is marked as an [@api ](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Product/Visibility.php#L14)

The second oddity: this model has special visibility getters
[getVisibleInCatalogIds](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Product/Visibility.php#L60)
[getVisibleInSearchIds](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Product/Visibility.php#L70)
[getVisibleInSiteIds](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Product/Visibility.php#L80)

But in some places modules[ use constants directly](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Indexer/Category/Product/AbstractAction.php#L319-L323)

And these getters are only for visible values, but we need not visible values in [some places](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/ResourceModel/Product.php#L408-L411)

No issues were found about this issue

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/Indexer/Category/Product/AbstractAction.php
+++ b/app/code/Magento/Catalog/Model/Indexer/Category/Product/AbstractAction.php
@@ -14,6 +14,7 @@ use Magento\Framework\DB\Query\Generator as QueryGenerator;
 use Magento\Framework\DB\Select;
 use Magento\Framework\EntityManager\MetadataPool;
 use Magento\Store\Model\Store;
+use Magento\Catalog\Model\Product\Visibility;
 
 /**
  * Class AbstractAction
@@ -113,6 +114,11 @@ abstract class AbstractAction
     protected $tableMaintainer;
 
     /**
+     * @var Visibility
+     */
+    protected $productVisibility;
+
+    /**
      * @var string
      * @since 101.0.0
      */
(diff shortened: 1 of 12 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=18364) <br/>All committers have signed the CLA.

orlangur:
Unfortunately such kind of changes is not allowed: https://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#prohibited-code-changes

PR #18385, opened 2018-10-04
Fix for #8854 - "'Add to cart' stuck 'adding'" on localhost

noahp78:
### Description
The new RFC for cookies that contain only 1 dot are invalid and should not be accepted on the browser side. however, there is a piece of client side javascript code that sets the cookie domain on some requests to "." + domain, this invalidates the cookies for localhost and causes the 'add to cart' button to fail due to the server not being informed about the new form_key.

This pull requests attempts to fix that by replacing all cookies that want to be for ".localhost" to "localhost"

1. magento/magento2#8854: 'Add to cart' stuck 'adding'

### Manual testing scenarios
1. Setup Magento 2 on 'localhost'.
2. Add a item
3. Add this item to cart
4. The item is successfully added to the cart.

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=18385) <br/>All committers have signed the CLA.

sivaschenko:
Hi, @noahp78 , thank you for contribution! However, I think don't think hardcoding the "localhost" in the code is a good practice.

By design, domain names must contain at least two dots; otherwise the browser will consider them invalid. (See reference on http://curl.haxx.se/rfc/cookie_spec.html)

Considering that the use case with "localhost" domain is actual only for development and can be easily worked around using the virtual host or IP address, I don't think it is necessary to include this change in the code base.

Please provide your feedback and/or close the pull request if you agree with the above.

noahp78:
While I agree on the fact that hard coding localhost may not be the best solution, I do think that it's a solution that needs to be clearly documentated or stated somewhere in the installation guides or fixed entirely.

I would also like to discus the particular reasoning behind overriding the form-key on the client side, is there any documentation for doing this? Since the whole idea of a CSRF token is that the server provides it and has authority over it. Would love to get some more info on this, since from my point of view it currently looks pretty pointless and it would solve this issue (and probably a whole lot of issues) completely. 

Also, the reference you listed is specificly on cookies. There is no (official) authority stating that a domain should contain however many dots. The RFC for cookies however does contain this clause. (See https://tools.ietf.org/html/rfc1034#section-3.5). 

Another solution for this issue would be to append ".local" to any domain that doesn't already contain a dot, since this is the recommended solution from the RFC
(https://tools.ietf.org/html/rfc2965 Page 2).

sivaschenko:
Thanks for the detailed response, @noahp78 !

I agree with you that the problem and solution should be addressed in the documentation. There is also an ability to contribute to documentation: (See https://github.com/magento/devdocs/blob/master/.github/CONTRIBUTING.md)

The good place to discuss the form-key behaviour or any other Magento functionality is [our forum](https://community.magento.com/). If you think the current form key functionality is causing an issue - it might be necessary to open an issue for the magento2 repository and discuss the possible solutions there.

sivaschenko:
Closing this pull request due to inactivity. Please feel free to contact me should you have any queries.

PR #18444, opened 2018-10-07
Issue #17508 - Change response redirect method to reference the corre…

jasonevans1:
### Description
Magento\Store\App\Response\Redirect::redirect method has the wrong interface declared for the response argument. The problem is that the 'setRedirect()' is not specified in \Magento\Framework\App\ResponseInterface but is instead not available until the next level of inheritance Magento\Framework\App\Response\HttpInterface.

Changed Magento\Store\App\Response\Redirect::redirect method so the $response argument is the type of \Magento\Framework\App\Response\HttpInterface. This fixes fixes the issue with the 'setRedirect()' not existing. Changed the unit tests.

1. magento/magento2#17508: Response/Redirect referring to wrong interface layer

### Manual testing scenarios
1. Tested that redirects are still working.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Store/App/Response/Redirect.php
+++ b/app/code/Magento/Store/App/Response/Redirect.php
@@ -177,12 +177,12 @@ class Redirect implements \Magento\Framework\App\Response\RedirectInterface
     /**
      * Set redirect into response
      *
-     * @param \Magento\Framework\App\ResponseInterface $response
+     * @param \Magento\Framework\App\Response\HttpInterface $response
      * @param string $path
      * @param array $arguments
      * @return void
      */
-    public function redirect(\Magento\Framework\App\ResponseInterface $response, $path, $arguments = [])
+    public function redirect(\Magento\Framework\App\Response\HttpInterface $response, $path, $arguments = [])
     {
         $arguments = $this->updatePathParams($arguments);
         $response->setRedirect($this->_urlBuilder->getUrl($path, $arguments));
--- a/app/code/Magento/Braintree/Test/Unit/Controller/Paypal/SaveShippingMethodTest.php
+++ b/app/code/Magento/Braintree/Test/Unit/Controller/Paypal/SaveShippingMethodTest.php
@@ -6,6 +6,7 @@
 
 namespace Magento\Braintree\Test\Unit\Controller\Paypal;
 
+use Magento\Framework\App\Response\HttpInterface;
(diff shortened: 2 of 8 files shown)


josefbehr:
@jasonevans1 Thank you for your contribution.

@maghamed The only possible problem I see is that we can still not be sure that the method actually gets called with a HttpInterface. I found 11 occurrences in the code base, and most often it's called with something like $this->getResponse(), which is not guaranteed to be a HttpInterface, but could in theory be any implementation of ResponseInterface, not providing the "setRedirect()" method. I think we should either add runtime tests or refactor those cases to prevent errors. What do you think?

jasonevans1:
I agree we should look at ways of preventing errors in these cases. Can you give me an example of how you like to handle this with either runtime tests or refactoring?

app/code/Magento/Store/App/Response/Redirect.php
```diff
@@ -177,12 +177,12 @@ public function updatePathParams(array $arguments)
...
      * Set redirect into response
      *
-     * @param \Magento\Framework\App\ResponseInterface $response
+     * @param \Magento\Framework\App\Response\HttpInterface $response
      * @param string $path
      * @param array $arguments
      * @return void
      */
-    public function redirect(\Magento\Framework\App\ResponseInterface $response, $path, $arguments = [])
+    public function redirect(\Magento\Framework\App\Response\HttpInterface $response, $path, $arguments = [])
```
sidolov, line 185: Unfortunately, we can't change the public method signature according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

sidolov (changes requested):

app/code/Magento/Store/App/Response/Redirect.php
jasonevans1: Since public method signatures cannot be changed should I close this issue and pull request?

sivaschenko: @jasonevans1 unfortunately those changes cannot be accepted as they are not backward compatible

sivaschenko:
Closed due to inactivity and backward incompatible changes that cannot be accepted to codebase (see [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/))

PR #18480, opened 2018-10-09
Fix newsletter subscription multiple callings

arendarenko:
### Description
When working with newsletter module I found when customers update their subscription, Magento did it two and even three times (I mean by counting calling the same method in Subscribe model). My PR fixes that, so it may improve performance and stability.

#### 1. When customers manages his subscription status from private office
1.1. Login as customer
1.2. Go to private office and click Edit on Newsletters block
1.3. Update customer subscription (if customer was subscribed then unsubscribe, otherwise subscribe)

**Expected result:**
Subscription status will be updated 1 time

**Actual result:**
Subscriber method "_updateCustomerSubscription" was called 3 times:
* 2 times from customer plugin (because in newsletter manage controller there is a customer saving method) - _without any changes_
* 1 times actually from newsletter manage controller - _with changes, the way it should be_

#### 2. When admin manages subscriber status from backend
2.1. Go to Customers and edit customer
2.2. Update customer subscription (if customer was subscribed then unsubscribe, otherwise subscribe)

**Expected result:**
Subscription status will be updated 1 time

**Actual result:**
Subscriber method "_updateCustomerSubscription" was called 2 times:
* 2 times from customer plugin - _during first call it's just updating current status, without any changes, and the second call is with changes, the way it should be_

### Contribution checklist

### DIFF
--- a/app/code/Magento/Newsletter/Controller/Manage/Save.php
+++ b/app/code/Magento/Newsletter/Controller/Manage/Save.php
@@ -75,13 +75,19 @@ class Save extends \Magento\Newsletter\Controller\Manage
             try {
                 $customer = $this->customerRepository->getById($customerId);
                 $storeId = $this->storeManager->getStore()->getId();
-                $customer->setStoreId($storeId);
+                $isNeedToSaveCustomer = false;
+                if($customer->getStoreId() !== $storeId) {
+                    $customer->setStoreId($storeId);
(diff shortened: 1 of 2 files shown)


arendarenko:
As we discussed with @ihor-sviziev this solution is not proper. Better solution is #18488.

PR #18814, opened 2018-10-25
Replace erroneous usage of unsData method

t-richards:
### Description (*)

This pull request replaces incorrect usages of the magic `unsData` method. Note that there is nothing inherently wrong with magic methods like `unsData()`, but calling the `uns` or `has` variants with any arguments is a mistake.

For example,

```php
// Syntactically correct, but probably doesn't do what you think it does
$object->unsData('_form_key');
//          ^^^^
// __call method takes the key from the method name, not from the method argument
```

The above statement is functionally equivalent to the following code:

```php
// We're unsetting the "data" key which is not what we wanted.
// This usually results in no data being unset from the object 
$object->unsetData('data');
```

Since the `unsData` method is being called with an argument, it is almost certainly the case that the original author(s) intended to use the `unsetData` method instead.

```php
// Corrected code
$object->unsetData('_form_key');
```

N/A

1. Expect for all test results to stay exactly the same.
2. If any test results change, then the tests should be re-examined for correctness.

### DIFF
--- a/app/code/Magento/GroupedProduct/Model/Product/Type/Grouped.php
+++ b/app/code/Magento/GroupedProduct/Model/Product/Type/Grouped.php
@@ -236,7 +236,7 @@ class Grouped extends \Magento\Catalog\Model\Product\Type\AbstractType
      */
     public function flushAssociatedProductsCache($product)
     {
-        return $product->unsData($this->_keyAssociatedProducts);
+        return $product->unsetData($this->_keyAssociatedProducts);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=18814) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @t-richards. Thank you for your contribution 

orlangur:
@t-richards nice catch! Need to check if there are similar occurrences in non-public code.

slavvka:
@orlangur Probably it makes sense to add some logic which checks if `uns***()` method is called with arguments? And preferably to fallback to `unset('***')` method?

slavvka:
@orlangur and I see no reason on holding this PR on since it is a fix anyway :)

orlangur (approved):
No such occurrences in B2B/EE code.

orlangur:
@slavvka I would rather throw an `InvalidArgumentException` but this error is quite rare thus I don't think we really need to change `DataObject`.

magento-engcom-team:
[ENGCOM-3401](https://jira.corp.magento.com/browse/ENGCOM-3401) has been created to process this Pull Request

magento-engcom-team:
Hi @t-richards. Thank you for your contribution. 

PR #18908, opened 2018-10-28
[Backport] fixed - Unable to select payment method according to country of the address at checkout time

ghost:
### Description (*)

I think we should remove check !quote.billingAddress() in file Magento/Checkout/view/frontend/web/js/model/shipping-save-processor/default.js

Due to this we are getting address which is selected first time.

1. magento/magento2#18907: Unable to select payment method according to country of the address at checkout time

1. Allow 2 payment methods from backend. like Bank Transfer and Purchase Order
2. Allow Bank Transfer method only for United States.
3. Allow Purchase Order method only for India
4. In frontend, login with customer account.
5. Add 2 address in customer account. one address with country India and other with country United States.
6. Now add product in Cart.
7. Go to Checkout page.
8. Select address with country United States. payment methods will appear
9. Now click on shipping section.
10. Select address with country India

Backport for https://github.com/magento/magento2/pull/18909

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/js/model/shipping-save-processor/default.js
+++ b/app/code/Magento/Checkout/view/frontend/web/js/model/shipping-save-processor/default.js
@@ -35,9 +35,8 @@ define([
         saveShippingInformation: function () {
             var payload;
 
-            if (!quote.billingAddress()) {
-                selectBillingAddressAction(quote.shippingAddress());
-            }
+            /* Assign selected address every time buyer selects address*/
+            selectBillingAddressAction(quote.shippingAddress());
 
             payload = {
                 addressInformation: {


magento-engcom-team:
Hi @rahulwebkul. Thank you for your contribution 

PrithweemaD:
I was facing the same issue in Magento 2.2.1. The basic issue is solved, but there is another issue, If I click on other addresses or input new shipping address, the payment methods are not reloading i.e. required for changing the available methods. Please help.

ghost:
@PrithweemaD - I will look into the issue.

VladimirZaets (approved):

magento-engcom-team:
[ENGCOM-3348](https://jira.corp.magento.com/browse/ENGCOM-3348) has been created to process this Pull Request

magento-engcom-team:
Hi @rahulwebkul. Thank you for your contribution. 

luukschakenraad:
@VladimirZaets @rahulwebkul This pull requests causes the billing address to be reset to the shipping address after every shipping information save.

In our case for example we build a store pick up function. After you select a store, we do a saveShippingInformation-call and set the address of the store as shipping address. 

After this fix, every time we do this, the billing address of the customer is set to the shipping address of the store. 

I think the original code is correct. You only want to set the billing address to be the same as the shipping address if no billing address is set. If a billing address is set, it should not change to be the same as the shipping address.

@VladimirZaets I would suggest to revert this change. 
@rahulwebkul I would suggest to find another solution for your issue.

vovayatsyuk:
Agree with @luukschakenraad. 

It's not convenient for customers in current implementation. Take a look at how it's looks like in 2.2.8 (the only Magento version with this patch):

1. Fill shipping info
2. Fill different billing address
3. Return to the shipping info step, don't change anything. Just press "Continue".
4. Billing address is equal will shipping :no_good_man:  

I think the correct way of this PR is something like this (didn't test it, though):

```js
if ((!quote.billingAddress() ||
        quote.billingAddress().isAddressSameAsShipping()) && 
    quote.shippingAddress().canUseForBilling()) {

    selectBillingAddressAction(quote.shippingAddress());
}
```

PR #19047, opened 2018-11-02
Refactoring for Customer\Controller\Address\FormPost

mrVrAlex:
### Description (*)
Initially this should be fix for #5013 
But current implementation saving customer addresses by storefront is different then REST API or Adminhtml controller:  uses AddressRepository::save instead if CustomerRepository::save as aggregate root saving.

1. #5013 
2. New found issue: 
During saving address with VAT & group changing implemented in https://github.com/arijit-de/magento2/blob/2.3-develop/app/code/Magento/Customer/Observer/AfterAddressSaveObserver.php observer was run code `$customer->save()` which have OLD address collection which loaded in memory before new address data persist to database. So customer address re-saved with old data on `$customer->save()` call.

    1. First pull a customer with using REST API and note down the updated time.
    2. Login to Frontend with the same account and edit the addresses.
    3. Pull the same again and see the updated time also got changed.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=19047) <br/>All committers have signed the CLA.

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -59,6 +65,7 @@ class FormPost extends \Magento\Customer\Controller\Address implements HttpPostA
      * @param PageFactory $resultPageFactory
      * @param RegionFactory $regionFactory
      * @param HelperData $helperData
+     * @param CustomerRepositoryInterface
```
lbajsarowicz: Missing variable name with the parameter

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -197,8 +206,24 @@ public function execute()
...
+            $addressId = $this->getRequest()->getParam('id');
+            $customerId = $this->_getSession()->getCustomerId();
+
+            $customer = $this->customerRepository->getById($customerId);
+
             $address = $this->_extractAddress();
-            $this->_addressRepository->save($address);
+
+            $addresses = [];
+            foreach ($customer->getAddresses() as $customerAddress) {
```
lbajsarowicz: How about using `array_filter()` in this specific case?

mrVrAlex:
We are not done with this on contribution day, so we will continue work on this in next 2-3 days: 
- need still rewrite unit tests
- refactor `$this->_addressRepository` usage (not needed inject anymore)
- fixes after review

app/code/Magento/Customer/Controller/Address/FormPost.php
mrVrAlex: Make sense, thanks!

dmitrijivanenko:
as it was mentioned before, during the investigation proccess of initial problem, were found several bugs: 

- The customer's `updated_at` field was not be changed not only by editing the address, but deleting too. So in order to fix the both issues, we refactored the `Magento\Customer\Controller\Address`, `Magento\Customer\Controller\Address\FormPost` and`Magento\Customer\Controller\Address\Delete` actions, now they use `CustomerRepository` to save the changes in address
-  Delete proccess was broken, couse the `Magento\Customer\Controller\Address\Delete` class was implementing `Magento\Framework\App\Action\HttpPostActionInterface`, so the delete action has to be initiated by `POST` call, but the FE `magento/module-customer/view/frontend/web/address.js` is not supporting it, so in order to keep this disscussion opened and fix the issue, temperally this implemntation was reverted as it was in previous versions, so now `Magento\Customer\Controller\Address\Delete` extends `\Magento\Customer\Controller\Address` only. 
- All needed implemantations for Unit Test suite were made.

mrVrAlex:
Ping @magento-engcom-team

app/code/Magento/Customer/Controller/Address.php
```diff
@@ -69,7 +69,7 @@
      * @param \Magento\Customer\Model\Session $customerSession
      * @param \Magento\Framework\Data\Form\FormKey\Validator $formKeyValidator
      * @param \Magento\Customer\Model\Metadata\FormFactory $formFactory
-     * @param \Magento\Customer\Api\AddressRepositoryInterface $addressRepository
```
sivaschenko: Removing the class constructor parameters and properties are backward incompatible changes, please add a new required parameter as optional to the end of the parameters list without removing existing ones.

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -96,11 +96,28 @@ public function __construct(
...
+        }
+
+        foreach ($customer->getAddresses() as $customerAddress) {
+            if ($customerAddress->getId() == $addressId) {
+                $existingAddressData = $this->getCustomerAddressMapper()->toFlatArray($customerAddress);
+            }
+        }
+
+        if (empty($existingAddressData) && $addressId) {
+            throw new \Exception();
```
sivaschenko: Please use the \Magento\Framework\Message\ManagerInterface to display the error message properly in the UI and provide a meaningful exception message

app/code/Magento/Customer/Controller/Address/Index.php
```diff
@@ -7,54 +7,45 @@
...
 
 /**
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 class Index extends \Magento\Customer\Controller\Address implements HttpGetActionInterface
 {
-    /**
-     * @var CustomerRepositoryInterface
-     */
-    protected $customerRepository;
```
sivaschenko, line 20: Renaming a property is backward incompatible change, it is ok to move the protected property to parent class, but please keep the name of the property

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -119,32 +136,13 @@ protected function _extractAddress()
...
         return $addressDataObject;
     }
 
-    /**
-     * Retrieve existing address data
-     *
-     * @return array
-     * @throws \Exception
-     */
-    protected function getExistingAddressData()
```
sivaschenko, line 147: Considering the length of _extractAddress() method I would keep the corresponding code in the getExistingAddressData() method

app/code/Magento/Customer/Test/Unit/Controller/Address/DeleteTest.php
```diff
@@ -5,6 +5,7 @@
  */
```
sivaschenko, line 5: Please fix the code styles in the test files

sivaschenko (changes requested):
Please see the code review comments

dmitrijivanenko:
Thank You for the review, will fix soon all your requested changes

app/code/Magento/Customer/Controller/Address.php
dmitrijivanenko: Have returned the `AddressRepository`, made the `CustomerRepository` optional one, as you asked. Also, improved the backward incompatible by adding the check for `CustomerRepository`, if it is not set, it uses the old (bugged) logic. The improved logic is covered with unit tests.

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: There is no message here, couse of security reason, this case is specific one, it means, that someone asked for change (`addressId` is set but no address with it in customer addresses were found), so in such case we are not showing any message (Such behavior was before this refactoring)

app/code/Magento/Customer/Controller/Address/Index.php
dmitrijivanenko: the old naming was returned

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: This function was returned, and was supplemented with logic which returns the `addressData` based on `AddressRepository` or `CustomerRepository`

app/code/Magento/Customer/Test/Unit/Controller/Address/DeleteTest.php
dmitrijivanenko: Sorry, did not get which styles should be fixed, the php-cs-fixer found anything based on `php.cs.dist` file in the root like the `Codacy/PR Quality Review` did too

dmitrijivanenko:
Requested changes were made

app/code/Magento/Customer/Controller/Address.php
```diff
@@ -89,12 +95,14 @@ public function __construct(
         \Magento\Framework\Reflection\DataObjectProcessor $dataProcessor,
         \Magento\Framework\Api\DataObjectHelper $dataObjectHelper,
         \Magento\Framework\Controller\Result\ForwardFactory $resultForwardFactory,
-        \Magento\Framework\View\Result\PageFactory $resultPageFactory
+        \Magento\Framework\View\Result\PageFactory $resultPageFactory,
+        ?\Magento\Customer\Api\CustomerRepositoryInterface $customerRepository
```
sivaschenko: I don't think this syntax (question mark) will work as intended. It will allow passing the argument of type `null`, but result in "Too few arguments" fatal error if no argument is passed. Please specify the default `null` value to make the new added dependency optional

app/code/Magento/Customer/Controller/Address/Delete.php
```diff
@@ -6,9 +6,7 @@
  */
 namespace Magento\Customer\Controller\Address;
 
-use Magento\Framework\App\Action\HttpPostActionInterface;
-
-class Delete extends \Magento\Customer\Controller\Address implements HttpPostActionInterface
+class Delete extends \Magento\Customer\Controller\Address
```
sivaschenko: Removing the interface is backward incompatible change. That might result in error on instances where this class is passed to a method that require `HttpPostActionInterface` parameter

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -89,18 +92,21 @@ public function __construct(
...
-            $resultPageFactory
+            $resultPageFactory,
+            $customerRepository
         );
     }
 
     /**
      * Extract address from request
      *
+     * @param null $customer
```
sivaschenko: Please specify correct parameter types

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -128,18 +134,36 @@ protected function _extractAddress()
...
+
+        if ($customer) {
+            foreach ($customer->getAddresses() as $customerAddress) {
+                if ($customerAddress->getId() == $addressId) {
+                    $existingAddressData = $this->getCustomerAddressMapper()->toFlatArray($customerAddress);
+                }
+            }
+
+            if (empty($existingAddressData) && $addressId) {
+                throw new \Exception();
```
sivaschenko: I still think there should be a meaningful exception message like: "The address does not exist anymore"

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -74,7 +76,8 @@ public function __construct(
         ForwardFactory $resultForwardFactory,
         PageFactory $resultPageFactory,
         RegionFactory $regionFactory,
-        HelperData $helperData
+        HelperData $helperData,
+        ?CustomerRepositoryInterface $customerRepository
```
sivaschenko: Please add a default value of `null` for this parameter

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -128,18 +134,36 @@ protected function _extractAddress()
 
     /**
      * Retrieve existing address data
+     * @param $customer
```
sivaschenko: Please specify parameter types

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -89,18 +92,21 @@ public function __construct(
...
 
     /**
      * Extract address from request
      *
+     * @param null $customer
      * @return \Magento\Customer\Api\Data\AddressInterface
+     * @throws \Exception
      */
-    protected function _extractAddress()
+    protected function _extractAddress($customer = null)
```
sivaschenko: Adding new parameters to protected methods is backward incompatible. Instead of adding parameters to protected methods, Create a new method with a new signature and deprecate the old method without changing it. https://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/

sivaschenko (changes requested):
Thanks for collaboration. Please see the review notes

app/code/Magento/Customer/Controller/Address/Delete.php
dmitrijivanenko: So, regarding this, do you think i have to report a new bug issue? i mean, i removed the interface, couse the current storefront use GET method (https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Customer/view/frontend/web/js/address.js#L65), or should i fix it with this PR?

sivaschenko: Using the controller to retrieve information via GET is correct behaviour and I believe the functionality will continue working even if Controller implements HttpPostActionInterface, however, removing an interface implementation is not backward compatible, so I would preserve it.

You could actually make this controller implement both HttpPostActionInterface and HttpGetActionInterface to mark it as get action

dmitrijivanenko:
code is updated

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3508](https://jira.corp.magento.com/browse/ENGCOM-3508) has been created to process this Pull Request

magento-engcom-team:
@shandyDev thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

app/code/Magento/Customer/Controller/Address/Delete.php
```diff
@@ -18,10 +18,16 @@ public function execute()
         $addressId = $this->getRequest()->getParam('id', false);
 
         if ($addressId && $this->_formKeyValidator->validate($this->getRequest())) {
+            $customer = $this->customerRepository->getById($this->_getSession()->getCustomerId());
```
sivaschenko, line 26: @shandyDev there is a fatal error - "Call to a member function getById() on null" that can be reproduced following the next steps:
* Create customer on frontend
* Go to My Account->Address Book
* Add customer address.
* Go to My Account->Address Book
* Click Add New Address.
* Fill required fields and click Save Address
* Click Delete Address in Additional Address Entries block.

dmitrijivanenko:

magento-engcom-team:
Hi @dmitrijivanenko. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @dmitrijivanenko, here is your [new Magento instance](https://pr-19047.instances.magento-community.engineering).
Admin access: [https://pr-19047.instances.magento-community.engineering/admin](https://pr-19047.instances.magento-community.engineering/admin)

app/code/Magento/Customer/Controller/Address/Delete.php
dmitrijivanenko: Reproduced on test instance, soon will be a fix

app/code/Magento/Customer/Controller/Address.php
```diff
@@ -89,12 +95,14 @@ public function __construct(
...
         \Magento\Framework\Controller\Result\ForwardFactory $resultForwardFactory,
-        \Magento\Framework\View\Result\PageFactory $resultPageFactory
+        \Magento\Framework\View\Result\PageFactory $resultPageFactory,
+        ?\Magento\Customer\Api\CustomerRepositoryInterface $customerRepository = null
     ) {
         $this->_customerSession = $customerSession;
         $this->_formKeyValidator = $formKeyValidator;
         $this->_formFactory = $formFactory;
         $this->_addressRepository = $addressRepository;
+        $this->customerRepository= $customerRepository;
```
sivaschenko: Please add a space before the "=" and get an instance of customer reporsitory via object manager
```suggestion
        $this->customerRepository = $customerRepository ?: \Magento\Framework\App\ObjectManager::getInstance()->get(\Magento\Customer\Api\CustomerRepositoryInterface::class);
```

dmitrijivanenko: isn't it too dirty? why my solution with `di.xml` doesnt fit?

sivaschenko: See https://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/#php for details

dmitrijivanenko: ok, i see.. sorry

sivaschenko: Thank you!

app/code/Magento/Customer/Controller/Address.php
```diff
@@ -29,6 +32,11 @@
      */
     protected $_addressRepository;
 
+    /**
+     * @var CustomerRepositoryInterface
+     */
+    protected $customerRepository;
```
sidolov, line 38: Please, make this property private

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -89,18 +96,44 @@ public function __construct(
...
+
+    /**
+     * Extract address from request
+     *
+     * @param CustomerInterface|null $customer
+     * @return AddressInterface
+     * @throws \Exception
      */
-    protected function _extractAddress()
+    protected function extractCurrentAddress(CustomerInterface $customer = null): AddressInterface
```
sidolov, line 134: Unfortunately, we can't change public and protected methods signature according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -89,18 +96,44 @@ public function __construct(
...
 
     /**
      * Extract address from request
      *
-     * @return \Magento\Customer\Api\Data\AddressInterface
+     * @return AddressInterface
+     * @throws \Exception
+     * @deprecated Use extractCurrentAddress instead
+     */
+    protected function _extractAddress(): AddressInterface
```
sidolov, line 111: Please, make this method private and avoid using underscores in method names

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -89,18 +96,44 @@ public function __construct(
...
+        return $this->extractCurrentAddress();
+    }
+    /**
+     * Retrieve existing address data
+     *
+     * @return array
+     * @throws \Exception
+     * @deprecated Use _getExistingAddressData instead
+     */
+    protected function getExistingAddressData(): array
```
sidolov, line 122: Please, make this method private

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -129,17 +162,35 @@ protected function _extractAddress()
     /**
      * Retrieve existing address data
      *
+     * @param CustomerInterface|null $customer
      * @return array
      * @throws \Exception
      */
-    protected function getExistingAddressData()
+    protected function _getExistingAddressData(CustomerInterface $customer = null): array
```
sidolov, line 179: Unfortunately, we can't change public and protected methods signature according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

app/code/Magento/Customer/Controller/Address/FormPost.php
```diff
@@ -221,14 +271,41 @@ public function execute()
         return $this->resultRedirectFactory->create()->setUrl($this->_redirect->error($url));
     }
 
+    /**
+     * Process address save, based on two repositories.
+     *
+     * @throws \Magento\Framework\Exception\LocalizedException
+     */
+    protected function saveAddress()
```
sidolov: Please, make this method private

sidolov (dismissed):

app/code/Magento/Customer/Controller/Address.php
dmitrijivanenko: this property is protected couse it is used in `Magento\Customer\Controller\Address\FormPost` and other which extends this base controller

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: there is nothing changed, the original `protected function _extractAddress()` is saved, it begins from 111 line. The `extractCurrentAddress` method is a new one.

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: Btw this is an original `protected function _extractAddress()` method

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: the same here, `getExistingAddressData` is original one `_getExistingAddressData` was added

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: the same here, `getExistingAddressData` is original one `_getExistingAddressData` was added

app/code/Magento/Customer/Controller/Address/FormPost.php
dmitrijivanenko: ok, will make a commit soon

dmitrijivanenko:
@sidolov You marked the PR as "Changes Requested", the conflict was solved, the Merging is not blocked anymore, what else should be changed?

sivaschenko:
Hi @dmitrijivanenko this PR does require changes, please take a look at code review comment by @sidolov

dmitrijivanenko:
Hi @sidolov , i've answered you on your comments, can you give any feedback?

app/code/Magento/Customer/Controller/Address.php
sidolov: We recommend using composition instead of inheritance according to our [Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html), that's why this property should be private. Please, make it private and define the new property in every child if you need it there.

sidolov (changes requested):
Please, see review comments

sidolov:
Hi @shandyDev , @dmitrijivanenko , I am closing this PR now due to inactivity.

PR #19176, opened 2018-11-12
Include custom source model options in attribute index - Up port #16727

rain2o:
### Description
This is an up-port of #16727 from 2.2 to 2.3

When a multiselect Product Attribute uses a custom source model those values are not included when indexing the attribute values for filters on the product grid. The result is those attributes cannot be used as a filter in the product grid. 

This was originally opened as [issue #417](https://github.com/magento/magento2/issues/417) but was closed as it was considered a feature request and moved to the forums [here](https://community.magento.com/t5/Magento-2-x-Feature-Requests-and/Attribute-Indexing-ignores-custom-source-model-options/idi-p/46851). The solution is based on @maderlock's comment [here](https://github.com/magento/magento2/issues/417#issuecomment-265146285) but replaced the use of `ObjectManager` with dependency injection of `AttributeRepositoryInterface`.

### Fixed Issues
1. magento/magento2#417 : Attribute Indexing ignores custom source model options

### Manual testing scenarios
1. Create a multiselect attribute with a custom source model. For a quick test I created this [sample module](https://github.com/rain2o/magento2-Rain2o_MultiselectSourceSample).
2. Within the attribute set Use in Layered Navigation to "Filterable (with results)" (if using the sample module this is already done.)
3. Select one or more values for the given attribute on two or more products in the same category.
4. View the category page. The new attribute should now be an available filter.

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=19176) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @rain2o. Thank you for your contribution 

rain2o:
The Travis CI build failed, but it looks like it just timed out. I don't see any errors. Is there a way to re-trigger the build?

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
```diff
@@ -297,6 +308,47 @@ protected function _prepareMultiselectIndex($entityIds = null, $attributeId = nu
...
+                ['attribute_id','entity_type_id', 'attribute_code']
+            )
+            ->where('attribute_id IN(?)', $attrIds)
+            ->where('source_model is not null');
+        $query = $select->query();
+
+        while ($row = $query->fetch()) {
+            try {
+                /** @var \Magento\Catalog\Model\ResourceModel\Eav\Attribute $attribute */
+                $attribute = $this->_attributeRepository->get($row['entity_type_id'], $row['attribute_code']);
```
sivaschenko: I think that loading each attribute one by one can have index performance implications on instances with a big amount of attributes that have custom source_models, is there a way to load them in bulk (i.e. using getList method of repository)?

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
```diff
@@ -297,6 +308,47 @@ protected function _prepareMultiselectIndex($entityIds = null, $attributeId = nu
...
+                $sourceModelOptions = $attribute->getOptions();
+                // Add options to list used below
+                foreach ($sourceModelOptions as $o) {
+                    $options[$row['attribute_id']][$o->getValue()] = true;
+                }
+            } catch (\BadMethodCallException $e) {
+                // Skip
+            } catch (\Magento\Framework\Exception\NoSuchEntityException $e) {
```
sivaschenko: Can you please clarify when the \BadMethodCallException and \Magento\Framework\Exception\NoSuchEntityException will arise, I think it's better to handle them somehow instead of just skipping (i.e. reporting to log file if the exception is thrown when instance data integrity is broken)

sivaschenko (changes requested):
Thanks for the contribution! Please take a look at the code review comments

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: @sivaschenko The `NoSuchEntityException` is to handle the potential of `$this->_attributeRepository->get()` throwing such an error. Would you recommend just logging an exception? 
The `BadMethodCallException` I think should be removed actually. I believe it was left over from the original code using `ObjectManager`. I don't see where that would get thrown anymore, so that's my bad. I can remove it and add some logging for the other.

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: @sivaschenko Good thinking. I'll see if I can change this around to use something like `getList` to accomplish the same filtering. Thanks for the suggestion!

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
sivaschenko: @rain2o thanks for the clarification, yes I do recommend logging NoSuchEntityException as the source of attribute information is internal and in this situation a missing attribute is an exceptional case

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: @sivaschenko Updated to use `getList`. That is so much better and cleaner.

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: @sivaschenko Due to the refactor to use `getList` this function no longer throws any exceptions that I'm aware of, so I have removed all try/catch blocks.

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
```diff
@@ -40,6 +53,8 @@ public function __construct(
         \Magento\Eav\Model\Config $eavConfig,
         \Magento\Framework\Event\ManagerInterface $eventManager,
         \Magento\Catalog\Model\ResourceModel\Helper $resourceHelper,
+        \Magento\Eav\Api\AttributeRepositoryInterface $attributeRepository,
```
sivaschenko: Please add new parameters to the end of the list and make them optional (= null) - see example here: https://github.com/magento/magento2/blob/1a805d05ccdffe5315b1fa1dd69dae46cc2a2af2/app/code/Magento/Cms/Controller/Adminhtml/Page/Save.php#L55-L64

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
```diff
@@ -234,6 +251,9 @@ protected function _prepareMultiselectIndex($entityIds = null, $attributeId = nu
             $options[$row['attribute_id']][$row['option_id']] = true;
         }
 
+        // Include custom source model options
+        $options = $this->getMultiSelectAttributeWithSourceModels($attrIds, $options);
```
sivaschenko: It will be better and more transparent to remove $options param for this function, retrieve the multiselet options and then merge these here

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
```diff
@@ -297,6 +317,39 @@ protected function _prepareMultiselectIndex($entityIds = null, $attributeId = nu
...
+        $criteria = $this->criteriaBuilder->create();
+        $attributes = $this->attributeRepository->getList(
+            ProductAttributeInterface::ENTITY_TYPE_CODE,
+            $criteria
+        )->getItems();
+        
+        foreach ($attributes as $attribute) {
+            $sourceModelOptions = $attribute->getOptions();
+            // Add options to list used below
+            foreach ($sourceModelOptions as $o) {
```
sivaschenko: Please use meaningful variable names, i.e. $option instead of $o

sivaschenko (changes requested):
Thanks for collaboration. Please see my review notes, also it would be great to ensure integration tests coverage of this change

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: @sivaschenko In order to do this the only option I can think of is to use [array union](http://php.net/manual/en/language.operators.array.php) to avoid a reindex of the `$options` array and maintain the keys as they are numeric. Do you have any other recommendations to approach this? If not then I'll push an update using union.

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: I understand the reason for this ask, but I was under the impression that the direct use of ObjectManager was discouraged as per the rules in the documentation [here](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/object-manager.html#usage-rules). And the warning indicates any usage in core classes is due to backwards compatibility and should be updated eventually. Is this an exception, or is there a particular reason why we should use it here?

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
sivaschenko: What do you think about `$options = array_replace_recursive($options, $this->getMultiSelectAttributeWithSourceModels($attrIds));`

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
sivaschenko: That is an exception metioned in the referenced guide: "You can use the ObjectManager to maintain backward compatibility for a constructor." We should use it here because there is no other way to ensure the properties are assigned and the implementation is backward compatible.

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: Ah, I was thinking that would result in the same issue as `array_merge` but it looks like it won't. I'll move forward with that. Thanks!

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Eav/Source.php
rain2o: Ok that makes sense, thanks for the explanation.

rain2o:
@sivaschenko Regarding the integration tests, do you have any suggestions for a source model? I can't think of any existing source models in the Catalog module that would work for the tests. Or would it be best to create a test source model in the test suite?

sivaschenko:
@rain2o if you cannot find appropriate source models for the test in the code base, a test source model would be fine

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3566](https://jira.corp.magento.com/browse/ENGCOM-3566) has been created to process this Pull Request

magento-engcom-team:
@rain2o thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

magento-engcom-team:
Hi @rain2o. Thank you for your contribution. 

PR #19184, opened 2018-11-13
Fixed magento text swatch switches product image even if attribute feature is disabled #16446

ravi-chandra3197:
### Description (*)

Fixed magento text swatch switches product image even if attribute feature is disabled #16446

1. magento/magento2#16446: Fixed magento text swatch switches product image even if attribute feature is disabled 

N/A

### DIFF
--- a/app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
+++ b/app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
@@ -192,6 +192,7 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
 
     /**
      * Get number of swatches from config to show on product listing.
+     *
      * Other swatches can be shown after click button 'Show more'
      *
      * @return string
@@ -231,6 +232,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * Get swatch attributes data.
+     *
      * @return array
      */
     protected function getSwatchAttributesData()
@@ -239,6 +242,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * Init isProductHasSwatchAttribute.
+     *
      * @deprecated 100.1.5 Method isProductHasSwatchAttribute() is used instead of this.
      *
      * @codeCoverageIgnore
@@ -367,6 +372,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * Get swatch product image.
+     *
      * @param Product $childProduct
      * @param string $imageType
      * @return string
@@ -387,6 +394,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * Check if product have image.
+     *
      * @param Product $product
      * @param string $imageType
      * @return bool
@@ -397,6 +406,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * Get configurable options ids.
+     *
      * @param array $attributeData
      * @return array
      * @since 100.0.3
@@ -456,8 +467,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * @inheritDoc
      * @deprecated 100.1.5 Now is used _toHtml() directly
-     * @return string
      */
     protected function getHtmlOutput()
     {
@@ -465,6 +476,8 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
     }
 
     /**
+     * Get media callback url.
+     *
      * @return string
      */
     public function getMediaCallback()
--- a/app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
+++ b/app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
@@ -182,6 +182,9 @@ class Configurable extends \Magento\ConfigurableProduct\Block\Product\View\Type\
                     $attributeDataArray
                 );
             }
+            if (isset($attributeDataArray['additional_data'])) {
+                $config[$attributeId]['additional_data'] = $attributeDataArray['additional_data'];
+            }
         }
 
         return $this->jsonEncoder->encode($config);
--- a/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
+++ b/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
@@ -753,7 +753,12 @@ define([
                     $widget.options.jsonConfig.optionPrices
                 ]);
 
-            $widget._loadMedia();
+            const checkAdditionalData = JSON.parse(this.options.jsonSwatchConfig[attributeId].additional_data);
+
+            if (checkAdditionalData.update_product_preview_image == 1) {
+                $widget._loadMedia();
+            }
+
             $input.trigger('change');
         },
 
--- a/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
+++ b/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
@@ -713,7 +713,8 @@ define([
                 $wrapper = $this.parents('.' + $widget.options.classes.attributeOptionsWrapper),
                 $label = $parent.find('.' + $widget.options.classes.attributeSelectedOptionLabelClass),
                 attributeId = $parent.attr('attribute-id'),
-                $input = $parent.find('.' + $widget.options.classes.attributeInput);
+                $input = $parent.find('.' + $widget.options.classes.attributeInput),
+                checkAdditionalData = JSON.parse(this.options.jsonSwatchConfig[attributeId]['additional_data']);
 
             if ($widget.inProductList) {
                 $input = $widget.productForm.find(
@@ -753,9 +754,7 @@ define([
                     $widget.options.jsonConfig.optionPrices
                 ]);
 
-            var checkAdditionalData = JSON.parse(this.options.jsonSwatchConfig[attributeId]['additional_data']);
-
-            if (1 == checkAdditionalData['update_product_preview_image']) {
+            if (checkAdditionalData['update_product_preview_image'] === '1') {
                 $widget._loadMedia();
             }
 
--- a/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
+++ b/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
@@ -753,9 +753,9 @@ define([
                     $widget.options.jsonConfig.optionPrices
                 ]);
 
-            const checkAdditionalData = JSON.parse(this.options.jsonSwatchConfig[attributeId].additional_data);
+            var checkAdditionalData = JSON.parse(this.options.jsonSwatchConfig[attributeId]['additional_data']);
 
-            if (checkAdditionalData.update_product_preview_image == 1) {
+            if (1 == checkAdditionalData['update_product_preview_image']) {
                 $widget._loadMedia();
             }
 
--- a/app/code/Magento/Catalog/Test/Mftf/Section/AdminCreateProductAttributeSection.xml
+++ b/app/code/Magento/Catalog/Test/Mftf/Section/AdminCreateProductAttributeSection.xml
@@ -13,6 +13,7 @@
         <element name="DefaultLabel" type="input" selector="#attribute_label"/>
         <element name="InputType" type="select" selector="#frontend_input"/>
         <element name="ValueRequired" type="select" selector="#is_required"/>
+        <element name="UpdateProductPreviewImage" type="select" selector="[name='update_product_preview_image']"/>
         <element name="AdvancedProperties" type="button" selector="#advanced_fieldset-wrapper"/>
         <element name="DefaultValue" type="input" selector="#default_value_text"/>
         <element name="Scope" type="select" selector="#is_global"/>
--- a/app/code/Magento/Swatches/Test/Mftf/Test/StorefrontFilterByImageSwatchTest.xml
+++ b/app/code/Magento/Swatches/Test/Mftf/Test/StorefrontFilterByImageSwatchTest.xml
@@ -41,6 +41,9 @@
         <!-- Select visual swatch -->
         <selectOption selector="{{AttributePropertiesSection.InputType}}" userInput="swatch_visual" stepKey="selectInputType"/>
 
+        <!-- Set Update Product Preview Image to Yes-->
+        <selectOption selector="{{AttributePropertiesSection.UpdateProductPreviewImage}}" userInput="Yes" stepKey="setUpdateProductPreviewImage"/>
+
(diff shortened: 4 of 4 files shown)


magento-engcom-team:
Hi @ravi-chandra3197. Thank you for your contribution 

app/code/Magento/Swatches/Helper/Data.php
```diff
@@ -117,7 +117,8 @@ public function __construct(
         UrlBuilder $urlBuilder,
         Json $serializer = null,
         SwatchAttributesProvider $swatchAttributesProvider = null,
-        SwatchAttributeType $swatchTypeChecker = null
+        SwatchAttributeType $swatchTypeChecker = null,
+        \Magento\Catalog\Model\ResourceModel\Eav\Attribute $attributeFactory
```
sivaschenko: Please add additional constructor parameters as options to keep backward compatibility

app/code/Magento/Swatches/Helper/Data.php
```diff
@@ -347,9 +349,26 @@ public function getProductMediaGallery(ModelProduct $product)
             return [];
         }
 
-        $resultGallery = $this->getAllSizeImages($baseImage);
+        $objectManager = \Magento\Framework\App\ObjectManager::getInstance();
+        $resource = $objectManager->get('Magento\Framework\App\ResourceConnection');
+        $attributeInfo = $this->_attributeFactory->getCollection();
+        if ($attributeInfo) {
+            $attributeInfo = $attributeInfo->join(array('catalog_eav_attribute' => $resource->getTableName('catalog_eav_attribute')), 'main_table.attribute_id = catalog_eav_attribute.attribute_id')
```
sivaschenko: Please avoid using objectManager to keep dependencies explicit. Please use \Magento\Catalog\Api\ProductAttributeRepositoryInterface::getList to retrieve attributes.

app/code/Magento/Swatches/Helper/Data.php
```diff
@@ -347,9 +349,26 @@ public function getProductMediaGallery(ModelProduct $product)
...
+                ->addFieldToFilter('catalog_eav_attribute.additional_data', array('like' => '%update_product_preview_image%'));
+
+            foreach ($attributeInfo as $attribute) {
+                $additional_data = json_decode($attribute['additional_data'], true);
```
sivaschenko: Please use camel case for variable names: $additionalData

sivaschenko (changes requested):
Please see the review comments. Also I think the attributes used to idetify "update_product_preview_image" should be retrived considering product context. Also, I believe it would be better to fix the issue in frontend implementation (See https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js)

ravi-chandra3197:
Hello @sivaschenko 
I have update change as per your suggestion can review it.

app/code/Magento/Swatches/Controller/Ajax/Media.php
```diff
@@ -56,21 +60,33 @@ public function __construct(
     public function execute()
     {
         $productMedia = [];
+        $searchCriteria = $this->searchCriteria->addFilter('additional_data', '%update_product_preview_image%', 'like')->create();
+        $attributeInfo = $this->attributeRepository->getList($searchCriteria);
+        $attributeInfo = $attributeInfo->getItems();
+        foreach ($attributeInfo as $attribute) {
+            $additionalData = json_decode($attribute['additional_data'], true);
```
sivaschenko: Please use serializer to decode the additional data

app/code/Magento/Swatches/Controller/Ajax/Media.php
```diff
@@ -39,11 +39,15 @@ public function __construct(
         Context $context,
         \Magento\Catalog\Model\ProductFactory $productModelFactory,
         \Magento\Swatches\Helper\Data $swatchHelper,
-        \Magento\PageCache\Model\Config $config
+        \Magento\PageCache\Model\Config $config,
+        \Magento\Framework\Api\SearchCriteriaBuilder $searchCriteria,
```
sivaschenko: Please add additional parameters as optional (= null)

app/code/Magento/Swatches/Controller/Ajax/Media.php
```diff
@@ -56,21 +60,33 @@ public function __construct(
     public function execute()
     {
         $productMedia = [];
+        $searchCriteria = $this->searchCriteria->addFilter('additional_data', '%update_product_preview_image%', 'like')->create();
```
sivaschenko: I cannot see the reference to specific attribute. Is this solution applicable in case with two swatch attributes, when the first swatch attribute should update the image and the second shouldn't?

sivaschenko (changes requested):
@ravi-chandra3197 thanks for your updates, please see the additional comments. Also can you please provide your feedback to my comment on the previous review

app/code/Magento/Swatches/Helper/Data.php
ravi-chandra3197: I have changed as per your suggestion

app/code/Magento/Swatches/Controller/Ajax/Media.php
ravi-chandra3197: If there are two swatch attributes when the first swatch attribute should update the image and second shouldn't then swatch image not change because one swatch attribute has update_product_preview_image = No

Any suggestion for it.

sivaschenko: I think that clicking the swatch of the first attribute, in this case, should still update the images because it has `update_product_preview_image` set to Yes.

ravi-chandra3197: so as per your suggestion if one attribute set update_product_preview_image to Yes then image change.
like size and color both have the attribute value update_product_preview_image now if any one set to yes then product image change.
if I am wrong then correct me.

sivaschenko: My understanding is the following:
* We have two swatch attributes (i.e. colour and size as in sample data) - there are two rows of swatches for a product on frontend.
* Colour attribute is configured `update_product_preview_image=yes`
* Size is configured `update_product_preview_image=no`

Then clicking any swatch in the first row (selecting colour) should change the product image, while clicking any swatch in the second row (selecting size) should not change the product image.

ravi-chandra3197: Now if we have color set update_product_preview_image=yes And Size set update_product_preview_image=No 
Now if click on color swatch then images change and if we click on size swatch then image not change.

sivaschenko: Do you mean it works that way on the current 2.3-develop ?

ravi-chandra3197: ok, I  have change my code as per your suggestion.

ravi-chandra3197: Hello @sivaschenko 
I have debugged code and I had found in ajax we do not pass any additional parameter for so any backend we can find ajax call comes from either size swatch or color swatch so it's complected to decide to update image preview or not
the solution has two possibility 
1. like And condition if   size swatch update_product_preview_image=yes and color swatch update_product_preview_image=yes,  all swatch attribute yes then update image
2. like Or condition if size swatch update_product_preview_image=Yes and color swatch update_product_preview_image=No,  either any swatch attribute yes then  update image 
so which solution is most effective?
or any additional suggestions?

sivaschenko: @ravi-chandra3197 Thanks for the information! I think it will be easier to implement the fix on the frontend side, we should not even perform a request for media if update_product_preview_image=No for the clicked swatch

ravi-chandra3197: @sivaschenko thanks for replay 
Front side, we have not additional information regarding swatch attribute additional data which is stored in searialized data so it's complected to get additional data of attribute in the front end.

sivaschenko: @ravi-chandra3197 is there a way to pass the required data to frontend side while initial rendering?

ravi-chandra3197: I have Try another way to achieve the same functionality

sivaschenko:
Hi @ravi-chandra3197 are you still working on this pull request?

ravi-chandra3197:
Hello @sivaschenko
I have try to find front-end solution.
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js)

sidolov:
Hi @ravi-chandra3197 , will you continue progress with the PR?

ravi-chandra3197:
Hello @sidolov 
I am working on it, try to find front-end solution using js.

https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js

sidolov:
@ravi-chandra3197 , I am closing this PR now due to inactivity.

app/code/Magento/Swatches/Controller/Ajax/Media.php
ravi-chandra3197: Hello @sivaschenko 
I have updated the code as per your suggestion and implement solution using swatch-renderer.js

https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js

ravi-chandra3197:
Hello @sivaschenko
I have updated the code as per your suggestion and implement solution using swatch-renderer.js

https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js

app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
```diff
@@ -183,6 +183,7 @@ public function getJsonSwatchConfig()
                 );
             }
         }
+        $config['additional_data'] = $attributesData;
```
sivaschenko: It's better to include only the necessary information (update_product_preview_image) to minimize the output size

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
```diff
@@ -1026,19 +1028,22 @@ define([
...
-                $widget.xhr = $.ajax({
-                    url: $widget.options.mediaCallback,
-                    cache: true,
-                    type: 'GET',
-                    dataType: 'json',
-                    data: mediaCallData,
-                    success: mediaSuccessCallback
-                }).done(function () {
+                var checkAdditionalData = this.options.jsonSwatchConfig.additional_data[currentAttributeId];
+                if (checkAdditionalData.update_product_preview_image == 1) {
```
sivaschenko: I think it's better to move this check to `_onClick` and `_onChange` (to handle select as well) functions and do not call `_loadMedia` at all in case product image update is set to No. (as there are two ways to load media - ajax and json config).
Also, with this approach, the `currentAttributeId` global variable is not needed.

sivaschenko (changes requested):
Hi @ravi-chandra3197 , thanks for update. Please see my review comments

app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
ravi-chandra3197: Hello @sivaschenko 
I have update app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
and pass the only additional_data to minimize output size

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
ravi-chandra3197: Hello @sivaschenko 
I have update app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js 
move check into _OnClick

ravi-chandra3197:
Hello @sivaschenko 
I have updated the code as your suggestion.

app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
```diff
@@ -182,6 +182,9 @@ public function getJsonSwatchConfig()
                     $attributeDataArray
                 );
             }
+            if (isset($attributeDataArray['additional_data'])) {
+            $config[$attributeId]['additional_data'] =  $attributeDataArray['additional_data'];
```
sivaschenko: Proper alignment makes code much easier to read

```suggestion
                $config[$attributeId]['additional_data'] = $attributeDataArray['additional_data'];
```

sivaschenko: Also, what do you think about assigning only 'update_product_preview_image' key?

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
```diff
@@ -697,6 +697,7 @@ define([
                 $label = $parent.find('.' + $widget.options.classes.attributeSelectedOptionLabelClass),
                 attributeId = $parent.attr('attribute-id'),
                 $input = $parent.find('.' + $widget.options.classes.attributeInput);
+                checkAdditionalData = JSON.parse(this.options.jsonSwatchConfig[attributeId].additional_data);
```
sivaschenko: It looks like an unnecessary variable. Even if it's used for readability it'd be better to move the declaration closer to the usage. Also, please use `const` statement to declare this variable

sivaschenko (changes requested):
@ravi-chandra3197 thanks for updates, please see my review comments

app/code/Magento/Swatches/Block/Product/Renderer/Configurable.php
ravi-chandra3197: Hello @sivaschenko 
in additional_data we get all data in serialize form
 {swatch_input_type: "text", update_product_preview_image: "1", use_product_image_for_swatch: 0}

app/code/Magento/Swatches/view/frontend/web/js/swatch-renderer.js
ravi-chandra3197: Hello @sivaschenko 
we can't use `const` because checkAdditionalData each time load only current attribute data only like if click on size swatch load only size data or click on color swatch load only color swatch data

sivaschenko: @ravi-chandra3197 I see only one assignment to this variable in the scope of the function, so the variable is immutable, `const` is fine from my point of view (even if was immutable the variable availability scope should be limited by `let` statement)
Please move the variable declaration to line 732 to be as close as possible to the usage of this variable

ravi-chandra3197:
Hello @sivaschenko
I have updated the PR code as per your suggestion can you review it.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-4825](https://jira.corp.magento.com/browse/ENGCOM-4825) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed 

**Before:**

before

**After:**

after

soleksii:
:heavy_check_mark: QA Passed

PR #19310, opened 2018-11-21
Framework Escaper - remove direct Object Manager usage

maciejslawik:
### Description (*)

This PR removes direct ObjectManager calls from ```\Magento\Framework\Escaper```. Additionally, it allows for injecting different implementations of ```\Psr\Log\LoggerInterface``` via the constructor.

### DIFF
--- a/lib/internal/Magento/Framework/Escaper.php
+++ b/lib/internal/Magento/Framework/Escaper.php
@@ -47,7 +47,6 @@ class Escaper
     private $escapeAsUrlAttributes = ['href'];
 
     /**
-     * Escaper constructor.
      * @param \Magento\Framework\ZendEscaper|null $escaper
      * @param \Psr\Log\LoggerInterface|null $logger
      */
--- a/lib/internal/Magento/Framework/Escaper.php
+++ b/lib/internal/Magento/Framework/Escaper.php
@@ -47,6 +47,21 @@ class Escaper
     private $escapeAsUrlAttributes = ['href'];
 
     /**
+     * Escaper constructor.
+     * @param \Magento\Framework\ZendEscaper|null $escaper
+     * @param \Psr\Log\LoggerInterface|null $logger
+     */
+    public function __construct(
+        \Magento\Framework\ZendEscaper $escaper = null,
+        \Psr\Log\LoggerInterface $logger = null
+    ) {
+        $this->escaper = $escaper ?? \Magento\Framework\App\ObjectManager::getInstance()
+                ->get(\Magento\Framework\ZendEscaper::class);
+        $this->logger = $logger ?? \Magento\Framework\App\ObjectManager::getInstance()
+                ->get(\Psr\Log\LoggerInterface::class);
+    }
+
+    /**
      * Escape string for HTML context.
      *
      * AllowedTags will not be escaped, except the following: script, img, embed,
@@ -84,7 +99,7 @@ class Escaper
                     );
                 } catch (\Exception $e) {
                     restore_error_handler();
-                    $this->getLogger()->critical($e);
+                    $this->logger->critical($e);
                 }
                 restore_error_handler();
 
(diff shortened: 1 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=19310) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @maciejslawik. Thank you for your contribution 

sivaschenko:
Thanks for your contribution @maciejslawik , however adding a constructor is backward incompatible change - it will result in issue on instances where this class is instantiated or extended. Please see https://devdocs.magento.com/guides/v2.0/contributor-guide/backward-compatible-development/

maciejslawik:
@sivaschenko Alright, I understand. 
How about adding this constructor and setting default values for those parameters to null and keeping those 2 getters? This should provide backward compatibility and still give a chance to inject custom implementations.

lib/internal/Magento/Framework/Escaper.php
```diff
@@ -37,6 +37,19 @@ class Escaper
...
+     * Escaper constructor.
+     * @param \Magento\Framework\ZendEscaper $escaper
+     * @param \Psr\Log\LoggerInterface $logger
+     */
+    public function __construct(
+        \Magento\Framework\ZendEscaper $escaper,
+        \Psr\Log\LoggerInterface $logger
+    ) {
+        $this->escaper = $escaper;
+        $this->logger = $logger;
```
orlangur: Such change is backward-incompatible. 

Please inject new dependencies in backward-compatible manner: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter

orlangur (changes requested):

lib/internal/Magento/Framework/Escaper.php
maciejslawik: I added null as default to keep all instances that are instantiated directly unbroken.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3518](https://jira.corp.magento.com/browse/ENGCOM-3518) has been created to process this Pull Request

magento-engcom-team:
@maciejslawik thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

lib/internal/Magento/Framework/Escaper.php
orlangur: @maciejslawik please remove private deprecated methods as you did before and add property initialization into constructor (as in devdoc I mentioned).

orlangur (changes requested):

maciejslawik:
@orlangur I pushed changes that should reflect your request.

orlangur:
@maciejslawik great! Changes look good now, please rebase against latest `2.3-develop` as there is a conflict.

maciejslawik:
@orlangur Rebase has been done.

orlangur (approved):

magento-engcom-team:
[ENGCOM-3518](https://jira.corp.magento.com/browse/ENGCOM-3518) has been created to process this Pull Request

magento-engcom-team:
Hi @maciejslawik. Thank you for your contribution. 

sivaschenko:
Hi @maciejslawik unfortunately we had to revert this PR changes because of backward compatibility concerns:
If the Escaper was extended and parent constructor was not called - that will result in `escaper` and `logger` properties not been inititalized

orlangur:
@sivaschenko mentioned use case does not seem valid to me. It means any other similar addition of dependency is backward incompatible.

First of all, we do not recommend inheritance. In the second place, it is not correct to not call parent constructor.

sivaschenko:
@orlangur in this specific case there is a high risk that if the class was extended the parent constructor was not called as there was no constructor implementation in place.

orlangur:
@slitvyachenko did somebody really extend this class having some troubles? I believe we should follow https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/ and nothing more than that. If doc is not sufficient, it should be extended with your note - "dependencies MUST NOT be added in case there were none before" or something.

sivaschenko:
@orlangur I believe that change is falling into the category of "adding public method" rather than adding more dependencies (each class does have a constructor, but in this case, as the constructor was not explicitly defined, it would have been treated as it does not exist).

PR #19585, opened 2018-12-05
Avoid duplicate loading of configuration files

ajardin:
### Description (*)

While I was investigating why some pages in back-office are extremely slow on my local environment, I found out that we are currently loading multiple times all `layouts.xml` files in the `\Magento\Theme\Model\PageLayout\Config\Builder` class.

Through this change, I propose to only load them one time by adding a new property to store the result and reuse it for the next calls. There is no structural change and the **loading time is reduced by almost 40%** on my local environment, as you can see with the screenshots from Blackfire below.

**Previously:** 
Blackfire screenshot

**Now:**
Blackfire screenshot

**Comparison:**
Blackfire screenshot

1. Log in to the back office.
2. Go to Content > Pages.
3. Look at the loading time... 😄 

Please note that I'm using Docker for Mac [with this environment](https://github.com/EmakinaFR/docker-magento2), and the application is in `developer` mode with all caches activated. The latest version is used for both Docker and Magento.


magento-engcom-team:
Hi @ajardin. Thank you for your contribution 

ajardin:

magento-engcom-team:
Hi @ajardin. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @ajardin, here is your [new Magento instance](https://pr-19585.instances.magento-community.engineering).
Admin access: [https://pr-19585.instances.magento-community.engineering/admin](https://pr-19585.instances.magento-community.engineering/admin)

ajardin:

magento-engcom-team:
Hi @ajardin. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @ajardin, here is your [Magento instance](https://i-19585-2-3-develop.instances.magento-community.engineering).
Admin access: [https://i-19585-2-3-develop.instances.magento-community.engineering/admin](https://i-19585-2-3-develop.instances.magento-community.engineering/admin)

app/code/Magento/Theme/Model/PageLayout/Config/Builder.php
```diff
@@ -56,11 +61,14 @@ public function getPageLayoutsConfig()
      */
     protected function getConfigFiles()
     {
-        $configFiles = [];
-        foreach ($this->themeCollection->loadRegisteredThemes() as $theme) {
-            $configFiles = array_merge($configFiles, $this->fileCollector->getFilesContent($theme, 'layouts.xml'));
+        if (empty($this->configFiles)) {
+            $configFiles = [];
```
sivaschenko, line 67: `$configFiles` variable is unnecessary, please use the `configFiles` property straight away.

app/code/Magento/Theme/Model/PageLayout/Config/Builder.php
```diff
@@ -27,6 +27,11 @@ class Builder implements \Magento\Framework\View\Model\PageLayout\Config\Builder
      */
     protected $themeCollection;
 
+    /**
+     * @var array
+     */
+    protected $configFiles = [];
```
sivaschenko: please change the access level to `private`

sivaschenko (changes requested):
@ajardin thanks for the contribution. Please see the review notes. Also, please consider covering the changed functionality with tests

app/code/Magento/Theme/Model/PageLayout/Config/Builder.php
ajardin: I'm not sure to understand correctly your comment. 😕 

Because putting an `array_merge` instruction directly inside the loop could impact the performances ([source](https://github.com/kalessil/phpinspectionsea/blob/master/docs/performance.md#slow-array-function-used-in-loop)). And I'm using a temporary variable to only modify the property once the value is fully calculated.

app/code/Magento/Theme/Model/PageLayout/Config/Builder.php
ajardin: For sure!

ajardin:
Hi @sivaschenko, thanks for taking time to review my pull request!

About your suggestion on the test coverage, as I've only slightly rewrote an existing code covered by `\Magento\Theme\Test\Unit\Model\PageLayout\Config\BuilderTest::testGetPageLayoutsConfig`, which kind of test should I write?

orlangur:
@ajardin a unit test which would check that files are accessed only once. An additional call to existing test can be added, I believe.

orlangur:
@ajardin one more thing. Could you tell more which pages are slow and in which scenario? I grepped all the usages and this data seems to be cached properly, for example

```
 public function getAllOptions()
    {
        if (!$this->_options) {
            $this->_options = $this->pageLayoutBuilder->getPageLayoutsConfig()->toOptionArray();
            array_unshift($this->_options, ['value' => '', 'label' => __('No layout updates')]);
        }
        return $this->_options;
    }
```

app/code/Magento/Theme/Model/PageLayout/Config/Builder.php
sivaschenko: Thanks for the clarification and sorry for the incorrect suggestion, please ignore my previous comment.

ajardin:
Sorry @sivaschenko, I misunderstood your comments about the test suite... I've updated the existing test to check that the registered themes are loaded only once.

@orlangur the scenario is described in the pull request description (log in to the back office and go to _Content > Pages_). All my application responds correctly except pages that contain grids. And by comparing with the _Cache Management_ page, I discovered that the layouts configuration are loaded multiple times in `\Magento\Theme\Model\PageLayout\Config\Builder`. Even if the change won't affect applications running on production mode, the impact is significant on our local environments.

I'm still not fully satisfied by the loading time, as we are still have to wait more than 20s, but I think it can be a first quick win. What's your opinion on that?

orlangur:
@ajardin is `\Magento\Cms\Block\Adminhtml\Page\Grid::_prepareColumns` a problematic method? I'm pretty sure that builder class is not a right place for caching considering all calls of `getPageLayoutsConfig`. Please add caching in place where the problem actually occurs. There could be even some unnecessary objects creation where could be used only on object.

ajardin:
Do you mean that using a property to store the result of `getFilesContent()` between several calls is a caching system that shouldn't be part of the builder class?
Or should we implement another thing in `\Magento\Cms\Block\Adminhtml\Page\Grid` to properly use Magento cache when available, in addition to what I already did?

orlangur:
@ajardin,

> is a caching system that shouldn't be part of the builder class?

Yep. Caching already occurs outside of builder class in most cases.

ajardin:
I think there is still a misunderstanding. 😄 

My current approach is to avoid duplicate calls to the filesystem within the same builder class by using a property which will save the result after the initial call. There is currently no usage of the Magento cache.

For several reasons in fact:
- as you said, it's not the right place to manage Magento cache.
- as the issue only occurs in development mode, it could be worthless to rely only on Magento cache since it's often deactivated in this context.

There are two different things to address from my point of view. First the builder class can be improved to mitigate its impact on the overall performances because it's the most time consuming part. And then, we'll look at the grid class to check how to take advantage of Magento cache when it's possible.

orlangur:
@ajardin by caching I mean storing in class property, but in another class, here was an example: https://github.com/magento/magento2/pull/19585#issuecomment-444823831

I would like to avoid storing data in property in multiple places, so, please find out where calls to builder are still not cached.

sivaschenko:
@orlangur I don't think the cache of client classes are related to this contribution.

The `Magento\Theme\Model\PageLayout\Config\Builder` is loading and merging file contents from the filesystem and should cache the result to avoid multiple resource consuming operations.

sivaschenko (approved):

orlangur:
@sivaschenko I don't see any scenario where this configuration is read more than once. Let's wait for the reply from contributor.

There is no need to introduce any such internal caching unless really necessary because it is quite memory consuming.

ajardin:
Sorry for the delay, but I thought the discussion was resolved after @sivaschenko's last comment.

@orlangur, there is already a "caching" system through to a property in `\Magento\Cms\Model\Page\Source\PageLayout`, but there is also an instance of `\Magento\Cms\Model\Page\Source\CustomLayout`. Since these two classes are calling directly the `\Magento\Framework\View\Model\PageLayout\Config\BuilderInterface::getPageLayoutsConfig` method, it makes their property inefficient to avoid duplicate calls to the filesystem.

Here is another screenshot from my initial Blackfire profile to illustrate this behavior.

Blackfire profile

orlangur:
@ajardin thanks, so there are actually two calls made in different classes within same scenario.

What do you think about removing unneeded internal caching of options so that we consume less memory?

ajardin:
In that specific case, the gain in memory consumption is almost inexistant.
```
Memory-3.68 kB (-0.008%)
47.6 MB → 47.5 MB
```
It's not a surprise since what is stored in `\Magento\Cms\Model\Page\Source\PageLayout::$options` is just an associative array of less than 10 elements. So, I can push the removing of that intermediate property if you want, but we won't gain anything from a performance perspective.

orlangur:
@ajardin thanks for checking. Yes, please do such removal for the sake of code simplicity.

orlangur (approved):

magento-engcom-team:
[ENGCOM-3634](https://jira.corp.magento.com/browse/ENGCOM-3634) has been created to process this Pull Request

sivaschenko:
@orlangur @ajardin removing protected properties is [backward incompatible change](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/). Such changes can break environments where these classes are extended and properties used.
Please keep the properties and assigned values (please mark them as deprecated if you don't think they are needed).

orlangur:
@sivaschenko good point, thanks.

app/code/Magento/Catalog/Model/Category/Attribute/Source/Layout.php
```diff
@@ -30,10 +30,9 @@ public function __construct(\Magento\Framework\View\Model\PageLayout\Config\Buil
...
     {
-        if (!$this->_options) {
-            $this->_options = $this->pageLayoutBuilder->getPageLayoutsConfig()->toOptionArray();
-            array_unshift($this->_options, ['value' => '', 'label' => __('No layout updates')]);
-        }
-        return $this->_options;
+        $options = $this->pageLayoutBuilder->getPageLayoutsConfig()->toOptionArray();
+        array_unshift($options, ['value' => '', 'label' => __('No layout updates')]);
+
+        return $options;
```
orlangur, line 43: Due to

> Continue storing the value in the property to preserve the old functionality.

Please mark property as deprecated but assign value to it each call.

app/code/Magento/Catalog/Model/Product/Attribute/Source/Layout.php
```diff
@@ -26,14 +26,13 @@ public function __construct(\Magento\Framework\View\Model\PageLayout\Config\Buil
     }
 
     /**
-     * @return array
+     * {@inheritdoc}
```
orlangur: Should be without braces, see https://devdocs.magento.com/guides/v2.0/coding-standards/docblock-standard-general.html#inheritdoc

app/code/Magento/Theme/Model/PageLayout/Config/Builder.php
```diff
@@ -56,11 +61,14 @@ public function getPageLayoutsConfig()
      */
     protected function getConfigFiles()
     {
-        $configFiles = [];
-        foreach ($this->themeCollection->loadRegisteredThemes() as $theme) {
-            $configFiles = array_merge($configFiles, $this->fileCollector->getFilesContent($theme, 'layouts.xml'));
+        if (empty($this->configFiles)) {
```
orlangur: `if(!...)` please

orlangur (changes requested):

ajardin:
Hi @orlangur, do you see something else to change?

orlangur (approved):

magento-engcom-team:
[ENGCOM-3634](https://jira.corp.magento.com/browse/ENGCOM-3634) has been created to process this Pull Request

app/code/Magento/Catalog/Model/Category/Attribute/Source/Layout.php
```diff
@@ -26,14 +32,14 @@ public function __construct(\Magento\Framework\View\Model\PageLayout\Config\Buil
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function getAllOptions()
     {
-        if (!$this->_options) {
```
sivaschenko, line 33: It looks a bit wired to set a value to the property and avoid using it. I think existing method body is more clear and avoids additional options processing.

sivaschenko (changes requested):
Hi @ajardin , thank you for your updates. Can you please keep the implementation of client methods as they are until the deprecated property will be removed.

app/code/Magento/Catalog/Model/Category/Attribute/Source/Layout.php
orlangur: There is no any additional "option processing".

> It looks a bit wired to set a value to the property and avoid using it

This is what current recommendation says. It would be much better to just remove property as probability of BC break is negligible, but if you insist on keeping it deprecated - let's just keep it prepared for removal.

ajardin: Hi guys, are you expecting something from me? 

Because I made what was the most logical thing from my point of view:
- deprecate the property
- keep assigning a value to the deprecated property 
- remove internal usages of the property to prepare its removal

But I'm not sure to understand exactly what you  want right now... 😄

orlangur: @ajardin yeah, everything is fine from my point of view. Remaining part is to convince Sergey, no code changes required until we agree upon them.

orlangur (approved):

magento-engcom-team:
[ENGCOM-3634](https://jira.corp.magento.com/browse/ENGCOM-3634) has been created to process this Pull Request

ajardin:
Hi @orlangur and @sivaschenko,
Do you have any news about this pull request?

(Happy New Year by the way)

sivaschenko:
Hi @ajardin , Happy New Year! This pull request is currently going throug internal verification process and is on it's way to mainline.

ajardin:
Hi @sivaschenko, thanks for the confirmation!
I was asking this question as the pull request is still tagged as `Progress: needs update`. 😉

sivaschenko:
Hi @ajardin sorry, and thanks for notification, the label was removed.

magento-engcom-team:
Hi @ajardin. Thank you for your contribution. 

PR #19756, opened 2018-12-13
Allow Custom Translation (__) Function

bentideswell:
### Description (*)
This pull request allows the definition of a custom translation function. If a custom translation function is not set then the default Magento translation function is set.

This does this by removing the include for Phrase/\__.php in the Magento Framework registration.php file. This gives other registration.php files in other modules a chance to define the function.

After app/autoload.php is loaded in app/bootstrap.php, a call is made to \Magento\Framework\Phrase::defineTranslationFunction and this includes Phrase/__.php. The definition for the __ function is wrapped in a call to function_exists so that the function is not created more than once.

This allows for integration with other online applications that define the __ translation function. This function is a common function in many popular online applications.

1. Apply pull request
2. Make a call to __('Any String Here');
3. String is translated as expected

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -28,6 +28,10 @@ HTML;
 }
 
 require_once __DIR__ . '/autoload.php';
+
+// Ensure that the translation function is defined
+\Magento\Framework\Phrase::defineTranslationFunction();
+
 // Sets default autoload mappings, may be overridden in Bootstrap::create
 \Magento\Framework\App\Bootstrap::populateAutoloader(BP, []);
 
--- a/lib/internal/Magento/Framework/Phrase.php
+++ b/lib/internal/Magento/Framework/Phrase.php
@@ -125,4 +125,13 @@ class Phrase implements \JsonSerializable
     {
         return $this->render();
     }
+    
+    /*
+	   * Register the translation function
+	   *
+	   */
+    public static function defineTranslationFunction()
+    {
+	    require_once __DIR__ . '/Phrase/__.php';
+    }
 }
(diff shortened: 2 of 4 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=19756) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @bentideswell. Thank you for your contribution 

orlangur (changes requested):
Such change is no different than previous implementation which included function in `app/functions.php`. Now instead of having translation function immediately available, instance have to make inclusion in bootstrap which goes against component approach.

As I mentioned,
```
if (!function_exists('__')) {	
    require 'Phrase/__.php';	
}
```
looks like this just for backward compatibility and not for replacement. Later is may be changed to

```
    require 'Phrase/__.php';	
```
without any condition.

Please simply rewrite core functionality like you did before with `app/functions.php` or use Composer patches.

bentideswell:
Surely the official advice of Magento isn't to modify core files? The number one rule for Magento development has always been "don't edit core files". In Magento 2, it's a tie between this and "never use the object manager!!".

Almost everything can be customised in Magento 2 without changing core files. Why is the translation function an exception? I have provided a simple way to make this possible. Why is this not acceptable?

orlangur:
> I have provided a simple way to make this possible. 

No.

> Why is this not acceptable?

Because it breaks modularity. Translation function must be available right after `Magento\Framework` is included, without any additional bootstrapping.

Also, order of components loading is not guaranteed. It's okay to run framework's `registration.php` before any other component or after any of them.

I would not recommend to have `__` function from different sources to coexist at the same time. It would be much more reliable to isolate Wordpress-specific code from Magento one.

Thanks @bentideswell, I'm closing this PR as it is not acceptable in it's current form, if you have some other implementation idea which do not contain mentioned flaws - feel free to share and after we agree on approach it can be actually implemented (so that your time is not wasted on another inappropriate implementation).

PR #19870, opened 2018-12-19
Added required error message.

vinogcs:
If any JSON error occurs user will be be shown error message if the app mode is not in Production.

### Description (*)

Added error description in JSON serialization. 


magento-engcom-team:
Hi @vinogcs. Thank you for your contribution 

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=19870) <br/>All committers have signed the CLA.

sivaschenko:
Hi @vinogcs thanks for your contribution! In my opinion the error message should be the same for production and dev environments. For production/frontend such errors should be properly handled on the higher layer catching that exception and adding a message that is understandable for user (as I think both error message and added description are not user-friendly).

vinogcs:
Hi @sivaschenko, I agree to your point that it has to be taken care by developers at their end and the error message I've added are not clearly understandable. But many non-technical people are adopting Magento for their E-Commerce and non E-Commerce application today. There are few places where these are directly shown to end users who don't have any idea what the error is.

Yesterday, one of my MIS team tried to import products from CSV, they got this error and had no idea what the actual problem was and internet and stack overflow failed when they searched with the error message "Unable to serialize value". 

So I thought it could be better to have little more description on error. The error I've added will tell that the problem was with input file, if it is(in my case it was malformed UTF characters).  And I think it will also help developers to include better response to their exception management without worrying about validating input file separately. 

Maybe I'm wrong, it's just my 2 cents.. :)

sivaschenko:
@vinogcs thanks for details. I completely agree with your point of view! I meant that I don't think `isOnProduction` check is necessary. Considering that the error message is anyway technical, I'd output the detailed error message on both dev and production environments.

If you don't think it's appropriate to output all the details on production environments I'd suggest to dump the details to the log.

Please let me know what do you think about it.

vinogcs:
@sivaschenko I think you're right. It would consume more resources to check the state of the app just to show the error message. We can just go for showing error irrespective of the environment.

lib/internal/Magento/Framework/Serialize/Serializer/Json.php
```diff
@@ -36,8 +36,8 @@ public function unserialize($string)
     {
         $result = json_decode($string, true);
         if (json_last_error() !== JSON_ERROR_NONE) {
-            throw new \InvalidArgumentException('Unable to unserialize value.');
+            throw new \InvalidArgumentException("Unable to unserialize value. Error: " . json_last_error_msg());
         }
         return $result;
-    }
+    }
```
sivaschenko: Removing trailing spaces
```suggestion
    }
```

sivaschenko (changes requested):
Hi @vinogcs , thanks for your update! Please consider removing of added trailing spaces as per my suggestio

vinogcs:
Oops!. My Bad.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3782](https://jira.corp.magento.com/browse/ENGCOM-3782) has been created to process this Pull Request

magento-engcom-team:
@vinogcs thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

magento-engcom-team:
Hi @vinogcs. Thank you for your contribution. 

Issue #7367, opened 2016-11-08
Magento2 multi-store price-per-website: Wrong price indexed for configurable products


dmitrisol:
I have a multi-website Magento 2.1.0 CE installation, with 2 websites and different prices per-website. Prices are note related to each other, for the same product across the two websites.

In product listing pages (PLP), I keep seeing "as low as $55.55" for configurable prices, along with the correct prices, in non-default website 2, even though prices have been set correctly, all configurable children have the same prices $55.55 in website 1 (default) and $77.77 in website 2 (non-default). Configurable product itself has been priced accordingly, there are no special nor tier prices applied.

I've checked price attribute values in database, all prices are save correctly per-website. Checking the price index table, though, returns correct min_price and max_price for simple products but prices from the wrong website for the configurable products, which is causing the "as low as" issue in product listing pages.

No 3rd-party modules are affecting indexing (vanilla Magento).

### Preconditions
1. Magento version CE 2.1.0 (or previous), not tested with more recent versions at the moment.

### Steps to reproduce
1. Install vanilla Magento 2.1.0 CE
2. Create a second website, store and store view
3. Configure price scope to be per-website
4. Create a configurable product, to be in both websites, with price 55.55 for website 1 (default) and price 77.77 for website 2 (non-default), set all prices for the children simple products to 55.55 for website 1 and to 77.77 for website 2, enable the product, make it visible in catalog and search, assign to at least one category and save;
5. Perform a full reindex

### Expected result
Prices should be shown as 55.55 in website 1 (store 1, default) and 77.77 in website 2 (store 2, non-default), in category page containing the product. No "as low as" prices is supposed to be shown.

### Actual result
1. Navigate to product listing (category) page containing the product just added in default store (default website 1), you'll see the price stated as 55.55;
2. Switch store to store 2 (website 2), navigate to product listing (category) page in store 2 (website 2), you will see the price stated as 77.77 (as low as 55.55) even though you have not defined any promotions, tier prices or discounts and all simple children products have price of 77.77 in website 2.
Select in database from `catalog_product_index_price` table for `website_id` IN (1,2) AND `entity_id` IN (1,2,3), where 1,2,3 are the IDs of the configurable product and its two children simple products. You should always see 55.55 for website 1 and 77.77 for website 2, in all `price`, `min_price`, `max_price` columns, instead you'll see 55.00 in `min_price` for website 2.

### Cause

[...]

PR #19878, opened 2018-12-19
Fixes price index on partial update

alepane21:
When $entityIds is passed to the function, the where does a BETWEEN, but that's wrong because it will set to zero all the products between the min($entityIds) and max($entityIds).

### Description (*)

I have just replaced the "BETWEEN" SQL condition to "IN", like has been suggested by gavinlimely in https://github.com/magento/magento2/issues/7367#issuecomment-447921233 and tested and verified it.

1. magento/magento2#19877: Wrong partial price reindex with two or more products 
2. magento/magento2#22028: Unable to update products via csv file, when products ids from file are from wide id range

1. put two or more products to be re-indexed in the catalog_product_price_cl, without contiguous ids;
2. launch indexer_update_all_views;
3. all the products between min($entityIds) and max($entityIds) will be set with price, min_price and max_price to zero.

### DIFF
--- a/app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Price/Query/BaseFinalPrice.php
+++ b/app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Price/Query/BaseFinalPrice.php
@@ -217,11 +217,7 @@ class BaseFinalPrice
         $select->where("e.type_id = ?", $productType);
 
         if ($entityIds !== null) {
-            if (count($entityIds) > 1) {
-                $select->where(sprintf('e.entity_id BETWEEN %s AND %s', min($entityIds), max($entityIds)));
-            } else {
-                $select->where('e.entity_id = ?', $entityIds);
-            }
+            $select->where('e.entity_id IN (?)', $entityIds);
         }
 
         /**


magento-engcom-team:
Hi @alepane21. Thank you for your contribution 

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=19878) <br/>All committers have signed the CLA.

alepane21:
Could you mark this pull request as made by Hevelop? Thanks!

alepane21:
> Could you mark this pull request as made by Hevelop? Thanks!

I already did as specified in https://devdocs.magento.com/guides/v2.2/contributor-guide/contributing.html#partners

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Price/Query/BaseFinalPrice.php
```diff
@@ -217,11 +217,7 @@ public function getQuery(array $dimensions, string $productType, array $entityId
         $select->where("e.type_id = ?", $productType);
 
         if ($entityIds !== null) {
-            if (count($entityIds) > 1) {
-                $select->where(sprintf('e.entity_id BETWEEN %s AND %s', min($entityIds), max($entityIds)));
-            } else {
-                $select->where('e.entity_id = ?', $entityIds);
-            }
+            $select->where('e.entity_id IN (?)', $entityIds);
```
kandy, line 220: I believe that from the performance point of view better to use two where condition, to allow MySQL use range optimization for primary index scan:
like
```suggestion
            $select->where(sprintf('e.entity_id BETWEEN %s AND %s', min($entityIds), max($entityIds)));
            $select->where('e.entity_id IN (?)', $entityIds);
```
and return case for exact one product

orlangur: @kandy wondering why MySQL cannot do such kind of optimization by itself...

alepane21: Hi, thanks for reviewing my pull request!
I made an EXPLAIN on the two different queries with MySQL 5.5 and I don't see any difference. I also saw that in all the other indexes/resource models this optimization with "BETWEEN" is not used.
So I'm not sure about having more complicated code that introduce no positive value...
You have also tested the performance fix you did or just assumed it should work? Maybe I'm doing something wrong!

Let me know, thanks 💯 

P.S.: If we want to keep this fix, I would re-add the if-else on the number of elements in $entityIds, so if we have only one element I could put e.entity_id = X and only if two or more elements I would add your code.

kandy: @orlangur, no, because it not common case and  base on specific usage in Magento Application

orlangur: I see, we know something about underlying data - like, range is not too big.

Still some heuristics could be applied on MySQL engine side.

orlangur: @kandy `e.entity_id IN (?)` I have serious doubts MySQL is not optimizing it well in case of only one ID.

osakoova: Using `BETWEEN` is wrong. Here is a real example. A catalog of 319,077 products, a product with an older `entity_id ` changes, as well as newer products. You get the following WHERE clause: `(e.entity_id BETWEEN 332999 AND 532434);`

In a multi-site installation etc... this SELECT returns nearly 2 million rows

```
mysql> SELECT COUNT(*) FROM `catalog_product_entity` AS `e`\n CROSS JOIN `customer_group` 
[...]
WHERE (ta_status.value = 1) AND (e.type_id = 'simple') AND (e.entity_id BETWEEN 332999 AND 532434);

+----------+
| COUNT(*) |
+----------+
|  1967373 |
+----------+
1 row in set (3.60 sec)
```

dudzio12: @orlangur - maybe because of the limit of query size:
https://stackoverflow.com/questions/4275640/mysql-in-condition-limit/4275704

It was hell to find the problem.

osakoova: Sure, there can be limits with SQL, MySQL. config parameters etc... but this is what code is there for, to do intelligent things even with possible limitations somewhere else...

orlangur (changes requested):
@alepane21 please check with `EXPLAIN` if `IN(only_one_id)` is ineffective.

If it is so, return `if` back as you suggested.

sidolov:
@alepane21 , I am closing this PR now due to inactivity.

app/code/Magento/Catalog/Model/ResourceModel/Product/Indexer/Price/Query/BaseFinalPrice.php
orlangur: @osakoova I don't see `e.entity_id IN (?)` part in your query.

PR #20087, opened 2019-01-05
[2.3] #20068 move caching of source models into abstract source

davidverholen:
### Description (*)

Moved caching of source models into abstract source (suggested here: https://github.com/magento/magento2/issues/20068)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=20087) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @davidverholen. Thank you for your contribution 

app/code/Magento/Eav/Model/Entity/Attribute/Source/AbstractSource.php
```diff
@@ -31,6 +31,22 @@
      */
     protected $_options = null;
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getAllOptions()
+    {
+        if (empty($this->_options)) {
```
sivaschenko: I'd suggest using the default _options value that is null in the condition to avoid multiple calls to loadOptions if no options are available.
```suggestion
        if (null === $this->_options) {
```

sivaschenko (changes requested):
Hi @davidverholen , thanks for the contribution please see my code review note

app/code/Magento/Eav/Model/Entity/Attribute/Source/AbstractSource.php
orlangur: @sivaschenko please avoid Yoda-style conditions

orlangur (changes requested):
I would suggest changing existing source models also to eliminate unneeded caching logic.

davidverholen:
@orlangur starting with that, I found a special case pretty fast. Should I just ignore them for now? FOr example \Magento\Eav\Model\Entity\Attribute\Source\Table has 2 additional method parameters for getAllOptions. Also the options are cached by store id and attribute id.
So I would assume in this case it makes sense that the implementation of getAllOptions is overridden with the custom caching mechanic

sivaschenko:
Hi @davidverholen thanks for your updates! I don't have any code review notes from my side, however, Travis tests should be fixed before we can process the pull request further.

davidverholen:
@sivaschenko right, but Im' really not sure, why they are failing.

The warnings state:
> Trying to configure method "getAllOptions" which cannot be configured because it does not exist, has not been specified, is final, or is static

but the getAllOptions is just a new public method in the abstract class, which was defined in the interface before, anyway. So did not even change the public api.

also the method definitely exists (since now defined in the abstract class) and is not static or final. I'll try to investigate further but it seems to be an issue with either the tests or with the phpunit mocking mechanism

sivaschenko:
@davidverholen thanks for the update. I will approve this issue to continue the processing internally.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-3886](https://jira.corp.magento.com/browse/ENGCOM-3886) has been created to process this Pull Request

sivaschenko:
Hi @davidverholen can you please update the branch and resolve conflicts?

sivaschenko (changes requested):
Hi @davidverholen Unfortunately, we were not able to approve the refactoring and adding new protected methods (that is against backward compatibility policy for patch releases).

We come up with a decision that caching should be implemented for each model separately.

Options caching for models that are missing that can be added in scope of this pull request.

orlangur:
Thanks for collaboration!

PR #20152, opened 2019-01-09
Add $data array as property to Magento\Backend\App\Action\Plugin\Authentication

nickolasburr:
### Description (*)

Add $data array property to Magento\Backend\App\Action\Plugin\Authentication. This makes it easier to use DI to extend the class.

1. N/A

1. Create basic module with class that extends `Magento\Backend\App\Action\Plugin\Authentication`
2. Add `output` method to subclass, which will `var_dump` the `$data` array
3. Add subclass `<preference>` in `etc/di.xml`
4. Add subclass `<type>` with arbitrary arguments given to `$data` in `etc/di.xml`
5. Call `output` method to display the given DI arguments

### DIFF
--- a/app/code/Magento/Backend/App/Action/Plugin/Authentication.php
+++ b/app/code/Magento/Backend/App/Action/Plugin/Authentication.php
@@ -69,6 +69,11 @@ class Authentication
     protected $formKeyValidator;
 
     /**
+     * @var array
+     */
+    protected $data;
+
+    /**
      * @param \Magento\Backend\Model\Auth $auth
      * @param \Magento\Backend\Model\UrlInterface $url
      * @param \Magento\Framework\App\ResponseInterface $response
@@ -78,6 +83,7 @@ class Authentication
      * @param \Magento\Framework\Controller\Result\RedirectFactory $resultRedirectFactory
      * @param \Magento\Backend\App\BackendAppList $backendAppList
      * @param \Magento\Framework\Data\Form\FormKey\Validator $formKeyValidator
+     * @param array $data
      */
     public function __construct(
         \Magento\Backend\Model\Auth $auth,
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=20152) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @nickolasburr. Thank you for your contribution 

kandy:
Your change doesn't make any sense for me. Can you explain what you try to achieve?
Also, pay attention that we do not recommend to extend core classes

nickolasburr:
In certain situations where you may need to change the mode of authentication for an admin user, it is necessary to override this class. The proposed change is to allow overriding this class without having to also override the constructor by allowing the passing of arbitrary data via the `$data` property.

kandy:
But why you do not want to override constructor?

nickolasburr:
Overriding the constructor unnecessarily requires additional overhead code, and makes the inheritance brittle and susceptible to breakage. Constructor method signatures change frequently in Magento, and the proposed changes provide two key benefits (especially for extension developers):

1. Allow arguments to be given via `di.xml`, making it more manageable to extend the class
2. Prevent subclasses of `Magento\Backend\App\Action\Plugin\Authentication` from breaking if there's a backward-incompatible change made in a future release

There are several examples of this done elsewhere in Magento, such as:

https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Tax/Model/Calculation.php#L232
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Shipping/Model/Order/Track.php#L54
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Shipping/Model/Order/Pdf/Packaging.php#L71

Albeit, each of these classes inherit from `Magento\Framework\DataObject`. If that's a more desirable approach, I'm more than happy to make the change and send another pull request.

PR #20777, opened 2019-01-29
Allow multiple conditions for expired quotes collection

sankalpshekhar:
### Description (*)
With Magento\Sales\Cron\CleanExpiredQuotes::setExpireQuotesAdditionalFilterFields() assigning values directly to the attribute and getExpireQuotesAdditionalFilterFields() being protected isn't callable outside the class, hence unable to add more to the existing condition, rather they end up being replaced with the new ones.

none

N/A

### DIFF
--- a/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
+++ b/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
@@ -30,12 +30,12 @@ class CleanExpiredQuotes
     /**
      * @var array
      */
-    private $expireQuotesFilterFields = [];
+    protected $expireQuotesFilterFields = [];
 
     /**
      * @var ExpireQuotesFilterFieldsProvider
      */
-    protected $expireQuotesFilterFieldsProvider;
+    private $expireQuotesFilterFieldsProvider;
 
     /**
      * @param StoresConfig $storesConfig
--- a/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
+++ b/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
@@ -61,6 +61,10 @@ class CleanExpiredQuotes
     public function execute()
     {
         $lifetimes = $this->storesConfig->getStoresConfigByPath('checkout/cart/delete_quote_after');
+        $fields = array_merge(
+            $this->getExpireQuotesAdditionalFilterFields(),
+            $this->expireQuotesFilterFieldsProvider->getFields()
+        );
         foreach ($lifetimes as $storeId => $lifetime) {
             $lifetime *= self::LIFETIME;
 
@@ -70,11 +74,6 @@ class CleanExpiredQuotes
             $quotes->addFieldToFilter('store_id', $storeId);
             $quotes->addFieldToFilter('updated_at', ['to' => date("Y-m-d", time() - $lifetime)]);
             $quotes->addFieldToFilter('is_active', 0);
-            
-            $fields = array_merge(
-                $this->getExpireQuotesAdditionalFilterFields(),
-                $this->expireQuotesFilterFieldsProvider->getFields()
-            );
 
             foreach ($fields as $field => $condition) {
                 $quotes->addFieldToFilter($field, $condition);
--- a/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
+++ b/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
@@ -20,12 +20,12 @@ class CleanExpiredQuotes
     /**
      * @var StoresConfig
      */
-    private $storesConfig;
+    protected $storesConfig;
 
     /**
      * @var QuoteCollectionFactory
      */
-    private $quoteCollectionFactory;
+    protected $quoteCollectionFactory;
 
     /**
      * @var array
@@ -35,7 +35,7 @@ class CleanExpiredQuotes
     /**
      * @var ExpireQuotesFilterFieldsProvider
      */
-    private $expireQuotesFilterFieldsProvider;
+    protected $expireQuotesFilterFieldsProvider;
 
     /**
      * @param StoresConfig $storesConfig
--- a/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
+++ b/app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
@@ -70,8 +70,13 @@ class CleanExpiredQuotes
             $quotes->addFieldToFilter('store_id', $storeId);
             $quotes->addFieldToFilter('updated_at', ['to' => date("Y-m-d", time() - $lifetime)]);
             $quotes->addFieldToFilter('is_active', 0);
+            
+            $fields = array_merge(
+                $this->getExpireQuotesAdditionalFilterFields(),
+                $this->expireQuotesFilterFieldsProvider->getFields()
+            );
 
-            foreach ($this->expireQuotesFilterFieldsProvider->getFields() as $field => $condition) {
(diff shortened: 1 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=20777) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @sankalpshekhar. Thank you for your contribution 

VladimirZaets:
Hi @sankalpshekhar, please, sign CLA, otherwise, we can't process your pull request

app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
```diff
@@ -85,6 +85,6 @@ protected function getExpireQuotesAdditionalFilterFields()
      */
     public function setExpireQuotesAdditionalFilterFields(array $fields)
```
VladimirZaets, line 104: Hi @sankalpshekhar. Thanks for collaboration. The `setExpireQuotesAdditionalFilterFields` sounds like this method should override value. In your case you case the method should be named something like `addExpireQuotesAdditionalFilterFields`, but according to our backward compatibility guide we can't introduce new public methods.

Can you please provide use case for this improvements?

sankalpshekhar: Hello @VladimirZaets 

Currently the method `setExpireQuotesAdditionalFilterFields()` is used in `Magento\PersistentHistory\Plugin\CleanExpiredQuotesPlugin` in _Commerce_ edition of Magento to specify additional condition which in turn is added to filter the quotes collection for deletion. The value passed to the method, overrides the one class attribute may already poses.

For e.g. If I have to add another condition to the filter collection, I can't without losing the previous set condition. If I add a new plugin and try to add a new condition, I either have to repeat the conditions previously added somewhere else in my module, or loose them as my plugin would override the previous values, depending on the sort order of the plugins being loaded.

The 2 possible solutions to remove this impediment (keeping backward compatibility in sight) are as follows:

1. Since `$expireQuotesFilterFields` is defined as an array, we could use the array_merge as I suggested in PR,
2. change `getExpireQuotesAdditionalFilterFields` to a _public_ method, so I could use this to retrieve & manipulate the array values within my plugin, but users could also end up creating plugins on top of this method.

sankalpshekhar: @VladimirZaets  Let me know, if additional information would be required here.

sankalpshekhar: any updates here? is there anything that i could help out with?

VladimirZaets:
Hi @sankalpshekhar. 
Sorry for the delay. I discussed this PR with my team and we made a decision to fix this issue in a more complex way. 

First of all, we should create DataProvider that should have the ability to be configured by DI declaration. 
This DataProvider should be injected to the constructor of the cron job model (CleanExpiredQuotes) and fields should be read from DataProvider. That's will give the ability to configure fields that you need directly by DI configuration and `set` and `get` method will no need at all, so they should be marked as deprecated (according to our [Backward Compatibility Guide](http://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/index.html).

If you have some concerns or you need help with it to feel free to ask, thanks.

sivaschenko:
Hi @sankalpshekhar , would you like to continue progress on this pull request?

VladimirZaets:
Hi @sankalpshekhar, I am closing this PR now due to inactivity.
Thank you for collaboration

sankalpshekhar:
Hello @VladimirZaets Could you provide more details?

As I see the stated approach has its limitations of specifying one field at a time, which could be a blocker if users wishes to use an OR statement.

sivaschenko (changes requested):
Hi @sankalpshekhar the pull request contains the 4575 changed files. Can you please ensure only necessary changes are included.

sivaschenko:
Hi @sankalpshekhar as mentioned by @VladimirZaets it wouldn't be the best solution to change setter behaviour, please see the suggested fix: https://github.com/magento/magento2/pull/20777#issuecomment-482153222

I case of any questions feel free to contact me on slack

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-5343](https://jira.corp.magento.com/browse/ENGCOM-5343) has been created to process this Pull Request

| Label | Description |

sivaschenko (changes requested):
PR is updated

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-5343](https://jira.corp.magento.com/browse/ENGCOM-5343) has been created to process this Pull Request

app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
```diff
@@ -17,28 +20,37 @@ class CleanExpiredQuotes
     /**
      * @var StoresConfig
      */
-    protected $storesConfig;
+    private $storesConfig;
```
sidolov: Please, revert to protected, we are not allowed to change vars visibility according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
```diff
@@ -17,28 +20,37 @@ class CleanExpiredQuotes
...
 
     /**
-     * @var \Magento\Quote\Model\ResourceModel\Quote\CollectionFactory
+     * @var QuoteCollectionFactory
      */
-    protected $quoteCollectionFactory;
+    private $quoteCollectionFactory;
```
sidolov: The same as above

app/code/Magento/Sales/Cron/CleanExpiredQuotes.php
```diff
@@ -17,28 +20,37 @@ class CleanExpiredQuotes
...
 
     /**
      * @var array
      */
-    protected $expireQuotesFilterFields = [];
+    private $expireQuotesFilterFields = [];
```
sidolov: The same as above

sidolov (changes requested):

sidolov (approved):

magento-engcom-team:
[ENGCOM-5343](https://jira.corp.magento.com/browse/ENGCOM-5343) has been created to process this Pull Request

engcom-Delta:
:heavy_check_mark: QA passed

sidolov:
Hi @sankalpshekhar , looks like the class was completely refactored and now it's possible to extend the behavior very easy. Could you please check and confirm that the current fix works for you?

sankalpshekhar:
@sidolov the implementation still has limitations, the current only allows you to add conditions on AND, but doesn't help with more flexible conditions.

Do we need to over-abstract the implementation for one class by creating a provider to manage the conditions? especially when this provider won't be reused elsewhere?

can't this be simply implemented using an array param to the class, and developers can add as many conditions to this through di.xml array argument?

sidolov:
Hi @sankalpshekhar , after refactoring you have an ability to add any filters by introducing the after plugin for `\Magento\Sales\Model\ResourceModel\Collection\ExpiredQuotesCollection`. 

Example: 
```php
public function afterGetExpiredQuotes(
        ExpiredQuotesCollection $subject,
        AbstractCollection $result
    ) : AbstractCollection {
        $result->addFieldToFilter('your_field', 'value');

        return $result;
    }
```

I'm closing this PR since the issue has been resolved.

PR #20950, opened 2019-02-03
magento/magento2#20773: Do not throw exception during autoload

Vinai:
### Description (*)

Avoid uncaught exceptions during autoloading. PSR-4 dictates that must not happen. Return false instead.
Instead, try to log the exception message.

magento/magento2#20773

Run `class_exists(FooFactory::class);`.
More information to reproduce the problem is in the issue thread.


magento-engcom-team:
Hi @Vinai. Thank you for your contribution 

davidverholen:
Hey @Vinai , since autoloader registration happens before the objectmanager  is fully initialized in [\Magento\Framework\App\ObjectManagerFactory::create](https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/App/ObjectManagerFactory.php#L106), it now (ironically) throws an exception during the construction of the autoloader.

Instead you would have to manually create the logger instance and inject it here [\Magento\Framework\ObjectManager\DefinitionFactory::createClassDefinition](https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/ObjectManager/DefinitionFactory.php#L55)

Not sure how complicated it is to initialize the logger without the OM. Unfortunately I did not find it in any of the calling classes to maybe pass it through.

Just another thing: Wouldn't it be more helpful (even if it's nor psr4 compliant) to keep the exceptions at least in developer mode so they are just displayed directly instead of "hidden" in a log file?

Vinai:
He, stupid of me, sorry. The contribution day ended so I ran out of time to test it.
I’m going to fix it tomorrow.
Thanks!

Vinai:
This was a typical case of Unit Tests green, missing integration tests.
Fixed now :)

Vinai:
@davidverholen I decided against letting the exception bubble up in developer mode, because that still would prohibit using another autoloader later in the autoloader chain (which is the original reason for the PR). Instead, the code now just tries to log the message if the `ObjectManager` has been initialized, otherwise there is nothing that can be done and the exception message is lost.
In almost all cases the logger should be available during a generation failure.

davidverholen (approved):

magento-engcom-team:
Hi @davidverholen, thank you for the review.
[ENGCOM-4104](https://jira.corp.magento.com/browse/ENGCOM-4104) has been created to process this Pull Request

Vinai:
Thanks for reviewing, @davidverholen! I just noticed some typos and couldn't resist fixing them and updating the PR. Sorry if that means you have to repeat the review - at least it's a small patch.

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Psr\Log\LoggerInterface;
+
+class AutoloaderTest extends TestCase
+{
+    /**
+     * @return ObjectManager|\Magento\Framework\App\ObjectManager
+     */
+    private function getTestFrameworkObjectManager()
+    {
+        return ObjectManager::getInstance();
```
orlangur, line 28: Could you please inline this method? I don't see how it improves readability and also `ObjectManager::getInstance()` is used as is at least once.

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
```diff
@@ -0,0 +1,77 @@
+<?php declare(strict_types=1);
+
+namespace Magento\Framework\Code\Test\Unit\Generator;
```
orlangur: Some copyright header should be here, not sure if Travis catches this.

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
```diff
@@ -0,0 +1,77 @@
...
+    public function setupLoggerTestDouble(): void
+    {
+        $loggerTestDouble = $this->createMock(LoggerInterface::class);
+        $this->getTestFrameworkObjectManager()->addSharedInstance($loggerTestDouble, MagentoMonologLogger::class);
+    }
+
+    /**
+     * @after
+     */
+    public function removeLoggerTestDouble(): void
```
orlangur, line 43: Maybe just call them `setUp`/`tearDown`? Why public?

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
```diff
@@ -0,0 +1,77 @@
...
+     * @param \RuntimeException $testException
+     * @return Generator|MockObject
+     */
+    private function createExceptionThrowingGeneratorTestDouble(\RuntimeException $testException)
+    {
+        /** @var Generator|MockObject $deneratorStub */
+        $deneratorStub = $this->createMock(Generator::class);
+        $deneratorStub->method('generateClass')->willThrowException($testException);
+
+        return $deneratorStub;
```
orlangur: -> `generator`

orlangur (changes requested):

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
Vinai: The return type hint fixes the type inference in PHPStorm (see screenshot).
typeinference

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
Vinai: Added.

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
Vinai: I prefer not to override parent class methods. The `@before` and `@after` annotations on public methods with descriptive names are preferable.

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
Vinai: Good catch, fixed 👍

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
Vinai: Apropos: I should make a PR to fix the return annotation on `getInstance` on the test framework ObjectManager - this has been annoying me for 2.5 years now.

Vinai:
The warnings that are causing Codacy to fail are unrelated to the PR, but because of outdated rulesets.
Screenshot of the warnings regarding outdated rulesets

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-4104](https://jira.corp.magento.com/browse/ENGCOM-4104) has been created to process this Pull Request

Vinai:
I think the failing test is unrelated to this PR.
Is there anything I can do to move this PR forward?

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
orlangur: @Vinai yep, there is no need to bloat each single test.

dev/tests/integration/testsuite/Magento/Framework/Code/Generator/AutoloaderTest.php
orlangur: Such name does not add any value compared to just `setUp`. I have nothing against such annotations in general but then we should make it an official recommendation, otherwise the code become inconsistent.

Vinai:
Thanks for merging 👍 👍 

I'd like to take the survey, but the app wants full write access to all my orgs:

bildschirmfoto 2019-02-25 um 06 17 34


This seems like a bad idea, so I'll give it a pass.

orlangur:
@Vinai LOOOL :) Thanks for pointing out, I'll pass it to EngCom guys.

Vinai:
The same survey app is used for devdoc PRs, too, by the way.

PR #21131, opened 2019-02-11
Fix Issue #19872 - checking if image is in media directory

Bartlomiejsz:
### Description (*)
Issue #19872 (closed) is still not fixed on develop branch.
On develop when image is added from media gallery while configuring category, it's path is saved as **/media/path_to_file**, but while Magento tries to check if this image is placed in media it compares it's path with **pub/media/** instead of **media/**.

1. magento/magento2#19872: Magento 2.3 category edit page "Select from gallery" button not working.
2. magento/magento2#22092

1. Go to category add/edit form
2. Edit **Category Image** attribute using **Select from Gallery** option
3. Save category
4. Check if **Category Image** is visible in backoffice for this category


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=21131) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @Bartlomiejsz. Thank you for your contribution 

Bartlomiejsz:

magento-engcom-team:
Hi @Bartlomiejsz. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @Bartlomiejsz, here is your [Magento instance](https://i-21131-2-3-develop.instances.magento-community.engineering).
Admin access: [https://i-21131-2-3-develop.instances.magento-community.engineering/admin](https://i-21131-2-3-develop.instances.magento-community.engineering/admin)

Bartlomiejsz:

magento-engcom-team:
Hi @Bartlomiejsz. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @Bartlomiejsz, here is your [new Magento instance](https://pr-21131.instances.magento-community.engineering).
Admin access: [https://pr-21131.instances.magento-community.engineering/admin](https://pr-21131.instances.magento-community.engineering/admin)

dmytro-ch:

magento-engcom-team:
Hi @dmytro-ch. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @dmytro-ch, here is your [new Magento instance](https://pr-21131.instances.magento-community.engineering).
Admin access: [https://pr-21131.instances.magento-community.engineering/admin](https://pr-21131.instances.magento-community.engineering/admin)

dmytro-ch:
Hi @Bartlomiejsz,
I colud not reproduce the issue on test instance in `2.3-develop` branch.
Also, the current solution does not work on test instance.

screen recording 2019-02-14 at 8 31 01 pm

Could you please make sure the issue is not related to your local envronment configuration?

Thank you!

Bartlomiejsz:
@dmytro-ch that's right, I also noticed, that the issue doesn't appear on github-generated instance of Magento.

But I think this is related to location of Magento instance. If it is located in **root** directory (which is not recommended due to security, but I think still this is used by github-generated instance) the issue doesn't occur. But if Magento is installed in recommended **pub/** directory, the issue occurs.
So it is my local environment configuration indeed, but also all environments that are installed according to Magento recommendations I think...

dmytro-ch:
@Bartlomiejsz I was also thinking about that and tested the issue locally in `2.3-develop` branch for both configurations /root and /pub.
As result, each time I got the same problem as you described, even when the document root is set to root Magento 2 directory. So I guess, this is not the reason of the issue.

However, after switching to PHP 7.2 the issue is not reproducible anymore.
Could you please check your current PHP version?

Thank you!

Bartlomiejsz:
@dmytro-ch I'm currently running on **7.1.25**, but I had same issue on instance running on **7.2.14**

Bartlomiejsz:

magento-engcom-team:
Hi @Bartlomiejsz. Thank you for your request. I'm working on Magento instance for you

Bartlomiejsz:

magento-engcom-team:
Hi @Bartlomiejsz. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @Bartlomiejsz, here is your [new Magento instance](https://pr-21131.instances.magento-community.engineering).
Admin access: [https://pr-21131.instances.magento-community.engineering/admin](https://pr-21131.instances.magento-community.engineering/admin)

Bartlomiejsz:
@dmytro-ch could you please recheck this one? Current solution should work no matter if installation is in root or pub directory

app/code/Magento/Catalog/Model/Category/FileInfo.php
```diff
@@ -169,14 +169,20 @@ public function isBeginsWithMediaDirectoryPath($fileName)
...
      * @return string
      */
-    private function getMediaDirectoryPathRelativeToBaseDirectoryPath()
+    private function getMediaDirectoryPathRelativeToBaseDirectoryPath(string $filePath = '')
     {
         $baseDirectoryPath = $this->getBaseDirectory()->getAbsolutePath();
         $mediaDirectoryPath = $this->getMediaDirectory()->getAbsolutePath();
 
         $mediaDirectoryRelativeSubpath = substr($mediaDirectoryPath, strlen($baseDirectoryPath));
+        $pubDirectory = 'pub/';
```
dmytro-ch: It would be better to use constant `\Magento\Framework\App\Filesystem\DirectoryList::PUB` in this line.

dmytro-ch (changes requested):
Hi @Bartlomiejsz,
could you please check the minor code review recommendation?
Thank you!

app/code/Magento/Catalog/Model/Category/FileInfo.php
Bartlomiejsz: @dmytro-ch done

dmytro-ch:

magento-engcom-team:
Hi @dmytro-ch. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @dmytro-ch, here is your [new Magento instance](https://pr-21131.instances.magento-community.engineering).
Admin access: [https://pr-21131.instances.magento-community.engineering/admin](https://pr-21131.instances.magento-community.engineering/admin)

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-4591](https://jira.corp.magento.com/browse/ENGCOM-4591) has been created to process this Pull Request

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-4591](https://jira.corp.magento.com/browse/ENGCOM-4591) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

app/code/Magento/Catalog/Model/Category/FileInfo.php
```diff
@@ -169,14 +169,20 @@ public function isBeginsWithMediaDirectoryPath($fileName)
...
      * @return string
      */
-    private function getMediaDirectoryPathRelativeToBaseDirectoryPath()
+    private function getMediaDirectoryPathRelativeToBaseDirectoryPath(string $filePath = '')
     {
         $baseDirectoryPath = $this->getBaseDirectory()->getAbsolutePath();
         $mediaDirectoryPath = $this->getMediaDirectory()->getAbsolutePath();
 
         $mediaDirectoryRelativeSubpath = substr($mediaDirectoryPath, strlen($baseDirectoryPath));
+        $pubDirectory = DirectoryList::PUB . DIRECTORY_SEPARATOR;
```
sivaschenko: `DirectoryList::PUB` is not a directory name but a key that stores configuration for pub directory, this key can be used to get actual directory path via '\Magento\Framework\Filesystem\DirectoryList::getPath`

sivaschenko (changes requested):
Hi @Bartlomiejsz Thanks for the pull request!

I think the root cause of the problem is that file path is saved inconsistently to category attribute:
 - file path relative to `media` directory is saved when uploading an image
 - file path relative to `pub` directory is saved when selecting an image from the gallery

I think the issue should be fixed by ensuring that file path relative to media directory is saved to category attribute when image is selected from the gallery.

I'd try to apply the fix during saving the category entity - validating that file path exists in relation to media directory, and if not - trying to get absolute path of the file via static content root (`pub` directory instance) and then relative path via media directory instance

Bartlomiejsz:
@sivaschenko ok, I will work on that

Bartlomiejsz:
> Hi @Bartlomiejsz Thanks for the pull request!
> 
> I think the root cause of the problem is that file path is saved inconsistently to category attribute:
> 
> * file path relative to `media` directory is saved when uploading an image
> * file path relative to `pub` directory is saved when selecting an image from the gallery
> 
> I think the issue should be fixed by ensuring that file path relative to media directory is saved to category attribute when image is selected from the gallery.
> 
> I'd try to apply the fix during saving the category entity - validating that file path exists in relation to media directory, and if not - trying to get absolute path of the file via static content root (`pub` directory instance) and then relative path via media directory instance

@sivaschenko I checked this and I believe that it is saved consistently. If user in MBO uploads new image it's path is relative to `media/catalog/category` where all category images are placed - it is same as for other image uploaders in MBO (`media/catalog/product` for instance).
If user selects image from media gallery it's path begins with `/media` or `/pub/media` (depends of location of document root in apache/nginx), because it's not connected to category entity but it's ordinary image in media, so it is same as all other paths for media files in whole system.
Could you confirm if this should be changed, because I think that such change would bring incompatibility here?

sivaschenko:
Hi @Bartlomiejsz Currently the category image path is stored in two formats (relative to `root` and relative to `media`). That is the root cause of the issue that should be fixed from my point of view.

Avoiding to use one of the currently used approaches of storing image path does not break backward compatibility (as image path will be still stored in one of the previously supported formats)

Bartlomiejsz:
Hi @sivaschenko, thank you for answer!
I don't want to argue, but this path is saved in two formats indeed (relative to `media/catalog/category` - when this is just category image and relative to root - if it is ordinary media image loaded for it) but I think there's good reason for this.
If we would remove storing it's path relative to `media/catalog/category` it probably wouldn't be backward compatible, because this is the way it was stored from M2 beginning.
And selecting image from media gallery was introduced later, so it is required to differentiate which image is category-only and which is ordinary media to load it's url correctly.
It's not possible to remove only path relative to root way of storing it, so I believe this way should be just fixed.

sivaschenko:
Thanks for clarification @Bartlomiejsz . I'd like to take some time to investigate this issue and discuss it with peers.

Bartlomiejsz:
Great, thank you @sivaschenko

sivaschenko:
Hi @Bartlomiejsz I discussed the pull request with maintainers and we agreed that it would be better to apply the fix during the save operation rather than during the rendering. That approach will ensure that consistent data is stored in the database and will not violate backward compatibility until one of the currently used image path formats is saved to the database.

Please let me know if you have any questions or need assistance.

Bartlomiejsz:

magento-engcom-team:
Hi @Bartlomiejsz. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @Bartlomiejsz, here is your [Magento instance](https://i-21131-2-3-develop.instances.magento-community.engineering).
Admin access: [https://i-21131-2-3-develop.instances.magento-community.engineering/admin](https://i-21131-2-3-develop.instances.magento-community.engineering/admin)

Bartlomiejsz:
Hi @sivaschenko. I think I understand you're point, but there is one problem I think. Sorry, I haven't thought about this before.
This PR fixes a bit different problem than we discussed. The problem is not caused by different ways of storing paths for images, but it is caused by different document root placement for magento instance.
You're saying about changing way that path is stored - so it would be relative to `media`. But the issue would still occur then, because if Magento would be installed in `pub/` directory media path would have to begin with `/media/`, and if it would be installed in root dir (one directory up from `pub/`) - media path would have to begin with `/pub/media/` - and it works this way in for all files in `pub/` - media, static etc.
And this `/pub` at the beginning of path is reason of issue here, not two different ways of storing this path.
So I assume you're point can be good - this could be changed, but this PR should be checked as fix for another issue in this area I believe.

Changing way of storing those paths is feature rather than fix in my opinion - I don't think this is the reason of any issue currently.

sivaschenko:
@Bartlomiejsz If only one way of storing images paths will be used, it will be possible to substantially simplify the algorithms of retrieving a file path (the algorithms that are updated in this pull request)

sivaschenko:
Hi @Bartlomiejsz , thanks for your response. Would you be able to add tests cases to `\Magento\Catalog\Model\Category\FileInfo` to cover both cases of paths for `isExist` and `getMediaDirectoryPathRelativeToBaseDirectoryPath` methods? I believe it will make things more clear.

Bartlomiejsz:
Hi @sivaschenko, will work on it!

Bartlomiejsz:
Hi @sivaschenko, added tests. Especially important here are paths beginning with `/pub/media` - for installation at root directory and beginning with `/media` - for installation at `/pub` directory

sivaschenko (changes requested):
Thanks for the tests @Bartlomiejsz 

The only thing that is not covered by the fix is the custom `pub` directory path.

Please use `$this->filesystem->getDirectoryRead(DirectoryList::PUB)->getRelativePath()` to retrive the correct pub directory path

Bartlomiejsz:
Hi @sivaschenko, done

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-4591](https://jira.corp.magento.com/browse/ENGCOM-4591) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

wescey:
Now when the selected image is saved (and is still there in backend), is it showing on frontend?

duckchip:
I have that same problem.
After applying this patch, i got a very weird path on the frontend. (Broken image)

PR #21181, opened 2019-02-13
Disable cron:run when maintenance mode is enabled

kassner:
### Description

I believe cron jobs should be disabled while `setup:upgrade` is being run, to avoid concurrent accesses, which can lead to table locks that can make setup upgrade fail, plus we'll be using stale/wrong data for the cron jobs.

Since the maintenance mode makes sure the frontend is inaccessible, avoiding any writes to the database through it, I believe we should disable crons as well, as they have a heavy impact on the database.

### Manual testing scenarios

1. Run `php bin/magento maintenance:enable`;
2. Run `php bin/magento cron:run`;
3. Cron jobs should not run;

### Contribution checklist

### DIFF
--- a/app/code/Magento/Cron/Console/Command/CronCommand.php
+++ b/app/code/Magento/Cron/Console/Command/CronCommand.php
@@ -10,6 +10,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Magento\Framework\App\MaintenanceMode;
 use Magento\Framework\App\ObjectManager;
 use Magento\Framework\App\ObjectManagerFactory;
 use Magento\Store\Model\Store;
@@ -44,16 +45,28 @@ class CronCommand extends Command
     private $deploymentConfig;
 
     /**
+     * Maintenance mode
+     *
+     * @var MaintenanceMode
+     */
+    private $maintenanceMode;
+
+    /**
      * @param ObjectManagerFactory $objectManagerFactory
      * @param DeploymentConfig $deploymentConfig Application deployment configuration
+     * @param MaintenanceMode $maintenanceMode
      */
     public function __construct(
         ObjectManagerFactory $objectManagerFactory,
-        DeploymentConfig $deploymentConfig = null
+        DeploymentConfig $deploymentConfig = null,
+        MaintenanceMode $maintenanceMode = null
     ) {
         $this->objectManagerFactory = $objectManagerFactory;
         $this->deploymentConfig = $deploymentConfig ?: ObjectManager::getInstance()->get(
             DeploymentConfig::class
+        );
+        $this->maintenanceMode = $maintenanceMode ?: ObjectManager::getInstance()->get(
+            MaintenanceMode::class
         );
         parent::__construct();
     }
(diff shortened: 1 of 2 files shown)


magento-engcom-team:
Hi @kassner. Thank you for your contribution 

miguelbalparda:
> Since the maintenance mode makes sure the frontend is inaccessible, **avoiding any writes to the database through it**, I believe we should disable crons as well, as they have a heavy impact on the database.

This is true only for some cases, I'm not sure if this should be included in the core. Maybe create a config instead of assuming this applies to everybody?

ihor-sviziev:
Hi @miguelbalparda,
Is there any use case when cron jobs should be running during enabled maintenance mode?

miguelbalparda:
It depends on how people is using maintenance mode. I've seen admins putting the site in that mode just to do changes on prices or CMS pages or even deploying new code changes while on maintenance, which sometimes require crons to be run.

ihor-sviziev:
So in this case would be better to add some parameter to env.php file, for
instance “disable_cron_running_in_maintenance_mode”. If this parameter
exists and it is set to 1 - than cron jobs should not be running.

What do you think?

miguelbalparda:
I think that’s no different than disabling Magento_Cron using the bin/magento tool and should have the same effect without adding anything to the core for an use case that’s not super intuitive.

kassner:
Disabling Magento_Cron with `module:disable` is not the same, as it will prevent future db upgrades while running `setup:upgrade`.

miguelbalparda:
Which is consistent with the description of this PR regarding no DB writes while in maintenance, achieving virtually the same objective.

ihor-sviziev:
With disabling cron module we’ll have few issues:
1. We’ll have errors that command cron:run doesn’t exists, these messages
will be sent to email. As for me - it’s not expected
2. Any amount of modules might have dependence on cron module, so we have
to disable all together, it means - not really useful

kassner:
@miguelbalparda no, as I am enabling maintenance mode just because I want to run `setup:upgrade` without anything else concurrently. I still want to run `Magento_Cron`'s UpgradeData/UpgradeSchema.

miguelbalparda:
Maintenance windows are very short usually, which means --force can be used to disable Magento_Cron just for the duration of the window without causing much issues. I still don't see why an edge case like this should be considered in the core, the reasoning behind an inclusion should be a broader benefit and I don't see that in this issue. In either case, this should be consulted with an architect IMO.

sivaschenko:
I think it is a good idea to disable cron by default during maintenance mode as it appears to be the safes approach. I'd add a command option to `maintenance:enable` to keep cron jobs running for those who are confident that cron jobs are not going to be affected during maintenance operations.

ihor-sviziev:
Hi @kandy @buskamuza,
What do you think about such proposal?

app/code/Magento/Cron/Console/Command/CronCommand.php
```diff
@@ -90,6 +103,10 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if ($this->maintenanceMode->isOn()) {
+            $output->writeln('<info>' . 'Cron is disabled because the maintenance mode is enabled.' . '</info>');
+            return;
+        }
         if (!$this->deploymentConfig->get('cron/enabled', 1)) {
```
NadiyaS, line 110: Please pay attention to this configuration. You already may disable cron. Add cron/enabled as 0 to configuration file. There is no special CLI command for this setting for now.

NadiyaS:
Hi,
First, there may be some possible cases when user enables maintenance mode and wants to have cron running. 
Also, there may be cases when you want just to disable cron without maintenance mode.
And we already have this possibility. I added comment to the review.
You may add *cron/enabled* config as 0 to configuration file (e.g. app/etc/env.php) and cron won't be run.
We did not implement CLI command to manage this configuration yet.

ihor-sviziev:
@NadiyaS good catch!
In this case I think would be better to add cli commands `cron:enable` and `cron:disable`. @kassner @buskamuza what do you think?

ihor-sviziev (changes requested):
Hi,

As we discussed with Architects in `#appdesign` - it's better to introduce `cron:enable` and `cron:disable` CLI commands, that will fit all needs and will not add backward incompatible changes.

@kassner  could you update your PR with these chagnes?

sidolov:
@kassner , I am closing this PR now due to inactivity.

PR #21331, opened 2019-02-19
Fix for bug HEAD request returns 404 #21299

Jitheesh:
### Description

After clearing Magento2.3 full page cache,if we try to access Magento 2.3 home via CURL HEAD requests, it returns 404 response and that response is cached for future requests.  This is because of new HttpMethodValidator

### Fixed Issues 

1. magento/magento2#21299: HEAD request returns 404

### Steps to reproduce
Use vanilla Magento 2.3 instance  
1. Clear Full page cache
2. Try to access home page via curl  curl -X HEAD -i http://shop.dev.magento.com/
3. 404 response will be return
4. Try to access home page via browser, 404 page will return
5. Clear full page cache and try to access home page - home page content will return

### debug log
[2019-02-19 09:23:52] main.DEBUG: URI '/'' cannot be accessed with HEAD method (Magento\Cms\Controller\Index\Index) [] []
[2019-02-19 09:23:52] main.DEBUG: Request validation failed for action "Magento\Cms\Controller\Index\Index\Interceptor" [] []

### Manual testing scenarios (*)

1. Clear Full page cache
2. Try to access home page via curl  curl -X HEAD -i http://shop.dev.magento.com/
3. 200 response will be return
4. Try to access home page via browser, it will render home page instead of 404 page

### Contribution checklist (*)
 - [x ] Pull request has a meaningful description of its purpose
 - [x ] All commits are accompanied by meaningful commit messages
 - [] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #21378, opened 2019-02-21
Fix for issue #21299. Change HEAD action mapping to GET action interface and add HEAD request handling

mattijv:
### Description
M2.3 changed the FrontController to explicitly require actions to specify if they respond to HEAD requests. Magento was then caching the 404 response resulting from these HEAD requests and serving them to subsequent GET requests even if they normally would have resulted in 200 response.

### Fixed Issues
1. magento/magento2#21299: HEAD request returns 404

### Manual testing scenarios
1. M2.3 store using the built-in full page cache
2. Make a HEAD request to for example the /customer/account/login route (`curl -I HEAD http://magento.local/customer/account/login/`)
3. Navigate to http://magento.local/customer/account/login/ in browser

#### Expected result
Customer login page

#### Actual result
404 page

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=21378) <br/>All committers have signed the CLA.

magento-engcom-team:
Hi @mattijv. Thank you for your contribution 

Nazar65:
Hi @mattijv same here -> https://github.com/magento/magento2/pull/21331

mattijv:
@Nazar65 https://github.com/magento/magento2/pull/21331 is not a fix to the underlying issue, it just makes one specific controller accept HEAD requests. A lot of other routes still wont (e.g. /customer/account/login).

The real issue is that Magento caches a 404 response for HEAD requests.

I could see it argued that all routes responding to GET requests should also respond to a HEAD request, but then the issue lies with how the FrontController decides whether a controller can handle a request or not.

If the core team decides that all GET routes should also respond to a HEAD request, then the interface mapping should be changed.

ihor-sviziev:
Hi @mattijv and @Nazar65,
Thank you for communication. We'll discuss those PRs with @swnsma and @buskamuza (Architect) and will decide which one is more correct.

mattijv:
@ihor-sviziev Thank you for your response. I have couple points I'd like to make, for your consideration.

To me it seems the issue is one of the two here:
1) Magento should not require GET routes to explicitly define that they also respond to a HEAD request, since conceptually they are very close to each other.

or

2) Magento is falsely caching the 404 response to a HEAD request and serving that to a later GET request.

In case you determine the problem to be number 1, then my pull request is not correct and the real fault lies in how the different requests are mapped to different intefaces in app/etc/di.xml. HEAD and GET should (in my opinion) both be mapped to the same interface.

If on the other hand you decide that routes should be able to respond to GET requests and ignore HEAD requests, then my pull request _might_ be a way to fix the issue.

No offence meant, but in neither case does https://github.com/magento/magento2/pull/21331 seem like the correct solution, as it is only making one affected controller respond to HEAD requests. This won't fix the issue for several other controllers that respond to GET requests, but not HEAD.

That said, I trust that a correct solution is implemented for this issue.

Nazar65:
@mattijv did you find why this issue not present in 2.2-develop ? just checked on 2.2-dev and works all perfectly. Maybe we need find why this works on 2.2-dev ?

mattijv:
@Nazar65 The problem doesn’t exist in M2.2 because M2.3 added validation that checks if a controller responds to a certain type of HTTP request (GET, HEAD, POST etc). In M2.2 GET and HEAD requests were treated the same, but now a controller must specify that it will respond to HEAD requests.

Nazar65:
@mattijv with your solution we always get 404 with curl i think this is not correct, if i do request like ` curl -X HEAD -i http://magento23dev.vg/index.php/` - response 404 
but when i go to web i have 200 ok 
Seems like not correct behavior
and all others like `/customer/account/create/ ` always return "404" on curl

mattijv:
@Nazar65 Yes, like I stated in https://github.com/magento/magento2/pull/21378#issuecomment-466366517, there are two ways this can be fixed. Either the routes SHOULD respond to HEAD requests, in which case my fix is wrong, or they can respond with a 404 to a HEAD request but 200 to a GET, in which case my fix should be something close to what should be implemented.

It's up to the core team to decide if GET routes should automatically respond to a HEAD request.

ihor-sviziev (changes requested):
Hi @mattijv,
We discussed this issue with Architects and here is result: https://github.com/magento/magento2/issues/21299#issuecomment-468298692
Could you update your PR based on following comment?

mattijv:
@ihor-sviziev Thank you for the message. I wrote some thoughts in that thread, but would be happy to update this PR if my proposed changes are deemed a suitable solution.

mattijv:
@ihor-sviziev I've updated the pull request based on conversations here: https://github.com/magento/magento2/issues/21299#issuecomment-468298692

The changes most likely require review.

The critical change for fixing the original issue is the change of the HEAD request map in the app/etc/di.xml to the GET request interface. I've left the now useless HttpHeadActionInterface for backwards compatibility, but marked it deprecated (hopefully I got that right).

I also added code to the framework that strips the body from HEAD requests as this was not present before (the HEAD requests were responded to with the full body).

lib/internal/Magento/Framework/App/Http.php
```diff
@@ -142,12 +142,26 @@ public function launch()
...
         return $this->_response;
     }
 
+    /**
+     * Handle HEAD requests by adding the Content-Length header and removing the body from the response.
+     * @return void
+     */
+    private function handleHeadRequest()
+    {
+        $contentLength = strlen($this->_response->getContent());
```
Vinai: The strlen function isn’t binary safe (i.e. images and compressed content) and also doesn’t handle multi-byte characters. I‘m not sure about the proper things to do here, but it would  be a good idea to do some more research and maybe add some tests for non-ascii response bodies.

mattijv: I was under the assumption that strlen specifically returns the number of bytes in the string, instead of the number of characters.[1] Isn't that what the Content-Length header should contain?

[1] http://php.net/manual/en/function.strlen.php#refsect1-function.strlen-notes

mattijv: I do agree on adding non-ascii test cases, that is a good idea.

mattijv: On further research it seems that some installations of PHP might have overloaded the strlen to use mb_strlen instead.[1] This would break the content length on multibyte characters as they would be counted as one, instead of the actual number of bytes.

Safer way seems to be to call mb_strlen explicitly with encoding that only supports byte characters.[2] This way we should always get the correct byte count for the body contents.

[1] http://php.net/manual/en/mbstring.overload.php
[2] https://stackoverflow.com/a/9718273

Vinai: I guess the important thing to check for would be if it is null byte safe `\0`. That might be PHP version dependent, hopefully PHP7 does not rely on low level C strings. Otherwise your suggested use of `mb_strlen` sounds good to me.

mattijv: As far as I can see, the null byte is treated as a one byte character. It won't terminate strings prematurely.[1]

[1] https://stackoverflow.com/a/32263348

mattijv: Updated the pull request to use mb_strlen instead of strlen and added test cases for some potentially anomalous content.

lib/internal/Magento/Framework/File/Transfer/Adapter/Http.php
```diff
@@ -19,13 +24,16 @@ class Http
...
+     * @param \Magento\Framework\App\Request\Http $request
      * @param \Magento\Framework\App\Response\Http $response
      * @param \Magento\Framework\File\Mime $mime
      */
     public function __construct(
+        \Magento\Framework\App\Request\Http $request,
         \Magento\Framework\HTTP\PhpEnvironment\Response $response,
         \Magento\Framework\File\Mime $mime
     ) {
+        $this->request = $request;
```
ihor-sviziev: 1. Please make this argument last one
2. Please make it optional & fallback to object manager to make this change backward compatible

ihor-sviziev (changes requested):

ihor-sviziev:
Hi @mattijv,
In general - looks good, but one small fix still needed

lib/internal/Magento/Framework/File/Transfer/Adapter/Http.php
mattijv: Done!

mattijv:
@ihor-sviziev It appears that after the requested changes this PR no longer passes the code quality check due to the use of ObjectManager. That's probably not an issue, right?

ihor-sviziev (approved):

ihor-sviziev:
> @ihor-sviziev It appears that after the requested changes this PR no longer passes the code quality check due to the use of ObjectManager. That's probably not an issue, right?

Yes, it's ok.

magento-engcom-team:
[ENGCOM-4474](https://jira.corp.magento.com/browse/ENGCOM-4474) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

lib/internal/Magento/Framework/File/Transfer/Adapter/Http.php
```diff
@@ -88,4 +99,22 @@ private function getFilePath($options): string
...
         return $filePath;
     }
+
+    /**
+     * Set and send all necessary headers.
+     *
+     * @param array $options
+     * @param string $filepath
+     */
+    protected function prepareResponse($options, string $filepath): void
```
YevSent: Why this method has `protected` visibility scope?

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
```diff
@@ -155,6 +155,7 @@ private function setUpLaunch()
     public function testLaunchSuccess()
     {
         $this->setUpLaunch();
+        $this->requestMock->expects($this->once())->method('isHead')->will($this->returnValue(false));
```
YevSent: Please, use `willReturn` instead of `returnValue`

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
```diff
@@ -174,13 +175,72 @@ public function testLaunchException()
...
+    /**
+     * Test that HEAD requests lead to an empty body and a Content-Length header matching the original body size.
+     * @dataProvider dataProviderForTestLaunchHeadRequest
+     * @param string $body
+     * @param int $expectedLength
+     */
+    public function testLaunchHeadRequest($body, $expectedLength)
+    {
+        $this->setUpLaunch();
+        $this->requestMock->expects($this->once())->method('isHead')->will($this->returnValue(true));
```
YevSent: Each method call should be on a separate line.

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
```diff
@@ -174,13 +175,72 @@ public function testLaunchException()
...
+                ['request' => $this->requestMock, 'response' => $this->responseMock]
+            );
+        $this->assertSame($this->responseMock, $this->http->launch());
+    }
+
+    /**
+     * Different test content for responseMock with their expected lengths in bytes.
+     * @return array
+     */
+    public function dataProviderForTestLaunchHeadRequest()
```
YevSent: The return type is missed.

lib/internal/Magento/Framework/File/Test/Unit/Transfer/Adapter/HttpTest.php
```diff
@@ -3,12 +3,21 @@
...
 
 use \Magento\Framework\File\Transfer\Adapter\Http;
 
+/**
+ * Tests http transfer adapter.
+ */
 class HttpTest extends \PHPUnit\Framework\TestCase
 {
+    /**
+     * @var \Magento\Framework\App\Request\Http|\PHPUnit_Framework_MockObject_MockObject
```
YevSent: Use `PhpUnit\Framework\MockObject` instead

lib/internal/Magento/Framework/File/Test/Unit/Transfer/Adapter/HttpTest.php
```diff
@@ -34,7 +43,11 @@ protected function setUp()
             ['setHeader', 'sendHeaders', 'setHeaders']
         );
         $this->mime = $this->createMock(\Magento\Framework\File\Mime::class);
-        $this->object = new Http($this->response, $this->mime);
+        $this->request = $this->createPartialMock(
+            \Magento\Framework\App\Request\Http::class,
```
YevSent: Please, avoid FQCN usage, use import instead.

YevSent (changes requested):

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
mattijv: Do you want me to change the rest of the test to use `willReturn` instead of `returnValue` also?

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
mattijv: Could you elaborate, please?

lib/internal/Magento/Framework/File/Test/Unit/Transfer/Adapter/HttpTest.php
mattijv: Would you like me to also change the rest of the test to use `PhpUnit\Framework\MockObject` instead?

lib/internal/Magento/Framework/File/Test/Unit/Transfer/Adapter/HttpTest.php
mattijv: Could you elaborate, please? I'm not sure what you mean by this.

YevSent: `use Magento\Framework\App\Request\Http` in the import section and just `Http` class name where you need it.

lib/internal/Magento/Framework/File/Test/Unit/Transfer/Adapter/HttpTest.php
YevSent: It's better to use the one namespace import for the same class instead of two. Please, replace other usages in this test.

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
YevSent: You specified the doc block return type but missed the method return type `dataProviderForTestLaunchHeadRequest(): array`.

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
YevSent: It's better to refactor all such places but I don't insist and you can replace only in your changes.

sidolov (approved):

magento-engcom-team:
[ENGCOM-4474](https://jira.corp.magento.com/browse/ENGCOM-4474) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

**Before:**

before1

**After:**

after

mattijv:
Thank you @o-iegorov for fixing the issues pointed out by @joni-jones. I was unfortunately unable to find the time to work on them.

SAN1TAR1UM:
I may be speaking out of turn here, but I created a composer patch from this PR diff, and the problem still exists. 

After successful deployment  on 2.3.1, and verifying the patch was applied properly I flushed the magento cache. Then ran `curl -I http://domain` which resulted in 404 header. Then when loading the site in a browser the homepage 404s.

mattijv:
@SAN1TAR1UM The actual fix for the 404s in this PR is the change to the app/etc/di.xml file. The PHP changes are mostly to handle the HEAD requests per spec. Can you make sure that the composer patch applied the changes to the app/etc/di.xml file?

mattijv:
@SAN1TAR1UM Just to confirm that the fix should work in M2.3.1 I made a clean install and changed the HEAD action mapping in app/etc/di.xml. Before the change the site was returning 404s on curl -I but after the change it correctly returned 200s.

lib/internal/Magento/Framework/File/Test/Unit/Transfer/Adapter/HttpTest.php
```diff
@@ -57,6 +74,9 @@ public function testSend(): void
             ->method('getMimeType')
             ->with($file)
             ->will($this->returnValue($contentType));
+        $this->request->expects($this->once())
+            ->method('isHead')
+            ->will($this->returnValue(false));
```
YevSent: Please, use `$this->willReturnValue` instead.

lib/internal/Magento/Framework/App/Test/Unit/HttpTest.php
```diff
@@ -171,33 +179,105 @@ public function testLaunchSuccess()
...
-            $this->returnCallback(
-                function () {
-                    throw new \Exception('Message');
-                }
-            )
-        );
+        $this->frontControllerMock->expects($this->once())
+            ->method('dispatch')
+            ->with($this->requestMock)->will(
+                $this->returnCallback(
```
YevSent: Please, use `willReturnCallback` or `willThrowException` instead

lib/internal/Magento/Framework/File/Transfer/Adapter/Http.php
```diff
@@ -19,15 +22,24 @@ class Http
...
+    /**
+     * @param \Magento\Framework\HTTP\PhpEnvironment\Response $response
      * @param \Magento\Framework\File\Mime $mime
+     * @param \Magento\Framework\App\Request\Http|null $request
      */
     public function __construct(
         \Magento\Framework\HTTP\PhpEnvironment\Response $response,
-        \Magento\Framework\File\Mime $mime
+        \Magento\Framework\File\Mime $mime,
+        \Magento\Framework\App\Request\Http $request = null
```
YevSent: Please, avoid usage of FQCN

SAN1TAR1UM:
@mattijv I confirmed the patch ran according to the jenkins logs, but upon review of the di.xml file it looks like the change is not there. Looks like I have a deployment issue, I will rectify it and re-test and advise.

SAN1TAR1UM:
Looks like composer install in a fresh directory overwrites the app/etc/di.xml file from the code repository. Does anyone know which composer dep is making this change?

YevSent (approved):

hostep:
@SAN1TAR1UM: `app/etc/di.xml` gets copied from the `magento/magento2-base` package using the [magento composer installer plugin](https://github.com/magento/magento-composer-installer), see the `extra => map` configuration in the `composer.json` file of the `magento/magento2-base` package to see which files gets copied.

It's best not to change the file `app/etc/di.xml` in your project directly, but patch the one in the `magento/magento2-base` package instead.

Hope this helps :)

PR #21458, opened 2019-02-26
Elasticsearch6 implementation.

romainruaud:
### Description (*)

Support Of Elasticsearch 6.

As discussed with @maghamed .

What have been done : 

- Removal of the `_all` field usage which is not supported anymore.
- Writing a new `Suggestions` data provider since the `suggest` endpoint of Elasticsearch is deprecated.
- Associated Unit Tests
- And allow to use ~6.1 version of the PHP client in composer.json

Imho it's enough to deal with the EOL of Elasticsearch 5 and support the 6.

However, the module is quite messy actually (contains many classes and namespaces named after the supported Elasticsearch version, like `Elasticsearch5`, when `Elasticsearch` is supposed to be a mixing between generic implementation and legacy 2.x).

The same apply for System configuration fields which are named according to the version of the engine, which is bad imho.

Most probably, it will need a huge refactoring on next stable (2.4) : 

- Cleanup the mess and re-organize the module structure.
- Deprecate the 5.x support
- Implement ES 7.x support that will be available soon.

1. https://github.com/magento/community-features/issues/141
2. ...

1. Install version 6 of Elasticsearch
2. Configure the Engine in the back-office (it uses new configuration fields, like it was done for 5.x support. Imho this is not good, but I implemented the same way).
3. Run a full reindex and navigate through pages in front, process searches, etc ...


magento-engcom-team:
Hi @romainruaud. Thank you for your contribution 

magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=21458) <br/>All committers have signed the CLA.

orlangur:
@romainruaud @maghamed current implementation approach does not look as a robust solution to me.

> Most probably, it will need a huge refactoring on next stable (2.4) :

It would be pretty hard if not impossible. Much better solution is to implement Elasticsearch 6 support as a new module, possibly disabled by default, old module may be deprecated as a whole when new one is ready.

romainruaud:
@orlangur I agree with you on some points.

It's the whole Elasticsearch "abstraction" **and** the `Search` part of Magento framework (if I understood properly, ES will be the only supported engine now that MySQL fulltext search implementation is facing deprecation too ?) that has to be revamped in my opinion : 

- Search Requests definitions (generic, to be able to query anything)
- Search Indices definition (generic, to be able to index anything)
- Search Query building with abstraction of all ES query and aggregation DSL types
- Properly customizable search relevance
- and so on

We can (and we'd enjoy to) help on these topics, since we already built everything listed above in our Open Source implementation of Elasticsearch (which is widely known and used by the community, even documented in the Magento2 devdocs) : https://github.com/Smile-SA/elasticsuite

But this is a huge amount of work to be done and to coordinate.

And for now, Elasticsearch 5 is EOL on 03/11 : https://www.elastic.co/support/eol

So the goal of this PR was to provide support of ES 6 to the community as soon as possible. 

Let me know how you wanna process on this PR.

In any cases, I'm not sure there should be a new module "dedicated to Elasticsearch 6", or we will face the same mess on a near future when we will begin to write another module "dedicated to Elasticsearch 7" which is inheriting/extending the other one. 

Imho there should be only one module that always handle the latest stable version of Elasticsearch, and that remains compatible with the previous "still maintained" versions with a Deprecation mechanism based on plugins. (So for now, module should fit with ES 6, and handle ES 5 specificity with plugins). 
Once a new version of ES is out, we flush all the deprecation plugins, update the default implementation, and build new deprecation plugins. (when ES 7 is out, we provide default implementation of Elasticsearch 7, implement ES 6 specifics as deprecation plugins, and get rid of 5.x completely, because it's EOL). And on and on.

Maybe this part should even be inside `Framework` directory, or managed on a separate lifecycle.

Product (and more) indexing & querying should be in a separate (most agnostic as possible) modules also.  

Let me know,

Best regards

orlangur:
@romainruaud https://github.com/magento/magento2/tree/2.3-develop/app/code/Magento/Tinymce3 is a good example of approach I described.

With all-in-one module it's pretty hard to determine which code parts are responsible for which Elasticsearch version and which versions are actually supported in each particular moment of time.

`Elasticsearch6` may be a new standalone module containing everything needed for v6 support. Depending on strategy chosen It may contain plugins for older versions support or such code should be placed in `Elasticsearch5` module only.

The same way, when `Elasticsearch7` module will be introduced it should provide full support when enabled by itself and may be based on some older code (but not inherit or extend it).

It's enough if only one of `Elasticsearch5` / `Elasticsearch6` / `Elasticsearch7` may be enabled at a time.

romainruaud:
But with such an approach you will end up with several modules actually sharing 95+% of their code, just like a big copy-paste, and will end up having to fix any issue on all of these nearly identical modules.

Why would you treat supported Elasticsearch versions differently than PHP versions (or any other component) ?

Supporting the latest and "old-stable" on a single module (with proper deprecation mechanism) is enough for me, and seems far easier to maintain.

But we are now a bit far from the initial topic.

romainruaud:
Discussed with @maghamed : 

- I'll rework this PR and provide it as a separated module (which will contain only ES6 related logic).

- Later on, we (or more exactly Magento people) will have to decide which support policy to have with ES versions (either support only latest & old-stable, like for PHP or support more versions with as many separated modules as needed).

I let Igor correct me if needed.

Regards

CajuCLC:
@romainruaud I agree with you that it should be only one module that supports what is still being supported.
I understand that having several modules allows end users to use older versions of ES, but that opens those to security issues in the future. And if end users are not updating ES, most likely they are not updating Magento either.

miguelbalparda:
@CajuCLC that's not always true since software versions are maintained and upgraded by hosting companies and Magento updates are up to the end users.

CajuCLC:
@miguelbalparda That goes back to what's the right way to maintain your app. IMHO the best way is to maintain everything update so there is minimal security risk.
Does Magento want to make everyone happy? Then why not support PHP 5.3 for all versions then?
You can't make everyone happy, but you can make sure your system is up to date in terms of security. Most of hosting companies will also support older versions of software for some time even if it's already EOL just so it gives their customers time to update.
It all comes down to security. Does Magento plan to leave support to software versions that can be exploited in the future or force customers to update to new versions to get best in class security?

romainruaud:
OK @maghamed I think I updated my PR before you drop your comment on the issue :)

It's now built as a separate module. Nothing altered in the old `Magento\Elasticsearch` module and everything specifically needed for ES6 in a new one.

Just one point : I'm not completely aware about what to do to ensure the new module will be properly created on the Magento2 composer private repo.

Feel free to ask anything you want more.

Regards

IvanPletnyov:
@romainruaud Hello. I've created simple product with sku = 'Simple product 1' http://prntscr.com/mqukzb
but in advanced search by sku product isn't founded http://prntscr.com/mqultq . 
Is it expected behavior ?

romainruaud:
@IvanPletnyov this use case does not seem to work on other Elasticsearch implementations (2.x or 5.x).

So most probably, it has never worked as expected. 

If it's easy to fix I can have a look but unfortunately, the main topic of this PR is not to fix all existing functional bugs on Elasticsearch implementation, but rather to ensure compatibility with current stable version of ES 6.

Regards

IvanPletnyov:
@romainruaud Okay, thanks for the answer.

app/code/Magento/Elasticsearch6/Model/Config.php
```diff
@@ -0,0 +1,57 @@
...
+    ) {
+        parent::__construct($scopeConfig, $clientResolver, $engineResolver, $prefix);
+        $this->engineResolver = $engineResolver ?: ObjectManager::getInstance()->get(EngineResolverInterface::class);
+    }
+
+    /**
+     * Return true if third party search engine is used
+     *
+     * @return bool
+     * @since 100.1.0
```
maghamed: probably this `@since 100.1.0` is excessive

app/code/Magento/Elasticsearch6/Model/Config.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param string|null $prefix
+     */
+    public function __construct(
+        \Magento\Framework\App\Config\ScopeConfigInterface $scopeConfig,
+        \Magento\AdvancedSearch\Model\Client\ClientResolver $clientResolver = null,
+        \Magento\Framework\Search\EngineResolverInterface $engineResolver = null,
+        $prefix = null
+    ) {
+        parent::__construct($scopeConfig, $clientResolver, $engineResolver, $prefix);
+        $this->engineResolver = $engineResolver ?: ObjectManager::getInstance()->get(EngineResolverInterface::class);
```
maghamed: taking into account that it's new module - we can just pass all dependencies as required.
No need to mulate backward compatibility

maghamed (approved):
In general, everything looks good. 
@romainruaud you did a great job, and that's a great catch regarding ES6 support in Magento 2.3.1 especially taking into account that EOL of Elasticseach 5 is March 11th of 2019!

We will do our best to make this PR a part of Magento 2.3.1 release

magento-engcom-team:
Hi @maghamed, thank you for the review.
[ENGCOM-4389](https://jira.corp.magento.com/browse/ENGCOM-4389) has been created to process this Pull Request

magento-engcom-team:
@romainruaud thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

app/code/Magento/Elasticsearch6/Model/Config.php
romainruaud: ok fixed.

app/code/Magento/Elasticsearch6/Model/Config.php
romainruaud: I just set them as required.

darksummenors:
So the ES6 will be supported with version 2.3.2? Thanks romainruaud for the great work! 👍

romainruaud:
@darksummenors I think Magento is working internally to make it reach the 2.3.1 release pipeline but it's out of my scope. Maybe I was too much "just-in-time" for 2.3.1 if they flagged it for the next release.

Probably @maghamed can clarify this point.

Regards

tylers-username:
@romainruaud - thanks for working on this.

darksummenors:
It has been added to 2.3.1 (noticed it release notes)!

tylers-username:
@darksummenors - also 2.2.8

PR #21773, opened 2019-03-15
Tax in sales order email is displayed after grand total row

Nazar65:
### Description (*)
Tax in sales order email is displayed after grand total row. If customers place order by non-base currencies, tax in total row is displayed in strange position.

1. magento/magento2#21768: Issue title
2. ...

1. Added any products into cart.
2. Proceed checkout.
3. Place order with non-base currency.
4. Receive order confirmation email.

### DIFF
--- a/app/code/Magento/Tax/Block/Sales/Order/Tax.php
+++ b/app/code/Magento/Tax/Block/Sales/Order/Tax.php
@@ -12,6 +12,8 @@ namespace Magento\Tax\Block\Sales\Order;
 use Magento\Sales\Model\Order;
 
 /**
+ * Tax totals modification block.
+ *
  * @api
  * @since 100.0.2
  */
@@ -84,7 +86,11 @@ class Tax extends \Magento\Framework\View\Element\Template
         $allowTax = $this->_source->getTaxAmount() > 0 || $this->_config->displaySalesZeroTax($store);
         $grandTotal = (double)$this->_source->getGrandTotal();
         if (!$grandTotal || $allowTax && !$this->_config->displaySalesTaxWithGrandTotal($store)) {
-            $this->_addTax();
+            $taxTotal = new \Magento\Framework\DataObject(['code' => 'tax', 'block_name' => $this->getNameInLayout()]);
+            $totals = $this->getParentBlock()->getTotals();
+            if ($totals['grand_total']) {
+                $this->getParentBlock()->addTotalBefore($taxTotal, 'grand_total');
+            }
         }
 
         $this->_initSubtotal();
@@ -118,6 +124,8 @@ class Tax extends \Magento\Framework\View\Element\Template
     }
 
     /**
+     * Init Subtotals.
+     *
      * @return $this
      * @SuppressWarnings(PHPMD.CyclomaticComplexity)
      */
@@ -199,6 +207,8 @@ class Tax extends \Magento\Framework\View\Element\Template
     }
 
     /**
+     * Init Shipping.
+     *
      * @return $this
      */
     protected function _initShipping()
@@ -260,6 +270,8 @@ class Tax extends \Magento\Framework\View\Element\Template
     }
 
     /**
+     * Init discount.
+     *
      * @return void
      */
     protected function _initDiscount()
@@ -267,6 +279,8 @@ class Tax extends \Magento\Framework\View\Element\Template
     }
 
     /**
+     * Initialization grand total.
+     *
      * @return $this
      */
     protected function _initGrandTotal()
@@ -311,6 +325,8 @@ class Tax extends \Magento\Framework\View\Element\Template
     }
 
     /**
+     * Get order.
+     *
      * @return Order
      */
     public function getOrder()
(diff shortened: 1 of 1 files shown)


magento-engcom-team:

app/code/Magento/Sales/Block/Order/Totals.php
```diff
@@ -198,7 +198,10 @@ public function addTotal(\Magento\Framework\DataObject $total, $after = null)
             }
             if (!$added) {
                 $last = array_pop($totals);
-                $totals[$total->getCode()] = $total;
+                // Push tax before grand total.
+                $totals =  array_slice($totals, 0,2, true) +
```
aleron75: please, insert a space after `0,` to comply with coding styles, thanks!

Nazar65: Done, in my IDE i didn't see that :)
DeepinScreenshot_select-area_20190318105921

Nazar65:
Hi @aleron75 any updates?

aleron75 (approved):

magento-engcom-team:
[ENGCOM-4642](https://jira.corp.magento.com/browse/ENGCOM-4642) has been created to process this Pull Request

app/code/Magento/Sales/Block/Order/Totals.php
```diff
@@ -198,7 +198,10 @@ public function addTotal(\Magento\Framework\DataObject $total, $after = null)
             }
             if (!$added) {
                 $last = array_pop($totals);
-                $totals[$total->getCode()] = $total;
+                // Push tax before grand total.
```
orlangur: This code is universal, it has nothing to do with tax or grand total.

app/code/Magento/Sales/Block/Order/Totals.php
```diff
@@ -198,7 +198,10 @@ public function addTotal(\Magento\Framework\DataObject $total, $after = null)
...
+                $totals =  array_slice($totals, 0, 2, true) +
+                    [$total->getCode() => $total] +
+                    array_slice($totals, 2, count($totals) -1, true);
```
orlangur: There is no need to change method, instead of this modify call so that tax  before grand total is explicit.

orlangur (changes requested):

Nazar65:
Hi @orlangur thank you for your review, that's what i need, now it's fixed correctly :+1:

app/code/Magento/Tax/Block/Sales/Order/Tax.php
```diff
@@ -97,13 +97,13 @@ public function initTotals()
     /**
      * Add tax total string
      *
-     * @param string $after
+     * @param string $before
      * @return \Magento\Tax\Block\Sales\Order\Tax
      */
-    protected function _addTax($after = 'discount')
+    protected function _addTax($before = 'grand_total')
```
orlangur: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#modifying-the-default-values-of-optional-arguments-in-public-and-protected-methods

Can we just fix a place where this method is called?

orlangur (changes requested):

app/code/Magento/Tax/Block/Sales/Order/Tax.php
Nazar65: ohh, now i see, how i can be so stupid,!) just change position string and this done.

Nazar65:
@orlangur done :heavy_check_mark:

Nazar65:
HI @orlangur is there something that i need to change ?

orlangur:
@Nazar65 I don't have clear understanding yet, the thing I don't like is that new logic is placed earlier than `_initGrandTotal` call.

Nazar65:
@orlangur, that's doesn't matter,issue comes when we have no discount, and push tax after discount, but discount doesn't exist so we need push before total.

Nazar65:
@orlangur any updates on this one ?

sivaschenko (changes requested):
@Nazar65 considering @orlangur concerns I'd recommend fixing the issue by ensuring that grand total is added last (on line 307), rather than trying to position tax.

Nazar65:
Hi @orlangur @sivaschenko but as you can see at gif, initTotal calls before _initTotal.  so we always get initialized totals in this case. And by default taxAdd($after = 'discount') but at this point we dont have discount, so tax added last.  If tax must be added before grant totals this fix correct, but if this not  required i woullld like just add $this->_addTax('shipping'); and we have the same result. but if we have discount then tax will be aded before discount.
deepin-screen-recorder_Select area_20190426110557

sivaschenko:
@Nazar65 considering that `initTotals` is a public function in API class, in my opinion, we shouldn't rely on assumption that totals were initialized before the call, all at least verify that required keys are present in the array.

Nazar65:
@sivaschenko done :heavy_check_mark:

sivaschenko:
@Nazar65 thanks for updates, but, unfortunately, we cannot process the contribution from the deleted source

PR #21876, opened 2019-03-21
$product->getUrlInStore() does not allow to override the scope in backend context

Nazar65:
### Description (*)
Magento always uses the 'admin' scope (and the default base url) to retrieve product url and returns incorrect url.

By calling the method $product->getUrlInStore(['_scope' => 2, '_nosid' => true]); I'm always retreiving the product's url with default scope in it. (for example http://default.website/catalog/product/view/...)

1. https://github.com/magento/magento2/issues/4247 getProductUrl does not allow to override the scope in backend context

On backend edit page try to get product url from second website
1. Create 2 websites
2. Set difference web urls for websites
3. get product Url on adminhtml in product Edit $product->getUrlInStore(['_scope' => 2, '_nosid' => true]); 
4. You must have url like http://second.website/product.html

Same steps on frontend.

### DIFF
--- a/app/code/Magento/Backend/Model/Url.php
+++ b/app/code/Magento/Backend/Model/Url.php
@@ -13,6 +13,7 @@ use Magento\Framework\App\ObjectManager;
  * Class \Magento\Backend\Model\UrlInterface
  *
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ * @SuppressWarnings(PHPMD.CookieAndSessionMisuse)
  * @api
  * @since 100.0.2
  */
@@ -367,6 +368,19 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
     }
 
     /**
+     * Set scope entity
+     *
+     * @param mixed $scopeId
+     * @return \Magento\Framework\UrlInterface
+     */
+    public function setScope($scopeId)
+    {
+        parent::setScope($scopeId);
+        $this->_scope = $this->_scopeResolver->getScope($scopeId);
+        return $this;
+    }
+
+    /**
      * Set custom auth session
      *
      * @param \Magento\Backend\Model\Auth\Session $session
@@ -402,13 +416,13 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
     }
 
     /**
-     * Retrieve action path.
-     * Add backend area front name as a prefix to action path
+     * Retrieve action path, add backend area front name as a prefix to action path
      *
      * @return string
      */
     protected function _getActionPath()
     {
+
         $path = parent::_getActionPath();
         if ($path) {
             if ($this->getAreaFrontName()) {
@@ -448,8 +462,7 @@ class Url extends \Magento\Framework\Url implements \Magento\Backend\Model\UrlIn
     }
 
     /**
-     * Get config data by path
-     * Use only global config values for backend
+     * Get config data by path, use only global config values for backend
      *
      * @param string $path
      * @return null|string
--- a/dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php
+++ b/dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php
@@ -42,6 +42,46 @@ class UrlTest extends \PHPUnit\Framework\TestCase
         $this->assertStringEndsWith('simple-product.html', $this->_model->getUrlInStore($product));
     }
 
+    /**
+     * @magentoDataFixture Magento/Store/_files/second_store.php
+     * @magentoConfigFixture default_store web/unsecure/base_url http://sample.com/
+     * @magentoConfigFixture default_store web/unsecure/base_link_url http://sample.com/
+     * @magentoConfigFixture fixturestore_store web/unsecure/base_url http://sample-second.com/
+     * @magentoConfigFixture fixturestore_store web/unsecure/base_link_url http://sample-second.com/
+     * @magentoDataFixture Magento/Catalog/_files/product_simple_multistore.php
+     * @dataProvider getUrlsWithSecondStoreProvider
+     * @magentoAppArea adminhtml
+     */
+    public function testGetUrlInStoreWithSecondStore($storeCode, $expectedProductUrl)
+    {
+        $repository = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()->create(
+            \Magento\Catalog\Model\ProductRepository::class
+        );
+        /** @var \Magento\Store\Model\Store $store */
+        $store = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()
+            ->create(\Magento\Store\Model\Store::class);
+        $store->load($storeCode, 'code');
+        /** @var \Magento\Store\Model\Store $store */
+
+        $product = $repository->get('simple');
+
+        $this->assertEquals(
+            $expectedProductUrl,
+            $this->_model->getUrlInStore($product, ['_scope' => $store->getId(), '_nosid' => true])
+        );
+    }
+
+    /**
+     * @return array
+     */
+    public function getUrlsWithSecondStoreProvider()
+    {
+        return [
+           'case1' => ['fixturestore', 'http://sample-second.com/index.php/simple-product-one.html'],
+           'case2' => ['default', 'http://sample.com/index.php/simple-product-one.html']
+        ];
+    }
+
     public function testGetProductUrl()
     {
         $repository = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()->create(


magento-engcom-team:

orlangur (changes requested):
Such change is not allowed, see https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#adding-parameters-in-protected-methods

Nazar65:
@orlangur thanks, i'm will refactor it.

Nazar65:
@orlangur wait a little bit i'm still working on this .

Nazar65:
Hi @orlangur all done :heavy_check_mark: i'm rechecked and run some test, seems like works correctly from frontend and adminhtml

app/code/Magento/Backend/Model/Url.php
```diff
@@ -436,6 +438,27 @@ protected function _getScope()
         return $this->_scope;
     }
 
+    /**
+     * Get scope for the url instance with provided parameter.
+     *
+     * @param $scope string
+     * @return \Magento\Store\Model\Store
+     */
+    protected function getScope($scope = null)
```
sivaschenko: Adding protected functions is backward incompatible and we cannot accept that to a patch release.

Also, the proposed implementation is against CQRS principle (getter should not change the state of the system)

sivaschenko (changes requested):
Thanks for the pull request @Nazar65 Please see my code review comments

Also, as for original issue - I believe it can be solved by setting current store to the store manager `\Magento\Store\Model\StoreManager::setCurrentStore`

Nazar65:
Hi @sivaschenko thanks for reviewing, but `setCurrentStore` dosent slove the issue. As when we call url in admin magento always use hardcoded admin scope.

Nazar65:
@sivaschenko with` \Magento\Store\Model\StoreManager::setCurrentStore` and my solution 
DeepinScreenshot_select-area_20190328123152
DeepinScreenshot_select-area_20190328123210

app/code/Magento/Backend/Model/Url.php
orlangur: @sivaschenko,

> Adding protected functions is backward incompatible

What? That's something new to me. Is it documented and then announced somewhere?

Nazar65:
@sivaschenko i'm found a better solution, give me some time to test it, thanks :+1:

lib/internal/Magento/Framework/Url.php
```diff
@@ -478,8 +483,9 @@ public function getBaseUrl($params = [])
         ) {
             $this->getRouteParamsResolver()->setType(UrlInterface::URL_TYPE_DIRECT_LINK);
         }
-
-        $result = $this->_getScope()->getBaseUrl($this->_getType(), $this->_isSecure());
+        $customScope
+            ? $result = $this->getData('scope')->getBaseUrl($this->_getType(), $this->_isSecure())
+            : $result = $this->_getScope()->getBaseUrl($this->_getType(), $this->_isSecure());
```
orlangur: Who taught you this?)

Please use ternary operator normally, `$result = $customScope ? ... : ...`

orlangur (changes requested):

Nazar65:
My observation, to get correct product url in adminhtml we need create the url object for the product ->

```
        $catalogUrl= $this->_objectManager->get(\Magento\Catalog\Model\ResourceModel\Url::class);
        $products = $catalogUrl->getRewriteByProductStore([$product->getId() => 2]);
        $urlData = new \Magento\Framework\DataObject($products[$product->getId()]);
        $prod->setUrlDataObject($urlData);
```
So after that we have correct url without admin in url. like this -> `http://default.website/product1.html`
But we not able to get second website url even if we set `setCurrentStoreId` becouse magento always use hardcoded admin scope.

lib/internal/Magento/Framework/Url.php
Nazar65: I'm just thinks that this more clearly, do i need to use "if else" ? at this point ?

Nazar65:
Hi @orlangur  Hi @sivaschenko  can you review new changes ?

app/code/Magento/Store/Url/Plugin/RouteParamsResolver.php
```diff
@@ -63,6 +63,7 @@ public function beforeSetRouteParams(
     ) {
         if (isset($data['_scope'])) {
             $subject->setScope($data['_scope']);
+            $subject->setRouteParam('_scope',$data['_scope']);
```
orlangur: There should be a space here according to PSR-2 but I'm not sure if Travis is not broken currently.

lib/internal/Magento/Framework/Url.php
orlangur: Still not fixed.

lib/internal/Magento/Framework/Url.php
```diff
@@ -478,8 +483,8 @@ public function getBaseUrl($params = [])
         ) {
             $this->getRouteParamsResolver()->setType(UrlInterface::URL_TYPE_DIRECT_LINK);
         }
-
-        $result = $this->_getScope()->getBaseUrl($this->_getType(), $this->_isSecure());
+        $result = $customScope ? $this->getData('scope')->getBaseUrl($this->_getType(), $this->_isSecure())
```
orlangur: Please explain the logic. Why we cannot always use `$this->getData('scope')`?

orlangur (changes requested):
@Nazar65 please check my comments. After we finalize the implementation, some tests would _really_ not hurt here.

Please use original ticket https://github.com/magento/magento2/issues/4247 on commit message and make it meaningful.

app/code/Magento/Store/Url/Plugin/RouteParamsResolver.php
Nazar65: ok

lib/internal/Magento/Framework/Url.php
Nazar65: @orlangur Because this metod have usages in many methods and not always sets "_scope"  to the dataObjects, and method _getScope() checks if _scope not exist in dataObject then they give us a hardcoded admin "_scope", so basically I added a 1 more scenario where we can get _scope from data object, and this scenario will not break some tests, i think

Nazar65:
@orlangur I'm tested this a couple day and this works perfect with my custom module, so i'm will write some tests, and will finish on this.

Nazar65:
HI @orlangur @sivaschenko i'm finished the implementation and cover with intergation tests, seems like works fine.

app/code/Magento/Store/Url/Plugin/RouteParamsResolver.php
sivaschenko:
```suggestion
            $subject->setRouteParam('_scope', $data['_scope']);
```

lib/internal/Magento/Framework/Url.php
```diff
@@ -457,8 +457,13 @@ public function getBaseUrl($params = [])
...
 
-        if (isset($params['_scope'])) {
-            $this->setScope($params['_scope']);
+        /**
+         * Custom scope
+         */
+        $customScope = $this->getRouteParamsResolver()->getRouteParam('_scope');
+
+        if (isset($params['_scope']) || $customScope) {
+            isset($params['scope']) ? $this->setScope($params['_scope']) : $this->setScope($customScope);
```
sivaschenko: Operating `$customScope`, `$params['scope']` and `$params['_scope']` is quite confusing. What do you think about extracting the scope setting login to a private method with a meaningful description and parameter names to make this code easier to understand?

dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php
```diff
@@ -42,6 +42,34 @@ public function testGetUrlInStore()
...
+        $store = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()
+            ->create(\Magento\Store\Model\Store::class);
+        $store->load('fixturestore', 'code');
+        $product = $repository->get('simple_2');
+
+        $this->assertEquals(
+            'http://sample.com/index.php/simple-product-2.html',
+            $this->_model->getUrlInStore($product, ['_scope'=> 1, '_nosid' => true])
+        );
+        $this->assertNotEquals(
```
sivaschenko: The changes are non-trivial, so I'd prefer strict comparisons to expected result for both scopes (ideally organized with data provider)

lib/internal/Magento/Framework/Url.php
sivaschenko: I think `_getScope` can be used instead of `getData('scope')` - it's basically the same plus handling the case when scope is not set.

sivaschenko (changes requested):
Hi @Nazar65 thanks for updates, please see my review notes

Nazar65:
Hi @sivaschenko thank you for you review notes that's what i needed. now finally bug in this method **Framework/Url ->** `protected function _getScope()` in some cases php use parent method in **Model/Ur**l `protected function _getScope()` ignore self method wich have scope in dataObject. 

seems like use `$tthis->_getScope() `or `self::_getScope() `fails integrations tests, that why i use `if `statment, and get scope on DataObject $this->getData('_scope');

lib/internal/Magento/Framework/Url.php
```diff
@@ -488,6 +497,21 @@ public function getBaseUrl($params = [])
         return $result;
     }
 
+    /**
+     * Set Requested scope, or custom scope if it sets in routeParams.
+     *
+     * @param array $params
+     */
+    private function setRequestScope($params) :void
```
sivaschenko:
```suggestion
    private function setRequestScope(array $params): void
```

lib/internal/Magento/Framework/Url.php
```diff
@@ -478,8 +488,7 @@ public function getBaseUrl($params = [])
         ) {
             $this->getRouteParamsResolver()->setType(UrlInterface::URL_TYPE_DIRECT_LINK);
         }
-
-        $result = $this->_getScope()->getBaseUrl($this->_getType(), $this->_isSecure());
+        $result = self::_getScope()->getBaseUrl($this->_getType(), $this->_isSecure());
```
sivaschenko: I don't think this change from `$this->_getScope()` to `self::_getScope()` is needed. Can you please provide more details on why do you think it is necessary

sivaschenko (changes requested):
Hi @Nazar65 thanks for the update, see my code review comments

lib/internal/Magento/Framework/Url.php
Nazar65: @sivaschenko   with `self::` we get ` _getScope()` in Framework/Url 
deepin-screen-recorder_Select area_20190425145526
2 - with `$this ` we get ` _getScope()` in Model/Url wich returns admin scope.
deepin-screen-recorder_Select area_20190425145648

Nazar65: But this is not good idea to use one method, when we need to call product url in admin scope, i thinks would be better to use previous solution with **$this->getData('_scope')** and `if `statement, because some tests will fails with this

Nazar65: I think this solution will be better and not break any functional
`$result  = $this->setRequestScope($params)->getBaseUrl($this->_getType(), $this->_isSecure());`
 
```
    private function setRequestScope($params)
    {
        $customScope = $this->getRouteParamsResolver()->getRouteParam('_scope');

        if (isset($params['_scope']) ) {
            $this->setScope($params['_scope']);
        } else if ($customScope) {
            $this->setScope($customScope);
            return $this->getData('scope');
        }
        return $this->_getScope();
    }
```

lib/internal/Magento/Framework/Url.php
```diff
@@ -488,6 +485,26 @@ public function getBaseUrl($params = [])
         return $result;
     }
 
+    /**
+     * Set Requested scope, or custom scope if it sets on params.
+     *
+     * @param array|null $params
+     * @return \Magento\Framework\Url\ScopeInterface
+     */
+    private function setRequestScope($params = null)
```
sivaschenko: This method is not following CQRS principle and therefore is too complicated. Setter should not be used as a getter.

lib/internal/Magento/Framework/Url.php
```diff
@@ -488,6 +485,26 @@ public function getBaseUrl($params = [])
...
+    {
+        $customScope = $this->getRouteParamsResolver()->getRouteParam('_scope');
+
+        if (isset($params['_scope'])) {
```
sivaschenko: What do you think about introducing a `setScope` method to `\Magento\Backend\Model\Url` to be able to assign custom scope to `_scope` property? (it's better to also keep parent method call for backward compatibility purposes)

sivaschenko (changes requested):
Hi @Nazar65 , please see my propositions the in review comments.

lib/internal/Magento/Framework/Url.php
Nazar65: @sivaschenko yes, that's what i needed, now it looks fine, i'm little bit confused by overriding method _getScope()

app/code/Magento/Backend/Model/Url.php
```diff
@@ -366,6 +366,18 @@ protected function _getMenu()
         return $this->_menu;
     }
 
+    /**
+     * Set scope entity
+     *
+     * @param null|int
+     * @return \Magento\Framework\UrlInterface
+     */
+    public function setScope($scopeId)
```
sivaschenko, line 376: Please call parent `setScope` as well to preserve existing functionality (and set `scope` DataObject key)

Also, as I can see \Magento\Framework\Url::setScope has `mixed $params` parameter

sivaschenko (changes requested):

Nazar65:
@sivaschenko done :heavy_check_mark:

app/code/Magento/Backend/Model/Url.php
```diff
@@ -366,6 +366,20 @@ protected function _getMenu()
...
+    /**
+     * Set scope entity
+     *
+     * @param mixed $scopeId
+     * @return \Magento\Framework\UrlInterface
+     */
+    public function setScope($scopeId)
+    {
+        parent::setScope($scopeId);
+        $this->setData('scope', $this->_scopeResolver->getScope($scopeId));
```
sivaschenko: Sorry, @Nazar65 , probably I made a bit confusing statement. I meant that it would be good to call parent `setScope` function so that `scope` data field is set. There is no need to set `scope` DataObject key twice as it's done in the parent function. 

This line can be removed

sivaschenko (changes requested):
Thanks for the quick update!

Please see my review comment

app/code/Magento/Backend/Model/Url.php
Nazar65: ok

Nazar65:
@sivaschenko done

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-4843](https://jira.corp.magento.com/browse/ENGCOM-4843) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

sivaschenko (changes requested):
Hi @Nazar65 can you please recheck the integration tests, they are falling with the following results

    Magento\Catalog\Model\Product\UrlTest::testGetUrlInStoreWithSecondStore
    Magento\Framework\Exception\NoSuchEntityException: The product that was requested doesn't exist. Verify the product and try again.
    
    /var/www/html/app/code/Magento/Catalog/Model/ProductRepository.php:276
    /var/www/html/dev/tests/integration/tmp/sandbox--3c335e531fab12e41ef1a983a92177df/generated/code/Magento/Catalog/Model/ProductRepository/Interceptor.php:24
    /var/www/html/dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php:68

dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php
```diff
@@ -42,6 +42,51 @@ public function testGetUrlInStore()
...
+            $this->_model->getUrlInStore($product, ['_scope'=> $store->getId(), '_nosid' => true])
+        );
+
+        $this->assertEquals(
+            $expected['default'],
+            $this->_model->getUrlInStore($product, ['_scope' => $storeDefault->getId(), '_nosid' => true])
+        );
+    }
+
+    public function testGetUrlsWithSecondStoreProvider()
```
sivaschenko: please remove test prefix from data provider

dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php
```diff
@@ -42,6 +42,51 @@ public function testGetUrlInStore()
...
+    {
+        return [
```
sivaschenko, line 79: the correct return format to have 2 arguments ($storeCode, $expectedProductUrl) in the test is:

    return ['iteration1' => ['fixturestore', 'http://...'], 'interation2' => ['default', 'http://...']];

dev/tests/integration/testsuite/Magento/Catalog/Model/Product/UrlTest.php
```diff
@@ -42,6 +42,51 @@ public function testGetUrlInStore()
...
+    /**
+     * @magentoDataFixture Magento/Store/_files/second_store.php
+     * @magentoConfigFixture default_store web/unsecure/base_url http://sample.com/
+     * @magentoConfigFixture default_store web/unsecure/base_link_url http://sample.com/
+     * @magentoConfigFixture fixturestore_store web/unsecure/base_url http://sample-second.com/
+     * @magentoConfigFixture fixturestore_store web/unsecure/base_link_url http://sample-second.com/
+     * @magentoDataFixture Magento/Catalog/_files/product_simple_multistore.php
+     * @magentoAppArea adminhtml
+     */
+    public function testGetUrlInStoreWithSecondStore()
```
sivaschenko:
```suggestion
    public function testGetUrlInStoreWithSecondStore($storeCode, $expectedPorductUrl)
```
Please correct the test to match the correct dataprovider usage and add @dataProvider <data provider method name> annotation to phpdoc

sivaschenko (changes requested):
Hi @Nazar65 thanks for the update!

Please correct the test to utilize the data provider functionality

Nazar65:
@sivaschenko done :heavy_check_mark:

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-4843](https://jira.corp.magento.com/browse/ENGCOM-4843) has been created to process this Pull Request

PR #22062, opened 2019-03-30
Fix typehint for ButtonList::update parameter

shochdoerfer:
Magento\Backend\Block\Widget\Button\Item::setData accepts mixed value parameter, so should Magento\Backend\Block\Widget\Button\ButtonList::update. 

Found this issue while experimenting with phpstan.

### DIFF
--- a/app/code/Magento/Backend/Block/Widget/Button/ButtonList.php
+++ b/app/code/Magento/Backend/Block/Widget/Button/ButtonList.php
@@ -81,7 +81,7 @@ class ButtonList
      *
      * @param string $buttonId
      * @param string|null $key
-     * @param string $data
+     * @param mixed $data
      * @return void
      */
     public function update($buttonId, $key, $data)


rogyar:
Hi @shochdoerfer. Thank you for your contribution. Your assumption would be correct in case if `Magento\Backend\Block\Widget\Button\ButtonList::update` method was a wrapper for `Magento\Backend\Block\Widget\Button\Item::setData`. But `Magento\Backend\Block\Widget\Button\ButtonList::update` also contains it's own logic and `$data` is supposed to be a string in this particular case. Please, take a look at the following lines

https://github.com/magento/magento2/blob/240e0b731ffebb87435f5cb3a61d83de69a8e493/app/code/Magento/Backend/Block/Widget/Button/ButtonList.php#L92-L94

where `$data` is used as `$this->buttons` array index. As result the method's logic expects $data to be a string (technically it might be INT but it's not the case).

shochdoerfer:
Thanks @rogyar, I missed $data in that part of the method. Well then I see 2 ways of going forward, both would be BC breaks, not sure how "open" you are for that. It clearly shows me the method is doing too much, limiting $data to just being an int does not make sense for the part dealing with the item. We could:
- split the method into two, keeping the update method to pass data on the item class and having a separate method for updating the button position.
- keep one method and add an additional type check in the part you quoted. Event today you could break the code by passing an array or object as $data which would break the array handling logic.

What do you think?

rogyar:
Hi @shochdoerfer. It depends on what problem we are going to solve here. As I mentioned above, the type description in DocBlock is correct for the current implementation. If we are going to add additional checks, we could use type hinting for the $data argument. However, we can bring additional issues with backward compatibility. So, I'm not sure that the change worth the effort. 
I see no problem in the current implementation. In the basic Magento code that goes from the initial 2.0 release type hinting is not widely used. This situation is being fixed in the new implementations with type hinting and strict types.  
You can refactor the method keeping in mind backward compatibility to improve the current implementation. But since we have no issues with the current version, I would just leave it as it is.

sidolov:
Hi @shochdoerfer , I am closing this PR now due to inactivity.

PR #22064, opened 2019-03-30
#22063 Category link store id fix

ilnytskyi:
### Description (*)
When there is a loop going through a few products and linking those products to categories. Some categories get data copied from another store.

Load a category from the default store instead of taking the last cached from CategoryRepository.

I think it would also be better to save the category via its repository. Unfortunately the CategoryRepository ignores posted products `$category->setPostedProducts($productPositions);`.
(probably a case for another bug)

1. https://github.com/magento/magento2/issues/22063

1. Create many stores 
2. Give different descriptions and short descriptions to a category in each store.
3. Loop through different stores e.g. with emulation or just load categories based on product store id
4. Link a product to a category.
5. Category has correct data (data was not copied from the category instance that loaded in different store)

### DIFF
--- a/app/code/Magento/Catalog/Model/CategoryLinkRepository.php
+++ b/app/code/Magento/Catalog/Model/CategoryLinkRepository.php
@@ -38,7 +38,7 @@ class CategoryLinkRepository implements \Magento\Catalog\Api\CategoryLinkReposit
      */
     public function save(\Magento\Catalog\Api\Data\CategoryProductLinkInterface $productLink)
     {
-        $category = $this->categoryRepository->get($productLink->getCategoryId());
+        $category = $this->categoryRepository->get($productLink->getCategoryId(), 0);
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/Model/CategoryLinkRepository.php
```diff
@@ -72,7 +72,7 @@ public function delete(\Magento\Catalog\Api\Data\CategoryProductLinkInterface $p
      */
     public function deleteByIds($categoryId, $sku)
     {
-        $category = $this->categoryRepository->get($categoryId);
+        $category = $this->categoryRepository->get($categoryId, 0);
```
orlangur, line 75: I believe there should be a constant for this value.

orlangur (changes requested):
Also, please implement an integration test revealing described issue.

orlangur:
Thanks for collaboration!

PR #22113, opened 2019-04-02
Attributes named like mysql reserved keywords will break flat catalog indexing

ghost:
### Description (*)
Having a product attribute named condition or some other mysql reserved keyword will result in an error while indexing flat catalog:

Ex:

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'condition IS NOT NULL)' at line 5, query was: UPDATE catalog_product_flat_1_tmp_indexer AS et
LEFT JOIN eav_attribute_option_value AS t0 ON t0.option_id = et.condition AND t0.store_id = 0
LEFT JOIN eav_attribute_option_value AS ts ON ts.option_id = et.condition AND ts.store_id = 1
SET et.condition_value = IFNULL(ts.value, t0.value)
WHERE (condition IS NOT NULL)

Fixes flat catalog indexing for that scenario
2. ...

### DIFF
--- a/app/code/Magento/Catalog/Model/Indexer/Product/Flat/FlatTableBuilder.php
+++ b/app/code/Magento/Catalog/Model/Indexer/Product/Flat/FlatTableBuilder.php
@@ -365,7 +365,7 @@ class FlatTableBuilder
                         []
                     )->columns(
                         [$columnName => $this->_connection->getIfNullSql('ts.value', 't0.value')]
-                    )->where($attributeCode . ' IS NOT NULL');
+                    )->where('et.' . $attributeCode . ' IS NOT NULL');
                     if (!empty($changedIds)) {
                         $select->where($this->_connection->quoteInto('et.entity_id IN (?)', $changedIds));
                     }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22113) <br/>All committers have signed the CLA.

orlangur:
@florinsith any contribution should be delivered to `2.3-develop` first. Is it reproducible there?

ghost:
The 2.3-develop code is different and weird but I dont think it has an issue. By looking at the code it should result in something like:

.....
SET et.condition_value = IFNULL(ts.value, t0.value)
WHERE (IFNULL(ts.value, t0.value) IS NOT NULL)
....

As you notice the attribute code is replaced by that ifnull function in the where condition.

So I guess its just the 2.2 branch,

orlangur:
@florinsith ok, then corresponding changes needs to be backported from `2.3-develop` (as full commits) instead of custom ones applied.

orlangur:
Thanks for collaboration!

ghost:
I dont really understand what you mean, I am not familiar with this flow. How would I make the change on 2.3-develop  since that has different code?

orlangur:
@florinsith you need to find commits where fixed logic was introduced for `2.3-develop` and cherry-pick them to `2.2-develop`.

PR #22131, opened 2019-04-03
Fixes Issue #20125. Set sortOrder of product_has_weight component in product form

jayankaghosh:
Set sortOrder of product_has_weight component in product form

### Description (*)
Set sortOrder of product_has_weight component in product form to one greater than weight field, so that the layout is not disrupted.

1. magento/magento2#20125: Design issue in product creation form

1. Login to admin panel
2. On left navigation goto Stores-> Attributes-> Attribute Set
3. Create a new attribute set on based on default attribute
4. Now Stores-> Attributes -> Product
5. Create an attribute of the text field
6. Now assign created attribute to previously created attribute set on step 3.
7. Now try to create a product with the above-created attribute set
8. The product form design should not be disrupted

### DIFF
--- a/app/code/Magento/Catalog/Ui/DataProvider/Product/Form/Modifier/General.php
+++ b/app/code/Magento/Catalog/Ui/DataProvider/Product/Form/Modifier/General.php
@@ -218,6 +218,12 @@ class General extends AbstractModifier
 
             $hasWeightPath = $this->arrayManager->slicePath($weightPath, 0, -1) . '/'
                 . ProductAttributeInterface::CODE_HAS_WEIGHT;
+
+            $sortOrder = $this->arrayManager->get(
+                $this->arrayManager->findPath('sortOrder', $meta, $weightPath),
+                $meta
+            );
+
             $meta = $this->arrayManager->set(
                 $hasWeightPath . static::META_CONFIG_PATH,
                 $meta,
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/Ui/DataProvider/Product/Form/Modifier/General.php
```diff
@@ -228,6 +234,7 @@ protected function customizeWeightField(array $meta)
                     'componentType' => Form\Field::NAME,
                     'dataScope' => 'product_has_weight',
                     'label' => '',
+                    'sortOrder' => $sortOrder + 1,
```
orlangur, line 237: Does this fix guarantee correct order if we have other elements with `sortOrder` and `sortOrder + 1` values?

orlangur (changes requested):

app/code/Magento/Catalog/Ui/DataProvider/Product/Form/Modifier/General.php
jayankaghosh: Usually when we drag and "sort" product attributes on the attribute set form, they are assigned sort orders with a difference of 10 (60, 70, 80,...). So unless some extension explicitly wrote a form modifier and assigned the sort order to be the same as weight or weight + 1, I think it shouldn't be a problem

orlangur: @jayankaghosh fix should guarantee order even when numbers are not consecutive.

sidolov:
@jayankaghosh , I am closing this PR now due to inactivity.

PR #22186, opened 2019-04-06
22127: check that products is set

davidverholen:
Magento 2.3.0: getSize call on configurable collection leads to exception, if no product filters are applied

### Description (*)
The newrelic cron tries to get the amount of configurable products (or at least the given children) calling the getCount function of the ConfigurableProductManagementInterface.
In this process a configurable collection object is created (magento/module-configurable-product/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php).
There the getSize function (in AbstractDB parent class) is called and as a consequence the _renderFilters function in the magento/module-configurable-product/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php class.

1. https://github.com/magento/magento2/issues/22127

1. Run magento_newrelicreporting_cron cron job

### DIFF
--- a/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
+++ b/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
@@ -12,6 +12,7 @@ namespace Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configura
  *
  * @api
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ * @SuppressWarnings(PHPMD.CookieAndSessionMisuse)
  * @since 100.0.2
  */
 class Collection extends \Magento\Catalog\Model\ResourceModel\Product\Collection
--- a/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
+++ b/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
@@ -26,7 +26,7 @@ class Collection extends \Magento\Catalog\Model\ResourceModel\Product\Collection
     /**
      * @var \Magento\Catalog\Model\Product[]
      */
-    private $products;
+    private $products = [];
 
     /**
      * Assign link table name


davidverholen:

magento-engcom-team:
Hi @davidverholen. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @davidverholen, here is your [new Magento instance](https://pr-22186.instances.magento-community.engineering).
Admin access: [https://pr-22186.instances.magento-community.engineering/admin](https://pr-22186.instances.magento-community.engineering/admin)

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
```diff
@@ -79,12 +79,12 @@ protected function _renderFilters()
...
         $parentIds = [];
-        foreach ($this->products as $product) {
-            $parentIds[] = $product->getData($metadata->getLinkField());
+        if ($this->products) {
+            foreach ($this->products as $product) {
+                $parentIds[] = $product->getData($metadata->getLinkField());
+            }
         }
-
         $this->getSelect()->where('link_table.parent_id in (?)', $parentIds);
```
orlangur: What if `$parentIds` is an empty array?

orlangur (changes requested):

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
davidverholen: not sure what select would do, but yes, it would totally make sense to move it into the if because it is empty anyway when the if is skipped

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
```diff
@@ -77,14 +77,14 @@ public function setProductFilter($product)
     protected function _renderFilters()
     {
         parent::_renderFilters();
-        $metadata = $this->getProductEntityMetadata();
-        $parentIds = [];
-        foreach ($this->products as $product) {
-            $parentIds[] = $product->getData($metadata->getLinkField());
+        if (!is_array($this->products)) {
```
orlangur:
```suggestion
        if (!$this->products) {
```

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
```diff
@@ -77,14 +77,14 @@ public function setProductFilter($product)
...
+            return $this;
         }
-
+        $productEntityMetadata = $this->getProductEntityMetadata();
+        $parentIds = array_map(function (Product $product) use ($productEntityMetadata) {
```
orlangur: No reason to use `array_map` here.

orlangur (changes requested):

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
davidverholen: array_map results in less verbose code when you want to create a new array of same size. I see no reason not to use it tbh

davidverholen: plus you gain type safety for array elements

orlangur: Code became much less simpler, please revert.

Type safety is not a real issue and does not justify decrease of readability.

orlangur (changes requested):

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
davidverholen: @orlangur  I disagree with that. It seems to be a very subjective opinion of you

1. array_map is much clearer in its intention: "map values of an array to a new array" .
2. type safety is always better in regards of potential future bugs
3. the type clearly states with which object your dealing (better readability)
4. with this awful value=[]; foreach... approach you have a code block that has to stay together, but is not bound together by syntax

orlangur:
```
        $parentIds = [];	           
        foreach ($this->products as $product) {	
            $parentIds[] = $product->getData($metadata->getLinkField());	
        }
```

is MUCH easier to comprehend and there is nothing subjective with that. 

Please just fix a bug and revert anything unrelated. For using `array_map` instead of such a simple loop you can create a proposal in https://github.com/magento/architecture.

orlangur (changes requested):

orlangur:
For type safety simply add type hint here: https://github.com/magento/magento2/blob/90d4bccced95603e1c380473ae0b4af904d8b03c/app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php#L66 It must not be enforced in some random places, check in setter in enough.

app/code/Magento/ConfigurableProduct/Model/ResourceModel/Product/Type/Configurable/Product/Collection.php
davidverholen: it's subjective as long as you cannot explain, why it is easier

davidverholen: plus it's not an architecture decision, it's just about clean, readable code

davidverholen: for further reading: https://szymonkrajewski.pl/why-you-should-use-array-functions-instead-of-basic-loops/

orlangur: @davidverholen `foreach` I mentioned is a clean, readable code. Please add at least SOME argument why such `foreach` is bad. Your code is not clean and not readable at all.

There is nothing new for me in that article, I know all about such functions and there are good use cases for them but this is not the one.

If you want to perform such kind of useless refactoring, you need to get an architectural approval first and add recommendation somewhere in https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html which would require any suitable loop to be written in form of `array_map`.

This PR is a bugfix and your "refactoring" is not related to achieving a goal. That's why I'm asking you to revert unrelated changes and deliver suggested improvements in separate PR once they get approved.

davidverholen: I provided several arguments why array_map is superior in that context, while you provided exactly 0

davidverholen:
@orlangur it's not a random place, it's where the object is actually used and where from a developer experience and stability perspective it's a total benefit to have type safety.

You still did not bring one real argument against using array_map here (besides you obviously not liking it).

So why do we have this discussion?

josevectorche:
Sure thanks for the event.

arnoudhgz:
@orlangur @davidverholen is it not more convenient and better to understand if you just set the `$this->products` variable default to an empty array since it is already a `private` variable?

Also on the `setProductFilter` method it expects already an array.

davidverholen:
@arnoudhgz I like the Idea, because it's very simple and maybe we can then finally close this pr. Thx for the suggestion, I will change it

orlangur (approved):
@davidverholen thanks for getting rid of unrelated code refactoring 👍 

@arnoudhgz nice catch, I overlooked such simplification 🙇

magento-engcom-team:
[ENGCOM-5099](https://jira.corp.magento.com/browse/ENGCOM-5099) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

**Before:**

before

**After:**

after

raghuiamsingh:
How can I apply this change to a Magento Installation on v2.3.2?

PR #22211, opened 2019-04-08
Show converted value for validateForRefund error message

kassner:
### Description

When refunding an order placed in a different currency than the default one (via admin), the refund validation is not taking into account this difference when showing the error message. The message is formatted with the correct currency, but the amount is wrong.

In my tests, with a product of $1.00 and exchange rate for USD-EUR of 0.88, I was getting `The most money available to refund is €1.00.` instead of `The most money available to refund is €0.88.`

### Manual testing scenarios

1. Place an order with at least 2 units of a product via Admin using a different currency than the default one;
2. Partially invoice this order (1 item only);
3. Create a credit memo for the entire order;

### Contribution checklist

### DIFF
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -1785,6 +1785,17 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
     }
 
     /**
+     * Retrieve text formatted base price value
+     *
+     * @param float $price
+     * @return string
+     */
+    public function formatBasePriceTxt($price)
+    {
+        return $this->getBaseCurrency()->formatTxt($price);
+    }
+
+    /**
      * Is currency different
      *
      * @return bool
--- a/app/code/Magento/Sales/Model/Order.php
+++ b/app/code/Magento/Sales/Model/Order.php
@@ -1503,7 +1503,7 @@ class Order extends AbstractModel implements EntityInterface, OrderInterface
      * Get item by quote item id
      *
      * @param mixed $quoteItemId
-     * @return \Magento\Framework\DataObject|null
+     * @return  \Magento\Framework\DataObject|null
      */
     public function getItemByQuoteItemId($quoteItemId)
     {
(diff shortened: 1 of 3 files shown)


rogyar:

magento-engcom-team:
Hi @rogyar. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @rogyar, here is your [new Magento instance](https://pr-22211.instances.magento-community.engineering).
Admin access: [https://pr-22211.instances.magento-community.engineering/admin](https://pr-22211.instances.magento-community.engineering/admin)

rogyar:
Hi @kassner. Thank you for your contribution. I believe we have a sort of bug here. An amount in the error message should be in the base currency (with the base currency symbol). So instead of converting base amount to the actual currency amount, we should just change the currency symbol to the base one. 
If you check the Refund Totals section, you can see that all values in the inputs are processed as base currency values. So if we enter some amount in base currency, it does not make sense to show an error with the amount converted to another currency.
The approach for Refund Shipping is correct in this case. If you put a value that is greater than a shipping cost, you will see an error message with the maximum value (in base currency, which is correct) that you can enter in this field.

kassner:
@rogyar I updated the code to fix it.

rogyar (approved):

magento-engcom-team:
[ENGCOM-5172](https://jira.corp.magento.com/browse/ENGCOM-5172) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

**Before:**

before

**After:**

after

Nazar65 (changes requested):
https://github.com/magento/magento2/blob/286009fe8e91c4a6463b61440750b4f39b741b87/app/code/Magento/Sales/Model/Order.php#L1793

Hi @kassner thank you for collaboration, such changes is Backward Incompatible, please see ->https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/

> Introduction of a method to a class or interface
> 
> The new interface may take over existing methods from the class if it makes sense to group them together. In this case, you must deprecate corresponding methods in the old interface/class with the @see annotation. The old methods should proxy the calls to the new interface instead of duplicating the logic.
> 
> For an example of an interface with an extracted method see the Magento\Catalog\Api\CategoryListInterface. This interface is responsible for the getList() method, but Magento\Catalog\Api\CategoryRepositoryInterface does not have that method.
> 
> For a PATCH product release, do NOT mark the new interface with @api.
> For a MINOR product release, an architect marks, or approves, the new interface with @api if applicable.

kassner:
@Nazar65 I changed the PR to not touch `\Magento\Sales\Model\Order`

Nazar65:
Hi @rogyar please review.

sidolov (approved):

magento-engcom-team:
[ENGCOM-5172](https://jira.corp.magento.com/browse/ENGCOM-5172) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

PR #18641, opened 2018-10-16
Set cache id prefix on installation

schmengler:
### Description (*)

The undocumented `id_prefix` option for the cache frontend is used to prefix cache keys. If it is not set, Magento uses the first 12 bits of the md5 hash of the absolute path to Magentos app/etc directory. But if this is not exactly the same on all web servers, cache invalidation does not work.

**Source:** https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/App/Cache/Frontend/Factory.php#L121-L130

To prevent this issue, the value shall be set on installation, so that the fall back on the fly does not happen anymore. Optionally, the value can be specified explicitly.

### Fixed Issues (if relevant)

1. magento/magento2#15828: Multisite installation, default website slow

### Manual testing scenarios (*)

1. Install Magento via CLI or Web Wizard
2. There should be new options "cache-id-prefix" and "page-cache-id-prefix" 
3. The value is saved in `env.php` at `cache/frontend/{default,page_cache}/id_prefix`

### Contribution checklist (*)
 - [X] Pull request has a meaningful description of its purpose
 - [X] All commits are accompanied by meaningful commit messages
 - [X] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

PR #22228, opened 2019-04-08
Fix loss of page_cache cache_dir setting from di.xml

Vinai:
### Description

In a native Magento 2 installation both the page_cache and the default cache is stored in the `var/cache` directory. The `var/page_cache` directory is not used, even, though it is [configured in di.xml](https://github.com/magento/magento2/blob/2.3-develop/app/etc/di.xml#L771).

This makes it impossible to use `bin/magento cache:flush full_page` to only clear the FPC, as it will also flush the cache directory.

The reason is any cache options configured in `di.xml` are lost if any configuration for a cache frontend is present in `app/etc/env.php`.

During the Magento installation a cache id_prefix for both the [default](\Magento\Setup\Model\ConfigOptionsList\Cache::createConfig) and the [page_cache](\Magento\Setup\Model\ConfigOptionsList\PageCache) frontends are written to the `app/etc/env.php`.

The bug was introduced by the PR: https://github.com/magento/magento2/pull/18641  
(note: the issue that was solved by that PR is also relevant for non-filesystem cache backends.) 

In the method body of `\Magento\Framework\App\Cache\Frontend\Pool::_getCacheSettings` the original comment read:

```
Merging is intentionally implemented through array_merge() instead of array_replace_recursive()
to avoid "inheritance" of the default settings that become irrelevant as soon as cache storage type changes
```

This reasoning of the original comment applies only if the backend type is changed, but not if an option is added to be used with the default.

### Fixed Issues

This PR will merge the deployment configuration and the `di.xml` configuration using `array_replace_recursive` instead of `array_merge`, as cache settings that are not used by other cache backends don't do any harm, and still can be changed by overwriting them when needed. 

An alternative implementation of the solution checked if the `page_cache` backend was set to a different value of the `default` cache frontend, but that added a lot of additional coupling and in favor of simplicity I believe the solution in this PR is better.

The above fix revealed a bug that was masked before, namely that the when the built in page cache was is used, moving a category does not clean the appropriate cache records.
With varnish this issue does not occur because there the category top navigation is rendered as an ESI block, which is associated with all category ID cache tags (`cat_c_<ID>`).  
The page cache records however are only associated with the generic category cache tag (`cat_c`).  
Moving a category only cleans records that are associated with the category ID cache tag.  
The solution is implemented as an event observer for the `category_move` event, which cleans the generic category cache tag if the built in page cache is used.

### Manual testing scenarios

1. download source Magento (tarball or composer or github clone)
2. install Magento (ui or cli)
3. access a frontend or adminhtml page in the browser
4. With the bug: `var/page_cache` is empty and `var/cache` is not empty
5. Run `bin/magento cache:flush full_page`
6. Check `var/cache` is empty

After the patch is applied `var/page_cache` will no longer be empty after a page is accessed and `bin/magento cache:flush full_page` will only flush `var/page_cache`

### Contribution checklist

### DIFF
--- a/dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
+++ b/dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
@@ -1,0 +1,55 @@
+<?php declare(strict_types=1);
+
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Framework\App\Cache\Frontend;
+
+use Magento\Framework\ObjectManager\ConfigInterface as ObjectManagerConfig;
+use Magento\TestFramework\ObjectManager;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * This superfluous comment can be removed as soon as the sniffs have been updated to match the coding guide lines.
+ */
+class PoolTest extends TestCase
+{
+    public function testPageCacheNotSameAsDefaultCacheDirectory(): void
+    {
+        /** @var ObjectManagerConfig $diConfig */
+        $diConfig = ObjectManager::getInstance()->get(ObjectManagerConfig::class);
+        $argumentConfig = $diConfig->getArguments(\Magento\Framework\App\Cache\Frontend\Pool::class);
+
+        $pageCacheDir = $argumentConfig['frontendSettings']['page_cache']['backend_options']['cache_dir'] ?? null;
+        $defaultCacheDir = $argumentConfig['frontendSettings']['default']['backend_options']['cache_dir'] ?? null;
+
+        $noPageCacheMessage = "No default page_cache directory set in di.xml: \n" . var_export($argumentConfig, true);
+        $this->assertNotEmpty($pageCacheDir, $noPageCacheMessage);
+
+        $sameCacheDirMessage = 'The page_cache and default cache storages share the same cache directory';
+        $this->assertNotSame($pageCacheDir, $defaultCacheDir, $sameCacheDirMessage);
+    }
+
+    /**
+     * @covers  \Magento\Framework\App\Cache\Frontend\Pool::_getCacheSettings
+     * @depends testPageCacheNotSameAsDefaultCacheDirectory
+     */
+    public function testCleaningDefaultCachePreservesPageCache()
+    {
+        $testData = 'test data';
+        $testKey = 'test-key';
+
+        /** @var \Magento\Framework\App\Cache\Frontend\Pool $cacheFrontendPool */
+        $cacheFrontendPool = ObjectManager::getInstance()->get(\Magento\Framework\App\Cache\Frontend\Pool::class);
+
+        $pageCache = $cacheFrontendPool->get('page_cache');
+        $pageCache->save($testData, $testKey);
+
+        $defaultCache = $cacheFrontendPool->get('default');
+        $defaultCache->clean();
+
+        $this->assertSame($testData, $pageCache->load($testKey));
+    }
+}
--- a/lib/internal/Magento/Framework/App/Cache/Frontend/Pool.php
+++ b/lib/internal/Magento/Framework/App/Cache/Frontend/Pool.php
@@ -3,6 +3,7 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+
 namespace Magento\Framework\App\Cache\Frontend;
 
 use Magento\Framework\App\Cache\Type\FrontendPool;
@@ -55,6 +56,7 @@ class Pool implements \Iterator
 
     /**
      * Create instances of every cache frontend known to the system.
+     *
      * Method is to be used for delayed initialization of the iterator.
      *
      * @return void
@@ -77,18 +79,21 @@ class Pool implements \Iterator
     protected function _getCacheSettings()
     {
         /*
-         * Merging is intentionally implemented through array_merge() instead of array_replace_recursive()
-         * to avoid "inheritance" of the default settings that become irrelevant as soon as cache storage type changes
+         * Merging is intentionally implemented through array_replace_recursive() instead of array_merge(), because even
+         * though some settings may become irrelevant when the cache storage type is changed, they don't do any harm
+         * and can be overwritten when needed.
+         * Also array_merge leads to unexpected behavior, for for example by dropping the
+         * default cache_dir setting from di.xml when a cache id_prefix is configured in app/etc/env.php.
          */
         $cacheInfo = $this->deploymentConfig->getConfigData(FrontendPool::KEY_CACHE);
         if (null !== $cacheInfo) {
-            return array_merge($this->_frontendSettings, $cacheInfo[FrontendPool::KEY_FRONTEND_CACHE]);
+            return array_replace_recursive($this->_frontendSettings, $cacheInfo[FrontendPool::KEY_FRONTEND_CACHE]);
         }
         return $this->_frontendSettings;
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      *
      * @return \Magento\Framework\Cache\FrontendInterface
      */
@@ -99,7 +104,7 @@ class Pool implements \Iterator
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function key()
     {
@@ -108,7 +113,7 @@ class Pool implements \Iterator
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function next()
     {
@@ -117,7 +122,7 @@ class Pool implements \Iterator
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function rewind()
     {
@@ -126,7 +131,7 @@ class Pool implements \Iterator
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function valid()
     {
--- a/app/code/Magento/Catalog/Observer/FlushCategoryPagesCache.php
+++ b/app/code/Magento/Catalog/Observer/FlushCategoryPagesCache.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types=1);
+/**
+ *
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Catalog\Observer;
+
+use Magento\Catalog\Model\Category;
+use Magento\Framework\Event\Observer as Event;
+use Magento\Framework\Event\ObserverInterface;
+use Magento\PageCache\Model\Cache\Type as PageCache;
+use Magento\PageCache\Model\Config as CacheConfig;
+
+/**
+ * Flush the built in page cache when a category is moved
+ */
+class FlushCategoryPagesCache implements ObserverInterface
+{
+
+    /**
+     * @var CacheConfig
+     */
+    private $cacheConfig;
+
+    /**
+     *
+     * @var PageCache
+     */
+    private $pageCache;
+
+    /**
+     * FlushCategoryPagesCache constructor.
+     *
+     * @param CacheConfig $cacheConfig
+     * @param PageCache $pageCache
+     */
+    public function __construct(CacheConfig $cacheConfig, PageCache $pageCache)
+    {
+        $this->cacheConfig = $cacheConfig;
+        $this->pageCache = $pageCache;
+    }
+
+    /**
+     * Clean the category page cache if built in cache page cache is used.
+     *
+     * The built in cache requires cleaning all pages that contain the top category navigation menu when a
+     * category is moved. This is because the built in cache does not support ESI blocks.
+     *
+     * @param Event $event
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function execute(Event $event)
+    {
+        if ($this->cacheConfig->getType() == CacheConfig::BUILT_IN && $this->cacheConfig->isEnabled()) {
+            $this->pageCache->clean(\Zend_Cache::CLEANING_MODE_MATCHING_ANY_TAG, [Category::CACHE_TAG]);
+        }
+    }
+}
--- a/app/code/Magento/Catalog/etc/adminhtml/events.xml
+++ b/app/code/Magento/Catalog/etc/adminhtml/events.xml
@@ -12,4 +12,7 @@
     <event name="catalog_category_change_products">
         <observer name="category_product_indexer" instance="Magento\Catalog\Observer\CategoryProductIndexer"/>
     </event>
+    <event name="category_move">
+        <observer name="clean_cagegory_page_cache" instance="Magento\Catalog\Observer\FlushCategoryPagesCache" />
+    </event>
 </config>
--- a/dev/tests/integration/testsuite/Magento/GraphQlCache/Controller/AbstractGraphqlCacheTest.php
+++ b/dev/tests/integration/testsuite/Magento/GraphQlCache/Controller/AbstractGraphqlCacheTest.php
@@ -7,9 +7,15 @@ declare(strict_types=1);
 
 namespace Magento\GraphQlCache\Controller;
 
-use PHPUnit\Framework\TestCase;
-use Magento\TestFramework\ObjectManager;
+use Magento\Framework\App\Request\Http as HttpRequest;
+use Magento\Framework\App\Response\HttpInterface as HttpResponse;
+use Magento\Framework\Registry;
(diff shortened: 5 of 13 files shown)


dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
```diff
@@ -0,0 +1,47 @@
+<?php declare(strict_types=1);
```
larsroettig, line 1:
```
/**
 */

```
Licence block is missing

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+        $pageCache = $cacheFrontendPool->get('page_cache');
+        $pageCache->save($testData, $testKey);
+
+        $defaultCache = $cacheFrontendPool->get('default');
+        $defaultCache->clean();
+
+        $this->assertSame($testData, $pageCache->load($testKey));
+    }
+}
```
larsroettig, line 55: Empty Line missing

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
larsroettig, line 1: Docblock missing 

```
/**
 */
```

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
```diff
@@ -0,0 +1,47 @@
+<?php declare(strict_types=1);
+
+namespace Magento\Framework\App\Cache\Frontend;
+
+use Magento\Framework\ObjectManager\ConfigInterface as ObjectManagerConfig;
+use Magento\TestFramework\ObjectManager;
+use PHPUnit\Framework\TestCase;
+
```
larsroettig, line 13: Class descriptions missing - code style will fail

larsroettig (changes requested):
Steps for quality:
- ✅ Review please see comments 
- ❗️ Travis build 
- ❗️ Manual Testing

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
Vinai: Fixed. But I'm curious what is the reason for this rule? It seems very arbitrary and useless to me...

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
Vinai: Added.

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
Vinai: Added.

Vinai:
Thanks for the review @larsroettig, sorry for being a little bitter sometimes. I really do appreciate you taking the time to look it over 👍

Vinai:
The Codacy warnings are related only to outdated rule names in the ruleset:

* **The "WordPress.XSS.EscapeOutput" sniff has been renamed to "WordPress.Security.EscapeOutput". Please update your custom ruleset.**
* **The "WordPress.CSRF.NonceVerification" sniff has been renamed to "WordPress.Security.NonceVerification". Please update your custom ruleset.**
* **The "WordPress.WP.PreparedSQL" sniff has been renamed to "WordPress.DB.PreparedSQL". Please update your custom ruleset.**

Bildschirmfoto 2019-04-09 um 04 50 49

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+namespace Magento\Framework\App\Cache\Frontend;
+
+use Magento\Framework\ObjectManager\ConfigInterface as ObjectManagerConfig;
+use Magento\TestFramework\ObjectManager;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * This is prime example for a senseless comment that only exists to satisfy the coding standard.
```
orlangur: I feel your pain, this is one of mine top priorities in phpcs-related stuff, annoys me since `@return void` was introduced.

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
orlangur: @larsroettig didn't get which line you talked about...

lib/internal/Magento/Framework/App/Cache/Frontend/Pool.php
```diff
@@ -77,12 +77,15 @@ protected function _initialize()
...
+         * Merging is intentionally implemented through array_replace_recursive() instead of array_merge(), because even
+         * though some settings may become irrelevant when the cache storage type is changed, they don't do any harm
+         * and can be overwritten when needed.
+         * Also array_merge leads to unexpected behavior, for for example by dropping the
+         * default cache_dir setting from di.xml when a cache id_prefix is configured in app/etc/env.php.
          */
         $cacheInfo = $this->deploymentConfig->getConfigData(FrontendPool::KEY_CACHE);
         if (null !== $cacheInfo) {
-            return array_merge($this->_frontendSettings, $cacheInfo[FrontendPool::KEY_FRONTEND_CACHE]);
+            return array_replace_recursive($this->_frontendSettings, $cacheInfo[FrontendPool::KEY_FRONTEND_CACHE]);
```
orlangur, line 90: Let's see if it won't break anything else.

orlangur (approved):
Thanks @Vinai!

Codacy can be completely ignored, I heard about plans to turn it off but not sure why it's still here.

magento-engcom-team:
[ENGCOM-4717](https://jira.corp.magento.com/browse/ENGCOM-4717) has been created to process this Pull Request

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
Vinai: Before I changed the commit line 59 was missing.

orlangur: @Vinai I don't think it's needed. Basically any missing line is not worth commenting - phpcs should catch all needed.

VasylShvorak:
:heavy_check_mark: QA passed

nmalevanec:
@Vinai, could you please check the following failures in unit tests?
https://travis-ci.org/magento/magento2/jobs/517594558#L1195

Vinai:
@nmalevanec Thank you for notifying me - I fixed and expanded the unit test cases.

nmalevanec:
@Vinai, unfortunately we've encountered some problem after your fix. Could you please take a look at such scenario:
1. Navigate to admin area
2. Create two categories with structure like:

image

3. Observe storefront index page

image

4. Move cat2 to the same level as cat1

image

5. Observe storefront index page

### Expected result

image

### Actual result

image

Vinai:
Good catch! My guess is the fix uncovered another bug that was masked by the regular cache and the page cache sharing the storage folder. I'll check what is causing the issue.

Vinai:
Confirmed, moving a category does not flush the full page cache, only category tags in the default cache storage (https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Category.php#L444)
Running `bin/magento cache:clean full_page` makes the new category structure appear on the frontend.
This is a separate bug, not related to this PR. Should I include a patch for it in this PR or is a separate PR better?

orlangur:
@Vinai please include it in the same PR but mention in "Fixed issues" section of description.

As separate commit would be the best option.

Vinai:
After investigating a bit I have a bit better understanding of what is happening.

When a category is moved, only the category ID based cache tag of the moved category is cleared (e.g. `cat_c_4`).

The issue only occurs with the **built in** full page cache.  

When **varnish** is configured, the menu block is rendered as an **ESI** block, which is associated with all category ID tags. That is why the cached menu is cleared when a category is moved, but the page itself is not.

However, with the **built in** full page cache, ESI is not supported.
The problem is that the page is not associated with the ID based category cache tags, only with the generic category tag (i.e. `cat_c`). So the page is not cleared from the full page cache when a category is moved.

This issue already existed before the bug this PR fixes was introduced.

Possible solutions:

1. Always clean the generic category cache tag `cat_c` when a category is moved.
   This is bad because it will clear unnecessary cache records when varnish is used.

2. Implement a special case for the built in cache via a plugin, so when a category is moved, not only the category ID cache tag (e.g. `cat_c_4`) is cleared, but also the generic category cache tag `cat_c`.
    This will effect a complete full page cache flush when the built in cache is used and a category is moved, since all pages contain the top navigation.
    Note to self: This could be implemented with an observer for the `category_move` event that checks if the built in cache is being used, and if so calls 
    `\Magento\Framework\App\CacheInterface::clean([\Magento\Catalog\Model\Category::CACHE_TAG])`.

3. Implement a special case for the built in cache via a plugin, so every page with the top menu is not only associated with the generic category ID tag, but with the tag of each category in the menu (e.g. `cat_c_3,cat_c_4`).
    This will also effect a complete full page cache flush when the built in cache is used and a category is moved.
   Note to self: after plugin on `\Magento\Catalog\Block\Navigation::getIdentities` that adds all category cache tags from categories in `\Magento\Catalog\Model\Layer`.

4. Implement ESI support for the built in cache.
    This is the best solution, however it's a lot of work and definitely outside of the scope of this PR.

If you have further ideas please let me know, otherwise I think I'll go with option 2 or 3.

Vinai:
Implemented option 2 above and added it to the "Fixed Issues" section in the PR description as requested by @orlangur.

Vinai:
To resolve the static test failures, I moved the observer into the catalog module, which already depends on `Magneto_PageCache`. Let's see if those tests now are green.

From a cursory glance at the integration test failures it seems like the bug fixed by this PR has found it's way into a few corners of the `Magento\GraphQlCache` module. It looks as if a bunch of tests expect the page cache to be cleared. My guess is only the regular is cleared, and the flushing of the page cache was a side effect of the bug fixed by this PR.

I haven't looked into the functional test failure. I hope that is fixed automatically once I've also fixed the affected `Magento\GraphQlCache` module.

Vinai:
After a chat with @cpartica it seems that there are some changes in the currently unmerged non-public `2.3.2-develop` branch.
I'll revisit and continue fixing the follow up issues this after it's merged.

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
cpartica: unfortunately Magento code is there for everybody to see and interpret and we have to do yet better, meaningful, rhetorical comments

IvanPletnyov:
@Vinai Hello! On you PR failed 6 integration tests. Could you please fix it ? 
https://testing-service.magento-community.engineering/reports/magento/magento2/pull/22228/ae2ff291-1805-46ad-b247-76a79b62e133/1627/Integration/allure-report-ce/index.html#suites

sivaschenko:
Hi @Vinai please let me know if you need assistance with the integration tests

Vinai:
@IvanPletnyov

In essence, the issues that currently fail in the integration tests is a problem in the GraphQl module.
The GET query caching does not work. My PR fixes the caching, so in theory it works, but now the bug in the GraphQl module is triggered and results in different queries being cached with the same cache record.

As I've written above 3 days ago:

> After a chat with @cpartica it seems that there are some changes in the currently unmerged non-public 2.3.2-develop branch.
> I'll revisit and continue fixing the follow up issues this after it's merged.

I don't think it's a good idea to work on code that already has pending changes.
Once the 2.3.2-develop branch is publicly available, probably when it's merged into 2.3-develop before the 2.3.2 release, then I'll revisit this PR.

Unfortunately this means this bug will still be present in 2.3.2, but better to do things right, rather than rush them.

Vinai:
@sivaschenko Thanks for your offer!
I don't think I need help though, what I need is the 2.3.2-develop branch (see my previous comment).

Thanks!

Vinai:
Quick note: now that 2.3.2 is release I'll pick the PR up again and get the tests to pass.

Vinai:
I've rebased my branch onto current 2.3-develop. Once the tests have completed I'll start fixing the failures.

Vinai:
I've fixed the integration and the static tests.
When looking at the build report for the functional tests there where no failures or errors, contrary to the summary here on github.
Let's see if they fail again. I hope this PR now will finally be ready for merging :)

dev/tests/integration/testsuite/Magento/Framework/App/Cache/Frontend/PoolTest.php
Vinai: Thankfully it's no longer necessary to add useless comments due to the updated phpdoc coding style guide!
I'll remove the docblock.

Vinai:
Force pushed in an attempt to get the stalled build to restart.

davidverholen:
@magento run all tests

Vinai:
@magento run all tests

Vinai:
@magento run all tests

Vinai:
Seems like the builds is still stalled. Do you have an ETA when they will work again?
Will the stalled builds be restarted automatically, or should I take some action?

davidverholen:
@Vinai  not sure yet, Austin is still sleeping I guess :P I hope it gets resolved soon, not sure if someone is already working on it.

Since the "run all tests" command seems to work only for maintainers, you can of course still force push to trigger a new build. But currently it will likely still not work (since all other new PR builds are also not finishing).

I guess, wait for updates in slack :o

Vinai:
Thanks for the answer @davidverholen, I'll just wait and keep an eye on slack 👍

davidverholen:
@magento run all tests

Vinai:
Whohoo, all green! Thanks for the help with the builds!

Vinai:
@cpartica and @sivaschenko 

I finally picked up this old PR regarding the FPC file storage and fixed the failing tests.

You offered to help me with the PR back when I opened it originally before 2.3.2 was released, and because it’s been so long, I wanted to ask you if you have time to help get it merged.

If it needs any more work please let me know - thank you.

Vinai:
Rebasing in progress...

Vinai:
The broken static test is a false positive on the sniff. It's not an unused local variable, it's a property of an internal class.
See https://testing-service.magento-community.engineering/reports/magento/magento2/pull/22228/ba64635c-3ec7-4107-932a-2bf9c9024190/3277/Statics/allure-report-ce/index.html#suites/51d3ab4d5d77a0bc6a7ea344e9e7e6be/bc36be418f95c625/

Will disable it with an annotation.

Vinai:
I've added a `@SuppressWarnings(PHPMD.UnusedLocalVariable)` annotation, but the rule still complains. Any idea how to disable the rule for that line?

EDIT: trying to get phpmd to stop complaining by adding the annotation to the method instead of the nested property.

Vinai:
Yay, static tests are green! Finally!
Now suddenly a functional test reported as failing, even though there has been only a change in a comment since the last run where they where all passing.
Also, when I try to see details about what is failing, all tests are reported as green.
https://testing-service.magento-community.engineering/reports/magento/magento2/pull/22228/64745c2f-7697-4182-9eb8-e78c793de276/9585/Functional/allure-report-ce/index.html#behaviors

Is this a bug in the report github integration?

Anyway, unless further change request are made, from my side I think the PR is ready to be merged.


EDIT: force pushed to trigger a rebuild. Curious if the functional tests are green this time.
EDIT 2: jup, they are green again now.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-4717](https://jira.corp.magento.com/browse/ENGCOM-4717) has been created to process this Pull Request

| Label | Description |

engcom-Delta:
:heavy_check_mark: QA passed

ihor-sviziev:
Hi @sidolov,
Is there any updates since August 9? Why it's still not merged?

sidolov:
Hi @ihor-sviziev , we faced the build failures with provided changes. I'll add more info about failures as soon as possible

Vinai:
@ihor-sviziev @sidolov Thanks for the follow up! I'm quite curious hear what issues occurred during the build and am happy to help fixing them if I can.

PR #22242, opened 2019-04-09
Dispatch adminhtml_controller_customer_group_prepare_save

kassner:
### Description

Adding a bit of extensibility to admin's customer group save, so the form can be extended and the dispatched event allows the developer to set the extensions attribute with new values before saving.

### Contribution checklist

### DIFF
--- a/app/code/Magento/Customer/Controller/Adminhtml/Group/Save.php
+++ b/app/code/Magento/Customer/Controller/Adminhtml/Group/Save.php
@@ -88,6 +88,11 @@ class Save extends \Magento\Customer\Controller\Adminhtml\Group implements HttpP
                 $customerGroup->setCode(!empty($customerGroupCode) ? $customerGroupCode : null);
                 $customerGroup->setTaxClassId($taxClass);
 
+                $this->_eventManager->dispatch('adminhtml_controller_customer_group_prepare_save', [
+                    'request' => $this->getRequest(),
+                    'customer_group' => $customerGroup,
+                ]);
+
                 $this->groupRepository->save($customerGroup);
 
                 $this->messageManager->addSuccess(__('You saved the customer group.'));


YevSent:
@kassner, could you more explain what kind of extension point you want to provide? From the changes in the code is not clear enough. In your changes event contains the entity which has attributes from request like `code` and `tax` but the event is dispatched before saving the entity.

If the main idea to set only extensions attributes it's better to use `afterSave` plugin for customer group repository instead of changing the original object in an observer.

app/code/Magento/Customer/Controller/Adminhtml/Group/Save.php
```diff
@@ -88,6 +88,11 @@ public function execute()
                 $customerGroup->setCode(!empty($customerGroupCode) ? $customerGroupCode : null);
                 $customerGroup->setTaxClassId($taxClass);
 
+                $this->_eventManager->dispatch('adminhtml_controller_customer_group_prepare_save', [
```
kandy, line 91: Instead of add event please introduce Request to CutomerGroup mapper that you can extend with plugins mechanism

kandy (changes requested):

app/code/Magento/Customer/Controller/Adminhtml/Group/Save.php
kassner: @kandy why the customer group mapper should know about the request object? What if I save a customer group via an upgrade script or a cron job? It won't have the request object.

kandy: @kassner  Lets me return your question to you. Why do you add request as an argument for the event?

kassner:
@joni-jones creating a plugin to the repository is not the best approach either, as there are situations when the Request object will not be available (cron/setup).

The other problem is that the object passed into the repository will not be consistent. At the point where the event is dispatched, `$customerGroup` will have the updated data, like name and tax class, but the `->getExtensionAttributes()` will still return the same information loaded from the database, and  developers have no way to set change it before sending the object to the repository.

This way, if you want to do some `beforeSave` validations, you have to get the Request object, check if you are in the correct route, then set the extension attributes before doing the validation. But, if you are calling it via cron/setup, the object passed to the repository will be already consistent (because you can assemble it as you want before calling `GroupRepository::save()`). So your plugin needs to know/check if the `save` was called via adminhtml or cron/setup, instead of just being agnostic to the caller.

Also, this brings a bit more consistency with the [Customer](https://github.com/magento/magento2/blob/d81b3999fc5d84eeca16322fa6aed62a6a60e6e1/app/code/Magento/Customer/Controller/Adminhtml/Index/Save.php#L318-L321) and [Category](https://github.com/magento/magento2/blob/d81b3999fc5d84eeca16322fa6aed62a6a60e6e1/app/code/Magento/Catalog/Controller/Adminhtml/Category/Save.php#L176-L179) save actions.

YevSent:
> there are situations when the Request object will not be available (cron/setup).

Repository plugins shouldn't work with request objects.
 
> The other problem is that the object passed into the repository will not be consistent. At the point where the event is dispatched, `$customerGroup` will have the updated data, like name and tax class, but the `->getExtensionAttributes()` will still return the same information loaded from the database, and developers have no way to set change it before sending the object to the repository.

That's how extension attributes work, you load them after the original entity is loaded and save them after the entity is saved. Extension attributes should be updated when the original entity is saved, can you provide a case when it should work vice versa?

> This way, if you want to do some `beforeSave` validations, you have to get the Request object, check if you are in the correct route, then set the extension attributes before doing the validation. But, if you are calling it via cron/setup, the object passed to the repository will be already consistent (because you can assemble it as you want before calling `GroupRepository::save()`). So your plugin needs to know/check if the `save` was called via adminhtml or cron/setup, instead of just being agnostic to the caller.

I do not clearly understand the case but the validation, which depends on Request data, should be done in the controller/WEB API endpoint but not in the application layer. And plugin (like other application code) definitely should not depend on the area.

> Also, this brings a bit more consistency with the Customer and Category save actions.

Mentioned controllers are legacy implementation and it's an example of bad design than good. I wouldn't recommend to follow it.

kassner:
> Repository plugins shouldn't work with request objects.

That's my point. There is no way to define the extension attributes for the Group object without adding a plugin to the repository that depends on the Request.

> Extension attributes should be updated when the original entity is saved, can you provide a case when it should work vice versa?

No, because that's not what I am trying to do. I'm trying to set the extension attributes in the object without having the repository plugin to depend on the Request.

> I do not clearly understand the case but the validation, which depends on Request data, should be done in the controller/WEB API endpoint but not in the application layer. And plugin (like other application code) definitely should not depend on the area.

Again, this is the whole purpose of this pull request. I have no way to do this validation right now, because there is nothing I can hook my own code between when the object is populated and when it is passed to the repository. Adding this event dispatch allows me to run validations that depend on the request exactly before calling the repository.

YevSent:
> Adding this event dispatch allows me to run validations that depend on the request exactly before calling the repository.

You can add a request validation in custom observer for `controller_action_predispatch_customer_group_save` event.

kassner:
@joni-jones this won't use the same object that will be sent to the repository, thus I still need to plugin into the repository on afterSave to populate the object with information from the Request object.

YevSent:
Probably, in this case, it's better to use `controller_action_postdispatch_customer_group_save`. The entity will be already saved, you can do a custom validation and update the extension attributes this data from the request.

kassner:
@joni-jones that doesn't allow us to do things inside the same DB transaction though.

YevSent:
@kassner how the approach with event allows doing it?

kassner:
@joni-jones I can persist the extra data using `after_save` event for that model, which is called inside the transaction.

VladimirZaets:
Hi @kassner, if I right understand you want to save entity and extension attributes in the one transaction. It is not fully correct because extension attributes aren't a part of the entity and should be saved in a separate transaction

kassner:
> extension attributes aren't a part of the entity and should be saved in a separate transaction

Is this a general rule or is specific to this PR?

VladimirZaets:
@kassner it is a general rule

kassner:
Seems like a very bad advice. Maybe not for my particular case, but a general rule like that affects extensions like payment methods and other things that handle order data. Not allowing them to save in the same transaction means no atomicity and potential to have data inconsistency.

VladimirZaets:
@kassner It's the general rule that we try to support, of course sometimes we have cases where we should break up this rule, usually this flow includes a lot of architects and product managers approvals. But really I think this doesn't case where we should  break up our rules

PR #22268, opened 2019-04-10
Resolved Customer Inline Edit from admin grid issue 

ansari-ziyaurrahman:
### Description (*)

Resolved Customer Inline Edit from admin grid issue 

2. ...

When update customer gender from admin customer grid it does not save the blank value.
1. Go to Admin
2. Customers >> All Customers 
3. From customer grid update any customer gender field to blank which already have value like male/female/NA 
4. Click on save
5. Display success message like "You have successfully saved your edits" but it does not save."


app/code/Magento/Customer/Controller/Adminhtml/Index/InlineEdit.php
```diff
@@ -124,6 +124,7 @@ public function execute()
         }
 
         foreach (array_keys($postItems) as $customerId) {
+            $postItems[$customerId]['gender'] = isset($postItems[$customerId]['gender']) ? $postItems[$customerId]['gender'] : '';
```
dmytro-ch: I would recommend using `if` statement in this case for better readability.

```
if (!isset($postItems[$customerId]['gender'])) {
    $postItems[$customerId]['gender'] = null;
}
```

I also think that setting the `null` value instead of an empty string should be enough.

dmytro-ch (changes requested):
Hi @ansari-krish,
there are couple of unit test failures with the current solution:

```
1) Magento\Customer\Test\Unit\Controller\Adminhtml\Index\InlineEditTest::testExecuteWithUpdateBilling
2) Magento\Customer\Test\Unit\Controller\Adminhtml\Index\InlineEditTest::testExecuteLocalizedException
3) Magento\Customer\Test\Unit\Controller\Adminhtml\Index\InlineEditTest::testExecuteException
```
Also, the provided fix does not seem to be resolving the actual reason of the issue. This is rather a hotfix for the particular case.
It would be better to investigate the real source of the bug: why actually the empty value is missing.
There might be some issue in customer_listing IU component configuration or even deeper, e.g in inline edit component implementation etc.

Thank you!

app/code/Magento/Customer/Controller/Adminhtml/Index/InlineEdit.php
ansari-ziyaurrahman: Hi @dmytro-ch 
Thanks for reviewing this issue.
I have tried below code which you have suggested
`if (!isset($postItems[$customerId]['gender'])) {
    $postItems[$customerId]['gender'] = null;
}`
unit test failures with this condition also.
I have already debugged `app/code/Magento/Customer/view/adminhtml/ui_component/customer_listing.xml` 
this file, but not getting success.
also Check app/code/Magento/Cms/Controller/Adminhtml/Page/InlineEdit.php File 
Line 116 where I found same condition.
I have updated code, please review it,
With current condition Unit Tests are passed.
Please give your thoughts.
Thanks

ansari-ziyaurrahman: Here I have set gender as an empty string 

`$customer->setGender('')`

when I use null Functionality not working.

`$customer->setGender(null);`

dmytro-ch:
Hi @ansari-krish,
please take consider my previous comment.

```
Also, the provided fix does not seem to be resolving the actual reason of the issue. This is rather a hotfix for the particular case.
It would be better to investigate the real source of the bug: why actually the empty value is missing.
There might be some issue in customer_listing IU component configuration or even deeper, e.g in inline edit component implementation etc.
```

I briefly checked the issue, and I think this empty option is even redundant. The default value for this attribute should be "Not Specified". There is no need to allow a user to set the empty value instead.

The proper solution is to set the "Not Specified" option as default value for this attribute and render only 3 options.
However, this is not allowed due to the backward compatibility requirements.
https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#db-schema

Closing this PR for now.

Thank you!

PR #22303, opened 2019-04-12
Old Price Not visible when one child product has both price higher th…

yurii-v-git:
…an configurable product #22257

### Description (*)

These changes allow considering all prices of child products(now only the lowest product price is used)

1. magento/magento2#22257: Old Price Not visible when one child product has both price higher than configurable product

1. Create a Configurable product with attributes color swatch and text swatch.
2. Set the price of the configurable product.
3. Create simple products with price same as the configurable product.
4. Change the price of one option product to be greater than all other products.
5. Change the special price of this option product to be greater than the normal price of all other products but not greater than its normal price.
6. On the frontend, when this option is selected, both the old price and normal price are visible.

### DIFF
--- a/app/code/Magento/ConfigurableProduct/Pricing/Render/FinalPriceBox.php
+++ b/app/code/Magento/ConfigurableProduct/Pricing/Render/FinalPriceBox.php
@@ -21,7 +21,6 @@ class FinalPriceBox extends \Magento\Catalog\Pricing\Render\FinalPriceBox
 {
     /**
      * @var LowestPriceOptionsProviderInterface
-     * @deprecated
      */
     private $lowestPriceOptionsProvider;
 
@@ -57,9 +56,6 @@ class FinalPriceBox extends \Magento\Catalog\Pricing\Render\FinalPriceBox
             $salableResolver,
             $minimalPriceCalculator
         );
-        /**
-        * @deprecated. Use product type implementation instead.
-        */
         $this->lowestPriceOptionsProvider = $lowestPriceOptionsProvider ?:
             ObjectManager::getInstance()->get(LowestPriceOptionsProviderInterface::class);
     }
@@ -72,7 +68,7 @@ class FinalPriceBox extends \Magento\Catalog\Pricing\Render\FinalPriceBox
     public function hasSpecialPrice()
     {
         $product = $this->getSaleableItem();
-        foreach ($product->getTypeInstance()->getUsedProducts($product) as $subProduct) {
+        foreach ($this->lowestPriceOptionsProvider->getProducts($product) as $subProduct) {
             $regularPrice = $subProduct->getPriceInfo()->getPrice(RegularPrice::PRICE_CODE)->getValue();
             $finalPrice = $subProduct->getPriceInfo()->getPrice(FinalPrice::PRICE_CODE)->getValue();
             if ($finalPrice < $regularPrice) {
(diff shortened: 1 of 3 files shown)


app/code/Magento/ConfigurableProduct/Pricing/Render/FinalPriceBox.php
```diff
@@ -68,7 +72,7 @@ public function __construct(
     public function hasSpecialPrice()
     {
         $product = $this->getSaleableItem();
-        foreach ($this->lowestPriceOptionsProvider->getProducts($product) as $subProduct) {
+        foreach ($product->getTypeInstance()->getUsedProducts($product) as $subProduct) {
```
kandy: lowestPriceOptionsProvider was introduced for improve performance to not iterate over all product (it can be 5k for example)

kandy:
I closed related issue as not issue, so  I propose to disease expected behavior for first

app/code/Magento/ConfigurableProduct/Pricing/Render/FinalPriceBox.php
yvechirko: Thank you for your review!
 You've closed the issue but pull request still open, it means I can change the solution of this problem (for example for faster getting the result we can use right direct SQL query likes Magento\ConfigurableProduct\Pricing\Price\LowestPriceOptionsProvider::getProducts()) or I should give up this issue?

yurii-v-git:
Thank you for your review!
You've closed the issue but pull request still open, it means I can change the solution of this problem (for example for faster getting the result we can use right direct SQL query likes Magento\ConfigurableProduct\Pricing\Price\LowestPriceOptionsProvider::getProducts()) or I should give up this issue?

kandy:
Let's discuss the expected behavior first

 Product | Price |Promotion Price
------------ | -------------|------
Pr1 | 8 | |
Pr2 | 10|  |
Pr3 | 14| 9|

Current Behavior: 
Pr: 8: Old Price: n/a 

Proposed Behavior:
Pr: 8: Old Price: ?

yurii-v-git:
It is a bit of incorrect question.
 Product view page of the configurable product, when we choose one option the option price is shown.  In your example, if Pr:3 is chosen it will be: Old price - 14 and Price - 9 
Now it doesn't work only because of is not rendered HTML for displaying the old price, JS works correctly.
The easiest way to fix it - always render HTML for the old price without any checking 
(app/code/Magento/ConfigurableProduct/view/base/templates/product/price/final_price.phtml  delete checking"<?php if (!$block->isProductList() && $block->hasSpecialPrice()): ?>")

kandy:
Sorry, I don't get, do you talk about PLP or PDP?

yurii-v-git:
Product view page (PDP)

kandy:
so, the issue is that on PDP page where option with minimum price does not show old price and you try to select other option that has it, the old price does not display

and now your proposal to always render block for special price on PDP page make sense for me, just need to check is it not break other behavior

yurii-v-git:
Ok, thank you

orlangur:
@kandy something very similar (an opposite condition) was proposed in https://github.com/magento/magento2/pull/20726/files causing a performance degradation.

kandy:
@orlangur this pr remove second part of "if" and it's run only on product page that in any case calculate prices for all child item. so if it will cause performance degradation we will see it on perf builds

orlangur:
@kandy ok, I would prefer something custom like https://github.com/magento/magento2/pull/20726#issuecomment-478544045 though.

orlangur (changes requested):
@yvechirko please resolve conflict.

sidolov:
@yvechirko , I am closing this PR now due to inactivity.

PR #22310, opened 2019-04-12
Patch #22309: Remove "null" value of not passed name attribute

cmuench:
### Description

If a category is updated via REST API in the "all" scope and the
category name is not sent, Magento use "null" as value for the name,
This leads into a validation error, because the category name is
required.
In this case we remove the null value out of the automatically
generated default value array. Magento will then skip the category,
because there is no change.

### Fixed Issues 
1. magento/magento2#22309: Category Update without "name" cannot be saved in scope "all" with REST API

### Manual testing scenarios

1. Install demo data in Magento 2.3.x
2. Try to save an existing category without the name attribute via REST API in scope "all".

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/Model/CategoryRepository.php
+++ b/app/code/Magento/Catalog/Model/CategoryRepository.php
@@ -80,6 +80,10 @@ class CategoryRepository implements \Magento\Catalog\Api\CategoryRepositoryInter
         $existingData = array_diff_key($existingData, array_flip(['path', 'level', 'parent_id']));
         $existingData['store_id'] = $storeId;
 
+        if (array_key_exists('name', $existingData) && $existingData['name'] === null) {
+            unset($existingData['name']);
+        }
+
         if ($category->getId()) {
             $metadata = $this->getMetadataPool()->getMetadata(
                 CategoryInterface::class


app/code/Magento/Catalog/Model/CategoryRepository.php
```diff
@@ -80,6 +80,10 @@ public function save(\Magento\Catalog\Api\Data\CategoryInterface $category)
         $existingData = array_diff_key($existingData, array_flip(['path', 'level', 'parent_id']));
         $existingData['store_id'] = $storeId;
 
+        if (array_key_exists('name', $existingData) && $existingData['name'] === null) {
```
orlangur, line 83: Instead of unsetting, please fix a place where name is set to `null`.

Are you sure such request in issue is valid?

orlangur (changes requested):

app/code/Magento/Catalog/Model/CategoryRepository.php
cmuench: The request is valid.
See screenshot in issue. This is a real world example. One of our customers wan't to change the the "is_active" flag. It is not possible to change the global value without to specify the name. If you send the same call with a store code it is OK.
Partial updates are supported by Magento.

cmuench: @orlangur I am also not satisfied by this solution but the problem is in the framework.
The "null" value is generated by this method: `\Magento\Framework\Reflection\DataObjectProcessor::buildOutputDataArray`.
Do you have any good idea to handle that?

cmuench:
Fixed by PR #22362

PR #22324, opened 2019-04-13
Adding a validation before adding or executing layout generator class.

tiagosampaio:
Before adding or executing any layout generator it's important to validate the class before.

It fixes no issues.

### DIFF
--- a/lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
+++ b/lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
@@ -35,9 +35,9 @@ class GeneratorPool
 
     /**
      * @param ScheduledStructure\Helper $helper
-     * @param ConditionFactory $conditionFactory
-     * @param \Psr\Log\LoggerInterface $logger
-     * @param array $generators
+     * @param ConditionFactory          $conditionFactory
+     * @param \Psr\Log\LoggerInterface  $logger
+     * @param array                     $generators
      */
     public function __construct(
         ScheduledStructure\Helper $helper,
@@ -69,8 +69,9 @@ class GeneratorPool
     /**
      * Traverse through all generators and generate all scheduled elements
      *
-     * @param Reader\Context $readerContext
+     * @param Reader\Context    $readerContext
      * @param Generator\Context $generatorContext
+     *
      * @return $this
      */
     public function process(Reader\Context $readerContext, Generator\Context $generatorContext)
@@ -86,11 +87,17 @@ class GeneratorPool
      * Add generators to pool
      *
      * @param GeneratorInterface[] $generators
+     *
      * @return void
      */
     protected function addGenerators(array $generators)
     {
         foreach ($generators as $generator) {
+            if (!$generator instanceof GeneratorInterface) {
+                throw new \InvalidArgumentException(
+                    sprintf('Generator class must be an instance of %s', GeneratorInterface::class)
+                );
+            }
             $this->generators[$generator->getType()] = $generator;
         }
     }
--- a/lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
+++ b/lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
@@ -35,9 +35,9 @@ class GeneratorPool
 
     /**
      * @param ScheduledStructure\Helper $helper
-     * @param ConditionFactory          $conditionFactory
-     * @param \Psr\Log\LoggerInterface  $logger
-     * @param array                     $generators
+     * @param ConditionFactory $conditionFactory
+     * @param \Psr\Log\LoggerInterface $logger
+     * @param array|null $generators
      */
     public function __construct(
         ScheduledStructure\Helper $helper,
@@ -67,11 +67,10 @@ class GeneratorPool
     }
 
     /**
-     * Traverse through all generators and generate all scheduled elements
+     * Traverse through all generators and generate all scheduled elements.
      *
-     * @param Reader\Context    $readerContext
+     * @param Reader\Context $readerContext
      * @param Generator\Context $generatorContext
-     *
      * @return $this
      */
     public function process(Reader\Context $readerContext, Generator\Context $generatorContext)
@@ -138,9 +137,9 @@ class GeneratorPool
     }
 
     /**
-     * Reorder a child of a specified element
+     * Reorder a child of a specified element.
      *
-     * @param ScheduledStructure $scheduledStructure,
+     * @param ScheduledStructure $scheduledStructure
      * @param Data\Structure $structure
      * @param string $elementName
      * @return void
@@ -234,6 +233,8 @@ class GeneratorPool
     }
 
     /**
+     * Check visibility conditions exists in data.
+     *
      * @param array $data
      *
      * @return bool


lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -238,4 +246,17 @@ protected function visibilityConditionsExistsIn(array $data)
             array_key_exists('visibilityConditions', $data['attributes']) &&
             !empty($data['attributes']['visibilityConditions']);
     }
+
+    /**
+     * @param GeneratorInterface $generator
+     * @return bool
+     */
+    private function validateGenerator($generator)
```
dmytro-ch: I think we may simplify this method in the following way:

```
private function validateGenerator($generator)
{
    return $generator instanceof GeneratorInterface;
}
```

dmytro-ch (changes requested):

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
tiagosampaio: Hi @dmytro-ch.

I agree with your suggestion and made the modification already.

Thanks.

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-4750](https://jira.corp.magento.com/browse/ENGCOM-4750) has been created to process this Pull Request

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -77,6 +77,10 @@ public function process(Reader\Context $readerContext, Generator\Context $genera
     {
         $this->buildStructure($readerContext->getScheduledStructure(), $generatorContext->getStructure());
         foreach ($this->generators as $generator) {
+            if (!$this->validateGenerator($generator)) {
+                continue;
```
orlangur: > Fixed Issues (if relevant)
> It fixes no issues.

What kind of issue it fixes then? What happened before if corresponding interface is not implemented?

orlangur (changes requested):

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
tiagosampaio: Hi @orlangur.

If any 3rd party module overrides or include a generator in GeneratorPool and doesn't declare the methods getType() and process() the user will be given an exception. That's why I think it's a good idea to force the implementation of `GeneratorInterface`.

Examples:

`Call to undefined method Any\Class::getType()`
`Call to undefined method Magento\Framework\View\Page\Config\Generator\Head::process()`

orlangur: @tiagosampaio so user already receives an exception and by this change you actually hide it. What should really be done is throwing exception in case of incorrect interface and only in `addGenerators` method.

tiagosampaio: @orlangur so I think it's a better idea to throw a specific error to guide the user better to solve this problem. What do you think?

tiagosampaio: @orlangur you're right. I'll do it.

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -69,14 +72,17 @@ public function getGenerator($type)
     /**
      * Traverse through all generators and generate all scheduled elements
      *
-     * @param Reader\Context $readerContext
+     * @param Reader\Context    $readerContext
```
orlangur: No need to revert but please avoid such changes in general. Previous formatting was just fine.

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -69,14 +72,17 @@ public function getGenerator($type)
...
      * @param Generator\Context $generatorContext
+     *
      * @return $this
+     * @throws LocalizedException
      */
     public function process(Reader\Context $readerContext, Generator\Context $generatorContext)
     {
         $this->buildStructure($readerContext->getScheduledStructure(), $generatorContext->getStructure());
         foreach ($this->generators as $generator) {
+            $this->validateGenerator($generator);
```
orlangur: Please validate in `addGenerators` only and get rid of this method.

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -238,4 +247,19 @@ protected function visibilityConditionsExistsIn(array $data)
...
+    /**
+     * @param GeneratorInterface $generator
+     *
+     * @return bool
+     * @throws LocalizedException
+     */
+    private function validateGenerator($generator)
+    {
+        if (!($generator instanceof GeneratorInterface)) {
+            throw new LocalizedException(__('Generator classes must be instances of %1', GeneratorInterface::class));
```
orlangur: Do we really use `LocalizedException` in such cases? Please grep for some similar check, I believe it should be `InvalidArgumentException` and no need to translate.

orlangur (changes requested):

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
tiagosampaio: It was my IDE default formatting process. I missed it, I'm sorry.

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
tiagosampaio: @orlangur I believe this validation should be there too since the argument `$generators` can be injected via DI as well as using `addGenerator` method. Please let me know what you think about it.

orlangur: How? As I see https://github.com/magento/magento2/blob/9c4e1758640ef6be4458929fc61d2619a78d9fc4/lib/internal/Magento/Framework/View/Layout/GeneratorPool.php#L51 it is still added via `addGenerators` this way.

tiagosampaio: Forget what I've said. It really doesn't make sense and I just made the commit.

orlangur (changes requested):
@tiagosampaio and now please inline the method as we need this check only once 😉

orlangur (changes requested):

tiagosampaio:
@orlangur it's done.

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -86,11 +87,17 @@ public function process(Reader\Context $readerContext, Generator\Context $genera
      * Add generators to pool
      *
      * @param GeneratorInterface[] $generators
+     *
      * @return void
      */
     protected function addGenerators(array $generators)
     {
         foreach ($generators as $generator) {
+            if (!($generator instanceof GeneratorInterface)) {
```
orlangur: Please get rid of unnecessary parentheses.

orlangur (changes requested):

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
tiagosampaio: Done.

orlangur (approved):

magento-engcom-team:
[ENGCOM-4750](https://jira.corp.magento.com/browse/ENGCOM-4750) has been created to process this Pull Request

VasylShvorak:
Hi @tiagosampaio during testing PR changes I got fatal error instead of exception error.
#22324WithFix
But I can get exception error if I add message to InvalidArgumentException as string: 
`throw new \InvalidArgumentException(
                    'Generator class must be an instance of '. GeneratorInterface::class
                );`
#22324WithEditedCode
Could you take a look?

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
```diff
@@ -86,11 +87,17 @@ public function process(Reader\Context $readerContext, Generator\Context $genera
...
      * @param GeneratorInterface[] $generators
+     *
      * @return void
      */
     protected function addGenerators(array $generators)
     {
         foreach ($generators as $generator) {
+            if (!$generator instanceof GeneratorInterface) {
+                throw new \InvalidArgumentException(
+                    'Generator class must be an instance of %1', GeneratorInterface::class
```
orlangur: Yep, better use concatenation here (looks like you intended to use `sprintf`).

Commit amend + force push, please.

orlangur (changes requested):

lib/internal/Magento/Framework/View/Layout/GeneratorPool.php
tiagosampaio: @orlangur I'm just using `sprintf()` now.

orlangur (approved):

magento-engcom-team:
[ENGCOM-4750](https://jira.corp.magento.com/browse/ENGCOM-4750) has been created to process this Pull Request

VasylShvorak:
:heavy_check_mark: QA passed

PR #22325, opened 2019-04-13
Added return false incase of exception

shikhamis11:
Added return false in case of exception at 
app/code/Magento/Integration/Model/OauthService.php
in createAccessToken function

### DIFF
--- a/app/code/Magento/Integration/Model/OauthService.php
+++ b/app/code/Magento/Integration/Model/OauthService.php
@@ -150,6 +150,7 @@ class OauthService implements \Magento\Integration\Api\OauthServiceInterface
                 unset($existingToken);
             }
         } catch (\Exception $e) {
+            return false;
         }
         if (!isset($existingToken)) {
             $consumer = $this->_consumerFactory->create()->load($consumerId);


dmytro-ch:
Hi @shikhamis11,

It looks like the `return false` is not an expected behavior in this case.
The first part of the method intended to remove an already existent token.
But if the token does not exist, the `\Magento\Integration\Model\Oauth\Token\Provider::getIntegrationTokenByConsumerId` method throws an exception `Magento\Framework\Oauth\Exception` that is being caught and ignored.
In this case, the expected behavior is to generate a new token and return a positive boolean value.

Please let me know if there is some other possible case.

Thank you!

shikhamis11:
@dmytro-ch Thank you for your review. 
yes it seems `return false` is not an expected behavior here but if try catch is used, exception must be thrown. Please let me know If we can throw exception in case control goes to catch

orlangur:
@shikhamis11,

> if try catch is used, exception must be thrown

why?

shikhamis11:
@orlangur what I mean here in case of exception nothing is returned not thrown .. please suggest if it is correct behavior or not .

dmytro-ch:
@shikhamis11, we should just ignore such an exception.
Please note, the exception message is:
```
A token with consumer ID %1 does not exist
```
In general, the `\Magento\Integration\Model\Oauth\Token\Provider` is built on exceptions, as well as the repositories "get" method is throwing `NoSuchEntity` exception, for example.
So I don't think need to fix something here.

Thank you!

PR #22362, opened 2019-04-16
[Fixed] Category Update without "name" cannot be saved in scope "all" with REST API

niravkrish:
### Description (*)
- Category update API threw error for required field "name".

1. magento/magento2#22309: Category Update without "name" cannot be saved in scope "all" with REST API

1. Manual Testing

### DIFF
--- a/app/code/Magento/Catalog/Api/Data/CategoryInterface.php
+++ b/app/code/Magento/Catalog/Api/Data/CategoryInterface.php
@@ -74,7 +74,7 @@ interface CategoryInterface extends \Magento\Framework\Api\CustomAttributesDataI
     /**
      * Get category name
      *
-     * @return string
+     * @return string|null
      */
     public function getName();
 


orlangur (changes requested):
@niravkrish why it works on scope "default" without such change and does not work on "all" scope?

niravkrish:
> @niravkrish why it works on scope "default" without such change and does not work on "all" scope?

Hello @orlangur ,
In this solution the key name is not sent.
If not set in data Body required to update.
So this solution is working fine with all and default as well.

orlangur:
@niravkrish why "default" scope update works _without_ such change?

We cannot add `null` as allowed value as proper category name cannot be null.

niravkrish:
> @niravkrish why "default" scope update works _without_ such change?
> 
> We cannot add `null` as allowed value as proper category name cannot be null.

This can not allow to add or set null value to Category name.
It is used to get Category Name as a string or if not then null.

This is not pass null on save the data to database or anywhere.

If we are not passing the name from request body the name key is set with null before and after applying this solution the key is set when there is name key is available with value.

please let me know for further changes.

Thanks 
Nirav Patel

orlangur (approved):
@niravkrish ow, we allow `null` value for any other getter here.

Still don't get why the bug is in "all" scope only but fix seems reasonable.

magento-engcom-team:
[ENGCOM-4771](https://jira.corp.magento.com/browse/ENGCOM-4771) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed 

**Before:**

before

**After:**

after

PR #22432, opened 2019-04-19
Directory-CountryCollection: loadByStore should reset previous added store-filter

jwundrak:
### Description (*)

`\Magento\Directory\Model\ResourceModel\Country\Collection::loadByStore` does not reset previous added country-filters, only appends the given country-Id(s).
We recognize this behaviour in a cron job, that sync customers by using emulationMode. Only first website (Website A) works, next coming (Website B) breaks with an ["invalid countryId-Error"](https://github.com/magento/magento2/blob/0eb8677b0b4e35606032e856cc1ef7c80e68829f/app/code/Magento/Customer/Model/Address/Validator/Country.php#L75).

1. See Description. 
2. Having 2 stores with different allowed countries and each has at least one customer
3. Emulation-Mode first store, load a customer, modify something and save
4. Emulation-Mode second store, load other customer and modify something and save -> should breaks with the countyId-Error in description

 - [ X] Pull request has a meaningful description of its purpose
 - [ X] All commits are accompanied by meaningful commit messages
 - [ X] All new or changed code is covered with unit/integration tests (if applicable)
 - [ X] All automated tests passed successfully (all builds on Travis CI are green)

### DIFF
--- a/app/code/Magento/Directory/Model/ResourceModel/Country/Collection.php
+++ b/app/code/Magento/Directory/Model/ResourceModel/Country/Collection.php
@@ -181,6 +181,7 @@ class Collection extends \Magento\Framework\Model\ResourceModel\Db\Collection\Ab
             ->getAllowedCountries(ScopeInterface::SCOPE_STORE, $store);
 
         if (!empty($allowedCountries)) {
+            $this->getSelect()->reset(\Magento\Framework\DB\Select::WHERE);
             $this->addFieldToFilter("country_id", ['in' => $allowedCountries]);
         }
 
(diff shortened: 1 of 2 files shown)


ihor-sviziev (changes requested):
Hi @jwundrak,
It looks like you’re using collection in incorrect way. If you want to use collection multiple times - it’s better to create each time new collection object.

Please add code example how are you using this collection, it will be much easier to understand full picture.

Is described issue inside magento or its inside your own module?

jwundrak:
Sure, but currently with an integration test i can't reproduced the issue in 2.3. Only in 2.2 (with multiple emulation-switches) cause of this commit: https://github.com/magento/magento2/commit/c23480acc89026c890e2ab6ad809cb7a0d30f9b6
I close this pull request an will create an issue for 2.2 (or maybe a pull request).
Thanks

Issue #1392, opened 2015-06-19
Could you add TYPE_VARCHAR


SKovbel:
Could you add new type
       \Magento\Framework\DB\Ddl\Table::TYPE_VARCHAR

orlangur:
`TYPE_TEXT` is used for this:
https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/DB/Adapter/Pdo/Mysql.php#L1685
https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/DB/Adapter/Pdo/Mysql.php#L2277

Mackovey:
So is `varchar` type deprecated in Magento2?
Or what the reason why i must use `text` instead of `varchar`

nevvermind:
@Mackovey - have you clicked the links? `varchar` wil be used if the specified col length is `<= 255`.

orlangur:
> So is `varchar` type deprecated in Magento2?

No, just one `TYPE_TEXT` constant is used to represent all following types:
```
            case 'tinytext':
            case 'char':
            case 'varchar':
            case 'text':
            case 'mediumtext':
            case 'longtext':
                return Table::TYPE_TEXT;
```
and the most suitable type of the DB engine will be used automatically.

Mackovey:
>> No, just one TYPE_TEXT constant is used to represent all following types:

ok, understood, thanx

vpelipenko:
@Mackovey, can we close this issue?

Mackovey:
You can.

![3013](https://cloud.githubusercontent.com/assets/813985/8287739/61925542-191a-11e5-89e3-ec9fc6b144a1.png)

orlangur:
Ok, now I understood you request.

https://github.com/magento/magento2/blob/develop/lib/internal/Magento/Framework/DB/Adapter/Pdo/Mysql.php#L2277 seems to be from ages when `VARCHAR` could not be longer than 255 characters.

After reading [this thread](http://stackoverflow.com/a/2023513) I'm not sure whether there is a practical reason to have `VARCHAR` column longer than 255 characters:
> Both VARCHAR and BLOB/TEXT are stored inline with other columns if the value on a given row fits in the page size

Thoughts?

akaplya:
The reason to keep varchar size less than 256 symbols -- give an ability to create index for single varchar field. http://dev.mysql.com/doc/refman/5.6/en/column-count-limit.html

PR #22442, opened 2019-04-21
Add support for char element to dto factory

wardcapp:
To avoid the notice "Types char is not declared" (which breaks setup:upgrade) when using the very type anywhere in the MySQL database, mapping for this specific type is added to the DI as well.

### Description
Adds the CHAR MySQL column type to the schema type declaration factory (di.xml). Which becomes required as soon as CHAR type has been used anywhere within the connected database - applies to non-Magento tables as well, since they are validated / scanned too upon executing setup:upgrade. 

### Fixed Issues
1. Was not able to find a report within the official GitHub repo on this, but still related is following community thread:
https://community.magento.com/t5/Magento-2-x-Version-Upgrades/Upgrade-Issue-2-2-7-to-2-3-0-Types-char-is-not-declared/td-p/114939

##### db_schema.xml
```
<?xml version="1.0"?>
<!--
/**
 */
-->
<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd">
    <table name="test_table" resource="default" engine="innodb" comment="Data">
        <column xsi:type="bigint" name="id" padding="20" unsigned="true" nullable="false" identity="true" comment="ID"/>
        <column xsi:type="char" name="dataChar" length="5" nullable="false" comment="data"/>
            <column name="id"/>
        </constraint>
    </table>
</schema>
```

##### Structure of created table 
```
mysql> show create table test_table\G
*************************** 1. row ***************************
       Table: test_table
Create Table: CREATE TABLE `test_table` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT 'ID',
  `dataChar` char(5) NOT NULL COMMENT 'data',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COMMENT='Data'
1 row in set (0.00 sec)
```

### Contribution checklist


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22442) <br/>All committers have signed the CLA.

orlangur (changes requested):
I believe this is slightly related to https://github.com/magento/magento2/issues/1392. Why do we need to add `CHAR` mapping if it's not used in Magento?

wardcapp:
This PR does slightly relate to #1392, which indeed explains why most of the users aren't applying the `CHAR` type throughout their MySQL DB over Magento setup scripts. Reasons to why `CHAR` should still be added to the DTO mapping / why the PR was created:
* Although Magento core MySQL tables are currently not mentioning the `CHAR` column type, allowing others to do so (within the Magento DB, using custom setup script, Enqueue integration, other reasons to define custom tables ...) seems only reasonable.
* Not having the mapping added to the ElementFactory and using `CHAR` anywhere in the DB breaks the setup:upgrade command, entirely (none of the module upgrades will be executed from then on).
* Traces within the Magento core elsewhere are giving the impression `CHAR` was intended to be supported at first, but never completed - some entry-points:
    * https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Eav/Model/Entity/Attribute/AbstractAttribute.php#L1033
    * https://github.com/magento/magento2/blob/2.3-develop/lib/internal/Magento/Framework/Setup/Test/Unit/Declaration/Schema/Db/MySQL/Definition/Columns/StringBinaryTest.php#L137
    * https://github.com/magento/magento2/blob/2.3-develop/app/etc/di.xml#L1470

Could be wrong still / assumptions made could be incorrect, so please feel free to provide feedback.

orlangur:
Thanks @wardcapp!

@skovalenk @rganin @akaplya is there any reason to disallow `CHAR` column type?

Seb33300:
I am encountering the issue `Types char is not declared` when installing Magento on a shared database with other applications. Some of them are using the `char` field type and this result in this error...

This pull request could solve my issue.
But I don't really understand why db tables not related to Magento can break Magento install...

Hanhan1989 (approved):

app/etc/di.xml
```diff
@@ -1435,6 +1435,7 @@
                 <item name="mediumtext" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\MediumText</item>
                 <item name="text" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\Text</item>
                 <item name="varchar" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\StringBinary</item>
+                <item name="char" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\StringBinary</item>
```
sidolov, line 1447: It's not enough to add char in di.xml, you need to fix db_scrma definition too

sidolov (changes requested):

app/etc/di.xml
wardcapp: First of all, thank you for the follow-up! Not quite sure if I am understanding this correctly - did you mean the DB schema type declarations (within the very same di.xml)? If so, char has already been defined there as well within the current 2.3-develop branch (see app/etc/di.xml:1471).

Apart from that, it is possible to add support for the char type to Magento 2 setup scripts. Which would require adding of a char schema setup declaration (similar to lib/internal/Magento/Framework/Setup/Declaration/Schema/etc/schema.xsd:22) indeed, but does not relate to the intent of this pull request (and could be set up over a separate PR or feature request). But I might still be overlooking another required type of definition?

orlangur: @sidolov please respond :)

akaplya: Hi @wardcapp This is a good idea extending compatible types, but unfortunately, this is not as easy as change one XSD. Here is an example of adding a new data type to a declarative schema. https://github.com/magento/magento2/pull/25479/files
In your case, it even could be more complicated because the CHAR type supports more options that JSON. Also, you can take a look at VARCHAR implementation as an example.

engcom-Golf:
i will provide support for db_schema

lenaorobei:
@wardcapp could you please sign CLA so our team will be able to finish this PR?

wardcapp:
Hi @lenaorobei, CLA has been signed and related check should pass now.

lib/internal/Magento/Framework/Setup/Declaration/Schema/etc/types/texts/char.xsd
```diff
@@ -0,0 +1,33 @@
...
+-->
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified">
+    <xs:include schemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/types/column.xsd"/>
+
+    <xs:complexType name="char">
+        <xs:complexContent>
+            <xs:extension base="abstractColumnType">
+                <xs:annotation>
+                    <xs:documentation>
+                        Here plain text can be persisted without trailing spaces. Length of this field cant be more than 255 characters
```
orlangur: Typo: "cannot" or "can't"

orlangur (changes requested):
@Nazar65 looks good to me except the small typo 👍 

Please fix it and squash your changes into a single commit.

lib/internal/Magento/Framework/Setup/Declaration/Schema/etc/types/texts/char.xsd
engcom-Golf: ok, thank you :+1:

engcom-Golf:
@orlangur thank you for you review, done :heavy_check_mark:

slavvka:
@orlangur Could you please complete the review?

akaplya (approved):

engcom-Alfa:
:heavy_check_mark: QA Passed

engcom-Golf:
@magento run all tests

lenaorobei (approved):

magento-engcom-team:
[ENGCOM-7016](https://jira.corp.magento.com/browse/ENGCOM-7016) has been created to process this Pull Request

Issue #22494, opened 2019-04-25
Magento\Email\Model\Transport only accepts Message implementation instead of interface


bcerban:
### Summary (*)

Class `Magento\Email\Model\Transport` requires a `Magento\Framework\Mail\Message` in the constructor ([see here](https://github.com/magento/magento2/blob/2.2-develop/app/code/Magento/Email/Model/Transport.php#L73)), but the `Magento\Framework\Mail\MessageInterface` should be used instead. The `TransportBuilder` object responsible for instantiating the transport requires and passes on a `MessageInterface` object ([see here](https://github.com/magento/magento2/blob/2.2-develop/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php#L108)). 

* This breaks extensibility since the message object can't be swapped. 
* This was already fixed in the 2.3 branch. 

### Examples (*)

To test this, define a custom `MessageInterface` implementation, and inject it via `di.xml` with 
```xml
<preference for="Magento\Framework\Mail\MessageInterface" type="MyCompany\MyModule\Model\Message"/>
```

When trying to send and email, you will get the following error:
```
Fatal error: Uncaught TypeError: Argument 1 passed to Magento\Email\Model\Transport::__construct() must be an instance of Magento\Framework\Mail\Message, instance of MyCompany\MyModule\Model\Message\Interceptor given
```

### Proposed solution

Change message dependency in constructor for `Magento\Email\Model\Transport`, from `Magento\Framework\Mail\Message` to `Magento\Framework\Mail\MessageInterface`.

engcom-Bravo:
This issue is fixed on 2.3.x branches(i tried to reproduce on 2.3-develop).

PR #22495, opened 2019-04-25
change dependency from message to messageinterface

bcerban:
### Description (*)

Replaces dependency in `Magento\Email\Model\Transport` from concrete message class to its interface. 

1. magento/magento2#22494: Magento\Email\Model\Transport only accepts Message implementation instead of interface

1. Define a custom `MessageInterface` implementation, and inject it via `di.xml` with 
```xml
<preference for="Magento\Framework\Mail\MessageInterface" type="MyCompany\MyModule\Model\Message"/>
```
2. Send an email (eg. for a new order).

3. When trying to send and email, you will get the following error:
```
Fatal error: Uncaught TypeError: Argument 1 passed to Magento\Email\Model\Transport::__construct() must be an instance of Magento\Framework\Mail\Message, instance of MyCompany\MyModule\Model\Message\Interceptor given
```

### DIFF
--- a/app/code/Magento/Email/Model/Transport.php
+++ b/app/code/Magento/Email/Model/Transport.php
@@ -7,7 +7,7 @@ namespace Magento\Email\Model;
 
 use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Exception\MailException;
-use Magento\Framework\Mail\Message;
+use Magento\Framework\Mail\MessageInterface;
 use Magento\Framework\Mail\TransportInterface;
 use Magento\Framework\Phrase;
 use Magento\Store\Model\ScopeInterface;
@@ -54,7 +54,7 @@ class Transport implements TransportInterface
     private $zendTransport;
 
     /**
-     * @var Message
+     * @var MessageInterface
      */
     private $message;
 
@@ -64,13 +64,13 @@ class Transport implements TransportInterface
     private $zendMessage;
 
     /**
-     * @param Message $message Email message object
+     * @param MessageInterface $message Email message object
      * @param ScopeConfigInterface $scopeConfig Core store config
      * @param ZendMessage $zendMessage
      * @param null|string|array|\Traversable $parameters Config options for sendmail parameters
      */
     public function __construct(
-        Message $message,
+        MessageInterface $message,
         ScopeConfigInterface $scopeConfig,
         ZendMessage $zendMessage,
         $parameters = null


orlangur:
@sivaschenko @sidolov do we allow such kind of changes? https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/ should become more specific then

sidolov (changes requested):
@bcerban , @orlangur unfortunately, such changes are backward incompatible. Please, consider injecting new dependency

bcerban:
> @bcerban , @orlangur unfortunately, such changes are backward incompatible. Please, consider injecting new dependency

Hi @sidolov, can you explain to me how this change is backward incompatible? Given that the default implementation for the `MessageInterface` is a `Message` object, any override of the object will continue to work after this change. 

I will not add another parameter, that will not only make things confusing, it just won't work since the internal implementation won't know which dependency to use. It also forces me to rewrite the `TransportBuilder` responsible for instantiating the `Transport` object. So this is not a solution.

orlangur:
@bcerban,

> Given that the default implementation for the MessageInterface is a Message object, any override of the object will continue to work after this change.

Will there be any notice/warning for class extending old constructor signature?

bcerban:
> Will there be any notice/warning for class extending old constructor signature?

No, there won't. 
* A child class' constructor signature does not need to match the parent class' constructor. 
* If the child class of `Transport` defines a new constructor and calls the parent constructor within it, everything is ok as long as the message object passed is an instance of `MessageInterface` which will always be the case for any class extending `Message` and `Message` itself.

bcerban:
Hey @sidolov @orlangur, any news on this?

sivaschenko:
@melnikovi can you please take a look if you can approve this change?

melnikovi:
`Magento\Framework\Mail\Message` extends `Zend_Mail` and has more methods compare to `\Magento\Framework\Mail\MessageInterface`. I would suggest not to do this change in patch release.  @bcerban can you workaround this issue in your extension?

bcerban:
> `Magento\Framework\Mail\Message` extends `Zend_Mail` and has more methods compare to `\Magento\Framework\Mail\MessageInterface`. I would suggest not to do this change in patch release. @bcerban can you workaround this issue in your extension?

@melnikovi I don't have many ways to work around this tbh. I tried changing the dependency, which doesn't work because of the reasons I explain above. I don't see an easy way to write a plugin, since the code is pretty closed in this section. 

You are right that `Magento\Framework\Mail\Message` has more methods than `\Magento\Framework\Mail\MessageInterface`, but
* none of them are used in the email sending process
* the transport builder **requires and passes on** a `MessageInterface`.

So, this is a moot point. We are arguing over the difference between an abstraction and its implementation, when in fact the only one that matters is the one passed on by the builder. 

For some context, I'm just trying to add attachments to order confirmation emails. Personally, I believe this functionality should be provided out of the box, since it is pretty basic. As it is not, we did this is our extension. The problem is that the code here keeps changing, and to make it worse, contracts are not adhered to. The argument that we need to keep this for compatibility seems wrong-headed to me, for this very reason.

bcerban:
@melnikovi @sidolov @orlangur  In fact, can you please take a look at the code that's been released? For example, compare [2.2.7](https://github.com/magento/magento2/blob/2.2.7/app/code/Magento/Email/Model/Transport.php#L60-L63) vs. [2.2.8](https://github.com/magento/magento2/blob/2.2.8/app/code/Magento/Email/Model/Transport.php#L72-L76). These changes don't make much sense, considering the arguments you posed to me.

VladimirZaets:
Hi @bcerban. The branch `2.2-develop` will be closed for contribution from **July 15, 2019**. Please finish all your Pull Requests till this date, otherwise, they will be closed.
Thanks for collaboration.

orlangur:
@melnikovi @sidolov please provide example illustrating how this change breaks backward compatibility.

According to https://github.com/magento/magento2/pull/22495#issuecomment-488098169 by author it shouldn't break anything.

VladimirZaets:
Hi @bcerban, 
The next 2.2 release will be the last of functional releases.
Due to it, we are closing 2.2-develop for contribution and all not ready for delivery Pull Requests. 
Thanks for collaboration.

melnikovi:
I'm checking if we can implement solution described here https://github.com/magento/architecture/pull/144.

orlangur:
@melnikovi ok, thanks. But still, was this change backward compatible or not?

sivaschenko:
@orlangur I believe it's not backward compatible taking into account third party code that may introduce a new implementation of `MessageInterface`. See https://github.com/magento/magento2/pull/22495#issuecomment-492415816

orlangur:
@sivaschenko you mean such implementation may unintentionally started to use instead of `Magento\Framework\Mail\Message`?

sivaschenko:
@orlangur I mean that presence of interface means that implementation of the interface can be safely replaced, which cannot be guaranteed in this case

Issue #22500, opened 2019-04-26
Cron Observer currentTime calculation change to make it cleanly extendable


gody01:
### Summary (*)
We have user story requirement to run Cron jobs in local time instead 'UTC' for synchronization issues with external processes.

As we tried to extend Observer we found that it is cumbersome, as private and protected functions are used. Function execute and saveSchedule has to be changed for customization which can brake functionalitiy changes in the future.

However, we found simple and clean solution which require minimal change in core code (described below)

### Examples (*)
Cron jobs are run at UTC time. However in DayLight saving time environement some customers needs to run cron jobs on local time or even on some other timezone.

### Proposed solution
We found out that if we introduce public method getCurrentTime and replace:
$currentTime = $this->dateTime->gmtTimestamp();
with 
$currentTime = $this->getCurrentTime();

Method getCurrentTime being just:

    public function getCurrentTime ()
    {
        return $this->dateTime->gmtTimestamp();
    }

Observer becomes cleanly extendable without any architectural change or risk of break on upgrades.

Extension for extending Observer to run jobs at localtime is then very simple and clean (will put example on github)

We also prepared pull request wich will be attached to this issue.

engcom-Charlie:
Hello @gody01 
I am closing this issue now due to inactivity.
Please reopen and update if you wish to continue.
Thank you for the collaboration!

PR #22501, opened 2019-04-26
For Issue #22500: Change currentTime calculation to public method to make Observer clea…

gody01:
…nly extendable.

 On branch 2.3-develop
 Your branch is up to date with 'origin/2.3-develop'.

 Changes to be committed:
	modified:   ProcessCronQueueObserver.php

Make Cron/Observer/ProcessCronQueueObserver.php cleanly extendable for changing time, when job runs to some other (e.g. local timezone) time.

### Description (*)
Bay introducing public method for currentTime calculation, one can develop clean extension to run jobs at specific timezone instead 'UTC'. 
And in the process there is also more clean OOP way to specify local property ...

1. magento/magento2#22500: Cron Observer currentTime calculation change to make it cleanly extendable

On issue #22500 will put github link to example extension using  this pull request changes.

### DIFF
--- a/app/code/Magento/Cron/Observer/ProcessCronQueueObserver.php
+++ b/app/code/Magento/Cron/Observer/ProcessCronQueueObserver.php
@@ -192,6 +192,15 @@ class ProcessCronQueueObserver implements ObserverInterface
     }
 
     /**
+     * return currentTime
+     * @return current timestamp
+    **/
+    public function getCurrentTime()
+    {
+        return $this->dateTime->gmtTimestamp() ;
+    }
+
+    /**
      * Process cron queue
      * Generate tasks schedule
      * Cleanup tasks schedule
@@ -205,7 +214,7 @@ class ProcessCronQueueObserver implements ObserverInterface
     public function execute(\Magento\Framework\Event\Observer $observer)
     {
 
-        $currentTime = $this->dateTime->gmtTimestamp();
+        $currentTime = $this->getCurrentTime();
         $jobGroupsRoot = $this->_config->getJobs();
         // sort jobs groups to start from used in separated process
         uksort(
@@ -546,7 +555,7 @@ class ProcessCronQueueObserver implements ObserverInterface
      */
     protected function saveSchedule($jobCode, $cronExpression, $timeInterval, $exists)
     {
-        $currentTime = $this->dateTime->gmtTimestamp();
+        $currentTime = $this->getCurrentTime();
         $timeAhead = $currentTime + $timeInterval;
         for ($time = $currentTime; $time < $timeAhead; $time += self::SECONDS_IN_MINUTE) {
             $scheduledAt = strftime('%Y-%m-%d %H:%M:00', $time);


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22501) <br/>All committers have signed the CLA.

gody01:
Example extension module with this pull request applied:

https://github.com/gody01/Agenda_CronAtLocalTime

orlangur (changes requested):
@gody01,

1) You can already customize this behavior by injecting different `$this->dateTime`
2) Generally it is not a good idea to create extension points in observers/plugins

gody01:
Hellom
> @gody01,
> 
>     1. You can already customize this behavior by injecting different `$this->dateTime`
I tried this but not succeeded. Not very into Magento2 code yet.
Can You point me to direction where can I find example for this. With this small change change is very small and elegant and it don't interfere with code or architecture in any way :)

>     2. Generally it is not a good idea to create extension points in observers/plugins
I know, but this one seems quite clean and also by general OOP practice it is good, to set variables via methods of the class instead of calling "outside" methods directly.

When I was searching for solution for our customer, who hes real need for this, i found quite a lot of posts asking the same question, so it is real need to be able to extend this Observer behavior.
I don't say, that any code for actual change of behavior should be in the core, just possibility to extend cleanly.

orlangur:
@gody01 it should probably be refactored somehow to move business logic into a separate model. Customizing observer / adding some public method is not a good idea as it can be even removed in future version and it is not a BC break.

Please check https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/di-xml-file.html#constructor-arguments and tell me back if you have some obstacles with such approach.

gody01:
> @gody01,
> 
>     1. You can already customize this behavior by injecting different `$this->dateTime`
I just checked this solution and it is not OK.

dateTime is used all over the Class ProcessCronQueueObserver.php and with injecting different $this->DateTime different time will be used everywhere. Even in writing schedule times in the database.

But we don't want that. It is OK that in database is used UTC time.
We just want, that Cron runs jobs at the localtime for example.

So it is dependent only on $currentTime which are calculated inside methods instead via getCurrentTime method.

So with $this->DateTime injection one will change architecture of Magento's Cron, which we don't want.

To be more specific. Fore example one global shop may have command database and many local webstores all over the world.

In current implementation all cronjobs are run at UTC. But than means, that for some part of the world that means, that some big slow integration job may be run in the middle of the working rush hours.

If we made possible that $currentTime is calculated via method one can write small extension/module to track and run cron jobs at local timezones of the app servers.

orlangur:
@gody01 you current implementation is missing the `$schedule->setExecutedAt(strftime('%Y-%m-%d %H:%M:%S', $this->dateTime->gmtTimestamp()))->save();` part.

I don't see how changing `$this->dateTime->gmtTimestamp()` all over the class consistently is not a good idea.

Anyway, such public method (and even protected one) like in this PR should not be introduced as customization via inheritance is not encouraged.

gody01:
> @gody01 it should probably be refactored somehow to move business logic into a separate model. Customizing observer / adding some public method is not a good idea as it can be even removed in future version and it is not a BC break.
> 
That is what I'm trying to achieve. It may be protected method, but not private, since private can't be overriden in extended object in PHP.

The main point is, that it should be possible to change currentTime in execute and in save..  via PHP class extension mechanizem, to make possible to move bussines logic out of core observer.
In issue I linked also example extension which shows, how easy and clean it is to do, if currentTime is fetched via method, as it should be by OOP priciniple anyway.

> Please check https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/di-xml-file.html#constructor-arguments and tell me back if you have some obstacles with such approach.
I will. Thanks for the info.

gody01:
> @gody01 you current implementation is missing the `$schedule->setExecutedAt(strftime('%Y-%m-%d %H:%M:%S', $this->dateTime->gmtTimestamp()))->save();` part.
> 
Exactly. That is the point. I don't want to change db UTC definition of cron job task.
I wan't to change interpertation, of when this task is runned in local enviroenement.

> I don't see how changing `$this->dateTime->gmtTimestamp()` all over the class consistently is not a good idea.
> 
Magento acctually has this implemented for admin and frontend. Exactly the same principle.
Dates are stored in DB as UTC, which is good and only sensible.
But displayed according to local timezone.

I would like to achieve, that one can make same for Cron jobs (not mandatory but via module).

You can see example in config for creating sitemap.xml.

In backend You select, that You want to have sitemap generated for example at:
12:15 In DB it is writen in UTC: 2019-04-26 14:15:00

Now:
sitemap.xml will be generated at 14:15:00

With this PR and exaple module from my github sitemap.xml will be generated at 12:15 as any admin would expected.

> Anyway, such public method (and even protected one) like in this PR should not be introduced as customization via inheritance is not encouraged.

Well there is difference between encouraged and prevented. At least protected methods and setting local variables via methods should be allowed :)

gody01:
> I don't see how changing `$this->dateTime->gmtTimestamp()` all over the class consistently is not a good idea.
Let me try to exmplain:

Customer has backend ERP systems in Slovenia and in USA.
During the day in ERP system they change prices etc.
Than at the midnight they want Magento store, to access ERP system and fetch new prices and other data.

So one creates two tasks:
fetch_si_prices
fetch_usa_prices

They both have to run at midnight. But midnight are different in Slovenia and in USA.
Slovenia has one set of servers with Timezone CET
USA has another set of servers with Timezone EST
But they share common database.

So if You want consistency (which You need, as ERP systems works in different time zones, and You don want that for example USA server fetches prices at 9 hours difference thinking, that it is in UTC time zone) you have to run jobs at current LocalTime.

And when You add Daylight savings time to that it is even more obvious, what I want to achieve with this simple infrastructure change, which will enable me to do so.

We had real issue with week of investigation after this Year Daylight change, when prices were not properly updated, as ERP system operated at CET timezone and magento cron not.

And it is not realistically to expect, that crontab.xml will be changed manually on every Daylight saving time.

orlangur:
Thanks for details @gody01! I'll try to understand what kind of change is sufficient to cover all described scenarios.

gody01:
> Thanks for details @gody01! I'll try to understand what kind of change is sufficient to cover all described scenarios.

Thanks,
I tried many options before and also asked my Magento programmers to come with some solution. 

However this approach with 2 line change and one line method is by far the cleanest, small and nonintrusive I saw. Even if it breaks political rules, that basic OOP inheritance principle is not good for Magento customization.

gody01:
> Please check https://devdocs.magento.com/guides/v2.3/extension-dev-guide/build/di-xml-file.html#constructor-arguments and tell me back if you have some obstacles with such approach.

I just consulted with my programmers and it seems that we don't nderstand above with suggestion from first reply:

You can already customize this behavior by injecting different $this->dateTime

How to inject different $this->dateTime. Does this means, that we have to extend/create new clas dateTime with different gmtTime method and inject this class somehow to Cron Observer ?

olgaIskra:
A very simple question is what the best and simple change will allow the Cron Jobs to run by the local time.

sidolov:
@gody01 , @olgaIskra observers and plugins are not a part of the API and can be removed/moved/changed at any time. We don't recommend to implement your logic on observer behavior. If you need to change observer it's possible to replace `ProcessCronQueueObserver` with your own by configuration or inject custom implementation to the observer via DI, as @orlangur mentioned above.

sidolov:
@gody01 , I am closing this PR now due to inactivity.

PR #22505, opened 2019-04-26
In edge cases product url rewrites got generated starting with a forw…

hostep:
…ard slash. This edge case gets fixed here.

### Description (*)
Not sure if this will get accepted but bear with me here.

I found an instance of a Magento2 shop where certain product url rewrites got generated which started with a forward slash.
This was caused because a category the product was in, somehow return an empty string from its rewritten url path ([probably because its parent was a root category](https://github.com/magento/magento2/blob/d3fdecf0d0f89d6c38ddb8109e5542d1a7b483eb/app/code/Magento/CatalogUrlRewrite/Model/CategoryUrlPathGenerator.php#L73-L75)).
It is expected that url rewrites don't start with a forward slash.

I don't know how to reproduce this problem on a vanilla Magento installation though. This is just an extra check which gets added to avoid this issue. In our case it is most likely caused by some 3rd party plugin to regenerate url rewrites.

The unit test added demonstrates what is expected, without this fix the unit test would result in: `/product-path` instead of `product-path`.

None I could find.

1. Manage to get the CategoryUrlPathGenerator->getUrlPath to return an empty string somehow.
2. Generate a url rewrite for one or multiple products in such a category somehow.
3. Expected url rewrite is `product-url-key.html`, but actual result was `/product-url-key.html`

### DIFF
--- a/app/code/Magento/CatalogUrlRewrite/Model/ProductUrlPathGenerator.php
+++ b/app/code/Magento/CatalogUrlRewrite/Model/ProductUrlPathGenerator.php
@@ -70,9 +70,15 @@ class ProductUrlPathGenerator
                 ? $this->prepareProductUrlKey($product)
                 : $this->prepareProductDefaultUrlKey($product);
         }
-        return $category === null
-            ? $path
-            : $this->categoryUrlPathGenerator->getUrlPath($category) . '/' . $path;
+
+        if ($category !== null) {
+            $categoryUrlPath = $this->categoryUrlPathGenerator->getUrlPath($category);
+            if ($categoryUrlPath !== '') {
+                return $categoryUrlPath . '/' . $path;
+            }
+        }
+
+        return $path;
     }
 
     /**
--- a/app/code/Magento/CatalogUrlRewrite/Test/Unit/Model/ProductUrlPathGeneratorTest.php
+++ b/app/code/Magento/CatalogUrlRewrite/Test/Unit/Model/ProductUrlPathGeneratorTest.php
@@ -173,6 +173,22 @@ class ProductUrlPathGeneratorTest extends \PHPUnit\Framework\TestCase
     /**
      * @return void
      */
+    public function testGetUrlPathWithCategoryHavingAnEmptyUrlPath(): void
+    {
+        $this->product->expects($this->once())->method('getData')->with('url_path')
+            ->will($this->returnValue('product-path'));
+        $this->categoryUrlPathGenerator->expects($this->once())->method('getUrlPath')
+            ->will($this->returnValue(''));
+
+        $this->assertEquals(
+            'product-path',
+            $this->productUrlPathGenerator->getUrlPath($this->product, $this->category)
+        );
+    }
+
+    /**
+     * @return void
+     */
     public function testGetUrlPathWithSuffix(): void
     {
         $storeId = 1;


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22505) <br/>All committers have signed the CLA.

orlangur (changes requested):
@hostep in your case, will it work if you just remove all leading slashes one time?

Why https://github.com/magento/magento2/blob/d3fdecf0d0f89d6c38ddb8109e5542d1a7b483eb/app/code/Magento/CatalogUrlRewrite/Model/CategoryUrlPathGenerator.php#L73-L75 does not lead to `''` returned in vanilla installation?

hostep:
In theory the bug should get triggered when a product is directly assigned to a root category, like the standard 'Default Category', but for some reason Magento doesn't trigger the bug either by saving the product or the category.
So it looks like this can only be triggered by code using the `ProductUrlPathGenerator` class.
I'll see if I can produce a script later on which can trigger this exact behaviour.

hostep:
Ok, after a couple of attempt of playing around in the backend of Magento, I was able to reproduce on a vanilla installation without having to write a php script. But you have to follow very specific steps, if you do something in a different order, it doesn't always trigger.

So:
1. Setup a vanilla installation of 2.3-develop (I used commit d3fdecf0d0f)
2. Create a new product, I called it 'Product', assign it to the 'Default Category' category. And make sure it is visible on the frontend.
3. Create a new category which has the 'Default Category' as a parent. I called it 'Sub'.
4. Using the category interface, select the 'Default Category', and remove the product from it.
5. Still using the category interface, select the 'Sub' category, and assign the product to it (you have to reset the filter before it shows up)

After these steps are done, you'll find the following url rewrites in the database:

| entity_type | entity_id | request_path     | target_path                          | metadata            |
| ----------- | --------- | ---------------- | ------------------------------------ | ------------------- |
| category    | 3         | sub.html         | catalog/category/view/id/3           | NULL                |
| product     | 1         | product.html     | catalog/product/view/id/1            | NULL                |
| product     | 1         | sub/product.html | catalog/product/view/id/1/category/3 | {"category_id":"3"} |
| product     | 1         | /product.html    | catalog/product/view/id/1/category/2 | {"category_id":"2"} |

The last one is incorrect here. Category with ID 2 is the 'Default Category' btw.

If you follow these more logical steps, you'll get the expected result:
1. Setup a vanilla installation of 2.3-develop (I used commit d3fdecf0d0f)
2. Create a new category which has the 'Default Category' as a parent. I called it 'Sub'.
3. Create a new product, I called it 'Product', assign it to the 'Sub' category. And make sure it is visible on the frontend.

Then these url rewrites are being generated:

| entity_type | entity_id | request_path     | target_path                          | metadata            |
| ----------- | --------- | ---------------- | ------------------------------------ | ------------------- |
| category    | 3         | sub.html         | catalog/category/view/id/3           | NULL                |
| product     | 1         | product.html     | catalog/product/view/id/1            | NULL                |
| product     | 1         | sub/product.html | catalog/product/view/id/1/category/3 | {"category_id":"3"} |



When attempting the first steps again, but with this fix in place, the expected request_paths are generated, but I can see that the target_path & metadata is different... Not sure how important this is though.

| entity_type | entity_id | request_path     | target_path                          | metadata            |
| ----------- | --------- | ---------------- | ------------------------------------ | ------------------- |
| category    | 3         | sub.html         | catalog/category/view/id/3           | NULL                |
| product     | 1         | product.html     | catalog/product/view/id/1/category/2            | {"category_id":"2"}                |
| product     | 1         | sub/product.html | catalog/product/view/id/1/category/3 | {"category_id":"3"} |

hostep:
Apparently using this fix and following those 5 steps causes the frontend to use the url `catalog/product/view/id/1/s/product/category/3/` instead of `product.html` for some strange reason, so this fix is causing new problems...


Pffff, I'm going to give it up for today, Magento really needs to throw away all this code and start from scratch, there are just way too many edge case problems happening all the time with these url rewrites. Why was that indexer even removed like we had in Magento 1, at least that one was much more stable than what we've been suffering from in the last 3,5 years on Magento 2 ... :(

hostep:
Hmm, according to the failing integration tests, it is expected that url rewrites can start with a `/`.
This was introduced in: https://github.com/magento/magento2/commit/450bec052d7c411b1d3d44d18a4fbc2cdb3d6398

I find it a bit strange that some actions cause a url rewrites to be generated starting with a `/` and others not (see scenario's above). So what is actually expected? In my opinion it should not be expected that the url rewrite `request_path` can start with a `/`, but maybe there is a good reason for this that I'm unaware of?

That's what I'm talking about that we no longer have the url rewrites indexer, there are so many different entry points which can generate url rewrites having all kinds of different contexts, and some of those cause slightly different results then others, having an indexer ensures a single entry point which creates it own context (what specific store view, which product, which category, ...) and would result in a much more stable system in that you can easily debug and fix issues. Now we have this giant mess which nobody seems to fully understand and where it becomes really hard to fix issues in.

Is there some info around those url rewrites starting with a `/` in the internal jira ticket MAGETWO-67087 by any chance?

hostep:
@orlangur: how can we get some information regarding if it is expected that the `request_path` of url rewrites can start with a `/` or not? Thanks!

orlangur:
@sidolov please add some details from Jira.

hostep:
@sidolov I am closing this PR now due to inactivity.
(sorry for the sarcasm 😉)

Anyway, closing since there is probably a good reason for these url rewrites to exist which start with `/`. I just don't know what the reason is.

PR #22514, opened 2019-04-26
Add actual meaning error to string

DominicWatts:
Adding meaningful error to string - I've seen a few different methods of generating this error online and I think it will help everyone if you see the actual error - not what's currently displayed

### Description (*)

Add meaning eror to end of string

Lots of reports online regarding this message

In my case I provided duplicate category IDs in an array to assignProductToCategories method as part of linking products to categories process.  Once I revealed the actual *unique key integrity* error it was much easier to understand what was happening.

```
public function __construct(
    \Magento\Catalog\Api\CategoryLinkManagementInterface $categoryLinkManagementInterface
) {
    $this->categoryLinkManagementInterface = $categoryLinkManagementInterface;
}

/**
 * Assign product to category
 * @param array $categoryIds
 * @return void
 */
public function assignProductToCategory($product, $categoryIds = [])
{
    if(!empty($categoryIds)) {
        $this->categoryLinkManagementInterface->assignProductToCategories(
            $product->getSku(),
            $categoryIds
        );
    }
```


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22514) <br/>All committers have signed the CLA.

orlangur (changes requested):
Low-level exception messages must never be exposed to end-user. Add logging of such exception instead.

DominicWatts:
Made suggested change

magento-engcom-team:
@DominicWatts thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

p-bystritsky:
@sivaschenko requested changes done.

app/code/Magento/Catalog/Model/CategoryLinkRepository.php
```diff
@@ -21,20 +26,28 @@ class CategoryLinkRepository implements \Magento\Catalog\Api\CategoryLinkReposit
...
     /**
      * @param \Magento\Catalog\Api\CategoryRepositoryInterface $categoryRepository
      * @param \Magento\Catalog\Api\ProductRepositoryInterface $productRepository
+     * @param LoggerInterface|null $logger
      */
     public function __construct(
         \Magento\Catalog\Api\CategoryRepositoryInterface $categoryRepository,
-        \Magento\Catalog\Api\ProductRepositoryInterface $productRepository
+        \Magento\Catalog\Api\ProductRepositoryInterface $productRepository,
+        ?LoggerInterface $logger
```
orlangur: Must not be nullable and should be `null` by default.

app/code/Magento/Catalog/Model/CategoryLinkRepository.php
```diff
@@ -87,6 +101,7 @@ public function deleteByIds($categoryId, $sku)
         try {
             $category->save();
         } catch (\Exception $e) {
+            $this->logger->critical($e->getMessage());
```
orlangur, line 104: Don't we have a separate method to log exception?

orlangur (changes requested):

app/code/Magento/Catalog/Model/CategoryLinkRepository.php
p-bystritsky: @orlangur is using nullable types prohibited?

app/code/Magento/Catalog/Model/CategoryLinkRepository.php
p-bystritsky: @orlangur seems like we don't.

app/code/Magento/Catalog/Model/CategoryLinkRepository.php
orlangur: @p-bystritsky this is just not needed in constructors as there is no sense to pass `null` explicitly. You either use default `null` value or pass a correct object.

app/code/Magento/Catalog/Model/CategoryLinkRepository.php
orlangur: Ok, will try to grep later.

sivaschenko:
Hi @akaplya what do you think about adding an optional constructor parameter to `CategoryLinkRepository` ?

sidolov:
@DominicWatts , I am closing this PR now due to inactivity.

Issue #22312, opened 2019-04-12
magento2 Customer unable to get token in customer/token rest api


uthegenius:
**Preconditions**
Magento 2.3 

**Steps to reproduce**
Customer is signup through rest api in magento 2.3 successfully. after signup, I am unable to get customer's token through customer/token rest api. Api returns following error:

 **Actual result**
![issue1](https://user-images.githubusercontent.com/6263757/56051857-617c5c00-5d69-11e9-8f8b-bd9c2278bb1f.PNG)

Although customer is successfully login through web but unable to get validate in rest api.

**Expected result**
token must be generated

magento-engcom-team:
:white_check_mark: Confirmed by @shikhamis11 
 Thank you for verifying the issue. Based on the provided information internal tickets `MAGETWO-99167`, `MAGETWO-99168` were created

**Issue Available**: @shikhamis11, _You will be automatically unassigned. Contributors/Maintainers can claim this issue to continue. To reclaim and continue work, reassign the ticket to yourself._

Shubham0492:
@magento-engcom-team give me 2.3-develop instance - upcoming 2.3.x release

magento-engcom-team:
Hi @Shubham-Webkul. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @Shubham-Webkul, here is your [Magento instance](https://i-22312-2-3-develop.instances.magento-community.engineering).
Admin access: [https://i-22312-2-3-develop.instances.magento-community.engineering/admin](https://i-22312-2-3-develop.instances.magento-community.engineering/admin)
Login: `admin` Password: `123123q`
Instance will be terminated in up to 3 hours.

Shubham0492:
Reproduced on 2.3.x
chrome-extension://fdpohaocaechififmbbbbbknoalclacl/capture.html?id=9&url=https%3A%2F%2Fi-22312-2-3-develop.instances.magento-community.engineering%2Fswagger%23%2FintegrationCustomerTokenServiceV1%2FintegrationCustomerTokenServiceV1CreateCustomerAccessTokenPost

ronak2ram:
Hie @uthegenius,
I think there is no issue with the API. Can you please remove `all` store code form URL.
so URL would be:  /rest/V1/integration/customer/token.
Please refer this for more information about store code `all` in URL: https://devdocs.magento.com/guides/v2.3/get-started/gs-web-api-request.html#endpoints

[...]

PR #22530, opened 2019-04-28
fixed issue #22312

Shubham0492:
### Description (*)
When we fetch customer token then it gives an error that this account sign-in was incorrect or your account is disabled temporarily due to there is a check of website id and we were not sending website id in API param. Now I have added website id as param now you have to fetch customer token accordingly website id. 

1. #22312 : magento2 Customer unable to get token in customer/token rest api

1. send website id also in the parameter.


app/code/Magento/Integration/Model/CustomerTokenService.php
```diff
@@ -84,12 +84,12 @@ public function __construct(
     /**
      * @inheritdoc
      */
-    public function createCustomerAccessToken($username, $password)
+    public function createCustomerAccessToken($username, $password, $websiteId)
```
dmytro-ch: The `$websiteId` parameter must be optional for backward compatibility.

dmytro-ch (changes requested):
Hi @Shubham-Webkul,

There are many unit and integration tests errors. Most of them are caused by backward compatibility issues. Could you please check and resolve them?

app/code/Magento/Integration/Model/CustomerTokenService.php
sidolov: We are not able to change public methods signature according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

Shubham0492: @dmytro-ch thanks for explaination i will update it according to Backward Compatible Development Guide.

Shubham0492: @dmytro-ch  & @sidolov could you help me to fix Semantic Version Checker build regarding issue. Where will i have to update for fixing this.

sidolov: Hi @Shubham-Webkul , Semantic Version Checker indicates the problem with the provided solution. Please, check my previous comment

Shubham0492: @sidolov I have made parameter optional still got an issue. Will i have to create another method. Can be update public method parameters?

PR #22553, opened 2019-04-30
Trim all import data

amenk:
### Description

Nobody wants whitespaces in import files. So we rim it.

no reported

* import data with whitespaces
* data is trimmed

### DIFF
--- a/app/code/Magento/ImportExport/Model/Import/AbstractSource.php
+++ b/app/code/Magento/ImportExport/Model/Import/AbstractSource.php
@@ -92,7 +92,10 @@ abstract class AbstractSource implements \SeekableIterator
             } else {
                 throw new \InvalidArgumentException(AbstractEntity::ERROR_CODE_COLUMNS_NUMBER);
             }
-        }
+	}
+
+	$row = array_map('trim', $row);
+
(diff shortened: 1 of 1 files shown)


amenk:
By the way: I tried to accomplish that via a Plugin, but [it is not possible](https://magento.stackexchange.com/questions/272593/i-cant-seem-to-write-an-plugin-for-magento-importexport-model-import-abstractso/272594#272594).

Then I thought, that this should better happen in the, as most data is anyways trimmed at various places.

orlangur:
@amenk this is not needed. Invalid data should be fixed instead (for example, by a custom script trimming all cells beforehand).

amenk:
@orlangur but why?

Spaces do cause problems and there are tons of trim calls in the import. But not everywhere.. so I thought it is good to clean it at the source.

Next, I cannot make a plug-in for the source class. So the other approach would be to plug-in-enable the import source classes. Shall we work towards such a solution instead? I thought it would be Overkill...

orlangur:
@amenk I saw your linked report, I believe instead of hiding a problem by trims here and there it would be nice to add appropriate validations.

PR #22560, opened 2019-04-30
Fix undefined methods 'addClass' and `removeClass` on a PrototypeJS Element

markvds:
### Description
This problem was caused by 5a283e3ff9 and reported in #20843. The mentioned commit adds two lines in `js.phtml` with non existing methods: `panel.addClass(activePanelClass);` and `panel.removeClass(activePanelClass);`. Both methods do not exist in PrototypeJS and should have been `addClassName()` and `removeClassName()` respectively.

The incorrect method names cause a Javascript error 'Uncaught TypeError: panel.addClass is not a function' in the console and prevents the Edit Attribute form from saving.

Enabling the Swatches module, as mentioned in the issue, only solves the issue because it replaces the phtml file with its own file.

1. magento/magento2#20843: Uncaught TypeError: panel.addClass is not a function when Swatches are disabled

### Manual testing scenarios
1. Disable Swatches modules (otherwise, the mentioned phtml file will not be used)
2. In admin, go to Stores > Attributes > Product
3. Click any attribute
4. Try to save the attribute
5. Nothing happens without the code change, but the attribute saves fine with the code change

### Contribution checklist

### DIFF
--- a/app/code/Magento/Catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
+++ b/app/code/Magento/Catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
@@ -45,11 +45,11 @@ function checkOptionsPanelVisibility(){
 
         if($('frontend_input') && ($('frontend_input').value=='select' || $('frontend_input').value=='multiselect')){
             panel.show();
-            panel.addClass(activePanelClass);
+            jQuery(panel).addClass(activePanelClass);
         }
         else {
             panel.hide();
-            panel.removeClass(activePanelClass);
+            jQuery(panel).removeClass(activePanelClass);
         }
     }
 }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22560) <br/>All committers have signed the CLA.

orlangur (changes requested):
@markvds please use `jQuery` equivalents instead.

markvds:
@orlangur I cannot imagine that it makes the situation any better; it's all surrounded by PrototypeJS code. I think it would even make readibility worse.

orlangur:
@markvds we don't need to add any new PrototypeJS-related code. Feel free to refactor any additional occurrences in file with jQuery.

barryvdh:
@markvds Thank you for your PR.

@orlangur I would keep the bugfix PR as minimal as possible to make it clear what the intent and actual fix of this PR is. Also to make it easier to avoid merge conflicts etc.Refactoring the prototype should happen in a seperate PR imho.

orlangur:
@barryvdh it will be minimal as I suggested since `jQuery` is already injected.

@markvds please apply necessary changes.

jissereitsma:
My 50 cents: The proposed fix of @markvds is a bugfix. However, the suggestion of @orlangur would lead to a code enhancement. They are different things. I'm not sure if postponing the bugfix is a good idea, waiting for the code enhancement. On the other hand, I'm not sure how much work the refactoring the jQuery code would be. Just my 50 cents.

app/code/Magento/Catalog/view/adminhtml/templates/catalog/product/attribute/js.phtml
```diff
@@ -45,11 +45,11 @@ function checkOptionsPanelVisibility(){
 
         if($('frontend_input') && ($('frontend_input').value=='select' || $('frontend_input').value=='multiselect')){
             panel.show();
-            panel.addClass(activePanelClass);
+            panel.addClassName(activePanelClass);
```
orlangur:
```suggestion
            jQuery(panel).addClass(activePanelClass);
```

orlangur (changes requested):
Initial code was jQuery-style. There is no need to step back. Please replace all `addClassName/removeClassName`.

orlangur (changes requested):

jissereitsma:
@orlangur You are awesome

markvds:
I changed the PR. 

@orlangur I finally get your point. You basically say the method name was correct, but the context was wrong. I assumed the context was right but the method name was wrong. It's just a different view on the situation. Understanding your thoughts helps me in future PRs. Thanks for explaining!

orlangur:
@markvds yep, sorry for confusion, will try to formulate better next time 🤝

orlangur (approved):

magento-engcom-team:
[ENGCOM-5020](https://jira.corp.magento.com/browse/ENGCOM-5020) has been created to process this Pull Request

magento-engcom-team:
@markvds thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

soleksii:
:heavy_check_mark: QA Passed 

**Before:**

before

**After:**
after

PR #22651, opened 2019-05-04
Fixed #22407 Top menu cache issue - 2.3.1

### LINKED ISSUE #22407: Top menu cache issue - 2.3.1
### Preconditions (*)

1.Magento 2.3.1 with all cache types enabled

### Steps to reproduce (*)

1. Create 2-3 categories, set include in menu = Yes
2. On frontend click on any category page in top menu
3. Flush magento cache
4. Open the same category page
5. Open any other category page

### Expected result (*)

Just opened category is marked as active in top menu

### Actual result (*)

Both categories in top menu are marked as active
test

gaurav-473:
Fixed #22407 Top menu cache issue - 2.3.1

### Preconditions (*)

1.Magento 2.3.1 with all cache types enabled

### Steps to reproduce (*)

1. Create 2-3 categories, set include in menu = Yes
2. On frontend click on any category page in top menu
3. Flush magento cache
4. Open the same category page
5. Open any other category page

### Expected result (*)

Just opened category is marked as active in top menu

### Actual result (*)

Both categories in top menu are marked as active
test

### DIFF
--- a/app/code/Magento/Theme/Block/Html/Topmenu.php
+++ b/app/code/Magento/Theme/Block/Html/Topmenu.php
@@ -394,4 +394,17 @@ class Topmenu extends Template implements IdentityInterface
         }
         return $this->_menu;
     }
+    
+    /**
+     * Get cache key informative items
+     *
+     * @return array
+     * @since 100.1.0
+     */
+    public function getCacheKeyInfo()
+    {
+        $keyInfo = parent::getCacheKeyInfo();
+        $keyInfo[] = $this->getUrl('*/*/*', ['_current' => true, '_query' => '']);
+        return $keyInfo;
+    }
 }
--- a/app/code/Magento/Theme/Block/Html/Topmenu.php
+++ b/app/code/Magento/Theme/Block/Html/Topmenu.php
@@ -399,7 +399,6 @@ class Topmenu extends Template implements IdentityInterface
      * Get cache key informative items
      *
      * @return array
-     * @since 100.1.0
      */
     public function getCacheKeyInfo()
     {


maaarghk:
your fix here would mean the block would be re-generated for _each_ non-category page whereas a single cached block would be sufficient for _all_ non-category pages.

app/code/Magento/Theme/Block/Html/Topmenu.php
```diff
@@ -394,4 +394,17 @@ public function getMenu()
...
+    /**
+     * Get cache key informative items
+     *
+     * @return array
+     * @since 100.1.0
+     */
+    public function getCacheKeyInfo()
+    {
+        $keyInfo = parent::getCacheKeyInfo();
+        $keyInfo[] = $this->getUrl('*/*/*', ['_current' => true, '_query' => '']);
```
sivaschenko, line 406: What do you think about using the current category in the cache key instead of the current URL?

app/code/Magento/Theme/Block/Html/Topmenu.php
```diff
@@ -394,4 +394,17 @@ public function getMenu()
         }
         return $this->_menu;
     }
+    
+    /**
+     * Get cache key informative items
+     *
+     * @return array
+     * @since 100.1.0
```
sivaschenko: This is not an introduction of the method, but just an override, so `@since` annotation is not necessary

sivaschenko (changes requested):
Thanks for the pull request @gauravagarwal1001 Please see my review comments

maaarghk:
https://github.com/magento/magento2/commit/61d81a8e07b627eb7f3850529767e1427461f52b#diff-454601ea9afd086413ac90923ae9ccaa

this is where the bug was introduced, looks like the test could be reintroduced (and then changed) as well.

app/code/Magento/Theme/Block/Html/Topmenu.php
hostep: @sivaschenko: do you mean the entire category object, or just some attribute of it? Which one then? We can't use the name, because the name isn't unique. Maybe the path and the store id?

maaarghk: primary key

sivaschenko:
@gauravagarwal1001 please let me know if any help is needed related to this pull request

ihor-sviziev:
Hi @gauravagarwal1001,
Why did you closed this issue? Was it fixed?

PR #22695, opened 2019-05-04
fixed vault payment method not selected after page refresh  issues on…

ghost:
… checkout

### Description (*)
If a payment method provides card save feature, and uses magento vault then this issue will come regardless of any payment method because magento remembers the payment method code of the respective payment method and then of page reload it checks that code is available in available payment methods, unfortunately vault can have multiple saved cards so it adds an integer token at the end of the payment code, so it is not found in available payment methods, there was no general solution for this it had to be fixed in app/code/Magento/Vault/view/frontend/web/js/view/payment/method-renderer/vault.js of I updated it. 

### Fixed Issues 
1. magento/magento2#22539: Saved Payment methods (Magento Vault) are not remembered on checkout

1. enable any online payment method that support vault(for example braintree)
2. now go to frontend and login as customer
3: add product to the cart
4: goto checkout and add a braintree card 
5: select save card and place order.
6: now again add any product in the cart 
7: go to checkout and select saved card
8: now refresh the page and you will see that saved card will be not selected it must be selected.

### DIFF
--- a/app/code/Magento/Ui/etc/di.xml
+++ b/app/code/Magento/Ui/etc/di.xml
@@ -7,7 +7,6 @@
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
     <preference for="Magento\Framework\View\Element\UiComponent\ContainerInterface" type="Magento\Ui\Component\Wrapper\UiComponent" />
-    <preference for="Magento\Framework\View\Element\UiComponent\ContainerInterface" type="Magento\Ui\Component\Wrapper\UiComponent" />
     <preference for="Magento\Framework\View\Element\UiComponent\Control\ActionPoolInterface" type="Magento\Ui\Component\Control\ActionPool" />
     <preference for="Magento\Framework\Data\Argument\InterpreterInterface" type="Magento\Framework\Data\Argument\Interpreter\Composite" />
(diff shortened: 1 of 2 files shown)


app/code/Magento/Ui/etc/di.xml
```diff
@@ -6,6 +6,7 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
+    <preference for="Magento\Framework\View\Element\UiComponent\ContainerInterface" type="Magento\Ui\Component\Wrapper\UiComponent" />
```
itmsenior: Duplicate line

ghost: don't know how it got there but I removed the duplicate line in latest commit thanks

itmsenior:
Hi @ashutoshwebkul, the fix you have provided only addresses the issue of the payment method being lost on refresh. If a customer chooses to change their delivery address and uses the top checkout navigation to do so, there is no hard refresh and thus upon returning to payment methods, the vault method is no longer selected.

Default Payment method are remembered regardless of if the customer refreshes, uses the AJAX checkout navigation or the back button.

sidolov:
Hi @ashutoshwebkul , please, review comment above

sidolov:
@ashutoshwebkul , I am closing this PR now due to inactivity.

PR #22726, opened 2019-05-04
show only current children categories

ankitsrivastavacedcoss:
Added check of parent_id, so that it will return the current categories of the called category id.
 
### Description (*)
 Currently, CategoryTree::getTree(..) returns all the enabled categories, so if the parent is disabled then it will also return the children of those parent categories, 
so I added a parent_id check so that it won't check and return the child categories of the disabled categories. 

 It was causing an issue in the PWA. It was returning the disabled categories index but values are null.

image

1. Disable any parent category.
2. In Graphql run the  PWA category list query.
<pre>
<code>
# TODO: get only active categories from graphql when it is ready
query categoryList($id: Int!) {
    category(id: $id) {
        id
        children {
            id
            name
            include_in_menu
            url_key
            url_path
            children_count
            path
            image
            productImagePreview: products(pageSize: 1) {
                items {
                    small_image {
                        url
                    }
                }
            }
        }
    }
}
</code>
</pre>
It will return the disabled category in index but the values are null(similar to pic).
 
### DIFF
--- a/app/code/Magento/CatalogGraphQl/Model/Resolver/Products/DataProvider/CategoryTree.php
+++ b/app/code/Magento/CatalogGraphQl/Model/Resolver/Products/DataProvider/CategoryTree.php
@@ -102,6 +102,7 @@ class CategoryTree
         $collection->addFieldToFilter('level', ['gt' => $level]);
         $collection->addFieldToFilter('level', ['lteq' => $level + $depth - self::DEPTH_OFFSET]);
         $collection->addAttributeToFilter('is_active', 1, "left");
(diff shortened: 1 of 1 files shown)


dmytro-ch:
Hi @ankitsrivastavacedcoss,

Please note, changes for GraphQL modules should be provided in the scope of [magento/graphql-ce](https://github.com/magento/graphql-ce/) repository.
Could you please create the corresponding issue and PR there?

The current pull request will be closed.

CC: @naydav

sidolov:
Hi @ankitsrivastavacedcoss , please, move the current pull request to the magento/graphql-ce project

PR #22733, opened 2019-05-04
Adds module:config:status command which checks if the module config i…

hostep:
…n the app/etc/config.php file is correct or not.

### Description (*)
This adds a new command to `bin/magento`: `module:config:status`.
This command checks if the modules in the file `app/etc/config.php` are correct or not.

I've created this command since I'm getting a bit tired of seeing incorrect `app/etc/config.php` files being added to our git repo's whenever somebody adds a new module using composer and forgets to run `bin/magento setup:upgrade`, or when some branch get merged which has merge conflicts in the `config.php` file and the conflict gets solved incorrectly.
Having this new command, would allow us to have a pre-commit githook which can check for this problem and then tell the developer to fix the `config.php` file before committing.

Feel free to let me know if the command name or description needs to be changed. And also if the output messages need to be changed.

None that I could find.

1. Have a correct Magento installation
2. Run `bin/magento module:config:status`, it should say: `The modules configuration is up to date.`
3. Edit the file `app/etc/config.php` and remove or add a module. Or just switch the sort order of 2 modules.
4. Run `bin/magento module:config:status`, it should say: `The modules configuration in the 'app/etc/config.php' file is outdated. Run 'setup:upgrade' to fix it.`
5. Run `bin/magento setup:upgrade`
6. Run `bin/magento module:config:status`, it should say: `The modules configuration is up to date.`

### DIFF
--- a/setup/src/Magento/Setup/Console/Command/ModuleConfigStatusCommand.php
+++ b/setup/src/Magento/Setup/Console/Command/ModuleConfigStatusCommand.php
@@ -1,0 +1,101 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+declare(strict_types=1);
+
+namespace Magento\Setup\Console\Command;
+
+use Magento\Framework\App\DeploymentConfig\Reader as ConfigReader;
+use Magento\Framework\Config\ConfigOptionsListConstants;
+use Magento\Framework\Config\File\ConfigFilePool;
+use Magento\Framework\Console\Cli;
+use Magento\Framework\Setup\ConsoleLogger;
+use Magento\Setup\Model\InstallerFactory;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Command to check if the modules config in app/etc/config.php matches with how Magento interprets it
+ */
+class ModuleConfigStatusCommand extends Command
+{
+    /**
+     * Deployment config reader
+     *
+     * @var ConfigReader
+     */
+    private $configReader;
+
+    /**
+     * Installer service factory.
+     *
+     * @var InstallerFactory
+     */
+    private $installerFactory;
+
+    /**
+     * @param ConfigReader     $configReader
+     * @param InstallerFactory $installerFactory
+     */
+    public function __construct(
+        ConfigReader     $configReader,
+        InstallerFactory $installerFactory
+    ) {
+        $this->configReader     = $configReader;
+        $this->installerFactory = $installerFactory;
+
+        parent::__construct();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function configure()
+    {
+        $this->setName('module:config:status')
+             ->setDescription('Checks the modules configuration in the \'app/etc/config.php\' file and reports if they are up to date or not');
+
+        parent::configure();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        try {
+            // the config as currently in app/etc/config.php
+            $currentConfig = $this->configReader->load(ConfigFilePool::APP_CONFIG);
+            if (!array_key_exists(ConfigOptionsListConstants::KEY_MODULES, $currentConfig)) {
+                throw new \Exception('Can\'t find the modules configuration in the \'app/etc/config.php\' file.');
+            }
+
+            $currentModuleConfig = $currentConfig[ConfigOptionsListConstants::KEY_MODULES];
+
+            $installer = $this->installerFactory->create(new ConsoleLogger($output));
+
+            // the module config as Magento calculated it
+            $correctModuleConfig = $installer->getModulesConfig();
+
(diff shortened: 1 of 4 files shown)


setup/src/Magento/Setup/Console/Command/ModuleConfigStatusCommand.php
```diff
@@ -0,0 +1,97 @@
...
+            $correctModuleConfig = $installer->getModulesConfig();
+
+            if ($currentModuleConfig !== $correctModuleConfig) {
+                throw new \Exception(
+                    'The current module config in the \'app/etc/config.php\' file is incorrect. Run \'setup:upgrade\' to fix it.'
+                );
+            }
+
+            $output->writeln(
+                '<info>Everything seems to be in order.</info>'
```
dmytro-ch: Could you please change the success message to something more certain? :)
e.g. `All modules configurations are up to date.`

setup/src/Magento/Setup/Console/Command/ModuleConfigStatusCommand.php
```diff
@@ -0,0 +1,97 @@
...
+            );
+        } catch (\Exception $e) {
+            $output->writeln('<error>' . $e->getMessage() . '</error>');
+
+            return \Magento\Framework\Console\Cli::RETURN_FAILURE;
```
dmytro-ch: Please move the `Magento\Framework\Console\Cli` class name to the imports as well.
```
use Magento\Framework\Console\Cli;
```
```
return Cli::RETURN_FAILURE;
```

setup/src/Magento/Setup/Model/Installer.php
```diff
@@ -1196,6 +1196,17 @@ public function updateModulesSequence($keepGeneratedFiles = false)
         $this->createModulesConfig([]);
     }
 
+    /**
+     * Get the modules config as Magento sees it
+     *
+     * @return array
+     * @throws \LogicException
+     */
+    public function getModulesConfig()
```
dmytro-ch, line 1209: Could you please avoid the creation of new public methods.
I would recommend adding this implementation as a private method of `Magento\Setup\Console\Command\ ModuleConfigStatusCommand` class.

dmytro-ch (changes requested):

setup/src/Magento/Setup/Console/Command/ModuleConfigStatusCommand.php
hostep: Thanks for the review, I'll try to address your comments later today!

setup/src/Magento/Setup/Model/Installer.php
hostep: The problem is that the `createModulesConfig` function is private, so I need to expose it somehow without copying it over completely.
According to the [semver checker result](https://results.solutionascode.com/results/magento/magento2/pull/22733/ea76f902-95f0-491c-82a7-5922f485cf57/10891/SemanticVersionChecker/report-magento2.html) it is ok to add a public function to this class.
What do you think?

dmytro-ch: > The problem is that the createModulesConfig function is private

You are right, I didn't notice this point.
So let's keep this method.
Thank you.

hostep:
@dmytro-ch: I've force pushed the requested changes.
I changed some other output messages and the description of the command as well, so they are more consistent with each other.

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-5080](https://jira.corp.magento.com/browse/ENGCOM-5080) has been created to process this Pull Request

dmytro-ch:
@hostep thank you!

VasylShvorak:
:heavy_check_mark: QA passed

hostep:
Thanks for the cleanup @nmalevanec! I'm curious which command you ran to detect these problems, so the next time I can try to fix these myself when creating a new PR.

nmalevanec:
Hi, @hostep. You can just run dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php test against modified php files.

sivaschenko:
@piotrekkaminski can you please take a look at this feature and provide your feedback

sivaschenko:
@magento run all tests

hostep:
@sivaschenko: is it ok if I still update my PR? There are some missing `declare(strict_types=1);` lines, it seems like I forgot to add these. And should I add these to new classes only, or also to other classes touched by this PR?

I'll then also do some git rebasing of this PR on top of the latest 2.3-develop, maybe that helps with these tests which seem to be not running somehow.

I also still don't understand how to execute static tests on my local, the `dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php` file just contains a class and no executable code, so I'm a bit confused about the statement above.

Thanks!

hostep:
Rebased PR on latest 2.3-develop branch and added `declare(strict_types=1);` to the 2 new classes.

sivaschenko:
@magento run all tests

hostep:
Force pushed again, which will hopefully satisfy the static tests (I've amended @nmalevanec's commit for this, as not to loose his authorship of the other fixes).

It would still be great if somebody can tell me how to execute these static tests on my local environment, so I don't have to wait 2 days for the results and can just fix them in one go, thanks! 🙂

PR #22815, opened 2019-05-10
MFTF: Update price column in checkout

magicsunday:
### Description (*)
In order to MFTF test the calculated shipping method price, its required to access the desired table price column. This requires the column to contain a identifier.

1. See price column in checkout shipping methods table
2. Should contain now a id-attribute

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/template/shipping-address/shipping-method-item.html
+++ b/app/code/Magento/Checkout/view/frontend/web/template/shipping-address/shipping-method-item.html
@@ -16,7 +16,7 @@
                     'checked': element.rates().length == 1 || element.isSelected" />
     </td>
     <td class="col col-price"
-        attr="'id': 'label_price_' + method.method_code + '_' + method.carrier_code"
+        attr="'data-table-column': 'label_price_' + method.method_code + '_' + method.carrier_code"
     >
         <each args="element.getRegion('price')" render="" />
     </td>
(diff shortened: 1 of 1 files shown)


orlangur (changes requested):
Functional tests must not rely on HTML ID, see [Code Demarcation Standard](https://devdocs.magento.com/guides/v2.3/coding-standards/code-standard-demarcation.html). Please add appropriate data attributes instead.

magicsunday:
I changed it, please take a look at it again.

app/code/Magento/Checkout/view/frontend/web/template/shipping-address/shipping-method-item.html
```diff
@@ -15,14 +15,17 @@
...
+    >
         <each args="element.getRegion('price')" render="" />
     </td>
     <td class="col col-method"
         attr="'id': 'label_method_' + method.method_code + '_' + method.carrier_code"
         text="method.method_title" />
     <td class="col col-carrier"
-        attr="'id': 'label_carrier_' + method.method_code + '_' + method.carrier_code"
+        attr="'id': 'label_carrier_' + method.method_code + '_' + method.carrier_code,
+             'data-table-column': 'label_carrier_' + method.method_code + '_' + method.carrier_code"
```
orlangur, line 28: Why do we need to duplicate it in both TDs?

`<tr data-container-for="shipping-method"` and using existing `ko-value="method.carrier_code + '_' + method.method_code"` should be sufficient.

orlangur (changes requested):

PR #22826, opened 2019-05-10
compiled interceptors module

### LINKED ISSUE #30311: [Issue] compiled interceptors module
This issue is automatically created based on existing pull request: magento/magento2#22826: compiled interceptors module

---------
### Description (*)

At Creatuity we have developed a module that generates Interceptors code using app config instead of a boilerplate that reads config at runtime.

https://github.com/creatuity/magento2-interceptors

We believe it changes performance and debugging magento significantly and it is worth to consider making it a part of the core witch this pull request does.

Any comments welcome.

to use in developer mode in `app/etc/di.xml` replace:
```
<item name="interceptor" xsi:type="string">\Magento\Framework\Interception\Code\Generator\Interceptor</item>
```
with:
```
<item name="interceptor" xsi:type="string">\Magento\Framework\CompiledInterception\Generator\CompiledInterceptor</item>
```
clear generated files and cache:

`rm -rf generated/* && rm -rf var/cache/* && bin/magento cache:clean`

generated interceptors instead of boilerplate that reads plugins config at runtime like this:

```
public function methodX($arg) {
    $pluginInfo = $this->pluginList->getNext($this->subjectType, 'methodX');
    if (!$pluginInfo) {
        return parent::methodX($arg);
    } else {
        return $this->___callPlugins('methodX', func_get_args(), $pluginInfo);
    }
}
```

should have a code built using config like:

```
public function methodX($arg) {
    switch(getCurrentScope()){
        case 'frontend':
            $this->_get_example_plugin()->beforeMethodX($this, $arg);
            $this->_get_another_plugin()->beforeMethodX($this, $arg);
            $result = $this->_get_around_plugin()->aroundMethodX($this, function($arg){
                return parent::methodX($arg);
            });
            return $this->_get_after_plugin()->afterMethodX($this, $result);
        case 'adminhtml':
            // ...
        default:
            return parent::methodX($arg);
    }
}
```

fsw:
### Description (*)

At Creatuity we have developed a module that generates Interceptors code using app config instead of a boilerplate that reads config at runtime.

https://github.com/creatuity/magento2-interceptors

We believe it changes performance and debugging magento significantly and it is worth to consider making it a part of the core witch this pull request does.

Any comments welcome.

to use in developer mode in `app/etc/di.xml` replace:
```
<item name="interceptor" xsi:type="string">\Magento\Framework\Interception\Code\Generator\Interceptor</item>
```
with:
```
<item name="interceptor" xsi:type="string">\Magento\Framework\CompiledInterception\Generator\CompiledInterceptor</item>
```
clear generated files and cache:

`rm -rf generated/* && rm -rf var/cache/* && bin/magento cache:clean`

generated interceptors instead of boilerplate that reads plugins config at runtime like this:

```
public function methodX($arg) {
    $pluginInfo = $this->pluginList->getNext($this->subjectType, 'methodX');
    if (!$pluginInfo) {
        return parent::methodX($arg);
    } else {
        return $this->___callPlugins('methodX', func_get_args(), $pluginInfo);
    }
}
```

should have a code built using config like:

```
public function methodX($arg) {
    switch(getCurrentScope()){
        case 'frontend':
            $this->_get_example_plugin()->beforeMethodX($this, $arg);
            $this->_get_another_plugin()->beforeMethodX($this, $arg);
            $result = $this->_get_around_plugin()->aroundMethodX($this, function($arg){
                return parent::methodX($arg);
            });
            return $this->_get_after_plugin()->afterMethodX($this, $result);
        case 'adminhtml':
            // ...
        default:
            return parent::methodX($arg);
    }
}
```

### Resolved issues:
1. [x] resolves magento/magento2#30311: compiled interceptors module


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22826) <br/>All committers have signed the CLA.

orlangur:
Hi @fsw,

> We believe it changes performance and debugging magento significantly

Do you have some benchmarks to share with us?

IvanChepurnyi:
@fsw This is a very nice module, do but did you consider compiling interceptors per area and specifying those on di for Magento to use, instead of using static calls to "getCurrentScope"?

fsw:
@orlangur, on a standard instance, simple ApacheBench tests are showing about ~10% gain on a request, I will create some more detailed benchmark for you.

@IvanChepurnyi, actually this was the original idea, but:
* we wanted this to be 100% backward compatible and there were issues with class per scope approach: scopes can be switched at runtime when interception should be working, compiled mode classmap would have to be hacked to use different class depending on scope etc. So it was simply easier to implement this way.
* in a lot of cases plugins are exactly the same for all scopes so they are being grouped in a single switch case so we end up having 6 times less classes and code in opcache in exchange for this one static switch.

However I and not 100% sure used approach is better, do you think there are other upsides other than removing the static switch?

IvanChepurnyi:
@fsw switch opcache argument sounds reasonable, but maybe you somehow could receive this current scope without doing this static call, like maybe pass it in the constructor, so DI container will automatically supply it for you?

fsw:
@IvanChepurnyi this should be easily doable as code generator uses reflection anyway so should be easy to override parent constructor. I will check how this would affect performance.

fsw:
@orlangur, trivial benchmark in developer mode:
```
bin/magento cache:clean > /dev/null
rm -rf generated/* && rm -rf var/cache/*
echo "warmup"
curl -so /dev/null -w '%{time_total}\n' $url
echo "cached"
ab -n 100 -c 4 $url | grep "Time per request:"
```

| request    |    | Interceptor (ms)          | CompiledInterceptor (ms) | gain  |
| ------------- |:-------------:| -----:| -----:| -----:|
| home page | warmup request | 15555 | 31342 | -101.49%
  | time per request | 176.439 | 166.831 | 5.45%
  | time per request (4 concurrent requests) | 44.11 | 41.708 | 5.45%
listing | warmup request | 14212 | 30557 | -115.01%
  | time per request | 1813.871 | 1661.836 | 8.38%
  | time per request (4 concurrent requests) | 453.468 | 415.459 | 8.38%
product page | warmup request | 15720 | 32143 | -104.47%
  | time per request | 177.388 | 172.602 | 2.70%
  | time per request (4 concurrent requests) | 44.347 | 43.15 | 2.70%
admin login page | warmup request | 16730 | 31671 | -89.31%
  | time per request | 439.592 | 422.958 | 3.78%
  | time per request (4 concurrent requests) | 109.898 | 105.74 | 3.78%

You can see that first request with generated code purged renders a lot slower as interceptors now read config to generate, but in exchange following requests are faster. This obviously depends on number of plugins and should be better in production mode where DI takes far less time.

fsw:
@IvanChepurnyi I have followed your suggestion and it seems to be a bit faster and code is much nicer indeed. 
examples of generated interceptors:

https://github.com/fsw/magento2/tree/2.3-develop/lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors

IvanChepurnyi:
@fsw Looks nice! But I just noticed, seems like you also use object manager directly for plugin instantiation, so maybe it makes sense to supply it as well as a dependency?

fsw:
@IvanChepurnyi already tried that :) but:

1) this would not be 100% backward compatible as currently plugins are "lazy loaded" as well and we can not be sure what happens in constructors. Here, we are doing stuff exactly in same order.
2) when I tried that (passing plugins in constructor) I got a lot of "circular dependency" issues as plugins were depending on objects that were intercepted and depended on other plugins etc.

IvanChepurnyi:
@fsw I mean passing object manager for instantiation of plugins later. Sure for performance reasons pre-instantiation of plugins does not make sense.

fsw:
@IvanChepurnyi oh yes, that makes sense. done.

Also I have 2 questions for general public:

1) Is slower generation time on first request a big problem? If so we can think about modifying class loader somehow so it would check file mtimes to cover most cases so developer would not have to purge var/generated/* each time plugin is added.
 
2) Is adding this as a separate generator with an option for a fallback a good approach? Because if this is 99.99% compatible (except edge cases like someone having plugins on config loaders) other approach would be to replace default interceptor generator with this?

IvanChepurnyi:
@kandy good point. 
@fsw how about creating a file watcher command line tool that will automatically clear generated code if plugin for that class is added in di.xml? I think it will greatly increase DevExperience, but I think it can be done a separate tool outside of this pull request, as it requires a dev to install an inotify extension, but it is worth it. Also, it is possible to use output from this tool instead of notifying, as it is much faster: https://docs.rs/crate/unison-fsmonitor/0.1.0

kandy:
I think it makes sense to include your code directly to Interceptor library. 
Also, we can implement a proxy that will select one of implementation depending on the mode. It may answer your question.


Also if this code will work in production mode we can make additional optimization like replace
 ```
$this->____om->get(\Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ItemPlugin\Advanced::class);
``` 
with
```
new \Magento\Framework\ClassReplacedByPrreferance(
 $this->____om->get(\Dep1),
 $this->____om->get(\De2),
 $this->____om->get(\Dep3),
)
```

**Comment by fsw** [comment]
@kandy I think plugin instances are shared between interceptors, plus I think there is a bit more logic in DI when it creates constructor arguments (constants/virtual types etc) so I am not sure if removing this single step of DI would be worth it.

**Comment by kandy** [comment]
@fsw, It's hard to predict the performance impact now, and in any case, I don't recommend doing it in the scope of this PR

**Comment by fsw** [comment]
Changes suggested by @IvanChepurnyi about adding ObjectManager and Scope to class constructor unfortunately are a bit trickier to implement in compiled mode. It seems DI is assuming interceptor constructor args are same as base class. I temporarily disabled this and will work on a solution.

File: lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php [hunk 1]
```diff
@@ -0,0 +1,399 @@
...
+            $resultClassName ,
+            $ioObject,
+            $classGenerator,
+            $definedClasses);
+
+        if ($plugins !== null) {
+            $this->plugins = $plugins;
+        } else {
+            $this->plugins = [];
+            foreach (['primary', 'frontend', 'adminhtml', 'crontab', 'webapi_rest', 'webapi_soap'] as $scope) {
```
maghamed: \Magento\Framework\App\AreaList::getCodes should be used
as now we lost GraphQL area

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+                $this->plugins[$scope] = new CompiledPluginList(ObjectManager::getInstance(), $scope);
```
maghamed: no business logic in constructors allowed, use only assignments

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+        //NOOP
+    }
+
+    protected function _generateCode()
+    {
+        $typeName = $this->getSourceClassName();
+        $reflection = new \ReflectionClass($typeName);
+
+        if ($reflection->isInterface()) {
+            return false;
```
maghamed, line 143: how does current approach support plugins for Interfaces? 

The intital return type of _generateCode() method is string

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+namespace Magento\Framework\CompiledInterception\Generator;
+
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\ObjectManagerInterface;
+use Magento\Framework\Code\Generator\EntityAbstract;
+use Magento\Framework\Config\ScopeInterface;
+use Magento\Framework\Interception\Code\Generator\Interceptor;
+use Magento\Framework\Interception\DefinitionInterface;
+
+class CompiledInterceptor extends Interceptor
```
maghamed: why do we need to extend from `\Magento\Framework\Interception\Code\Generator\Interceptor`
but not from `\Magento\Framework\Code\Generator\EntityAbstract` as all other code generation entities do if we are not using the logic from Magento\Framework\Interception\Code\Generator\Interceptor class? Is the reason just to substitute default implementation via DI preference on class ? 
Like :
`<preference for="Magento\Setup\Module\Di\Code\Generator\Interceptor" type="Creatuity\Interception\Generator\CompiledInterceptor" />`

We can easily provide another implementation and inject new `item name="interceptor" `
```
    <type name="Magento\Framework\Code\Generator">
        <arguments>
            <argument name="generatedEntities" xsi:type="array">
                <item name="extensionInterfaceFactory" xsi:type="string">\Magento\Framework\Api\Code\Generator\ExtensionAttributesInterfaceFactoryGenerator</item>
                <item name="factory" xsi:type="string">\Magento\Framework\ObjectManager\Code\Generator\Factory</item>
                <item name="proxy" xsi:type="string">\Magento\Framework\ObjectManager\Code\Generator\Proxy</item>
                <item name="interceptor" xsi:type="string">\Magento\Framework\Interception\Code\Generator\Interceptor</item>
                <item name="logger" xsi:type="string">\Magento\Framework\ObjectManager\Profiler\Code\Generator\Logger</item>
                <item name="mapper" xsi:type="string">\Magento\Framework\Api\Code\Generator\Mapper</item>
                <item name="persistor" xsi:type="string">\Magento\Framework\ObjectManager\Code\Generator\Persistor</item>
                <item name="repository" xsi:type="string">\Magento\Framework\ObjectManager\Code\Generator\Repository</item>
                <item name="convertor" xsi:type="string">\Magento\Framework\ObjectManager\Code\Generator\Converter</item>
                <item name="searchResults" xsi:type="string">\Magento\Framework\Api\Code\Generator\SearchResults</item>
                <item name="extensionInterface" xsi:type="string">\Magento\Framework\Api\Code\Generator\ExtensionAttributesInterfaceGenerator</item>
                <item name="extension" xsi:type="string">\Magento\Framework\Api\Code\Generator\ExtensionAttributesGenerator</item>
                <item name="remote" xsi:type="string">\Magento\Framework\MessageQueue\Code\Generator\RemoteServiceGenerator</item>
            </argument>
        </arguments>
    </type>
```

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+        $this->classMethods = [];
+        $this->classProperties = [];
+        $this->injectPropertiesSettersToConstructor($reflection->getConstructor(), [
+            ScopeInterface::class => '____scope',
+            ObjectManagerInterface::class => '____om',
+        ]);
+        $this->overrideMethodsAndGeneratePluginGetters($reflection);
+
+        //return parent::_generateCode();
+        return EntityAbstract::_generateCode();
```
maghamed: inheriting from `\Magento\Framework\Code\Generator\EntityAbstract` , you can just call parent's method

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+        }
+        foreach ($allPlugins as $plugins) {
+            foreach ($plugins as $plugin) {
+                $this->classMethods[] = $this->_getPluginGetterInfo($plugin);
+                $this->classProperties[] = $this->_getPluginPropertyInfo($plugin);
+            }
+        }
+    }
+
+    protected function injectPropertiesSettersToConstructor(\ReflectionMethod $parentConstructor = null, $properties = [])
```
maghamed: no need to add new methods as protected.
Let's keep them private , all the new variables should be also private and follow current standards

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+
+    protected function _getClassMethods()
+    {
+        return $this->classMethods;
+    }
+
+    protected function _getClassProperties()
+    {
+        return $this->classProperties;
+    }
```
maghamed, line 347: This looks like temporal coupling which should be avoided.

In a sense, that methods _getClassMethods and _getClassProperties could be called just after another method which will populate corresponding properties.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+            'parameters' => [],
+            'body' => implode("\n", $body),
+            //'returnType' => $class,
+            'docblock' => [
+                'shortDescription' => 'plugin "' . $plugin['code'] . '"' . "\n" . '@return \\' . $plugin['class']
+            ],
+        ];
+    }
+
+    protected function _getCompiledMethodInfo(\ReflectionMethod $method, $config)
```
maghamed: new methods should be private and without preceding underscore

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
```diff
@@ -0,0 +1,72 @@
...
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\CompiledInterception\Generator;
+
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\Interception\PluginList\PluginList;
+use Magento\Framework\Interception\ObjectManager\ConfigInterface;
+use Magento\Framework\ObjectManager\Config\Reader\Dom;
+
+class CompiledPluginList extends PluginList
```
maghamed: No need to inherit from PluginList, 
just implement the interface `\Magento\Framework\Interception\PluginListInterface`

lib/internal/Magento/Framework/CompiledInterception/Generator/FileCache.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        // TODO: Implement test() method.
+    }
+
+    /**
+     * Load cache record by its unique identifier
+     *
+     * @param string $identifier
+     * @return string|bool
+     * @api
```
maghamed: `@api` applied to a class level

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/CompiledInterceptorTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace Magento\Framework\CompiledInterception\Test\Unit\CompiledInterceptor;
+
+use Magento\Framework\Code\Generator\Io;
+use Magento\Framework\CompiledInterception\Generator\CompiledInterceptor;
+
+use Magento\Framework\CompiledInterception\Test\Unit\CompiledPluginList\CompiledPluginListTest;
+use \PHPUnit_Framework_MockObject_MockObject as MockObject;
+
+class CompiledInterceptorTest extends \PHPUnit\Framework\TestCase
```
maghamed: these tests are rather Integration than a unit one

maghamed (changes requested):

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+
+        if (isset($conf[DefinitionInterface::LISTENER_AFTER])) {
+            foreach ($conf[DefinitionInterface::LISTENER_AFTER] as $plugin) {
+                if ($returnVoid) {
+                    $chain[] = ["((\$tmp = \$this->" . $this->getGetterName($plugin) . "()->after$capName(\$this, \$result$extraParams)) !== null) ? \$tmp : \$result;"];
+                } else {
+                    $chain[] = ["\$this->" . $this->getGetterName($plugin) . "()->after$capName(\$this, \$result$extraParams);"];
+                }
+            }
+        }
```
maghamed, line 423: we can introduce either 3 corresponding methods or classes which will generate handlers before/around/after

avoiding introducing too many conditional logic to the current method.
Doing so we will make this method much easier to read , and remove annotation for 
```
     * @SuppressWarnings(PHPMD.NPathComplexity)
```

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+
+    /**
+     * @param \ReflectionMethod $method
+     * @param $conf
+     * @param $parameters
+     * @return array
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     * @SuppressWarnings(PHPMD.NPathComplexity)
+     */
+    protected function _getMethodSourceFromConfig(\ReflectionMethod $method, $conf, $parameters, $returnVoid)
```
maghamed: all the new methods should be private and should follow new naming standards (should not start from the underscore)

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,399 @@
...
+        return $ret;
+    }
+
+    protected function _getPluginsConfig(\ReflectionMethod $method, &$allPlugins)
+    {
+        $className = ltrim($this->getSourceClassName(), '\\');
+
+        $ret = array();
+        foreach ($this->plugins as $scope => $pluginsList) {
+            $p = $this->_getPluginsChain($pluginsList, $className, $method->getName(), $allPlugins);
```
maghamed: don't use meaningless variable names like `res`, `p`, `k` - all the . variables should have meaningful names and should be easy to read

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
```diff
@@ -0,0 +1,72 @@
...
+            new FileCache($cachePath),
+            new \Magento\Framework\ObjectManager\Relations\Runtime(),
+            $omConfig,
+            new \Magento\Framework\Interception\Definition\Runtime(),
+            $objectManager,
+            new \Magento\Framework\ObjectManager\Definition\Runtime(),
+            ['first' => 'global'],
+            'compiled_plugins_' . $scope,
+            new FileCache($cachePath)
+        );
```
maghamed: avoiding inheritance will make this class much nicer

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
```diff
@@ -0,0 +1,72 @@
...
+            new StaticScope($scope),
+            new FileCache($cachePath),
+            new \Magento\Framework\ObjectManager\Relations\Runtime(),
+            $omConfig,
+            new \Magento\Framework\Interception\Definition\Runtime(),
+            $objectManager,
+            new \Magento\Framework\ObjectManager\Definition\Runtime(),
+            ['first' => 'global'],
+            'compiled_plugins_' . $scope,
+            new FileCache($cachePath)
```
maghamed: it's undesirable to provide to for class initialization the instance of the same type twice , like 
` new FileCache($cachePath)` and  ` new FileCache($cachePath)`

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: all classes that implement an interface will have calls to plugins for this interface in the compiled interceptor. `CompiledPluginsList` class takes care of that using same logic as previous "runtime mode".

fsw: fixed return type

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: this is a good point. generatedEntities seems not to be working in production mode. However after removing inheritance from Generator\Interceptor using preferences for developer/production mode seems to be working fine. So this one is fixed.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
fsw: This inherits from `PluginList` not to duplicate code. It uses parent class logic to make sure plugins will be executed exactly in the same order and using same logic as they were previously.

lib/internal/Magento/Framework/CompiledInterception/Generator/FileCache.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
fsw: This class actually exists so we can use logic from `_loadScopedData` and `_inheritPlugins` methods of `PluginList` without modifying this class. It mostly removes few DI dependencies so interceptors can be compiled with as few dependencies as possible (because even core classes can have plugins in theory).

Alternatively I could just modify `PluginList` class but this way we keep original method intact.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        if (!$reader || !$omConfig) {
+            $reader = $objectManager->get(Dom::class);
+            $omConfig = $objectManager->get(ConfigInterface::class);
+        }
+        parent::__construct(
+            $reader,
+            new StaticScope($scope),
+            new FileCache($cachePath),
+            new \Magento\Framework\ObjectManager\Relations\Runtime(),
```
YevSent: Please, avoid FQCN usage

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/CompiledInterceptorTest.php
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+namespace Magento\Framework\CompiledInterception\Test\Unit\CompiledInterceptor;
+
+use Magento\Framework\App\AreaList;
+use Magento\Framework\Code\Generator\Io;
+use Magento\Framework\CompiledInterception\Generator\CompiledInterceptor;
+
+use Magento\Framework\CompiledInterception\Test\Unit\CompiledPluginList\CompiledPluginListTest;
+use \PHPUnit_Framework_MockObject_MockObject as MockObject;
```
YevSent: Please, use \PHPUnit\Framework\MockObject\MockObject instead

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
```diff
@@ -0,0 +1,193 @@
...
+        		}, $variadicValue);
+        	default:
+        		return parent::firstVariadicParameter($variadicValue);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function secondVariadicParameter($value, ... $variadicValue)
```
YevSent: Redundant empty space between `...` and argument name.

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/Custom/Module/Model/ComplexItemTyped.php
```diff
@@ -0,0 +1,85 @@
...
+// @codingStandardsIgnoreFile
+
+namespace Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model;
+
+class ComplexItemTyped
+{
+    private $value;
+    private $variadicValue;
+
+    public function returnVoid() : void
```
YevSent: Redundant empty space before `:`

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Interceptor class for @see \Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ComplexItem
+ */
+class Interceptor extends \Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ComplexItem
+{
+    /**
+     * @var ScopeInterface
+     */
+    private $____scope = null;
```
YevSent: I would suggest reducing the usage of underscore for 2 instead of 4.

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * @var \Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ItemPlugin\Advanced
+     */
+    private $____plugin_advanced_plugin = null;
+
+    /**
+     * @var \Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ItemPlugin\Complex
+     */
+    private $____plugin_complex_plugin = null;
```
YevSent: I would suggest using cameCase naming because the generated code looks messy.

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        switch($this->____scope->getCurrentScope()){
+        	case 'global':
+        	case 'frontend':
+        	case 'emptyscope':
+        		$this->____plugin_advanced_plugin()->beforeGetName($this);
```
YevSent: The mix of cameCase and underscore names looks messy.

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
```diff
@@ -0,0 +1,193 @@
+namespace Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ComplexItem;
+
+use Magento\Framework\Config\ScopeInterface;
+use Magento\Framework\ObjectManagerInterface;
+
+/**
+ * Interceptor class for @see \Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ComplexItem
+ */
+class Interceptor extends \Magento\Framework\CompiledInterception\Test\Unit\Custom\Module\Model\ComplexItem
```
YevSent: Please, avoid FQCN usage

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+use Magento\Framework\Config\ScopeInterface;
+use Magento\Framework\Interception\DefinitionInterface;
+use Magento\Framework\App\AreaList;
+
+class CompiledInterceptor extends EntityAbstract
+{
+    /**
+     * Entity type
+     */
+    const ENTITY_TYPE = 'interceptor';
```
YevSent: Constants should have a visibility scope.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+
+    protected $plugins;
+
+    protected $classMethods = null;
+
+    protected $classProperties = null;
+
+    protected $baseReflection = null;
```
YevSent: How this property is used?

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+use Magento\Framework\App\AreaList;
+
+class CompiledInterceptor extends EntityAbstract
+{
+    /**
+     * Entity type
+     */
+    const ENTITY_TYPE = 'interceptor';
+
+    protected $plugins;
```
YevSent: By default, all properties should be private.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+
+    protected $classMethods = null;
```
YevSent: Why do you need these null assignments, null is a default value.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+    public function __construct(
+        AreaList $areaList,
+        $sourceClassName = null,
+        $resultClassName = null,
+        Io $ioObject = null,
+        CodeGeneratorInterface $classGenerator = null,
+        DefinedClasses $definedClasses = null,
+        $plugins = null
+    )
+    {
```
YevSent: `{` should be on the same line as `)`

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+use Magento\Framework\Code\Generator\CodeGeneratorInterface;
+use Magento\Framework\Code\Generator\DefinedClasses;
+use Magento\Framework\Code\Generator\Io;
+use Magento\Framework\ObjectManagerInterface;
+use Magento\Framework\Code\Generator\EntityAbstract;
+use Magento\Framework\Config\ScopeInterface;
+use Magento\Framework\Interception\DefinitionInterface;
+use Magento\Framework\App\AreaList;
+
+class CompiledInterceptor extends EntityAbstract
```
YevSent: The class description is missed

lib/internal/Magento/Framework/CompiledInterception/Generator/FileCache.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Clean cache records matching specified tags
+     *
+     * @param string $mode
+     * @param array $tags
+     * @return bool
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     */
+    public function clean($mode = \Zend_Cache::CLEANING_MODE_ALL, array $tags = [])
```
YevSent, line 109: It's better to introduce class constant instead of using zend framework 1

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+     * @param $config
+     * @return array
+     */
+    private function getCompiledMethodInfo(\ReflectionMethod $method, $config)
+    {
+        $parameters = $method->getParameters();
+        $returnsVoid = ($method->hasReturnType() && $method->getReturnType()->getName() == 'void');
+
+        $body = [
+            'switch($this->____scope->getCurrentScope()){'
```
YevSent: 1. Should be empty space after `switch`. 
2. Should empty space between `)` and `{`.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+        return new \ReflectionClass($this->getSourceClassName());
+    }
+
+    /**
+     * Whether method is intercepted
+     *
+     * @param \ReflectionMethod $method
+     * @return bool
+     */
+    protected function isInterceptedMethod(\ReflectionMethod $method)
```
YevSent: Method should be private

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+    /**
+     * @throws \ReflectionException
+     */
+    private function generateMethodsAndProperties()
+    {
+        if ($this->classMethods === null) {
+            $this->classMethods = [];
+            $this->classProperties = [];
+
+            $this->injectPropertiesSettersToConstructor($this->getSourceClassReflection()->getConstructor(), [
```
YevSent: Breaks PSR rules, each argument should be on a separate line.

lib/internal/Magento/Framework/CompiledInterception/README.md
```diff
@@ -0,0 +1,88 @@
...
+The Interceptors generated by this plugin are 100% compatible with the ones generated by Magento by default, so there is no need to change anything in your plugins.
+
+### ENABLING
+
+To use compiled interceptors in developer mode please add following preference to your di.xml
+```
+<preference for="Magento\Framework\Interception\Code\Generator\Interceptor" type="Magento\Framework\CompiledInterception\Generator\CompiledInterceptor" />
+```
+
+To use compiled interceptors in production mode please add following preference to your di.xml
```
YevSent: This is needed for all modes where `php bin/magento setup:di:compile` is used

YevSent:
The important thing for `setup:di:compile` command usage. I've used my local environment:
Before changes:
Exec time: **62.41sec** Memory: **293MiB**

After changes:
Exec time: **133.78sec** Memory: **353MiB**

The code generation takes 2 times longer and consumes more memory.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+        return implode(', ', $ret);
+    }
+
+    /**
+     * @param $plugin
+     * @return string
+     */
+    private function getGetterName($plugin)
+    {
+        return '____plugin_' . $plugin['clean_name'];
```
YevSent, line 499: This getter should be private, not a public.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+                'tags' => [['name' => 'var', 'description' => '\\' . $plugin['class']]],
+            ]
+        ];
+    }
+
+    /**
+     * @param $plugin
+     * @return array
+     */
+    private function getPluginGetterInfo($plugin)
```
YevSent, line 508: It doesn't make sense to store a dependency on private properties because `ObjectManager::get` is used. It reduces code readability.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,532 @@
...
+            $this->addCodeSubBlock($body, $this->getMethodSourceFromConfig($method->getName(), $case['conf'], $parameters, $returnsVoid), 2);
+            if ($returnsVoid) {
+                $body[] = "\t\tbreak;";
+            }
+        }
+
+        $body[] = "}";
+        $returnType = $method->getReturnType();
+        $returnTypeValue = $returnType
+            ? ($returnType->allowsNull() ? '?' : '') .$returnType->getName()
```
YevSent: Should be whitespace between `.` and `$returnType`

YevSent (changes requested):

YevSent:
The names of generated methods should be more convenient.
Screen Shot 2019-05-24 at 1 19 47 PM

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
fsw: I think this is an issue of `Magento\Framework\Code\Generator\ClassGenerator` witch is not touched in this PR, please see `magento/framework/Interception/Test/Unit/Code/Generator/_files/TInterceptor.txt` it is working this way already.

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
fsw: As reading parent class fields would add even more to generation time this is to avoid name collision. I know 4 seem paranoid but on the other hand generated code clearly stands out witch is good i suppose. Do we have any guideline/standard for generated code in such case?

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
fsw: Should this apply to generated code? It is the easiest and cleanest way to avoid name collisions (Plugins can have last bit of FQCN same as intercepted class for example). Also this works this way currently and I assumed this is reasonable.

lib/internal/Magento/Framework/CompiledInterception/Generator/FileCache.php
fsw: This comes from `Magento\Framework\Cache\FrontendInterface` interface.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/CompiledInterceptorTest.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/Custom/Module/Model/ComplexItemTyped.php
fsw: fixed. However I would like to point out that this is weird that we are using this convention but code generated by `Magento\Framework\Code\Generator\ClassGenerator` is different.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: good catch. This was supposed to be a cache for Reflection class to make generation slightly more faster. Fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
fsw: I will elaborate in PR comment to open discussion about this.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: Is there any guideline for this? It seems like most constants in magento code (including this one in parent class) do not have a scope hence are public?

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: When I bench-marked this, using property gave a tiny speed benefit. 
Also without this property this getter is moot because `ObjectManager` could be used inline.
Isn't this getter with a cleanly defined return type readable enough?

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

fsw:
@joni-jones thanks for your review.
Re getter names:

To generate getter name I am using name attribute from `di` config:

`<plugin name="admingws_category_resource_plugin" type="Magento\AdminGws\Model\Plugin\CategoryResource"/>`

I see there are different conventions for value of this attribute used (camelCase/underscore etc). Is there a place witch says what format should be used?

I think upside of this is that you can run a text search for "admingws_category_resource_plugin" in code and see where it is used. But on the other hand this method already have this key in its doc block so maybe it is redundant indeed.

But if this is not the way then what value would be best here. Should I just convert this attribute to camel case like: `admingwsCategoryResourcePlugin` ?

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: this was a good idea. fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/CompiledInterceptorTest.php
fsw: fixed, splited tests into unit and integration

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: fixed

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
fsw: this is just to be consistent with Interceptor class

lib/internal/Magento/Framework/CompiledInterception/README.md
fsw: fixed

fsw:
Changes to interceptor constructors (adding objects with DI) were fixed in compiled mode.
One extra preference needs to be added for it to work.

fsw:
@joni-jones @maghamed all your remarks from your reviews were either fixed or require further discussion. Not sure if I shall take any steps to push those back to you.

davidverholen:
@fsw is [this commit](https://github.com/magento/magento2/pull/22826/commits/e57df097ad16ef43e6bfd080797c85cce2ea37ec) a fix for [this issue](https://github.com/magento/magento2/issues/22769) ?

fsw:
@davidverholen

This issue is with the default method and I was not aware it exists.
This commit fixes general issues with methods returning "self" so it fixes this #11905 for compiled mode.
However I am using `getDeclaringClass` to substitute for self so probably #22769 wont apply in compiled interceptors and this bit:

        if ($returnTypeValue === 'self') {
            $returnTypeValue = $method->getDeclaringClass()->getName();
        }

can be also used to fix #22769 in default mode.

maghamed:
@fsw 
The corresponding PR with Architectural proposal is created - https://github.com/magento/architecture/pull/185

Currently, I run voting regarding the incorporation of this PR in #appdesign Slack channel, please join there to Vote
https://magentocommeng.slack.com/archives/CBSL1DF8B/p1559858324023000

joshuaadickerson:
Use type hints.

lib/internal/Magento/Framework/CompiledInterception/Test/Unit/CompiledInterceptor/_out_interceptors/ComplexItem.txt
YevSent: We have a rule according to PSR-2 (https://www.php-fig.org/psr/psr-2/#42-properties) that properties shouldn't have an underscore in names, I believe, even `__` usage is a very rare case.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
YevSent: The method by itself is readable enough (if not to take into account the method name and FQCN for return type) but the method body looks noisy, the properties names too long.

fsw:
I have pushed a fix to optimize config loading and memory usage to improve `setup:di:compile` performance as it was affected as pointed out by @joni-jones 

compilation timing:

without module: 2m39.824s

with module before fix: 5m47.604s

with module after fix: 2m33.752s



Also I have discovered that `warmup` request takes so long only because we generate code for all scopes at once. Hence when you clear generated and go to fronted, even admin config XMLs are red/parsed etc.

I think that if it makes such a significant difference maybe it will be worth reconsidering the idea of writing a separate interceptor class for each scope. I did not like the idea at first because I am not sure how hard would it be and how it will affect the system to modify OM to load different classes depending on scope.

Another solution would be to have a 'intermidiate' `Interceptor` class that calls `FrontendInterceptor` or `AdminInterceptor` class. Not sure how this will affect code readability though.

fsw:
Further investigation showed that the issue is not exactly reading config for all scopes but the fact that `PluginList` does not cache data at scope level hence:
* generator was loading config for `general|frontend` 
* *.xml files for `general` and for `frontend` scopes were loaded and merged
* generator was loading config for `general|adminhtml` 
* *.xml files for `general` and for `adminhtml` scopes were loaded and merged
* and so on
Hence general scope has lots of xml parsing them for each scope was the bottleneck.

Last commit fixed this issue. Here is the updated benchmark:

reqest |   | Interceptor | CompiledInterceptor | gain
-- | -- | -- | -- | --
home page | warmup request | 18498 | 18383 | 0.62%
  | time per request | 217.1 | 207.97 | 4.21%
  | time per request (4 concurrent requests) | 54.275 | 51.993 | 4.20%
listing | warmup request | 18747 | 18206 | 2.89%
  | time per request | 2878.525 | 2727.68 | 5.24%
  | time per request (4 concurrent requests) | 719.631 | 681.92 | 5.24%
product page | warmup request | 18833 | 18784 | 0.26%
  | time per request | 217.022 | 206.669 | 4.77%
  | time per request (4 concurrent requests) | 54.255 | 51.667 | 4.77%
admin dashboard page | warmup request | 22955 | 20965 | 8.67%
  | time per request | 1363.842 | 1201.763 | 11.88%
  | time per request (4 concurrent requests) | 340.96 | 300.441 | 11.88%

@maghamed  I guess as slower first request time was the biggest downside here and now it is even a bit faster, the poll should be rebooted with different question? Sorry for the confusion, I was under the wrong impression reflection is the bottleneck here.

amenk:
@fsw Thanks for the awesome work. I can't wait to get this merged and I hope it will (I am a big fan of the improvements in the debugging process)

Can you fix the static tests to get everything green?

lib/internal/Magento/Framework/CompiledInterception/Generator/AreasPluginList.php
```diff
@@ -0,0 +1,65 @@
...
+ * Class AreasPluginList
+ */
+class AreasPluginList
+{
+    /** @var Scope */
+    private $scope;
+
+    /** @var array */
+    private $plugins;
+
```
gamort, line 35: Why use private?  Your just destroying the ability to subclass properly.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledPluginList.php
gamort: If you inherit from PluginList then you have to implement anything which is private.  
Properties: $logger, $serializer
Methods: trimInstanceStartingBackslash, filterPlugins, getLogger

Make life easier for anyone who wants to subclass from you and define them as protected.  You are allowed to raise the visibility of a private method in a child since the parent and any siblings don't know about it.

lib/internal/Magento/Framework/CompiledInterception/Generator/AreasPluginList.php
YevSent: All properties by default should be private. The composition should be used instead of inheritance. Also, the list of plugins can be customized via a constructor.

fsw: Both approaches has pros and cons but I guess consistency is most important and I assume "private by default " is M2 new standard.

fsw:
We have found some backward incompatible cases in witch module was overriding a plugin via preference (instead of redefining it). 
This made plugin getter throw runtime exception as returned class did not match.
That's why return type was removed from plugin getter and now this value is in phpdoc.

buskamuza:
Just FYI: proposal (https://github.com/magento/architecture/pull/185) has been just merged.

sidolov:
@magento run all tests

lib/internal/Magento/Framework/CompiledInterception/Generator/AreasPluginList.php
```diff
@@ -0,0 +1,64 @@
...
+    public function __construct(
+        Scope $scope,
+        $plugins = null
+    ) {
+        $this->scope = $scope;
+        $this->plugins = $plugins;
+    }
+
+    /**
+     * Get array of plugns config indexed by scope code
```
YevSent: Typo in "plugins" word.

lib/internal/Magento/Framework/CompiledInterception/Generator/AreasPluginList.php
```diff
@@ -0,0 +1,64 @@
...
+        if ($this->plugins === null) {
+            $this->plugins = [];
+            //this is to emulate order M2 is reading scopes config to use scope cache
+            //"global|primary" should be loaded first and then "global|primary|frontend" etc.
+            $defaultScopePluginList = $defaultScope = null;
+            foreach ($this->scope->getAllScopes() as $scope) {
+                if ($defaultScopePluginList === null) {
+                    $defaultScopePluginList = new CompiledPluginList(
+                        ObjectManager::getInstance(),
+                        new StaticScope($scope)
```
YevSent: I would suggest using a factory for object creation instead of the new to provide scope replaceability and better testing possibilities.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
YevSent: Old constants do not have visibility scope as it was introduced only PHP 7.1. In general a rule, the same for the methods - by default everyting is private. Existing, inherited constants should have public visibility scope.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,720 @@
...
+class CompiledInterceptor extends EntityAbstract implements InterceptorInterface
+{
+    /**
+     * Entity type
+     */
+    const ENTITY_TYPE = 'interceptor';
+
+    private $classMethods;
+
+    private $classProperties;
```
YevSent, line 33: Doc block is missed.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    /**
+     * Unused function required by production mode interface
+     *
+     * @param mixed $interceptedMethods
+     */
+    public function setInterceptedMethods($interceptedMethods)
+    {
+        //this is not used as methods are read from reflaction
```
YevSent: The typo in "reflection" word.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
YevSent: I don't see reasons to don't make it private. Also, it's an alternative realization and doesn't inherit default implementation.

lib/internal/Magento/Framework/CompiledInterception/Generator/StaticScope.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Class StaticScope
+ */
+class StaticScope implements ScopeInterface
+{
+    /**
+     * @var string
+     */
+    protected $scope;
```
YevSent: The visibility scope should be private.

engcom-Foxtrot:
@magento run Integration-Tests

engcom-Foxtrot:
@joni-jones updated.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,723 @@
...
+            $this->overrideMethodsAndGeneratePluginGetters($this->getSourceClassReflection());
+        }
+    }
+
+    /**
+     * Get properties to be injected from DI
+     *
+     * @return array
+     */
+    public static function propertiesToInjectToConstructor()
```
YevSent: The public function should not be mixed with private, it decreases the class readability.

lib/internal/Magento/Framework/CompiledInterception/Test/Integration/CompiledInterceptor/CompiledInterceptorTest.php
```diff
@@ -0,0 +1,170 @@
...
+        $readerMock->expects($this->any())->method('read')->will($this->returnValueMap($readerMap));
+
+        $omMock = $this->createMock(ObjectManager::class);
+        $omMock->method('get')->with(\Psr\Log\LoggerInterface::class)->willReturn(new NullLogger());
+
+        $omConfigMock =  $this->getMockForAbstractClass(
+            \Magento\Framework\Interception\ObjectManager\ConfigInterface::class
+        );
+
+        $omConfigMock->expects($this->any())->method('getOriginalInstanceType')->will($this->returnArgument(0));
```
YevSent: `expects($this->any())` can be omitted as it's a default behavior.

YevSent: Please, use `willReturn` instead of `will`

lib/internal/Magento/Framework/CompiledInterception/Test/Integration/CompiledInterceptor/CompiledInterceptorTest.php
```diff
@@ -0,0 +1,170 @@
...
+    {
+        $readerMap = include __DIR__ . '/../_files/reader_mock_map.php';
+        $readerMock = $this->createMock(\Magento\Framework\ObjectManager\Config\Reader\Dom::class);
+        $readerMock->expects($this->any())->method('read')->will($this->returnValueMap($readerMap));
+
+        $omMock = $this->createMock(ObjectManager::class);
+        $omMock->method('get')->with(\Psr\Log\LoggerInterface::class)->willReturn(new NullLogger());
+
+        $omConfigMock =  $this->getMockForAbstractClass(
+            \Magento\Framework\Interception\ObjectManager\ConfigInterface::class
```
YevSent: Please, avoid FQCN usage.

engcom-Foxtrot:
@joni-jones updated, please review.

lib/internal/Magento/Framework/CompiledInterception/Generator/AreasPluginList.php
```diff
@@ -0,0 +1,93 @@
...
+            foreach ($this->scope->getAllScopes() as $scope) {
+                $configScope = $this->scopeInterfaceFactory->create(
+                    [
+                        'scope' => $scope,
+                    ]
+                );
+                if ($defaultScopePluginList === null) {
+                    $defaultScopePluginList = $this->compiledPluginListFactory->create(
+                        [
+                            'objectManager' => ObjectManager::getInstance(),
```
YevSent: Why do you need to get the object manager instance inside the loop? You can get it once and re-use it.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,723 @@
...
+
+    /**
+     * @var AreasPluginList
+     */
+    private $areasPlugins;
+
+    /**
+     * CompiledInterceptor constructor.
+     * @param AreasPluginList $areasPlugins
+     * @param null|string $sourceClassName
```
YevSent: The `null` should not be specified as a first argument.

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
```diff
@@ -0,0 +1,723 @@
...
+     * @param string $method
+     * @param array $allPlugins
+     * @param string $next
+     * @return array
+     */
+    private function getPluginsChain(CompiledPluginList $plugins, $className, $method, &$allPlugins, $next = '__self')
+    {
+        $result = $plugins->getNext($className, $method, $next);
+        if (!empty($result[DefinitionInterface::LISTENER_BEFORE])) {
+            foreach ($result[DefinitionInterface::LISTENER_BEFORE] as $k => $code) {
```
YevSent, line 641: Looks like these two conditions blocks are similar and the difference only in the `$result` index. Maybe it makes sense to extract this logic to a separate method.

setup/src/Magento/Setup/Module/Di/Compiler/Config/Chain/CompiledInterceptorSubstitution.php
```diff
@@ -0,0 +1,57 @@
...
+    public function modify(array $config)
+    {
+        $config = $this->interceptorSubstitution->modify($config);
+
+        foreach ($config['arguments'] as $instanceName => &$arguments) {
+            if (substr($instanceName, -12) === '\Interceptor') {
+                foreach (CompiledInterceptor::propertiesToInjectToConstructor() as $type => $name) {
+                    $preference = isset($config['preferences'][$type]) ? $config['preferences'][$type] : $type;
+                    foreach ($arguments as $argument) {
+                        if (isset($argument['_i_']) && $argument['_i_'] == $preference) {
```
YevSent: Could you explain why it doesn't use strict comparison?

lib/internal/Magento/Framework/CompiledInterception/README.md
```diff
@@ -0,0 +1,92 @@
+### ABOUT
```
YevSent, line 1: Please, make sure the latest README is updated according to the latest changes.

lib/internal/Magento/Framework/CompiledInterception/Test/Integration/CompiledInterceptor/_out_interceptors/Item.txt
```diff
@@ -0,0 +1,87 @@
...
+        return $this->____plugin_simple_plugin;
+    }
+
+    /**
+     * plugin "advanced_plugin"
+     * @return \Magento\Framework\CompiledInterception\Test\Integration\CompiledInterceptor\Custom\Module\Model\ItemPlugin\Advanced
+     */
+    private function ____plugin_advanced_plugin()
+    {
+        if ($this->____plugin_advanced_plugin === null) {
```
YevSent: To avoid issues with state modification by plugins in a stateful mode like Magento + Swoole, we need to remove class level caching for plugins references via private properties. So instead of `$this->____plugin_advanced_plugin = $this->____om->get(ClassNamespace::class)` it should be `$____plugin_advanced_plugin = $this->____om->get(ClassNamespace::class)`. All plugins references should be local variables.

YevSent (changes requested):

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
engcom-Foxtrot: @joni-jones this method would have five arguments with some of them passed by reference. I think such an update would only decrease code readability.

engcom-Foxtrot:
@joni-jones updated.

VladimirZaets:
@magento run all tests

lib/internal/Magento/Framework/CompiledInterception/Generator/CompiledInterceptor.php
YevSent: How it would have five arguments if in both `if` conditions it has four arguments and the difference only in `DefinitionInterface::LISTENER_BEFORE` and `DefinitionInterface::LISTENER_AFTER`.

ihor-sviziev:
@engcom-Foxtrot I see you were committing some changes last time. 
Is there any updates on this PR?

ihor-sviziev:
@VladimirZaets @slavvka any updates?

ihor-sviziev:
@magento run all tests

sidolov:
@magento run all tests

sidolov:
@magento run all tests

sidolov:
@magento run Semantic Version Checker

sidolov:
@magento run all tests

sidolov:
@magento run all tests

okorshenko:
@magento create issue

sivaschenko:
@ihor-sviziev the pull request is targeting 2.5 release line, it's currently on hold until the corresponding branch is going to be created.

ihor-sviziev:
Hi @sivaschenko @sidolov,
2.5-develop branch was already created. Could you re-target this PR to it and resolve all merge conflicts?

sivaschenko:
The base branch was changed to 2.5-develop, @fsw can you please resolve the conflicts

mrtuvn:
CC @fsw you you update with resolve conflicts files

fsw:
Hi @mrtuvn @sivaschenko,
I have resolved conflicts but was not sure about them as I am a bit out of the loop here, I see some fixes beeing pushed here and I am not completely clear about reasons. I am using and focusing on module version. 
If someone needs access to source repo to help move this further please just let me know.
cheers

mrtuvn:
@magento run all tests

thlassche:
What’s needed to get this merged? I’d happily help out on it, we all benefit from a faster core I’d say

ihor-sviziev:
@sidolov @sivaschenko ?

kandy:
I believe that PR still has failed builds
@magento run all tests

kandy:
@magento run all tests

ihor-sviziev:
As discussed with @sidolov in Slack, he should ping the team working on finalizing this PR and write some updates here.

ihor-sviziev:
As we discussed with @sidolov in Slack - he'll try to allocate time next week to go through this PR and figure out what's left.

fsw:
@ihor-sviziev @sidolov this version and module version have diverged a bit but maybe it will be useful for you to take a look at fixes to module version to see if they are useful here:
https://github.com/creatuity/magento2-interceptors/pulls?q=is%3Apr+

engcom-Hotel:
Hello @fsw,

Thanks for the contribution!

Currently, we are not processing the PRs targeted for 2.5-develop. Hence we are moving this PR `on hold` for now.

We will pick this PR in the future whenever we start working on 2.5-develop.

Thanks

renttek:
As far as I can see, there is not even a timeline for 2.5, the next release in 2024 is 2.4.7.
So the earliest possible release after this (according the the plan) would be somewhere 2025.

So my honest (yes or no) question: will this PR ever happen? (it's been 4+ years now)

loid345:
> As far as I can see, there is not even a timeline for 2.5, the next release in 2024 is 2.4.7. So the earliest possible release after this (according the the plan) would be somewhere 2025.
> 
> So my honest (yes or no) question: will this PR ever happen? (it's been 4+ years now)
This will probably happen when shopware 8 is released and everyone will forget about magento)

sinhaparul:
We are currently processing PRs targeted to 2.4-develop only. So, removing this from the community picked backlog for now.

memegento-dev:
you were supposed to fight evil not join it

image

Den4ik:
@sinhaparul I believe that we can update PR with changes for Magento 2.4 from https://github.com/creatuity/magento2-interceptors and move forward. What do you think?

Aquive:
Any updates?

MagePsycho:
This feature has already been incorporated into Mage-OS and is still under review for Magento Open Source.
It's time for Magento to catch up!

choussamaster:
any update !

lbajsarowicz:
I'd love to hear some updates here.

PR #16445, opened 2018-06-28
getRegionHtmlSelect does not have configuration

kandarp26:
### Description

When i use the getRegionHtmlSelect function in custom module there is not way to pass argument for create region drop down.  This arguments available in getCountryHtmlSelect in same way need to implement in getRegionHtmlSelect .

### Fixed Issues 
magento/magento2#7491:Method does not have configuration

### Manual testing scenarios
1. Created the address from customer account. 
2. Created the Order.
3. Created and Edit the customer from admin with address details. 

### Contribution checklist
   [X] Pull request has a meaningful description of its purpose
   [X] All commits are accompanied by meaningful commit messages
   [ ] All new or changed code is covered with unit/integration tests 
  [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #22834, opened 2019-05-11
#16445 - getRegionHtmlSelect does not have configuration - resolved

nikunjskd20:
### Description (*)

To give region field a name like array Ex. name="somearray[region]"  and getting data as array value. 

1. magento/magento2#16445: IgetRegionHtmlSelect does not have configuration

1. ... create form with getRegionHtmlSelect() you will get a static name 'region'. developer can not pass any other value.
2. ...

### DIFF
--- a/app/code/Magento/Directory/Block/Data.php
+++ b/app/code/Magento/Directory/Block/Data.php
@@ -3,6 +3,8 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+declare(strict_types=1);
+
 namespace Magento\Directory\Block;
 
 /**
@@ -173,10 +175,33 @@ class Data extends \Magento\Framework\View\Element\Template
      * Returns region html select
      *
      * @return string
+     * @deprecated
+     * @see getRegionSelect() method for more configurations
      */
     public function getRegionHtmlSelect()
     {
+        return $this->getRegionSelect();
+    }
+
+    /**
+     * Returns region html select
+     *
+     * @param null|int $value
+     * @param string $name
+     * @param string $id
+     * @param string $title
+     * @return string
+     */
+    public function getRegionSelect(
+        ?int $value = null,
+        string $name = 'region',
+        string $id = 'state',
+        string $title = 'State/Province'
+    ): string {
         \Magento\Framework\Profiler::start('TEST: ' . __METHOD__, ['group' => 'TEST', 'method' => __METHOD__]);
+        if ($value === null) {
+            $value = (int)$this->getRegionId();
+        }
         $cacheKey = 'DIRECTORY_REGION_SELECT_STORE' . $this->_storeManager->getStore()->getId();
         $cache = $this->_configCacheType->load($cacheKey);
         if ($cache) {
@@ -188,15 +213,15 @@ class Data extends \Magento\Framework\View\Element\Template
         $html = $this->getLayout()->createBlock(
             \Magento\Framework\View\Element\Html\Select::class
         )->setName(
-            'region'
+            $name
         )->setTitle(
-            __('State/Province')
+            __($title)
         )->setId(
-            'state'
+            $id
         )->setClass(
             'required-entry validate-state'
         )->setValue(
-            (int)$this->getRegionId()
+            $value
         )->setOptions(
             $options
         )->getHtml();


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22834) <br/>All committers have signed the CLA.

app/code/Magento/Directory/Block/Data.php
```diff
@@ -172,11 +172,18 @@ public function getRegionCollection()
...
      * Returns region html select
      *
+     * @param null|string $defValue
+     * @param string $name
+     * @param string $id
+     * @param string $title
      * @return string
      */
-    public function getRegionHtmlSelect()
+    public function getRegionHtmlSelect($defValue = null,$name = 'region', $id = 'state', $title = 'State/Province')
```
swnsma: Adding new parameters to public method is violate backward compatibility.
I think we can maximize backward compatibility with:
- mark method **getRegionHtmlSelect** as deprecated
- create new method with parameters (e.g. **getRegionSelect** or **getRegionHtml**)
- add `@see` section and call newly created method in **getRegionHtmlSelect**

swnsma: Please rename parameter `$defValue` to `$defaultValue` or just `$value`. 
It is better to avoid reductions.

swnsma (changes requested):
Hi @nikunjskd20,
Your PR looks good, but there are some things which need to be adjusted.

Please check my comment.
Please check failed static tests.

app/code/Magento/Directory/Block/Data.php
nikunjskd20: > Adding new parameters to public method is violate backward compatibility.
> I think we can maximize backward compatibility with:
> 
>     * mark method **getRegionHtmlSelect** as deprecated
> 
>     * create new method with parameters (e.g. **getRegionSelect** or **getRegionHtml**)
> 
>     * add `@see` section and call newly created method in **getRegionHtmlSelect**

As per this solution:

How can I pass parameter to **getRegionSelect** if it's parent function (e.g. **getRegionHtmlSelect** ) don't have parameter.

and as all parameters are optional it will work if we don't pass any parameter.

swnsma: The idea that for everyone who used `getRegionHtmlSelect` before, for them nothing will be changed after they will upgrade to a new version of Magento 2 and receive your changes.
Their project will continue working without any troubles and they will be able to refactor their code and use new method `getRegionSelect` which will be introduced by you.

I agree that in general adding optional parameters will work in case of extending of the current class, but this is a violation to [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/).
So, I need to ask you to proceed with the new method and deprecation of the old one.

nikunjskd20: I have done the required changes and pushed them to 2.3-develop branch.

app/code/Magento/Directory/Block/Data.php
```diff
@@ -168,11 +168,12 @@ public function getRegionCollection()
...
+    
     /**
-     * Returns region html select
-     *
      * @return string
+     * 
+     * @deprecated
+     * @see getRegionSelect() method for more configuration
      */
     public function getRegionHtmlSelect()
```
swnsma, line 181: Could you please replace the body of the method with 
`return $this->getRegionSelect();`

app/code/Magento/Directory/Block/Data.php
```diff
@@ -196,7 +197,49 @@ public function getRegionHtmlSelect()
...
+    /**
+     * Returns region html select
+     *
+     * @param null|string $value
+     * @param string $name
+     * @param string $id
+     * @param string $title
+     * @return string
+     */
+    public function getRegionSelect($value = null, $name = 'region', $id = 'state', $title = 'State/Province')
```
swnsma: Could you please add typehints for method arguments and return typehint ?

app/code/Magento/Directory/Block/Data.php
```diff
@@ -168,11 +168,12 @@ public function getRegionCollection()
         }
         return $collection;
     }
-
+
```
swnsma: Please remove these redundant spaces.

app/code/Magento/Directory/Block/Data.php
```diff
@@ -168,11 +168,12 @@ public function getRegionCollection()
...
     }
-
+    
     /**
-     * Returns region html select
-     *
      * @return string
+     * 
+     * @deprecated
+     * @see getRegionSelect() method for more configuration
```
swnsma: Please, `* @see getRegionSelect() method for more configuration` -> `     * @see getRegionSelect() method for more configurations `

swnsma (changes requested):
Hi @nikunjskd20,

Thank you for your changes!

We almost finish, but beforoe we here there are few thing which need to be done.
So, could you please:
- add `declare(strict_types=1);` to the beginning of the file;
- check my comments about minor changes request;
- check and fix failed tests;

app/code/Magento/Directory/Block/Data.php
nikunjskd20: sure

app/code/Magento/Directory/Block/Data.php
nikunjskd20: Is it necessary to add `declare(strict_types=1); ` at the beginning of file?

swnsma: According to [Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html#1-basic-programming-principles) `#1.3.1`, this is *should* be done.
But as we are introducing a new method, it is better to use strict types, as is described in `#1.3` and where PHP is aiming ([PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md)).

swnsma: I think this is better to use `?int $value = null,` instead of `string $value = null,` to avoid double conversion to int.

nikunjskd20: yeah, sure.

swnsma (approved):

magento-engcom-team:
[ENGCOM-5141](https://jira.corp.magento.com/browse/ENGCOM-5141) has been created to process this Pull Request

magento-engcom-team:
@nikunjskd20 thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

nikunjskd20:
Hi @swnsma, do I need to do anything or this pull request will merge automatically?

swnsma:
Hi @nikunjskd20.
This PR will be checked with Magento QA and if everything is OK it will be merged by Magento.
But if there will be any problems, I will let you know :)

VasylShvorak:
:heavy_check_mark: QA passed

p-bystritsky:
@nikunjskd20, @swnsma please reduce the number of commits.

PR #22835, opened 2019-05-11
issue #22832 fixed Should not hide element in page as In customer account register page a element hide from css, But we should properly remove from html

cedarvinda:
issue #22832 fixed Should not hide element in page as In customer account register page a element hide from css, But we should properly remove from html

### Description (*)
Should not hide element in page as In customer account register page a element hide from css, But we should properly remove from html


Karlasa:
@cedarvinda 
Will close this PR as it should not be removed as css style is part of Luma Theme and we should not remove functionality that is part of Blank Theme and is visible there. 

Cheers,

Issue #22833, opened 2019-05-11
Short-term admin accounts


piotrekkaminski:
### Description (*)
Merchants often create admin accounts for allowing extension vendors to support their extensions remotely. Such admin accounts often have simple, easy to guess passwords - and they are never removed even when no longer used.

### Expected behavior (*)

Ability to define how long given account is enabled. The account will be disabled after the specified time passes. There should be ability to reopen the account for another time frame or make it permanent. Additionally, system should propose high complexity password during account creation to encourage using strong passwords.

### Benefits

Given that majority of attacks on Magento installations include getting admin access and installing JavaScript malware, all features allowing to limit the admin accounts could help protect systems.

lfolco:
Question about password complexity: should that be included in this issue or should it be moved into a separate issue? The two issues are related but are distinct functionality.

PR #22837, opened 2019-05-11
Short-term admin accounts #22833

lfolco:
### Description (*)
Provides the ability to set an expiration date on admin user accounts, after which the account will be disabled.

1. magento/magento2#22833: Short-term admin accounts

#### Manual testing scenario 1
1. Create an admin user
2. Set their expiration date to 5 minutes from now
3. Login as that user
5. Wait 5 minutes
4. Navigate to another page
5. User should be logged out

#### Manual testing scenario 2
1. Create an admin user
2. Set their expiration date to 5 minutes from now
3. In 5 minutes, try to login as that user
4. Login should fail with message: "The account sign-in was incorrect or your account is disabled temporarily. Please wait and try again later"

#### Manual testing scenario 3
1. Create an admin user
2. Set their expiration date to 5 minutes from now
3. In 5 minutes, run the Magento cron manually
4. Check the admin user: the account should be disabled and the expiration date should now be empty


sidolov:

magento-engcom-team:
Hi @sidolov. Thank you for your request. I'm working on Magento instance for you

magento-engcom-team:
Hi @sidolov, here is your [new Magento instance](https://pr-22837.instances.magento-community.engineering).
Admin access: [https://pr-22837.instances.magento-community.engineering/admin](https://pr-22837.instances.magento-community.engineering/admin)

phoenix128:
Hello @lfolco ,
are you still on this PR?

lfolco:
Yes, I'm running into issues with my tests failing here but not locally. Could use some help!

lfolco:
I know that this PR has issues with the semantic version checker build, but it's also failing on the static tests build and the functional tests build for stuff that I haven't changed.

piotrekkaminski:
@szurek ^^

piotrekkaminski:
@okorshenko can someone from the team help here? it's a very useful feature for us

ishakhsuvarov:
@lfolco This can happen with the Static tests, as they are executed only for the changed files. In some situations you may be editing a legacy class, which haven't had an update in ages. Generally, the best practice is to take care of the static test failures even if it's not your exact changes which caused the failure, this is what most people do.

I'd say that the bigger issue in this case is backward incompatible changes to the API interface, caused by introducing additional methods. This is a kind of change which cannot be accepted, as it may affect other already existing customizations. I'm not entirely sure what would be the best way to refactor these methods, but you may check if it is something that can be extracted into a separate entity, which is solely responsible for the admin locking dates. 

Please let us know if you have any specific questions on this or need any help!

lfolco:
Okay I'll go back and see what I can refactor to get it backward-compatible.

app/code/Magento/Security/Model/AdminSessionsManager.php
```diff
@@ -27,6 +27,11 @@ class AdminSessionsManager
      */
     const LOGOUT_REASON_USER_LOCKED = 10;
 
+    /**
+     * User has been logged out due to an expired user account
+     */
+    const LOGOUT_REASON_USER_EXPIRED = 11;
```
AlexMaxHorkun: that's a backward incompatible change, you could make it private or better just given them AdminSessionInfo::LOGGED_OUT. That way we wouldn't have to use a new status ID which is better since some extensions may have used it already. Also it doesn't hurt to obfuscate account lock reason

app/code/Magento/User/Block/User/Edit/Tab/Main.php
```diff
@@ -170,6 +170,19 @@ protected function _prepareForm()
             );
         }
 
+        $baseFieldset->addField(
```
AlexMaxHorkun: You are treating this value as a GMT date, please convert it from Magento configured timezone into GMT before saving it and vise versa when showing it on the user form

app/code/Magento/Security/Model/AdminSessionsManager.php
```diff
@@ -353,4 +373,21 @@ private function getIntervalBetweenConsecutiveProlongs()
             )
         );
     }
+
+    /**
+     * Check if the current user is expired and, if so, revoke their admin token.
+     */
+    private function revokeExpiredAdminUser()
+    {
+        $expiresAt = $this->dateTime->gmtTimestamp($this->authSession->getUser()->getExpiresAt());
```
AlexMaxHorkun: To keep the expiration logic in one place please create a UserExpirationManager with expireUsers(?array $users = null):void which will be used here and inside the cron job. Here we will pass a certain user with cron job not passing any arguments and expiring all users.

app/code/Magento/Security/Model/AdminSessionsManager.php
```diff
@@ -144,6 +153,11 @@ public function processProlong()
             $this->getCurrentSession()->save();
         }
 
+        // todo: don't necessarily have a user here
```
AlexMaxHorkun: This function is invoked in the plugin for Magento\Backend\Model\Auth::login() - it's supposed to return an AuthenticationException with a message detailing what went wrong, not just quietly go on

app/code/Magento/User/Cron/DisableExpiredUsers.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->securityConfig = $securityConfig;
+    }
+
+    /**
+     * Disable all expired user accounts and invalidate their sessions.
+     */
+    public function execute()
+    {
+        /** @var \Magento\User\Model\ResourceModel\User\Collection $users */
+        $users = $this->userCollectionFactory->create()
```
AlexMaxHorkun: Make it use UserExpirationManager so we'd have expiration logic in one place

dev/tests/integration/testsuite/Magento/User/Model/ResourceModel/User/CollectionTest.php
```diff
@@ -0,0 +1,43 @@
...
+    protected function setUp()
+    {
+        $this->collection = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()->create(
+            \Magento\User\Model\ResourceModel\User\Collection::class
+        );
+    }
+
+    public function testSelectQueryInitialized()
+    {
+        static::assertContains(
```
AlexMaxHorkun: this is whitebox testing that is not supposed to be present inan integration test

dev/tests/integration/testsuite/Magento/User/Cron/DisableExpiredUsersTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $user = Bootstrap::getObjectManager()->create(\Magento\User\Model\User::class);
+        $user->loadByUsername($adminUserNameFromFixture);
+
+        /** @var \Magento\Integration\Model\Oauth\Token $tokenModel */
+        $tokenModel = Bootstrap::getObjectManager()->get(\Magento\Integration\Model\Oauth\Token::class);
+        $token = $tokenModel->loadByAdminId($user->getId());
+
+        static::assertEquals(0, $user->getIsActive());
+        static::assertNull($user->getExpiresAt());
+        static::assertEquals(null, $token->getId());
```
AlexMaxHorkun: please test that you cannot create a token for an expired user as well

app/code/Magento/User/Api/Data/UserInterface.php
```diff
@@ -162,4 +162,19 @@ public function getInterfaceLocale();
      * @return $this
      */
     public function setInterfaceLocale($interfaceLocale);
+
+    /**
+     * Get user expiration date.
+     *
+     * @return string
+     */
+    public function getExpiresAt();
```
AlexMaxHorkun: There are couple of problems with this approach:
1. It's backward incompatible
2. Seems a bit weird to create this property in User module and only using it in Security module, these modules are interconnected enough

I recommend creating a new table with user_id as primary key and expires_at columns and do it within Security module then use it inside UserExpirationManager

AlexMaxHorkun (changes requested):
It's an important feature and I've added some comments I think would help to merge it

app/code/Magento/Security/Model/AdminSessionsManager.php
lfolco: That works; I had originally looked at `AdminSessionInfo::LOGGED_OUT` but assumed that admin users would want to know why they can't login; had't thought about the security implications of hiding that reason.

app/code/Magento/User/Api/Data/UserInterface.php
lfolco: Should I put UserExpirationManager in the Security module as well?

app/code/Magento/Security/Model/AdminSessionsManager.php
lfolco: Do you mean `\Magento\Security\Model\AdminSessionsManager::processLogin` or `\Magento\Security\Model\AdminSessionsManager::processProlong`? From what I can see, `processProlong` is called in `\Magento\Backend\App\Action\Plugin\Authentication::aroundDispatch`

lfolco: Are you talking about the potential exception thrown when saving the user? Or should I throw an `AuthenticationException` when expiring the user, indicating the reason that they cannot login?

lfolco: Also, assume you're talking about `\Magento\Security\Model\AdminSessionsManager::processLogin` not `\Magento\Security\Model\AdminSessionsManager::processProlong`.

AlexMaxHorkun: Correct, I am talking about processLogin and yes, I think the AuthenticationException must be thrown after a user is expired

lfolco:
I'm running into an issue around sessions in integration tests. I am using a data fixture to create admin users, and logging them in and out to run tests.

The users are setup here: https://github.com/lfolco/magento2/blob/project_pepe/dev/tests/integration/testsuite/Magento/Security/_files/expired_users.php
```
$userModelNotExpired = $objectManager->create(\Magento\User\Model\User::class);
$userModelNotExpired->setFirstName("John")
    ->setLastName("Doe")
    ->setUserName('adminUserNotExpired')
    ->setPassword(\Magento\TestFramework\Bootstrap::ADMIN_PASSWORD)
    ->setEmail('adminUserNotExpired@example.com')
    ->setRoleType('G')
    ->setResourceId('Magento_Adminhtml::all')
    ->setPrivileges("")
    ->setAssertId(0)
    ->setRoleId(1)
    ->setPermission('allow')
    ->setIsActive(1)
    ->save();
```

In my test, I am logging the user in, running tests, then logging them out (https://github.com/lfolco/magento2/blob/project_pepe/dev/tests/integration/testsuite/Magento/Security/Model/UserExpirationManagerTest.php#L81):

But, I am getting a foreign key constraint error when I run several of these tests in the same test class (i.e., _test1_ logs the same user in and then out, _test2_ does the same thing, etc.).

What's happening is that when `\Magento\Security\Model\AdminSessionsManager::createNewSession` is called, the user returned from `$this->authSession->getUser()` is the user from the previous test.

```
protected function createNewSession()
    {
        $this->adminSessionInfoFactory
            ->create()
            ->setData(
                [
                    'session_id' => $this->authSession->getSessionId(),
                    'user_id' => $this->authSession->getUser()->getId(),
                    'ip' => $this->remoteAddress->getRemoteAddress(),
                    'status' => AdminSessionInfo::LOGGED_IN
                ]
            )->save();

        return $this;
    }
```

To make sure it isn't my code, I switched to 2.3-develop and modified a different test that does something similar. `\Magento\Security\Model\AdminSessionsManagerTest` also logs users in and out, but it uses the default admin user. I updated all tests that login a user to load a user from a data fixture and ran the tests:

```
  /**
     * Test if the admin session is created in database
     *
     * @magentoDataFixture Magento/User/_files/user_with_role.php
     */
    public function testIsAdminSessionIsCreated()
    {
        $this->auth->login(
            'adminUser',
            \Magento\TestFramework\Bootstrap::ADMIN_PASSWORD
        );
        $sessionId = $this->authSession->getSessionId();
        $this->adminSessionInfo->load($sessionId, 'session_id');
        $this->assertGreaterThanOrEqual(1, (int)$this->adminSessionInfo->getId());
        $this->auth->logout();
    }
```

And the same error occurs:
```Zend_Db_Statement_Exception : SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`m2contrib_tests`.`admin_user_session`, CONSTRAINT `ADMIN_USER_SESSION_USER_ID_ADMIN_USER_USER_ID` FOREIGN KEY (`user_id`) REFERENCES `admin_user` (`user_id`) ON DELETE CASCADE), query was: INSERT INTO `admin_user_session` (`session_id`, `user_id`, `status`, `ip`) VALUES (?, ?, ?, ?)```

Not sure if this is a bug, or an issue in how I'm creating users for tests (I followed the example in `Magento/User/_files/user_with_role.php`).

**Comment by AlexMaxHorkun** [comment]
@lfolco 
seems like session related classes like AdminSessionManager and other have trouble managing state. Do you recreate them each time in you setUp method like it's done in AdminSessionsManagerTest ?

**Comment by lfolco** [comment]
Yes, I'm creating them each time. If I use the default user (`\Magento\TestFramework\Bootstrap::ADMIN_NAME`), then my session status for the second test is `\Magento\Security\Model\AdminSessionInfo::LOGGED_OUT_BY_LOGIN` instead of `\Magento\Security\Model\AdminSessionInfo::LOGGED_OUT`, so calling `$this->auth->logout();` doesn't seem to actually log the user out?

**Comment by AlexMaxHorkun** [comment]
@lfolco I think the exception you're getting doesn't expose the original issue - you are not successfully creating an admin user in your fixture because of the unique constraints. At least that's what I assume. You do not have a rollback for your expired_users.php fixture so you are not deleting the users created between tests which makes the second test to attempt to recreate the same user which fails. Try to add @magentoDbIsolation enabled to your test

**Comment by lfolco** [comment]
Same issue (I thought that `@magentoDataFixture` automatically enabled db isolation?).

I stepped through it with a debugger and verified that it's loading the users in the database each time, and on the second test it's loading the previous user ID.

There definitely seems to be something happening around `\Magento\Backend\Model\Auth::logout`: If you update the `\Magento\Security\Model\AdminSessionsManagerTest::testIsAdminSessionIsCreated` test to add a check for the logged out status, it returns `\Magento\Security\Model\AdminSessionInfo::LOGGED_OUT_BY_LOGIN` instead of `\Magento\Security\Model\AdminSessionInfo::LOGGED_OUT`:

```
    public function testIsAdminSessionIsCreated()
    {
        $this->auth->login(
            \Magento\TestFramework\Bootstrap::ADMIN_NAME,
            \Magento\TestFramework\Bootstrap::ADMIN_PASSWORD
        );
        $sessionId = $this->authSession->getSessionId();
        $this->adminSessionInfo->load($sessionId, 'session_id');
        $this->assertGreaterThanOrEqual(1, (int)$this->adminSessionInfo->getId());
        $this->auth->logout();
// Add these lines
        $this->adminSessionInfo->load($sessionId, 'session_id');
        $this->assertEquals($this->adminSessionInfo->getStatus(), AdminSessionInfo::LOGGED_OUT);
    }
```

This test occurs right after `\Magento\Security\Model\AdminSessionsManagerTest::testProcessLogout`, which does a login and a logout.

What seems to be happening is the previous user is only logged out fully when you log them in again.

**Comment by AlexMaxHorkun** [comment]
@lfolco Found the problem - AdminSessionsManager has the old currentSession when you login the second time which makes it give wrong statuses to wrong sessions. The manager is not being recreated inside the test since it's already initialized as a plugin for Backend session object and that's why the old session remains.

**Comment by lfolco** [comment]
Still looking into this, but in the meantime the checks seem to be stuck? 
image

**Comment by ishakhsuvarov** [comment]
@magento run all tests

**Comment by ishakhsuvarov** [comment]
@lfolco Not sure why it was stuck that time. Hope this time it would work

**Comment by piotrekkaminski** [comment]
@lfolco  @avstudnitz the tests were unblocked but failed. 
HealthIndex at least shows a problem, haven't reviewed all of them in detail.
can you also review the changes request and approve/dismiss as needed?

**Comment by lfolco** [comment]
Yep, working on those changes. Thanks for checking!

**Comment by lfolco** [comment]
I'm still getting failed tests, but they seem to not have anything to do with my code per se. Any pointers on how to get them to pass?

**Comment by lfolco** [comment]
Also would love some guidance on getting MFTF tests setup for this.

**Comment by josefbehr** [comment]
> I'm still getting failed tests, but they seem to not have anything to do with my code per se. Any pointers on how to get them to pass?
@sidolov ?

> Also would love some guidance on getting MFTF tests setup for this.
I'd try the MFTF slack channel and official devdocs.

**Comment by lfolco** [comment]
I've finished adding tests and everything is passing except for the Magento Health Index:

```
Overall Status | Difference with mainline detected
Magento_Security | Failure (79%)

Details
Current file(s) contain protected modifier for method or property.
(phpProtectedModifier)Magento/Security/Model/ResourceModel/UserExpiration.phpMagento/Security/Model/ResourceModel/UserExpiration/Collection.phpMagento/Security/Model/UserExpiration.phpFile has calls static method. 
(phpStaticMethodCalls)Magento/Security/Model/UserExpiration/Validator.php
```

Not sure how to resolve this. What is _difference with mainline detected_ mean? And for the other issues, I'm using core Magento methods here and not sure how to get past this.

Otherwise this should be ready to go.

File: app/code/Magento/User/Test/Mftf/ActionGroup/AdminFillNewUserFormRequiredFieldsActionGroup.xml [hunk 9]
```diff
@@ -13,7 +13,7 @@
             <description>Fills in the provided User details on the New User creation page.</description>
         </annotations>
         <arguments>
-            <argument name="user" type="entity"/>
+            <argument name="user" defaultValue="NewAdminUser"/>
```
AlexMaxHorkun, line 16: I'm not that experienced with MFTF but why remove "entity" attr?

app/code/Magento/User/etc/db_schema_whitelist.json
```diff
@@ -42,4 +42,4 @@
             "ADMIN_PASSWORDS_USER_ID_ADMIN_USER_USER_ID": true
         }
     }
-}
\ No newline at end of file
+}
```
AlexMaxHorkun: can be reverted

app/code/Magento/User/Block/User/Edit/Tab/Main.php
```diff
@@ -170,6 +170,25 @@ protected function _prepareForm()
...
+            \IntlDateFormatter::MEDIUM
+        );
+        $timeFormat = $this->_localeDate->getTimeFormat(
+            \IntlDateFormatter::MEDIUM
+        );
+        $baseFieldset->addField(
+            'expires_at',
+            'date',
+            [
+                'name' => 'expires_at',
```
AlexMaxHorkun: how does a block of User module knows about a field from Security module?

lfolco:
Also please take a look at my MFTF tests, it's a first attempt there.

app/code/Magento/Security/Model/UserExpirationManager.php
```diff
@@ -0,0 +1,129 @@
...
+            $currentSessions = $this->adminSessionInfoCollectionFactory->create()
+                ->addFieldToFilter('user_id', ['in' => $expiredRecords->getAllIds()])
+                ->filterExpiredSessions($this->securityConfig->getAdminSessionLifetime());
+            /** @var \Magento\Security\Model\AdminSessionInfo $currentSession */
+            $currentSessions->setDataToAll('status', \Magento\Security\Model\AdminSessionInfo::LOGGED_OUT)
+                ->save();
+        }
+
+        // delete expired records
+        $expiredRecordIds = $expiredRecords->getAllIds();
```
AlexMaxHorkun, line 124: I'd swap deleting expired records and deactivating users - since there's no transaction in case execution of this method would be terminated right after deactivating all users leaving their expired records in DB would be ok, but if we delete expired records 1st and then script execution is terminated expired users would not get deactivated in consecutive cron command run

app/code/Magento/Security/Model/UserExpirationManager.php
```diff
@@ -0,0 +1,129 @@
...
+        $users->setDataToAll('is_active', 0)->save();
+    }
+
+    /**
+     * Check if the given user is expired.
+     *
+     * @param \Magento\User\Model\User $user
+     * @return bool
+     */
+    public function userIsExpired(\Magento\User\Model\User $user): bool
```
AlexMaxHorkun: I think isUserExpired would be better

app/code/Magento/Security/Model/UserExpiration.php
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Security\Model;
+
+/**
+ * Admin User Expiration model.
+ * @method string getExpiresAt()
```
AlexMaxHorkun: can you please add getUserId() and setUserId() ?

app/code/Magento/Security/Model/ResourceModel/UserExpiration/Collection.php
```diff
@@ -0,0 +1,78 @@
...
+        );
+    }
+
+    /**
+     * Filter for expired, active users.
+     *
+     * @param string $now
+     * @return $this
+     */
+    public function addActiveExpiredUsersFilter($now = null): Collection
```
AlexMaxHorkun: can you please add ?string type hint. Also wouldn't hurt to explain what kind of date (timezone, format) is expected as the parameter

app/code/Magento/Security/Model/ResourceModel/UserExpiration/Collection.php
```diff
@@ -0,0 +1,78 @@
...
+        return $this;
+    }
+
+    /**
+     * Filter collection by user id.
+     *
+     * @param int[] $userIds
+     * @return Collection
+     */
+    public function addUserIdsFilter($userIds = []): Collection
```
AlexMaxHorkun: please add array typehint

app/code/Magento/Security/Model/ResourceModel/UserExpiration.php
```diff
@@ -0,0 +1,88 @@
...
+     * UserExpiration constructor.
+     *
+     * @param \Magento\Framework\Model\ResourceModel\Db\Context $context
+     * @param \Magento\Framework\Stdlib\DateTime\TimezoneInterface $timezone
+     * @param string $connectionName
+     */
+    public function __construct(
+        \Magento\Framework\Model\ResourceModel\Db\Context $context,
+        \Magento\Framework\Stdlib\DateTime\TimezoneInterface $timezone,
+        $connectionName = null
```
AlexMaxHorkun: please add ?string typehint

app/code/Magento/Security/Model/UserExpirationManager.php
```diff
@@ -0,0 +1,129 @@
...
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Security\Model;
+
+use Magento\Security\Model\ResourceModel\UserExpiration\Collection as ExpiredUsersCollection;
+
+/**
+ * Class to handle admin user expirations.
```
AlexMaxHorkun: Can you please add proper description for this class? I think this would be a good place to describe what "User expiration" is and how it works

app/code/Magento/Security/Model/UserExpirationManager.php
AlexMaxHorkun: Also why do we need the whole model? ID would be enough as the parameter

app/code/Magento/Security/Observer/AdminUserAuthenticateBefore.php
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @param UserExpirationManager $userExpirationManager
+     * @param \Magento\User\Model\User $user
+     */
+    public function __construct(
+        \Magento\Security\Model\UserExpirationManager $userExpirationManager,
+        \Magento\User\Model\User $user
+    ) {
+        $this->userExpirationManager = $userExpirationManager;
+        $this->user = $user;
```
AlexMaxHorkun: Please use the UserFactory instead of reusing one model for different sets of data, User is not meant to be a singleton

dev/tests/integration/testsuite/Magento/Security/Model/UserExpirationManagerTest.php
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * @var \Magento\Security\Model\UserExpirationManager
+     */
+    private $userExpirationManager;
+
+    protected function setUp()
+    {
+        $this->objectManager = \Magento\TestFramework\Helper\Bootstrap::getObjectManager();
+        $this->objectManager->get(\Magento\Framework\Config\ScopeInterface::class)
+            ->setCurrentScope(\Magento\Backend\App\Area\FrontNameResolver::AREA_CODE);
```
AlexMaxHorkun: official way to emulate an area in an integration test is to use magentoAppArea annotation

AlexMaxHorkun (changes requested):

app/code/Magento/User/Test/Mftf/ActionGroup/AdminFillNewUserFormRequiredFieldsActionGroup.xml
lfolco: Because it's `entity` be default, unless you specify otherwise.

app/code/Magento/Security/Model/ResourceModel/UserExpiration/Collection.php
lfolco: Actually that parameter is left over from a refactoring; there are no methods passing in the time so I'll remove it and default to the current time.

app/code/Magento/User/Block/User/Edit/Tab/Main.php
lfolco: I could use an observer or an around plugin to get that field on the form, which would be better? The observer would listen to the `adminhtml_block_html_before` event and check the block type; the around plugin would wrap around `getFormHtml`.

AlexMaxHorkun: I think event listener is better

AlexMaxHorkun (approved):
Everything looks good to me

AlexMaxHorkun:
Magento Health Index fails on false-positives

sidolov:
Hi @lfolco it's incredible work! Thank you for your involvement! We will proceed with the delivery as soon as possible!

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

lfolco:
Any chance I could get the award label for the MFTF tests? :)

engcom-Alfa:
Hi @lfolco 

During testing we faced with some issues

_**Case 1**_

**_Problem:_** Problem with validation when entering incorrect data

**Manual testing scenario:**
1. Go to System->Permissions->All Users->Add New User;
2. Fill all required fields;
3. In field "Expiration Date" enter "test" value for example;
4. Click on "Save User" button;

**Actual Result:**
The admin homepage was opened without any notice and in the logs we received an error message
after

**_Case 2_**

It would be nice to add a phrase "**Expiration Date**" to the translation file

@lfolco Could you take a look, please?

Thanks!

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

engcom-Alfa:
Hi @sidolov,
could you review the last commit?

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

engcom-Alfa:
Hi @lfolco 

During testing, we faced the issue

**_Problem:_** After saving the user, the value of the "Expiration Date" field is not displayed.

**Manual testing scenario:**

1. Go to System->Permissions->All Users->Add New User;
2. Fill all required fields;
3. In field "Expiration Date" enter valid value;
4. Click on "Save User" button;

**Actual Result:**
Field "Expiration Date" is empty
after2

but in DB the value is present
after3

@lfolco Could you take a look?

Thanks!

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

lfolco:
@engcom-Alfa that issue is fixed.

engcom-Alfa:
Hi @sidolov . Could you review the last commit?

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

engcom-Alfa:
Hi @lfolco 

During testing, we came across another problem.

**_Problem:_** Updating the user status in the grid occurs only after disabled user trying to log in

**Steps to reproduce:**

1. Log in to Admin;
2. Go to System->Permissions->User Roles->Create New Role ;
3. Create a new role ("test" for example) with some resources;
4. Go to System->Permissions->All Users->Add New User;
5. Fill all required fields and select created before "test" user role ;
6. Set the expiration date to 5 minutes from now;
7. Save ("tester" for example) and Login as that user ;
8. Wait 5 minutes;
9. Navigate to another page;
10. User logged out;
11. Log in to the system as "admin" main user;
12. Go to System->Permissions->All Users;

**Actual Result:** :x: user ("tester") status still remains "Active"
after4
13. Sign out from the admin;
14. Try to log in as "tester" user;
15. We received the error message : "The account sign-in was incorrect or your account is disabled temporarily. Please wait and try again later."
16. Log in to the system as "admin" main user again;
17. Go to System->Permissions->All Users;

**Actual Result:** :heavy_check_mark: user ("tester") status is "Inactive"
after5

@lfolco Could you take a look?

Thanks!

sidolov:
@lfolco looks like Magento has indexer for customer grid that should be updated

lfolco:
@engcom-Alfa This is fixed. @sidolov Not sure how the customer grid is affected with this, can you elaborate?

sidolov:
@lfolco nevermind, I thought that bug described by @engcom-Alfa caused by outdated grid index.

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

engcom-Alfa:
Hi @lfolco

During testing, we came across other problems.

### Case 1

**_Problem:_**  Updating the status of a user in the grid does not happen if the user has never been logged in

**Manual testing scenario:**
1. Log in to Admin;
2.Go to System->Permissions->All Users->Add New User;
3. Fill all required fields and select user role ;
4. Set the expiration date to 5 minutes from now;
5. Save ("tester" for example) user;
6. Wait 5 minutes;
7. Go to System->Permissions->All Users;

**Actual Result:** :heavy_multiplication_x:  user ("tester") status still remains "Active"
after8

8. Sign out from the admin;
9. Try to log in as "tester" user;
10. We received the error message: "The account sign-in was incorrect or your account is disabled temporarily. Please wait and try again later."
11. Log in to the system as "admin" main user again;
12. Go to System->Permissions->All Users;

**Actual Result:** :heavy_check_mark: user ("tester") status is "Inactive"
after9


### Case 2

_**Problem:**_ Cron job `security_deactivate_expired_users` finished with the error message

**Manual testing scenario:**
1. Create an admin user;
2. Set their expiration date to 5 minutes from now;
3. In 5 minutes, run the Magento cron manually;

**Actual Result:**
after6

In **cron.log** file:
after7

@lfolco Could you take a look?

Thanks!

sidolov (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed

slavvka:
Hey @lfolco, the BICs are approved by the architects but there arose another concern:
> Specifically, Magento\User\Model\User class is used as a DTO. Ideally, we need to have new DTO (or no DTO) for expiration property.

Could you please consider changing your PR to satisfy their wishes?

slavvka:
@lfolco Also please take a look on the failing tests

lfolco:
The functional test failures have nothing to do with the code changes for this PR; what's the protocol for that scenario?

And it looks like there are different static test standards for B2B than for everything else; which ruleset should I be using? (Note that I do not have access to B2B or EE).

slavvka:
I restarted Functional tests as it seems to be random failure. But please look the the rest of them and also to the comment from our architects

slavvka:
@lfolco Are you going to proceed with this PR?

lfolco:
Yes I am, sorry but I've been slammed with work the last week or so. I'll be working on it this week.

lfolco:
Can I get more information on this:

> Specifically, Magento\User\Model\User class is used as a DTO. Ideally, we need to have new DTO (or no DTO) for expiration property.

app/code/Magento/Security/Model/UserExpiration.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Magento\Security\Model;
+
+/**
+ * Admin User Expiration model.
+ * @method string getUserId()
+ * @method \Magento\Security\Model\UserExpiration setUserId($userId)
+ * @method string getExpiresAt()
+ * @method \Magento\Security\Model\UserExpiration setExpiresAt($value)
+ */
+class UserExpiration extends \Magento\Framework\Model\AbstractModel
```
kokoc: This model has no interface. The fields are defined in phpdoc. Suggest to introduce and use the interface for expiration model

kokoc: @lfolco

lfolco: Do you mean create an interface class or define the methods in `UserExpiration`?

kokoc: Just a regular php interface with implementation inside UserExpiration class.

E.g. 
```
interface UserExpirationInterface {
    public function getExpiresAt(): string, date or timestamp type
}

class UserExpiration implements UserExpirationInterface {
public function getExpiresAt() {
 return 'something';
}


//Some other place

class OtherClass {
    public function save(UserExpirationInterface $expiration) {
         $expiration->getExpiresAt();
    }

}
```

lfolco: How would I use that in `\Magento\Security\Observer\AfterAdminUserSave`, for example? This relies on the abstract model/resource model:

```
public function execute(Observer $observer)
    {
        /* @var $user \Magento\User\Model\User */
        $user = $observer->getEvent()->getObject();
        if ($user->getId()) {
            $expiresAt = $user->getExpiresAt();
            /** @var \Magento\Security\Model\UserExpiration $userExpiration */
            $userExpiration = $this->userExpirationFactory->create();
            $this->userExpirationResource->load($userExpiration, $user->getId());

            if (empty($expiresAt)) {
                // delete it if the admin user clears the field
                if ($userExpiration->getId()) {
                    $this->userExpirationResource->delete($userExpiration);
                }
            } else {
                if (!$userExpiration->getId()) {
                    $userExpiration->setId($user->getId());
                }
                $userExpiration->setExpiresAt($expiresAt);
                $this->userExpirationResource->save($userExpiration);
            }
        }
    }
```
Unless I create a repository as well? Or should I not use the new interface in the scenario where I need to load or save the UserExpiration object directly?

lfolco: So I guess I don't need to do this anymore? I was only asking for clarification, I would have preferred to implement it. I never heard back about my question.

VladimirZaets: Hi @lfolco. Sorry for the delay and this accident, I start working on this PR and really want to delivery it. 
That is ok that you extend \Magento\Framework\Model\AbstractModel, all that new need it create the interface for this class. We should have a strict contract, currently, contract provided just in phpdoc block, it isn't strict and we have just one option to provide strict contract - it's an interface

lfolco:
@magento run all tests

VladimirZaets:
@magento run all tests

VladimirZaets:
@magento run all tests

VladimirZaets:
@magento run all tests

VladimirZaets (approved):

magento-engcom-team:
[ENGCOM-6014](https://jira.corp.magento.com/browse/ENGCOM-6014) has been created to process this Pull Request

PR #22898, opened 2019-05-14
Fixed #6955 Admin: While product which has images deleted. It's not deleting from Folder

### LINKED ISSUE #6955: Admin: While product which has images deleted. It's not deleting from Folder.
### Preconditions

1. I'm using Magento 2 CE Version 2.1.1
### Steps to reproduce

1.  Go to Admin -> Products -> Catalog
2.  When Product has image & if you delete from listing.
3.  It will not delete product images from Folder.
### Expected result

1.  When product is deleted, It should delete all product images from Folder as well.
### Actual result

1. Currently it's not deleting product Images.
2. Due to Image Not Deleted, there will be lots of space on server will be used.
3. This is the only case for Products, May be we have same issue of other Admin module as well where image is uploaded.
4. Need to check whole Admin where image is related while Delete.

amitvishvakarma:
Fixed #6955 Admin: While product which has images deleted. It's not deleting from Folder
### Preconditions

1. I'm using Magento 2 CE Version 2.1.1
### Steps to reproduce

1.  Go to Admin -> Products -> Catalog
2.  When Product has image & if you delete from listing.
3.  It will not delete product images from Folder.
### Expected result

1.  When product is deleted, It should delete all product images from Folder as well.
### Actual result

1. Currently it's not deleting product Images.
2. Due to Image Not Deleted, there will be lots of space on server will be used.
3. This is the only case for Products, May be we have same issue of other Admin module as well where image is uploaded.
4. Need to check whole Admin where image is related while Delete.

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/MassDelete.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/MassDelete.php
@@ -12,7 +12,7 @@ use Magento\Backend\App\Action\Context;
 use Magento\Ui\Component\MassAction\Filter;
 use Magento\Catalog\Model\ResourceModel\Product\CollectionFactory;
 use Magento\Catalog\Api\ProductRepositoryInterface;
-
+use Magento\Catalog\Model\ProductFactory;
 class MassDelete extends \Magento\Catalog\Controller\Adminhtml\Product implements HttpPostActionInterface
 {
     /**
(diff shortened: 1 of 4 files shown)


orlangur (changes requested):
Please implement logic as a plugin for `$this->productRepository->delete($product);` instead.

Is there a smarter way to remove images maybe?

amitvishvakarma:
> Please implement logic as a plugin for `$this->productRepository->delete($product);` instead.
> 
> Is there a smarter way to remove images maybe?

I had reverted previous changes and used event "catalog_product_delete_after_done" now so media images will be deleted once product is deleted.

sidolov:
Hi @amitvishvakarma , please, implement logic in the plugin for productRepository as @orlangur mentioned and revert productRepository usage instead of model.

sidolov:
@amitvishvakarma , I am closing this PR now due to inactivity.

PR #22920, opened 2019-05-16
Mview Indexers getList should return integer values of id's and not strings

mhodge13:
### Description (*)
When using Maria DB I have noticed that when the `indexer_update_all_views` cron job runs that some of the indexer queries have a mix of entity_id values that are strings and some that are integers.  When running a describe on the query I have confirmed that Maria DB will look at every row in the table and not use the entity_id values that are in the WHERE clause.  This does not seem to be an issue when using MySql but is in an issue when using Maria DB.

When reviewing the queries for the `catalog_product_attribute` indexer I saw a mix of string and integer id values.

1. Run a describe on the query, notice mix of string and integer
    DESC DELETE FROM `catalog_product_index_eav` WHERE (entity_id IN('2764369', 2766213));
    rows returned shows all rows in table: 9,133,650 and Extra shows "Using Where"

2. Run the same query but change the entity_ids to be integers
    DESC DELETE FROM `catalog_product_index_eav` WHERE (entity_id IN(2764369, 2766213));
    rows returned shows only 23 rows and Extran shows "Using Where"
    

Maria DB version tested on: 10.1.38-MariaDB-1~xenial mariadb.org binary distribution

### DIFF
--- a/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
+++ b/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
@@ -215,10 +215,10 @@ class ChangelogTest extends \PHPUnit\Framework\TestCase
         $this->connectionMock->expects($this->once())
             ->method('fetchCol')
             ->with($selectMock)
-            ->will($this->returnValue(['some_data']));
+            ->will($this->returnValue([1]));
 
         $this->model->setViewId('viewIdtest');
-        $this->assertEquals(['some_data'], $this->model->getList(1, 2));
+        $this->assertEquals([1], $this->model->getList(1, 2));
     }
 
     public function testGetListWithException()
--- a/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
+++ b/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
@@ -6,6 +6,11 @@
 
 namespace Magento\Framework\Mview\Test\Unit\View;
 
+/**
+ * Class ChangelogTest
+ *
+ * @package Magento\Framework\Mview\Test\Unit\View
+ */
 class ChangelogTest extends \PHPUnit\Framework\TestCase
 {
     /**
--- a/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
+++ b/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
@@ -7,10 +7,9 @@
 namespace Magento\Framework\Mview\Test\Unit\View;
 
 /**
- * Test Changelog View Class
+ * Test Coverage for Changelog View.
  *
  * @see \Magento\Framework\Mview\View\Changelog
- * @package Magento\Framework\Mview\Test\Unit\View
  */
 class ChangelogTest extends \PHPUnit\Framework\TestCase
 {
--- a/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
+++ b/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
@@ -7,8 +7,9 @@
 namespace Magento\Framework\Mview\Test\Unit\View;
 
 /**
- * Class ChangelogTest
+ * Test Changelog View Class
  *
+ * @see \Magento\Framework\Mview\View\Changelog
  * @package Magento\Framework\Mview\Test\Unit\View
  */
 class ChangelogTest extends \PHPUnit\Framework\TestCase
--- a/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
+++ b/lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
@@ -50,7 +50,7 @@ class ChangelogTest extends \PHPUnit\Framework\TestCase
     }
 
     /**
-     * @expectedException \Exception
+     * @expectedException \Magento\Framework\DB\Adapter\ConnectionException
      * @expectedExceptionMessage The write connection to the database isn't available. Please try again later.
      */
     public function testCheckConnectionException()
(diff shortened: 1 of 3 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22920) <br/>All committers have signed the CLA.

swnsma (changes requested):
Hi @mhodge13,

I want to say that you are digging really interesting topic.
However I would like to ask you clarify exact case how described by you issue can be reproduced.
Could you also clarify which version of Magento you have used to reproduce this issue?

I have made small research and I have found that since 2.2.8 return of this function is converted to int in https://github.com/magento/magento2/blob/0c786907ffe03d0e2990612eec16ee58b00379c5/lib/internal/Magento/Framework/Mview/View.php#L288

Also these changes are present in 2.3.1:
https://github.com/magento/magento2/blob/36097739bbb0b8939ad9a2a0dadee64318153dca/lib/internal/Magento/Framework/Mview/View.php#L288

But your fix I like more your fix, because it force \Magento\Framework\Mview\View\ChangelogInterface::getList to follow described contract in PHPDOC.

Lets check together:
1. Have your issue was reproduced (or can be reproduced) on version which I mentioned?
2. If your issue was fixed in 2.2.8/2.3.1 - I would like propose to replace fix which was done in 
lib/internal/Magento/Framework/Mview/View.php:288 with your changes. Then please revert changes in this line and fix all failed tests in build.

mhodge13:
Hi @swnsma 

The version that was causing this issue is in Magento 2.2.7.  It seems that the changes in 2.2.8 and 2.3.1 do in fact fix this issue.  I was checking the PHPDOC as well and saw that the return value of the function should in fact be integers.  I will revert the other changes and recommit to trigger the checks again.

Some of the issues found during the static tests were not changes made by me.  Do you want me to fix those as well?

swnsma:
Hi @mhodge13,
According to Magento refactoring policy, we need to have a green build to approve the pull request.
So, it will be nice if you fix static tests as well

mhodge13:
@swnsma Everything is passing now except a banned keyword in the Static Test Build

https://testing-service.magento-community.engineering/reports/magento/magento2/pull/22920/623b69a1-db69-4947-8ed6-3856492c9548/30865/Statics/report-sanity-ce.html

Can you let me know how you want me to fix this one please?

lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -53,12 +60,14 @@ public function __construct(\Magento\Framework\App\ResourceConnection $resource)
...
      *
      * @return void
-     * @throws \Exception
+     * @throws \Magento\Framework\Exception\SessionException
      */
     protected function checkConnection()
     {
         if (!$this->connection) {
-            throw new \Exception("The write connection to the database isn't available. Please try again later.");
+            throw new SessionException(
```
swnsma: I am not sure this is the best place for SessionException.
I think, \Magento\Framework\DB\Adapter\ConnectionException will be more suitable.

lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -173,7 +183,9 @@ public function getVersion()
         if (isset($row['Auto_increment'])) {
             return (int)$row['Auto_increment'] - 1;
         } else {
-            throw new \Exception("Table status for `{$changelogTableName}` is incorrect. Can`t fetch version id.");
+            throw new LocalizedException(
+                new Phrase("Table status for `{$changelogTableName}` is incorrect. Can`t fetch version id.")
```
swnsma: Please replace {$changelogTableName} with placeholder like in lib/internal/Magento/Framework/Mview/View/Changelog.php:104

Also, I would like to suggest use `\Magento\Framework\Exception\RuntimeException` instead of `\Magento\Framework\Exception\LocalizedException`.

lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -182,13 +194,15 @@ public function getVersion()
...
      *
-     * @throws \Exception
+     * @throws \Magento\Framework\Exception\LocalizedException
      * @return string
      */
     public function getName()
     {
         if (strlen($this->viewId) == 0) {
-            throw new \Exception("View's identifier is not set");
+            throw new LocalizedException(
```
swnsma: I think here is more suitable `\DomainException`.

lib/internal/Magento/Framework/Mview/Test/Unit/View/ChangelogTest.php
```diff
@@ -6,6 +6,11 @@
 
 namespace Magento\Framework\Mview\Test\Unit\View;
 
+/**
+ * Class ChangelogTest
+ *
+ * @package Magento\Framework\Mview\Test\Unit\View
```
swnsma: I think this information is redundant.

Better to add comment about general purposes of this class and leave with `@see` reference to class which it cover.

lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -6,8 +6,15 @@
...
 
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Exception\SessionException;
 use Magento\Framework\Phrase;
 
+/**
+ * Class Changelog
+ *
+ * @package Magento\Framework\Mview\View
+ */
```
swnsma, line 15: I think this information is redundant.

Could you change with comment about general purposes of this class?

swnsma (changes requested):
Hi @mhodge13,

We almost done.
Please check and fix my minor notices.

mhodge13:
@swnsma Can you please review the Unit Test build issue since it is calling out a js error and the static test issue is because of a banned word ("changelog").

Thanks, Mark

swnsma (approved):

magento-engcom-team:
[ENGCOM-5140](https://jira.corp.magento.com/browse/ENGCOM-5140) has been created to process this Pull Request

magento-engcom-team:
@mhodge13 thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

Nazar65:
QA Passed :heavy_check_mark:

p-bystritsky:
This PR should be approved.

PR #22947, opened 2019-05-20
phpcs error on rule classes - must be of the type integer

Nazar65:
### Description (*)
Unfortunately https://github.com/magento/magento2/pull/22081 does in no way fix the real issue in the Sniff, but pushes the problem a few lines down in the code.

The issue is that the `PHP_CodeSniffer\Files\File::findPrevious` method returns either a pointer to the previous token of a type **or `false` if none is found**. This case is triggered when the CodeSniffer is running on a file with no closing comment tags.

Earlier the issue was because this `false` was passed to a function which expected an `int` instead. What  https://github.com/magento/magento2/pull/22081 does is to cast the boolean into an integer, **which obviously is not correct** (as the Sniff now treats the code like there is a comment closing token at position 0 when in fact there is none in the file).

Now the Sniff returns a new error as it's trying to get the closing comment from the token list:
```
An error occurred during processing; checking has been aborted.
The error message was: Undefined index: comment_closer in
dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php on line 102
```

magento/magento2#20186: Issue title

1. clean Magento installation on 2.3-develop branch
2. create a Foo.php file with the following contents

```php
<?php

class Foo{}
```

3. run the phpcs on that file with the Magento ruleset `./vendor/bin/phpcs --standard=./dev/tests/static/framework/Magento/ Foo.php`

### DIFF
--- a/dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
+++ b/dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
@@ -97,8 +97,12 @@ class ClassAnnotationStructureSniff implements Sniff
     {
         $tokens = $phpcsFile->getTokens();
         $previousCommentClosePtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPtr - 1, 0);
-        $this->validateAnnotationBlockExists($phpcsFile, (int)$previousCommentClosePtr, (int)$stackPtr);
-        $commentStartPtr = (int)$phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
+        if (!$previousCommentClosePtr) {
+            $phpcsFile->addError('Comment block is missing', $stackPtr -1, 'MethodArguments');
+            return;
+        }
+        $this->validateAnnotationBlockExists($phpcsFile, $previousCommentClosePtr, $stackPtr);
+        $commentStartPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
         $commentCloserPtr = $tokens[$commentStartPtr]['comment_closer'];
         $emptyTypeTokens = [
             T_DOC_COMMENT_WHITESPACE,
@@ -111,7 +115,7 @@ class ClassAnnotationStructureSniff implements Sniff
         } else {
             $this->annotationFormatValidator->validateDescriptionFormatStructure(
                 $phpcsFile,
-                (int)$commentStartPtr,
+                $commentStartPtr,
                 (int) $shortPtr,
                 $previousCommentClosePtr,
                 $emptyTypeTokens
--- a/dev/tests/static/framework/Magento/Sniffs/Annotation/MethodAnnotationStructureSniff.php
+++ b/dev/tests/static/framework/Magento/Sniffs/Annotation/MethodAnnotationStructureSniff.php
@@ -45,6 +45,10 @@ class MethodAnnotationStructureSniff implements Sniff
         $tokens = $phpcsFile->getTokens();
         $commentStartPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, ($stackPtr), 0);
         $commentEndPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, ($stackPtr), 0);
+        if (!$commentStartPtr) {
+            $phpcsFile->addError('Comment block is missing', $stackPtr, 'MethodArguments');
+            return;
+        }
         $commentCloserPtr = $tokens[$commentStartPtr]['comment_closer'];
         $functionPtrContent = $tokens[$stackPtr+2]['content'] ;
         if (preg_match('/(?i)__construct/', $functionPtrContent)) {
--- a/dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
+++ b/dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
@@ -550,8 +550,12 @@ class MethodArgumentsSniff implements Sniff
         $numTokens = count($tokens);
         $previousCommentOpenPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
         $previousCommentClosePtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPtr - 1, 0);
-        if (!$this->validateCommentBlockExists($phpcsFile, $previousCommentClosePtr, $stackPtr)) {
-            $phpcsFile->addError('Comment block is missing', $stackPtr, 'MethodArguments');
+        if ($previousCommentClosePtr && $previousCommentOpenPtr) {
+            if (!$this->validateCommentBlockExists($phpcsFile, $previousCommentClosePtr, $stackPtr)) {
+                $phpcsFile->addError('Comment block is missing', $stackPtr, 'MethodArguments');
+                return;
+            }
+        } else {
             return;
         }
         $openParenthesisPtr = $phpcsFile->findNext(T_OPEN_PARENTHESIS, $stackPtr + 1, $numTokens);
@@ -663,6 +667,9 @@ class MethodArgumentsSniff implements Sniff
         foreach ($argumentPositions as $index => $argumentPosition) {
             $commentPosition = $commentPositions[$index];
             $type = $paramDefinitions[$index]['type'];
+            if ($type === null) {
+                continue;
+            }
             $paramName = $paramDefinitions[$index]['paramName'];
             if (($argumentPosition !== strlen($type) + 1) ||
                 (isset($commentPosition) && ($commentPosition !== $argumentPosition + strlen($paramName) + 1))) {


dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
```diff
@@ -97,8 +97,12 @@ public function process(File $phpcsFile, $stackPtr)
     {
         $tokens = $phpcsFile->getTokens();
         $previousCommentClosePtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPtr - 1, 0);
-        $this->validateAnnotationBlockExists($phpcsFile, (int)$previousCommentClosePtr, (int)$stackPtr);
-        $commentStartPtr = (int)$phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
+        if (!$previousCommentClosePtr) {
+            $phpcsFile->addError('Comment block is missing', $stackPtr -1, 'MethodArguments');
+            return;
+        }
+        $this->validateAnnotationBlockExists($phpcsFile, $previousCommentClosePtr, (int)$stackPtr);
```
orlangur: Is this cast to `(int)` still necessary?

dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -663,6 +663,10 @@ private function noneParamsAligned(array $argumentPositions, array $commentPosit
         foreach ($argumentPositions as $index => $argumentPosition) {
             $commentPosition = $commentPositions[$index];
             $type = $paramDefinitions[$index]['type'];
+            if ($type == null) {
```
orlangur: Why need this?

orlangur (changes requested):

dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
Nazar65: @orlangur Found one new issue in dev branch
DeepinScreenshot_select-area_20190520153623

dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
Nazar65: @orlangur Nope,  can i remove it ?

dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
orlangur: Could you add more info - I'm more suspicious about the `unset`

Nazar65: @orlangur well, unset is not needed, when we have type null, we can just skip iteration.

orlangur: Ok, please do so and also make a check strict (`=== null`).

orlangur (changes requested):

Nazar65:
@orlangur  done :heavy_check_mark:

orlangur (approved):

magento-engcom-team:
[ENGCOM-5139](https://jira.corp.magento.com/browse/ENGCOM-5139) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

mattijv:
@Nazar65 The exact same problem is present in the `Magento\Sniffs\Annotation\MethodArgumentsSniff` class lines 551 to 553.

You can use this class to test.

```php
<?php

class Foo {

    public function bar()
    {
        return 0;
    }

}
```

The run the sniffs with `./vendor/bin/phpcs --standard=./dev/tests/static/framework/Magento/ Foo.php`, which leads to the following error:

```
PHP Fatal error:  Uncaught TypeError: Argument 2 passed to
Magento\Sniffs\Annotation\MethodArgumentsSniff::validateCommentBlockExists()
must be of the type integer, boolean given
```

dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
```diff
@@ -97,8 +97,12 @@ public function process(File $phpcsFile, $stackPtr)
     {
         $tokens = $phpcsFile->getTokens();
         $previousCommentClosePtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPtr - 1, 0);
-        $this->validateAnnotationBlockExists($phpcsFile, (int)$previousCommentClosePtr, (int)$stackPtr);
-        $commentStartPtr = (int)$phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
+        if (!$previousCommentClosePtr) {
```
mattijv, line 100: This check should probably be `$previousCommentClosePtr === false` as I believe 0 is a valid return value from the `findPrevious` method, but should be treated differently to `false`.

mattijv (changes requested):

dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
Nazar65: ok, but what the difference with " !" and "=== false" ?

Nazar65: https://www.php.net/manual/en/language.operators.comparison.php this must be the same

Nazar65: @mattijv ok i agree.

p-bystritsky (approved):

dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
orlangur: @mattijv I had this concern during review but note that it cannot be `0` here (this is a CLOSING comment pointer). Nothing to change.

dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
```diff
@@ -97,8 +97,12 @@ public function process(File $phpcsFile, $stackPtr)
     {
         $tokens = $phpcsFile->getTokens();
         $previousCommentClosePtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPtr - 1, 0);
-        $this->validateAnnotationBlockExists($phpcsFile, (int)$previousCommentClosePtr, (int)$stackPtr);
-        $commentStartPtr = (int)$phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
+        if ($previousCommentClosePtr === false) {
```
orlangur: Revert.

dev/tests/static/framework/Magento/Sniffs/Annotation/MethodAnnotationStructureSniff.php
```diff
@@ -45,6 +45,10 @@ public function process(File $phpcsFile, $stackPtr)
         $tokens = $phpcsFile->getTokens();
         $commentStartPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, ($stackPtr), 0);
         $commentEndPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, ($stackPtr), 0);
+        if ($commentStartPtr === false) {
```
orlangur: Revert.

dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -550,8 +550,12 @@ public function process(File $phpcsFile, $stackPtr)
         $numTokens = count($tokens);
         $previousCommentOpenPtr = $phpcsFile->findPrevious(T_DOC_COMMENT_OPEN_TAG, $stackPtr - 1, 0);
         $previousCommentClosePtr = $phpcsFile->findPrevious(T_DOC_COMMENT_CLOSE_TAG, $stackPtr - 1, 0);
-        if (!$this->validateCommentBlockExists($phpcsFile, $previousCommentClosePtr, $stackPtr)) {
-            $phpcsFile->addError('Comment block is missing', $stackPtr, 'MethodArguments');
+        if ($previousCommentClosePtr !== false && $previousCommentOpenPtr !== false) {
```
orlangur: Revert.

orlangur (changes requested):

Nazar65:
@orlangur done :heavy_check_mark:

orlangur (approved):

magento-engcom-team:
[ENGCOM-5139](https://jira.corp.magento.com/browse/ENGCOM-5139) has been created to process this Pull Request

dev/tests/static/framework/Magento/Sniffs/Annotation/ClassAnnotationStructureSniff.php
mattijv: @orlangur I agree that in all practical situations the return value probably won't be 0, but personally I prefer being explicit over implicit when handling edge cases. The strict comparison with false was also the style the rest of the sniffs seemed to be favoring. However, in practice it most like won't make a difference.

mattijv (approved):

soleksii:
:heavy_check_mark: QA Passed

PR #22953, opened 2019-05-20
module-media-storage throws type mismatch exception.

Roger-Keulen:
Added null check for function that uses "$this->_scopeConfig->getValue" that will return 'null' when no configuration is found.

When your table core_config_data does not have a configuration path: "general/file/protected_extensions". Then you get the next warning in your logging and add-to-cart does not work any more. Also you get a message, like: "Something went wrong".

[2019-05-20 13:42:57] report.CRITICAL: Warning: Invalid argument supplied for foreach() in /private/var/www/drinkies/vendor/magento/module-media-storage/Model/File/Validator/NotProtectedExtension.php on line 84 {"exception":"[object] (Exception(code: 0): Warning: Invalid argument supplied for foreach() in /private/var/www/drinkies/vendor/magento/module-media-storage/Model/File/Validator/NotProtectedExtension.php on line 84 at /private/var/www/drinkies/vendor/magento/framework/App/ErrorHandler.php:61)"}

### Description (*)

I added a simple null check.

### DIFF
--- a/app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
+++ b/app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
@@ -78,6 +78,9 @@ class NotProtectedExtension extends \Zend_Validate_Abstract
     {
         if (!$this->_protectedFileExtensions) {
             $extensions = $this->getProtectedFileExtensions();
+            if (is_null($extensions)) {
+                $extensions = [];
+            }
             if (is_string($extensions)) {
                 $extensions = explode(',', $extensions);
             }
@@ -93,7 +96,7 @@ class NotProtectedExtension extends \Zend_Validate_Abstract
      * Return list with protected file extensions
      *
      * @param \Magento\Store\Model\Store|string|int $store
-     * @return string|string[]
+     * @return mixed
      */
     public function getProtectedFileExtensions($store = null)
     {
--- a/app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
+++ b/app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
@@ -96,7 +96,7 @@ class NotProtectedExtension extends \Zend_Validate_Abstract
      * Return list with protected file extensions
      *
      * @param \Magento\Store\Model\Store|string|int $store
-     * @return mixed
+     * @return string|null
      */
     public function getProtectedFileExtensions($store = null)
     {
--- a/app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
+++ b/app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
@@ -78,7 +78,7 @@ class NotProtectedExtension extends \Zend_Validate_Abstract
     {
         if (!$this->_protectedFileExtensions) {
             $extensions = $this->getProtectedFileExtensions();
-            if (is_null($extensions)) {
+            if ($extensions === null) {
                 $extensions = [];
             }
             if (is_string($extensions)) {


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=22953) <br/>All committers have signed the CLA.

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
```diff
@@ -93,7 +96,7 @@ protected function _initProtectedFileExtensions()
      * Return list with protected file extensions
      *
      * @param \Magento\Store\Model\Store|string|int $store
-     * @return string|string[]
```
Roger-Keulen, line 96: This function will NEVER return an array of strings. The only function used in this will return mixed.

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
```diff
@@ -78,6 +78,9 @@ protected function _initProtectedFileExtensions()
     {
         if (!$this->_protectedFileExtensions) {
             $extensions = $this->getProtectedFileExtensions();
+            if ($extensions === null) {
+                $extensions = [];
+            }
             if (is_string($extensions)) {
```
Roger-Keulen, line 84: if(is_null($extension)) is not allowed. Then why can i see _is_string($extension)_ over here ? 
Are there special rules for Magento Core developers ?

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
```diff
@@ -78,6 +78,9 @@ protected function _initProtectedFileExtensions()
     {
         if (!$this->_protectedFileExtensions) {
             $extensions = $this->getProtectedFileExtensions();
```
Roger-Keulen, line 80: The function called here is public and can not be changed without braking parts of Magento. We assume that when nill/null is returned we going to use an empty array as muted value. Ps: With Magento 3 it's better to learn programming without null checks. Thus the getProtectedFileExtensions should have returned ALWAYS an array. That list can then be empty as "null".

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
```diff
@@ -78,6 +78,9 @@ protected function _initProtectedFileExtensions()
...
+            if ($extensions === null) {
```
Roger-Keulen: is_null is not allowed. And there for we have multi programming styles now. The Magento style is not allowed for me.

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
```diff
@@ -93,7 +96,7 @@ protected function _initProtectedFileExtensions()
      * Return list with protected file extensions
      *
      * @param \Magento\Store\Model\Store|string|int $store
-     * @return string|string[]
+     * @return mixed
```
orlangur: `mixed` must not be used, please specify exact types.

orlangur (changes requested):
Please report an issue and specify it in PR description.

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
orlangur: > why can i see is_string($extension) over here ?

I don't get your logic. What would be a replacement for `is_string`?

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
Roger-Keulen: Ok, no problem. The underlaying function is returning "mixed" according to its doc type. But i will change it to "string|null".

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
Roger-Keulen: Then why change is_null then ? Thus is_string is only good because you can't change it. Also the code is not cleaner. You should try to code cleanly, not like your sniffer wants to.

app/code/Magento/MediaStorage/Model/File/Validator/NotProtectedExtension.php
Roger-Keulen: Changed to "string|null"

Roger-Keulen:
Ah, it has to be an issue first. I just wanted to make Magento less brittle.

Because the configuration of 'general/file/protected_extensions' used in Magento\MediaStorage\Model\File\Validator\NotProtectedExtension is in module Magento_Store and not in Magento_MediaStorage. But according to module.xml it's not depended on that module and there for the configuration can be empty.

So, it's not really an issue.

Issue #22808, opened 2019-05-09
php bin/magento catalog:image:resize error if image is missing


ioweb-gr:
### Preconditions (*)
1. Magento 2.3.1 

### Steps to reproduce (*)
1. Delete manually a product image from pub/media/catalog folder 
2. Execute catalog:image:resize 

### Expected result (*)
1. The system creates the resized images for all products that the images exist and lists the errors.

### Actual result (*)
1. The execution halts immediately at the first missing image. 

![image](https://user-images.githubusercontent.com/20220341/57453750-d7dc8300-726f-11e9-9a9b-f40b3784830c.png)

This doesn't affect on the fly generation of images when browsing the products but it's good to be able to execute and test.

nikunjskd20:
Hi, I am facing the same issue in **Magento Commerce ver. 2.3.1**

ioweb-gr:
I forgot to mention this can also happen after migrating from magento 1 because the data migration tool does not take into account a few things regarding media galleries like cleaning up catalog_entity_varchar table and magento 2 cannot handle product images where the path is not like /x/x/filename.ext but just /filename.ext. the latter can happen in older versions of m1 using csv import so this can be a problem for a lot of users who migrated

gwharton:
How about database media storage mode, when it is valid for pub/media to be completely wiped. Make sure any solution works in this mode too.

amitvishvakarma:
#imagine2019

magento-engcom-team:
:white_check_mark: Confirmed by @sdzhepa 
 Thank you for verifying the issue. Based on the provided information internal tickets `MC-17217`, `MC-17218` were created

**Issue Available**: @sdzhepa, _You will be automatically unassigned. Contributors/Maintainers can claim this issue to continue. To reclaim and continue work, reassign the ticket to yourself._

magento-engcom-team:
Hi @ioweb-gr, @k4emic, @tdgroot, @gulshanchitransh. 

Thank you for your report and collaboration!

The issue was fixed by Magento team. The fix was delivered into `magento/magento2:2.3-develop` branch(es).
Related commit(s):
 - [6a694a9bab15ab20531ffb2f70a744127611a1e3](https://github.com/magento/magento2/commit/6a694a9bab15ab20531ffb2f70a744127611a1e3)
[...]

PR #23005, opened 2019-05-24
Improve command catalog:images:resize

tdgroot:
- Only resize images that are actually in use.
- Improve code on several places

### Description (*)
The resizer uses table `catalog_product_entity_media_gallery` as source table. These records contain the paths to the actual images. The problem is, these records can exist, even when they are not being used by any products. I added an inner join to table `catalog_product_entity_media_gallery_value`, to make sure that there's a link between the image and a product.

This improves performance and is less error prone, because the orphan images might not have a physical presence on the filesystem. Now the latter is an other issue, but hardening the resizer won't do any harm.

### Fixed Issues 

1. N/a
1. Create product with image.
2. Insert a dummy row into `catalog_product_entity_media_gallery`.
3. Run `bin/magento catalog:images:resize`.
4. Output should look like this:
```
Product images resized successfully

1/1 [============================] 100% < 1 sec 44.2 MiB 	| /t/e/testupload.jpg
```

### DIFF
--- a/app/code/Magento/MediaStorage/Console/Command/ImagesResizeCommand.php
+++ b/app/code/Magento/MediaStorage/Console/Command/ImagesResizeCommand.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 namespace Magento\MediaStorage\Console\Command;
 
 use Magento\Framework\App\Area;
+use Magento\Framework\App\ObjectManager;
 use Magento\Framework\App\State;
 use Magento\MediaStorage\Service\ImageResize;
 use Symfony\Component\Console\Helper\ProgressBar;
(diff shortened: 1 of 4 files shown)


app/code/Magento/MediaStorage/Console/Command/ImagesResizeCommand.php
```diff
@@ -28,24 +28,24 @@ class ImagesResizeCommand extends \Symfony\Component\Console\Command\Command
...
         State $appState,
         ImageResize $resize,
-        ObjectManagerInterface $objectManager
+        ProgressBarFactory $progressBarFactory
     ) {
         parent::__construct();
         $this->resize = $resize;
         $this->appState = $appState;
-        $this->objectManager = $objectManager;
+        $this->progressBarFactory = $progressBarFactory;
```
dmytro-ch: Unfortunately, according to the Contribution Guide, we may not remove constructor params.
The new constructor parameters should be optional in order to avoid backward compatibility issues.

To preserve the constructor signature:
- Remove type hinting for unused parameters to remove dependency on their type.
- Add `@SuppressWarnings(PHPMD.UnusedFormalParameter)` for unused parameters.

dmytro-ch (changes requested):
Hi @tdgroot,

Could you please also check and fix the automated tests failures.

app/code/Magento/MediaStorage/Console/Command/ImagesResizeCommand.php
tdgroot: Thank you for the review! I updated the changeset, fixed coding style errors and fixed the BIC.

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-5173](https://jira.corp.magento.com/browse/ENGCOM-5173) has been created to process this Pull Request

soleksii:
:heavy_check_mark: QA Passed

**Steps to reproduce:**

1. Create product with image;
2. Insert a dummy row into catalog_product_entity_media_gallery;
3. Run bin/magento catalog:images:resize

**Before:**

before1

**After:**

after1

**Note:** I unassigned the issue #22808 because PR changes do not fix this problem!

**Thanks!**

PR #23008, opened 2019-05-24
Improve ComponentRegistrar with new register methods

tdgroot:
- Add register methods for each component type
- Add strict types

### Description (*)
To improve developer experience and code readability, I added register methods for each component type to the `ComponentRegistrar`.

Now people can register modules as following:
``` php
<?php

use \Magento\Framework\Component\ComponentRegistrar;

ComponentRegistrar::registerModule('Acme_SampleModule', __DIR__);
```

Just a simple addition to make things easier. 

Aside from that, I also added strict types to all the methods in `ComponentRegistrar` and `ComponentRegistrarInterface`

1. Create a module and register it using `registerModule`.
2. It should be loaded.

### DIFF
--- a/lib/internal/Magento/Framework/Component/ComponentRegistrar.php
+++ b/lib/internal/Magento/Framework/Component/ComponentRegistrar.php
@@ -3,6 +3,8 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+declare(strict_types=1);
+
 namespace Magento\Framework\Component;
 
 /**
@@ -15,7 +17,7 @@ namespace Magento\Framework\Component;
 class ComponentRegistrar implements ComponentRegistrarInterface
 {
     /**#@+
-     * Different types of components
+     * Available component types.
      */
     const MODULE = 'module';
     const LIBRARY = 'library';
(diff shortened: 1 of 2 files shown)


orlangur:
@tdgroot we are going to get rid of `registration.php` files, you know. Is this aligned?

tdgroot:
> @tdgroot we are going to get rid of `registration.php` files, you know. Is this aligned?

I know, but are there any concrete plans yet? Until `registration.php` gets removed from the framework (and actually released), I think this change can create a better developer experience in at least 2.3 and 2.4.

orlangur:
See https://github.com/magento/architecture/pull/63#issuecomment-492804323. I don't think it's a good idea to force developer learn something new and then remove it.

I would leave type hintings without adding methods but it does not pass SemVer check: https://testing-service.magento-community.engineering/reports/magento/magento2/pull/23008/cf23c537-870a-49af-95e7-f8744143785d/23495/SemanticVersionChecker/report-magento2.html

tdgroot:
@orlangur I don't think you're right. That proposal was accepted, but isn't nearly done. See https://github.com/magento/architecture/pull/162. The discussion is still in favor of keeping `registration.php` for non-composer modules. 

Also, I think closing this is too early, because there's still discussion going on.

orlangur:
@tdgroot there is no reason to keep it open until we cannot merge it. https://github.com/magento/architecture/pull/63 is already merged, until compatible with this PR approach is approved. Also, value of introducing such methods is quite doubtful.

PR #23040, opened 2019-05-29
Don't create a new account-nav block - use existing instead.

vovayatsyuk:
### Description
Account navigation block is already created by Magento_Customer module in
[Magento/Customer/view/frontend/layout/customer_account.xml](https://github.com/magento/magento2/blob/2.3-develop/app/design/frontend/Magento/luma/Magento_Customer/layout/customer_account.xml#L16) layout update file.
So instead of creating a new block that will contain another one, we can
use the same block.

1. Navigate into customer account page and inspect sidebar with browser dev tools
2. You'll see an empty block below `block-collapsible-nav` block:

   image

3. Apply the patch and empty block will disappear.

### DIFF
--- a/app/design/frontend/Magento/luma/Magento_Customer/layout/customer_account.xml
+++ b/app/design/frontend/Magento/luma/Magento_Customer/layout/customer_account.xml
@@ -7,52 +7,12 @@
 -->
 <page layout="2columns-left" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
     <body>
-        <referenceContainer name="sidebar.main">
-            <block class="Magento\Framework\View\Element\Template" name="customer_account_navigation_block" template="Magento_Theme::html/collapsible.phtml" before="-">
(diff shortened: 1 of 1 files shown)


orlangur (approved):
@vovayatsyuk wow. Would be nice to write a dynamic test to catch such declarations.

magento-engcom-team:
[ENGCOM-5180](https://jira.corp.magento.com/browse/ENGCOM-5180) has been created to process this Pull Request

dmytro-ch:
@vovayatsyuk nice catch!

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-5180](https://jira.corp.magento.com/browse/ENGCOM-5180) has been created to process this Pull Request

dmytro-ch:
The only thing I'm wondering about is backward compatibility.
Basically, we are removing the `customer_account_navigation_block` block provided by Luma theme.
Thus all the links added to this block in the scope of a child theme will disappear after an upgrade.

Nevertheless, the current issue definitely should be fixed.

soleksii:
:heavy_check_mark: QA Passed

**Before:**

before

**After:**

after

PR #23064, opened 2019-05-30
Added boundary checking for the rawResponse array

ghost:
Added boundary checking for the rawResponse array

### Description (*)
This is just a boundary check for an array in \Magento\Framework\Search\Adapter\Mysql\ResponseFactory. The behavior will still be the same for anything that was valid before because all this code change does is a boundary check. If this should still throw an error, I personally believe that the error should be caught and handled explicitly as to more explain and track what actually went wrong.

Accessing any page that utilized the search (categories or the search page itself) will demonstrate the effect of this code change. However, if you wanted to work with a base installation, it would be.
1. Add a product
2. Open up the homepage of your Magento installation.
3. Type a search query that would normally bring up the product such as the name.

### Questions or comments
This fixed an issue I had when migrating from 1.9.3.8 to 2.3. The issue occurred when trying to access the search after migrating the database. All the products had migrated, and the autofill behavior for the search bar itself had actually worked, but nothing was showing up under search pages. The error was simply 
 Undefined index: total in /var/www/lib/internal/Magento/Framework/Search/Adapter/Mysql/ResponseFactory.php
which shook me for a second. Again, if this is really supposed to still throw an error, that also makes sense, but I feel like it should be explicitly caught and explained when it does happen.

### DIFF
--- a/lib/internal/Magento/Framework/Search/Adapter/Mysql/ResponseFactory.php
+++ b/lib/internal/Magento/Framework/Search/Adapter/Mysql/ResponseFactory.php
@@ -62,15 +62,10 @@ class ResponseFactory
             /** @var \Magento\Framework\Api\Search\Document[] $documents */
             $documents[] = $this->documentFactory->create($rawDocument);
 	}
-
 	$total = 0;
-	if(array_key_exists('total',$rawResponse))
-	{
-
+	if(array_key_exists('total',$rawResponse)){
 		$total = $rawResponse['total'];	
-
 	}
-
 	/** @var \Magento\Framework\Search\Response\Aggregation $aggregations */
         $aggregations = $this->aggregationFactory->create($rawResponse['aggregations']);
         return $this->objectManager->create(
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23064) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dev.magento.com/magento/magento2?pullRequest=23064) before we can accept your contribution.<br/>

orlangur:
@emmahasabadtime please reproduce this on a vanilla instance and report an issue first.

ghost:
I thought about this for a second, but then, realized that I can't actually replicate the system exactly how it is on my machine due to the steps involving data that is sensitive, and I don't really want to dedicate the amount of time required to replicate a 1.9.3.8 installation with all the same configurations without any sensitive data, and then, use the migration tool to migrate it again. I'll just keep the boundary check locally. Really, I just think that the data structure used here in this function is just bad practice regardless of if it causes any issues. If there are three required fields, the parameter should either require those to be set from its definition, or the function should initialize the values separately from operating on them, or the function should throw a more verbose error when one of those values is not set.

orlangur:
@emmahasabadtime I was just trying to understand whether the problem is in core code or in migration tool.

PR #23100, opened 2019-06-02
Resolve issue with improper EAV attribute join statement

udovicic:
### Description (*)

When using flat tables for products in combination with on save indexers, there is an issue with product save. On smaller data sets this is not obvious, but on larger data sets, save action may need few GB of ram to complete due to improper join of status attribute in indexer.

1. n/a

1. Enable product flat tables and reindex
2. Set indexing mode for flat tables to "on save"
3. Create new product from admin and hit save
4. Either set your debugger [here](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php#L99) or turn on query log, but [affected statement](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php#L84-L97) will return multiple rows for single product where it should have returned one. And rows are exactly the same.

### Questions or comments

Not sure that `$statusAttribute->getEntityIdField()` is correct choice for both tables, so correct me if I am wrong.

### DIFF
--- a/app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
+++ b/app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
@@ -9,6 +9,7 @@ namespace Magento\Catalog\Model\Indexer\Product\Flat\Action;
 
 use Magento\Framework\App\ResourceConnection;
 use Magento\Catalog\Model\Product\Attribute\Source\Status;
+use Magento\Framework\EntityManager\MetadataPool;
 use Magento\Store\Model\Store;
 
 /**
@@ -32,18 +33,27 @@ class Eraser
     protected $storeManager;
 
     /**
+     * @var MetadataPool
+     */
+    private $metadataPool;
+
+    /**
      * @param \Magento\Framework\App\ResourceConnection $resource
      * @param \Magento\Catalog\Helper\Product\Flat\Indexer $productHelper
      * @param \Magento\Store\Model\StoreManagerInterface $storeManager
+     * @param MetadataPool|null $metadataPool
      */
     public function __construct(
         \Magento\Framework\App\ResourceConnection $resource,
         \Magento\Catalog\Helper\Product\Flat\Indexer $productHelper,
-        \Magento\Store\Model\StoreManagerInterface $storeManager
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
+        MetadataPool $metadataPool = null
     ) {
         $this->productIndexerHelper = $productHelper;
         $this->connection = $resource->getConnection();
         $this->storeManager = $storeManager;
+        $this->metadataPool = $metadataPool ?:
+            \Magento\Framework\App\ObjectManager::getInstance()->get(MetadataPool::class);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
```diff
@@ -85,13 +85,17 @@ public function removeDisabledProducts(array &$ids, $storeId)
         $select->joinLeft(
             ['status_global_attr' => $statusAttribute->getBackendTable()],
             ' status_global_attr.attribute_id = ' . (int)$statusAttribute->getAttributeId()
-            . ' AND status_global_attr.store_id = ' . Store::DEFAULT_STORE_ID,
+            . ' AND status_global_attr.store_id = ' . Store::DEFAULT_STORE_ID
+            . ' AND status_global_attr.' . $statusAttribute->getEntityIdField() . '='
+            . 'product_table.' . $statusAttribute->getEntityIdField(),
```
dmytro-ch: I guess we should not use the `$statusAttribute->getEntityIdField()` method's call in order to get the link field name of `catalog_product_entity` DB table. This is rather a coincidence that linked fields have the same names. Please use the same approach as we already have in `\Magento\Catalog\Model\Indexer\Product\Flat\Action\Eraser::getSelectForProducts`.

dmytro-ch (changes requested):

app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
udovicic: Could you please elaborate? I do not see anything in the [referenced function](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php#L118-L125).

While I do agree that fetching entity id field from `$statusAttribute` is wrong, wouldn't proper way than be to get `Magento\Catalog\Api\Data\ProductInterface` metadata from `Magento\Framework\EntityManager\MetadataPool` and call `getLinkField()` on it?

dmytro-ch: I mean, that using the hardcoded field name `product_table.entity_id` is just enough in this case.

dmytro-ch: I would recommend removing MetadataPool dependency.

dmytro-ch:
@udovicic, could you please squash all commits into a single one in order to clean up history, after the changes are applied and automated test failures are fixed.

Thank you!

app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
udovicic: This pull request is coming from bugfix I did for our EE client. Hardcoding `entity_id` there would cause incorrect relation as EAV attributes are joined by `row_id` instead. Hence the usage of MetadataPool to make it compatible for both.

Correct me if I am missing something.

dmytro-ch: Thank you for the update!
Agreed.

app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
```diff
@@ -31,6 +31,11 @@ class Eraser
      */
     protected $storeManager;
 
+    /**
+     * @var \Magento\Framework\EntityManager\MetadataPool
+     */
+    protected $metadataPool;
```
dmytro-ch: Please add new properties with the `private` scope of visibility, according to the contributor's guide.

app/code/Magento/Catalog/Model/Indexer/Product/Flat/Action/Eraser.php
```diff
@@ -39,11 +44,13 @@ class Eraser
     public function __construct(
         \Magento\Framework\App\ResourceConnection $resource,
         \Magento\Catalog\Helper\Product\Flat\Indexer $productHelper,
-        \Magento\Store\Model\StoreManagerInterface $storeManager
+        \Magento\Store\Model\StoreManagerInterface $storeManager,
+        \Magento\Framework\EntityManager\MetadataPool $metadataPool
```
dmytro-ch: The new constructor parameters should be optional in order to avoid backward compatibility issues.

dmytro-ch (changes requested):

udovicic:
Hi @dmytro-ch , all requested changes have been done.

Out of curiosity, could you explain why is it requested for commits to be already squashed on PR? This can also be done at the merge time through GitHub interface. Obviously not a problem, I am just curious.

dmytro-ch:
@udovicic thank you for the update!

> This can also be done at the merge time through GitHub interface.

This is a good point. Magento Community Engineering team is responsible for merging the PRs. I'll check and clarify if it may be done at the merge time.

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-5307](https://jira.corp.magento.com/browse/ENGCOM-5307) has been created to process this Pull Request

| Label | Description |

engcom-Delta:
:heavy_check_mark: QA Passed
Screenshots attached: 
1.On 2.3-develop

2. This PR

engcom-Foxtrot:
QA passed.
Mainline:
index
PR:
index

PR #23101, opened 2019-06-03
per PSR2 recommendation, ElseIf should be used instead of else-if 

sankalpshekhar:
### Description (*)

This per updates usage of ELSE-IF to ElseIf, per PSR2 recommendation.

1. ...
2. ...

### DIFF
--- a/app/code/Magento/CatalogImportExport/Model/Import/Product/Validator.php
+++ b/app/code/Magento/CatalogImportExport/Model/Import/Product/Validator.php
@@ -72,7 +72,7 @@ class Validator extends AbstractValidator implements RowValidatorInterface
         $val = $this->string->cleanString($this->_rowData[$attrCode]);
         if ($type == 'text') {
             $valid = $this->string->strlen($val) < Product::DB_MAX_TEXT_LENGTH;
-        } else if ($attrCode == Product::COL_SKU) {
+        } elseif ($attrCode == Product::COL_SKU) {
             $valid = $this->string->strlen($val) <= SKU::SKU_MAX_LENGTH;
         } else {
             $valid = $this->string->strlen($val) < Product::DB_MAX_VARCHAR_LENGTH;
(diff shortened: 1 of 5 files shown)


dmytro-ch:
Hi @sankalpshekhar,
Could you please check and resolve the static test failures?

orlangur:
Thanks @sankalpshekhar, we don't need a manual PR for such change though as all such occurrences will be fixed automatically.

sankalpshekhar:
@orlangur Could you help on the steps laid out for this to be fixed automatically?

From what I see the last commit on these files, the oldest being almost a year old and they haven't been fixed yet.

orlangur:
@sankalpshekhar `phpcbf` will be applied all over the code base. Coding standard is being finalized currently.

sankalpshekhar:
@orlangur I see that the standards are still being finalized, including versioning. Do we have a release date on when this would be applied to the framework?

orlangur:
@sankalpshekhar I believe we can do it during June and target `2.3.3` release, here is a corresponding issue: https://github.com/magento/magento2/issues/8612#issuecomment-448558741

hostep:
@orlangur: can we attempt to put this code standard changes in as few commits as possible, even maybe dedicate a separate release for this? [I don't want to go figuring out again](https://github.com/magento/magento2/commit/160331e273e738193ef1b6b4e79ad5bb5f1081b2#commitcomment-28953613) what actually changed in the code besides the code standards changes. Having a dedicated release number for only code standards applying would make it more clear and will help a lot with upgrading shops to a newer version where we can then easily see the more "significant" changes in the codebase (and ignore the code standard changes).

orlangur:
@hostep thanks, will keep this in mind. I remember such approach used only once for copyright year update, probably makes sense if we will update PHPDocs massively.

**UPD**:

> With every update of Magento, I go through all changed files manually to see what has changed in between them. Especially the frontend files (phtml, html, js, xml, ...) as they tend to not be backwards compatible

Wow, this is really awkward use case, people usually rely on automation these days. I don't think it justifies a separate release, ignoring minimalistic amount of CS-related commit should be sufficient (and this is much better than hundreds of manual fixes in different commits being contributed currently).

hostep:
> people usually rely on automation these days

If you know of such an automation tool, feel free to share. The best one we found this far is just detecting changes, but not applying those changes automatically, this is https://github.com/AmpersandHQ/ampersand-magento2-upgrade-patch-helper

Sorry for going completely off topic here :)

orlangur:
@hostep no-no, I meant it's quite strange to look into diffs, just write enough automated tests and run them whenever you upgrade :)

hostep:
Magento tends to fix security related issues in phtml files by introducing escape functions for example. I'm not sure if tests can catch these changes in overwritten phtml files in a custom theme? Since the output will be exactly the same if no weird data was passed. While your custom theme then might contain vulnerable code which was fixed in core Magento.

PR #23106, opened 2019-06-03
Payment Failed Email not working as intended

krnshah:
### Description (*)

I made a couple of tests and in my environment, trying a couple of different payment methods, Magento never run those controllers. So, even if the payment fails, the email notification is not sent.

This behaviour is very different from Magento 1, where the merchant receives effective notifications and can monitor checkout/payment problems.

1. magento/magento2#11347: Payment Failed Email not working as intended

1. N/A

### DIFF
--- a/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
+++ b/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
@@ -114,9 +114,7 @@ class GuestPaymentInformationManagement implements \Magento\Checkout\Api\GuestPa
             try {
                 $orderId = $this->cartManagement->placeOrder($cartId);
             } catch (\Magento\Framework\Exception\LocalizedException $e) {
-                /** @var $quoteIdMask QuoteIdMask */
                 $quoteIdMask = $this->quoteIdMaskFactory->create()->load($cartId, 'masked_id');
-                /** @var \Magento\Quote\Model\Quote $quote */
                 $quote = $this->cartRepository->getActive($quoteIdMask->getQuoteId());
                 $this->checkoutHelper->sendPaymentFailedEmail($quote, __($e->getMessage()));
                 throw new CouldNotSaveException(
--- a/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
+++ b/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
@@ -77,22 +77,22 @@ class GuestPaymentInformationManagement implements \Magento\Checkout\Api\GuestPa
         \Magento\Quote\Api\GuestPaymentMethodManagementInterface $paymentMethodManagement,
         \Magento\Quote\Api\GuestCartManagementInterface $cartManagement,
         \Magento\Checkout\Api\PaymentInformationManagementInterface $paymentInformationManagement,
-        \Magento\Quote\Model\QuoteIdMaskFactory $quoteIdMaskFactory,
+        \Magento\Quote\Model\MaskedQuoteIdToQuoteIdInterface $MaskedQuoteIdToQuoteIdInterface,
         CartRepositoryInterface $cartRepository,
+        ResourceConnection $connectionPool = null,
         \Magento\Checkout\Helper\Data $checkoutHelper = null,
-        \Psr\Log\LoggerInterface $logger = null,
-        ResourceConnection $connectionPool = null
+        \Psr\Log\LoggerInterface $logger = null
     ) {
         $this->billingAddressManagement = $billingAddressManagement;
         $this->paymentMethodManagement = $paymentMethodManagement;
         $this->cartManagement = $cartManagement;
         $this->paymentInformationManagement = $paymentInformationManagement;
-        $this->quoteIdMaskFactory = $quoteIdMaskFactory;
+        $this->MaskedQuoteIdToQuoteIdInterface = $MaskedQuoteIdToQuoteIdInterface;
         $this->cartRepository = $cartRepository;
+        $this->connectionPool = $connectionPool ?: ObjectManager::getInstance()->get(ResourceConnection::class);
         $this->checkoutHelper = $checkoutHelper ?:
             ObjectManager::getInstance()->get(\Magento\Checkout\Helper\Data::class);
         $this->logger = $logger ?: ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
-        $this->connectionPool = $connectionPool ?: ObjectManager::getInstance()->get(ResourceConnection::class);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -86,6 +89,9 @@ public function __construct(
         $this->paymentInformationManagement = $paymentInformationManagement;
         $this->quoteIdMaskFactory = $quoteIdMaskFactory;
         $this->cartRepository = $cartRepository;
+        $this->checkoutHelper = $checkoutHelper ?:
+            ObjectManager::getInstance()->get(\Magento\Checkout\Helper\Data::class);
+        $this->logger = $logger ?: ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
```
Stepa4man, line 101: Please add these parameters to the end of the parameters list. [(Adding a constructor parameter)](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#adding-a-constructor-parameter)

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -108,12 +114,17 @@ public function savePaymentInformationAndPlaceOrder(
             try {
                 $orderId = $this->cartManagement->placeOrder($cartId);
             } catch (\Magento\Framework\Exception\LocalizedException $e) {
+                /** @var $quoteIdMask QuoteIdMask */
+                $quoteIdMask = $this->quoteIdMaskFactory->create()->load($cartId, 'masked_id');
```
Stepa4man, line 124: "load()" method is deprecated in AbstractModel. Please use repos or Resource models to retrieve model. This is general comment.
 In this particular case,  I believe the best solution would be to use: \Magento\Quote\Model\MaskedQuoteIdToQuoteIdInterface

Stepa4man (changes requested):

Stepa4man:
Hi @krnshah, thank you for your contribution!
Please check my comments and provide fixes.

krnshah:
Hi @Stepa4man here i provide fixes as per your comment can you please look into.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -108,12 +114,15 @@ public function savePaymentInformationAndPlaceOrder(
             try {
                 $orderId = $this->cartManagement->placeOrder($cartId);
             } catch (\Magento\Framework\Exception\LocalizedException $e) {
+                $quoteIdMask = $this->MaskedQuoteIdToQuoteIdInterface->create()->load($cartId, 'masked_id');
```
Stepa4man: Did you check this code? Are you sure it will work?
Please check how to use MaskedQuoteIdToQuoteIdInterface. It doesn't have a method "create".
MaskedQuoteIdToQuoteId which implements MaskedQuoteIdToQuoteIdInterface has methos "execute" and check returning type. It doesn't return  QuoteIdMask model.

app/code/Magento/Checkout/Model/PaymentInformationManagement.php
```diff
@@ -85,12 +94,15 @@ public function savePaymentInformationAndPlaceOrder(
         try {
             $orderId = $this->cartManagement->placeOrder($cartId);
         } catch (\Magento\Framework\Exception\LocalizedException $e) {
+            /** @var \Magento\Quote\Model\Quote $quote */
```
Stepa4man, line 102: Please get rid of this unnecessary doc block. We have return type at doc block for "cartRepository->getActive" method. It's more than enough for this method.

Stepa4man (changes requested):
Hi @krnshah, thanks for your fixes. Please check 2 more comments and tests. Looks like some tests failed.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -76,17 +77,22 @@ public function __construct(
         \Magento\Quote\Api\GuestPaymentMethodManagementInterface $paymentMethodManagement,
         \Magento\Quote\Api\GuestCartManagementInterface $cartManagement,
         \Magento\Checkout\Api\PaymentInformationManagementInterface $paymentInformationManagement,
-        \Magento\Quote\Model\QuoteIdMaskFactory $quoteIdMaskFactory,
+        \Magento\Quote\Model\MaskedQuoteIdToQuoteIdInterface $MaskedQuoteIdToQuoteIdInterface,
```
Stepa4man: Please use variables with names which starts from lower-case letter

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -76,17 +77,22 @@ public function __construct(
...
+        ResourceConnection $connectionPool = null,
+        \Magento\Checkout\Helper\Data $checkoutHelper = null,
+        \Psr\Log\LoggerInterface $logger = null
     ) {
         $this->billingAddressManagement = $billingAddressManagement;
         $this->paymentMethodManagement = $paymentMethodManagement;
         $this->cartManagement = $cartManagement;
         $this->paymentInformationManagement = $paymentInformationManagement;
-        $this->quoteIdMaskFactory = $quoteIdMaskFactory;
+        $this->MaskedQuoteIdToQuoteIdInterface = $MaskedQuoteIdToQuoteIdInterface;
```
Stepa4man: Same here. Please use variables with names which starts from the lower-case letter.
Also please add a private property that you inject here to the list of properties before the constructor

Stepa4man (changes requested):

sidolov:
@krnshah , I am closing this PR now due to inactivity.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
krnshah: @Stepa4man Can you please suggest any other way for this?

app/code/Magento/Checkout/Model/PaymentInformationManagement.php
```diff
@@ -85,12 +99,15 @@ public function savePaymentInformationAndPlaceOrder(
...
+            $quote = $this->cartRepository->getActive($cartId);
+            $this->checkoutHelper->sendPaymentFailedEmail($quote, __($e->getMessage()));
```
sidolov, line 104: Could you please explain why we should send an email with payment failed status every time when placeOrder method returns exception? The exception can be caused by different reasons, not only if payment if failed.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
Stepa4man: @krnshah Why do you need another way? I suggested the right way.

Stepa4man (changes requested):
@krnshah, thanks for your contribution.
Please resolve all comments.

Thanks.

sidolov:
@krnshah , I am closing this PR now due to inactivity.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
indefinitedevil: I agree in principle, but the entire class loads $quoteIdMask this way, and there is no relevant interface for it.

"class QuoteIdMask extends \Magento\Framework\Model\AbstractModel"

In this case, if you want someone to use a repo, you're going to have to convert it to use repos first.

app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
```diff
@@ -166,20 +183,6 @@ public function getPaymentInformation($cartId)
...
-     * @deprecated 100.2.0
-     */
-    private function getLogger()
-    {
-        if (!$this->logger) {
-            $this->logger = \Magento\Framework\App\ObjectManager::getInstance()->get(\Psr\Log\LoggerInterface::class);
-        }
-        return $this->logger;
-    }
-
```
indefinitedevil, line 182: Not entirely sure why rewriting the logger functionality came into this?

Issue #23122, opened 2019-06-04
Error handling causes customer data leak


ryanpalmerweb:
### Preconditions (*)
1. Magento 2.3.1

### Steps to reproduce (*)
1. Asynchronous email sending must be enabled in global settings
2. Backlog of unsent emails (e.g in Magento\Sales\Model\Order\Email\Sender\CreditmemoSender)
3. An error must be encountered during the getTransport method (e.g a plugin that references this tries to call an undefined function)
4. The application must run via cron so that any \Errors are automatically converted to \Exceptions by Magento\Framework\App\ErrorHandler.php

### Expected result (*)
The class Magento\Framework\Mail\Template\TransportBuilder is implemented using the Singleton pattern, so should reset its internal variables after an error to prevent customers from seeing other customers order details.

### Actual result (*)
The class Magento\Framework\Mail\Template\TransportBuilder is not reset, causing the customer who experienced the original issue to receive a copy of the next customers order/invoice/credit memo email.

### Background information ###

One of our clients was receiving complaints from customers that they were being emailed the wrong customers orders, and were therefore concerned about their own privacy.

We isolated the error and found that the Fooman EmailAttachments module was referencing the createAttachment function which existed in Magento 2.2 but was removed in Magento 2.3

In Magento 2.2 :
class Message extends \Zend_mail implements MessageInterface

In Magento 2.3 :
class Message implements MailMessageInterface

Searching the web for "Call to undefined method Magento\Framework\Mail\Message\Interceptor::createAttachment()" showed 68 results, meaning that this is not an isolated fault.

The probability of this happening to other installations as a result of breaking changes to the core is very high

### Suggested fix ###

The potential for a third party module to cause Magento to email the wrong customer is caused by the error handling in the SenderBuilder not resetting the TransportBuilder.

The send() function of the SenderBuilder should catch any errors, and reset the TransportBuilder to prevent this from happening.

Alternatively the TransportBuilder could be converted from a Singleton into a Factory.

ryanpalmerweb:
Just to add to this, one of our clients was receiving complaints from customers that they were being emailed other customers orders, and were concerned about their own privacy.

We isolated the error and found that the Fooman EmailAttachments module was referencing the createAttachment function which existed in Magento 2.2 but was removed in Magento 2.3

In Magento 2.2 : 
class Message extends \Zend_mail implements MessageInterface

In Magento 2.3 :  
class Message implements MailMessageInterface

[...]

PR #23139, opened 2019-06-05
SenderBuilder now implements TransportBuilder as a factory

ryanpalmerweb:
### Description (*)
The SenderBuilder class now implements the TransportBuilder as a factory rather than a singleton to prevent the possibility of customer data persisting and leaking, as it was previously reliant on a specific execution path to reset the state in between uses.

1. magento/magento2#23122 : Error handling causes customer data leak

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
+++ b/app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
@@ -50,7 +50,9 @@ class SenderBuilder
         $this->templateContainer = $templateContainer;
         $this->identityContainer = $identityContainer;
         $this->transportBuilder = $transportBuilder;
-        $this->transportBuilderFactory = $transportBuilderFactory ? $transportBuilderFactory : \Magento\Framework\App\ObjectManager::getInstance()->get(\Magento\Framework\Mail\Template\TransportBuilderFactory::class);
+        $this->transportBuilderFactory = $transportBuilderFactory
+          ? $transportBuilderFactory
+          : \Magento\Framework\App\ObjectManager::getInstance()->get(TransportBuilderFactory::class);
     }
 
     /**
@@ -72,7 +74,7 @@ class SenderBuilder
     public function send()
     {
         $this->resetTransportBuilder();
-
+        
         $this->configureEmailTemplate();
 
         $this->transportBuilder->addTo(
@@ -104,7 +106,7 @@ class SenderBuilder
         if (!empty($copyTo) && $this->identityContainer->getCopyMethod() == 'copy') {
             foreach ($copyTo as $email) {
                 $this->resetTransportBuilder();
-                
+
                 $this->configureEmailTemplate();
 
                 $this->transportBuilder->addTo($email);
--- a/app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
+++ b/app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
@@ -6,6 +6,7 @@
 namespace Magento\Sales\Model\Order\Email;
 
 use Magento\Framework\Mail\Template\TransportBuilder;
+use Magento\Framework\Mail\Template\TransportBuilderFactory;
 use Magento\Framework\Mail\Template\TransportBuilderByStore;
 use Magento\Sales\Model\Order\Email\Container\IdentityInterface;
 use Magento\Sales\Model\Order\Email\Container\Template;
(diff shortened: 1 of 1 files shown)


orlangur (changes requested):
Such change is not allowed, see https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/

app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
```diff
@@ -37,25 +38,41 @@ class SenderBuilder
...
         $this->transportBuilder = $transportBuilder;
+        $this->transportBuilderFactory = $transportBuilderFactory ? $transportBuilderFactory : \Magento\Framework\App\ObjectManager::getInstance()->get(\Magento\Framework\Mail\Template\TransportBuilderFactory::class);
     }
 
+    /**
+     * Create a new transport builder to stop data persisting between customers
+     *
+     * @return void
+     */
+    protected function resetTransportBuilder()
```
orlangur, line 63: Please make methods private whenever possible.

app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
```diff
@@ -86,6 +103,8 @@ public function sendCopyTo()
 
         if (!empty($copyTo) && $this->identityContainer->getCopyMethod() == 'copy') {
             foreach ($copyTo as $email) {
+                $this->resetTransportBuilder();
```
orlangur, line 108: For both occurrences: we must reset only after `TransportBuilder` is used for a first time.

orlangur (changes requested):
@ryanpalmerweb please check my comments, also, please provide actionable steps to reproduce.

ryanpalmerweb:
@orlangur 

I've broken down the problem here :

1. Magento\Sales\Model\Order\Email\Sender::checkAndSend calls SenderBuilder::send()
2. SenderBuilder::send() uses TransportBuilder::addTo() to populate the email addresses
3. Later in the above send() function, if TransportBuilder::getTransport() is called, the TransportBuilder is then reset
4. However if any exception is encountered, this is caught by the Sender::checkAndSend() function meaning that the **TransportBuilder is never reset**.
5. If we are processing a batch of invoices/credit memos/etc (e.g via cron), the next one to be processed will retain the email address populated in step 2, as well as the correct email address, causing a copy of the email to be sent to the original customer.

I appreciate that the core cannot be responsible for third party code, however as developers we cannot be certain that an exception will never occur, no matter how well written our code is, and we also know that given the sequence above, a data leak can easily happen if one is thrown.

It would surely make sense to rewrite this area defensively - I mean if the code can *never* go wrong, then why is there even a try/catch statement in there?  (@SerhiyShkolyarenko - do you have any input?)

Are you able to see the logic flow that I'm talking about here, and how any intermittent Exception within this area could lead to a customer data leak? 

You mentioned "we must reset only after TransportBuilder is used for a first time.", however the TransportBuilder class does not have any methods available to poll its internal state.

One possible solution is to reset the TransportBuilder class from the Sender::checkAndSend function whenever an exception has been caught, however as the reset method is private I assume we would either have to reinstantiate the entire object, or make the reset function public.

app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
ryanpalmerweb: Done

app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
ryanpalmerweb: I've requested guidance on this

ryanpalmerweb:
Just to add to the above, I'm looking at the technical guidelines

2.12. “Getters” SHOULD NOT change the state of an object.

However calling getTransport() on Magento\Framework\Mail\Template\TransportBuilder changes the state of the object by explicitly resetting it - so perhaps the existing implementation is already breaking your guidelines?

app/code/Magento/Sales/Model/Order/Email/SenderBuilder.php
sidolov: The method still protected, please, make it private

sidolov (changes requested):
Please, cover changes with automated tests

ihor-sviziev:
Hi @ryanpalmerweb,

ryanpalmerweb:
@ihor-sviziev Sorry, the core team don't really seem to acknowledge the issue and blamed a third party extension, so I'm not interested in fixing it.

ihor-sviziev:
@ryanpalmerweb this is not looks expected behavior as for me. Yeah, extensions definitely should be tested carefully, but if we could prevent such situation from happening - that would be good.
I just asked QAs to re-check if the issue is still valid, or it might be already fixed somehow

PR #23155, opened 2019-06-06
Fix for item quantity value in transactional emails when partially or fully canceled

kl3m3n:
### Description (*)
When you partially or fully cancel items from order then the wrong item quantity is shown in transactionals emails. The price is correct but the quantity is not. This fix adjust item quantity to correct amount in all transactional emails that uses layout handle "sales_email_order_items".

1. Create order with multiple items with 1 or more quantity
2. Cancel some items fully or cancel some items partially
3. Send order email from admin (or programatically) and you'll get original item quantity despite being canceled.

### Questions or comments
Affected line is 35 changed from 
`<td class="item-qty"><?= /* @escapeNotVerified */  $_item->getQtyOrdered() * 1) ?></td>`
to
`<td class="item-qty"><?= /* @escapeNotVerified */  ($_item->getQtyOrdered() * 1) - ($_item->getQtyCanceled() * 1) ?></td>`

### DIFF
--- a/app/code/Magento/Sales/view/frontend/templates/email/items/order/default.phtml
+++ b/app/code/Magento/Sales/view/frontend/templates/email/items/order/default.phtml
@@ -32,7 +32,7 @@ $_order = $_item->getOrder();
         <?php endif; ?>
         <?= $block->escapeHtml($_item->getDescription()) ?>
     </td>
-    <td class="item-qty"><?= /* @escapeNotVerified */  $_item->getQtyOrdered() * 1 ?></td>
+    <td class="item-qty"><?= /* @escapeNotVerified */  ($_item->getQtyOrdered() * 1) - ($_item->getQtyCanceled() * 1) ?></td>
     <td class="item-price">
         <?= /* @escapeNotVerified */  $block->getItemPrice($_item) ?>
     </td>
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23155) <br/>All committers have signed the CLA.

davidverholen:
Hi @kl3m3n ,

thank you for your contribution. In this repository the Magento code resides in the app/code directory and so the change has to be made there (That's why there is a whole new file instead of just a diff of the one line in the changes)

So for us to be able to process this pr, please make the change here:
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/view/frontend/templates/email/items/order/default.phtml

If you need help you can also have a look at the relatively new contribution beginners guide: https://github.com/magento/magento2/wiki/Getting-Started

orlangur:
@davidverholen why we need `on hold` here?

davidverholen:
@orlangur just wanted to make sure its not processed further until the change is in the correct path

orlangur:
@davidverholen it does not work this way yet, single approval will remove both labels :(

davidverholen:
@orlangur yes, I know :/ My hope was, that people are more careful with approving when the on hold label is set

davidverholen (changes requested):
move changes to correct file location at https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Sales/view/frontend/templates/email/items/order/default.phtml

davidverholen:
@orlangur but right, I can request changes in a review which will block it also when someone else approves it. I will hide our comments now to reduce noise here

PR #23188, opened 2019-06-10
Fix for item quantity value in transactional emails when partially or fully canceled

kl3m3n:
### Description (*)
When you partially or fully cancel items from order then the wrong item quantity is shown in transactionals emails. The price is correct but the quantity is not. This fix adjust item quantity to correct amount in all transactional emails that uses layout handle "sales_email_order_items".

1. Create order with multiple items with 1 or more quantity
2. Cancel some items fully or cancel some items partially
3. Send order email from admin (or programatically) and you'll get original item quantity despite being canceled.

### Questions or comments
Affected line is 35 changed from 
`<td class="item-qty"><?= /* @escapeNotVerified */  $_item->getQtyOrdered() * 1) ?></td>`
to
`<td class="item-qty"><?= /* @escapeNotVerified */  ($_item->getQtyOrdered() * 1) - ($_item->getQtyCanceled() * 1) ?></td>`

### DIFF
--- a/app/code/Magento/Sales/view/frontend/templates/email/items/order/default.phtml
+++ b/app/code/Magento/Sales/view/frontend/templates/email/items/order/default.phtml
@@ -32,7 +32,7 @@ $_order = $_item->getOrder();
         <?php endif; ?>
         <?= $block->escapeHtml($_item->getDescription()) ?>
     </td>
-    <td class="item-qty"><?= /* @escapeNotVerified */  $_item->getQtyOrdered() * 1 ?></td>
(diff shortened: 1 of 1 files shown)


phoenix128:
Hello @kl3m3n,
thank you for your contribution, but I will close this PR because the behavior you are trying to fix is the expected one. Order email should be presenting the information of the original user's request.

If you are canceling one or more items from one order, the communication should be handled by credit memo email template in the `sales_email_order_creditmemo_items` layout section.

kl3m3n:
I disagree with yours reason for closure, phoenix128. The quantity is original bit the price will change? Why is the price set to 0,00 (or reduced) if user should see original request? This is just confusing for customers and they expect products for free. That's why we had to fix this part to let them know they'll not get products for free. In case of just reducing the item numbers on order they need to know exact amount of products they'll get, not the original one.

Issue #15656, opened 2018-06-01
Duplicate product


rommelfreddy:
### Preconditions

1.  Magento 2.2.2
2. Existing product with different values in different StoreViews (e.g StoreView UK, NL, DE, ...)

### Steps to reproduce

1. open the product
2. select one StoreView (e.g. UK - not "All StoreViews"!)
3. click on "Save product & duplicate"

### Expected result

1. the product got be duplicated with all (different) values in each storeView

### Actual result

1. the default values of the **new** product are not the default values from the **original** product. Instead, the values from the selected storeView will be used to create the new product.

Example (Productname):

_Original product:_
StoreView Default: "My default name"
StoreView UK: "My UK name"
StoreView NL: "My NL name"
...

_Duplicated product:_ (before clicking on duplicate, I switch to the storeView "UK")
StoreView Default: "My **UK** name"
StoreView UK: "My UK name"
StoreView NL: "My NL name"
...

rogyar:
I believe that's expected behavior. You have a possibility to choose the exact product's variant to duplicate. If you need default values - choose the global scope, otherwise choose the variant that should be duplicated.

ghost:
@Fredwak, thank you for your report.
We've acknowledged the issue and added to our backlog.

magento-engcom-team:
Hi @shikhamis11. Thank you for working on this issue. 
Looks like this issue is already verified and confirmed. But if your want to validate it one more time, please, go though the following instruction:
- [ ] 1. Add/Edit `Component: XXXXX` label(s) to the ticket, indicating the components it may be related to.

- [ ] 2. Verify that the issue is reproducible on `2.3-develop` branch<details><summary>Details</summary>- Add the comment `@magento-engcom-team give me 2.3-develop instance` to deploy test instance on Magento infrastructure. <br/> - If the issue is reproducible on `2.3-develop` branch, please, add the label `Reproduced on 2.3.x`.<br/>- If the issue is not reproducible, add your comment that issue is not reproducible and close the issue and **_stop verification process here_**!
</details>

- [ ] 3. Verify that the issue is reproducible on `2.2-develop` branch. <details><summary>Details</summary>- Add the comment `@magento-engcom-team give me 2.2-develop instance` to deploy test instance on Magento infrastructure. <br/>- If the issue is reproducible on `2.2-develop` branch, please add the label `Reproduced on 2.2.x`</details>

- [ ] 4. If the issue is not relevant or is not reproducible any more, feel free to close it.

[...]

PR #23209, opened 2019-06-11
Load product in default store before duplicating

Jeroenll:
### Description (*)
Duplicating a product will use the current scope data as default values instead of copying it from the original product.

1. magento/magento2#15656: Duplicate product

1. Create product with different values in different StoreViews (e.g StoreView UK, NL, DE, ...)
2. Open product and select one StoreView (e.g. UK - not "All StoreViews"!)
3. Click on "Save product & duplicate"

### Questions or comments
Not really happy with reloading the product while we have it already as method parameter. Maybe deprecate the `copy()` method and create a `copyBySku()` method?

Will update tests if implementation is approved by Magento.

### DIFF
--- a/app/code/Magento/Bundle/Controller/Adminhtml/Bundle/Product/Edit/Duplicate.php
+++ b/app/code/Magento/Bundle/Controller/Adminhtml/Bundle/Product/Edit/Duplicate.php
@@ -6,6 +6,9 @@
  */
 namespace Magento\Bundle\Controller\Adminhtml\Bundle\Product\Edit;
 
+/**
+ * @deprecated
+ */
 class Duplicate extends \Magento\Catalog\Controller\Adminhtml\Product\Duplicate
 {
 }
--- a/app/code/Magento/Catalog/Block/Adminhtml/Product/Edit.php
+++ b/app/code/Magento/Catalog/Block/Adminhtml/Product/Edit.php
@@ -242,6 +242,7 @@ class Edit extends \Magento\Backend\Block\Widget
 
     /**
      * @return string
+     * @deprecated
      */
     public function getDuplicateUrl()
     {
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Duplicate.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Duplicate.php
@@ -9,6 +9,9 @@ namespace Magento\Catalog\Controller\Adminhtml\Product;
 use Magento\Backend\App\Action;
 use Magento\Catalog\Controller\Adminhtml\Product;
 
+/**
+ * @deprecated
+ */
 class Duplicate extends \Magento\Catalog\Controller\Adminhtml\Product
 {
     /**
--- a/app/code/Magento/Downloadable/Controller/Adminhtml/Downloadable/Product/Edit/Duplicate.php
+++ b/app/code/Magento/Downloadable/Controller/Adminhtml/Downloadable/Product/Edit/Duplicate.php
@@ -6,6 +6,9 @@
  */
 namespace Magento\Downloadable\Controller\Adminhtml\Downloadable\Product\Edit;
 
+/**
+ * @deprecated
+ */
 class Duplicate extends \Magento\Catalog\Controller\Adminhtml\Product\Duplicate
 {
 }
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Save.php
@@ -157,9 +157,19 @@ class Save extends \Magento\Catalog\Controller\Adminhtml\Product implements Http
                 );
 
                 if ($redirectBack === 'duplicate') {
-                    $product->unsetData('quantity_and_stock_status');
-                    $newProduct = $this->productCopier->copy($product);
-                    $this->checkUniqueAttributes($product);
+                    if ($product->getStoreId() === \Magento\Store\Model\Store::DEFAULT_STORE_ID) {
+                        $productToDuplicate = $product;
+                    } else {
+                        $productToDuplicate = $this->productRepository->get(
+                            $product->getSku(),
+                            false,
+                            \Magento\Store\Model\Store::DEFAULT_STORE_ID,
+                            true
+                        );
+                    }
+
+                    $productToDuplicate->unsetData('quantity_and_stock_status');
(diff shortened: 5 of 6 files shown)


dmytro-ch (changes requested):
Hi @JeroenVanLeusden!

> Not really happy with reloading the product while we have it already as method parameter. Maybe deprecate the copy() method and create a copyBySku() method?

I also agree, it would be better to pass the already correct product data as `copy` method parameter.
So we may try to load the product via repository in case the `current_store_id` value of the product is not equal to  `Store::DEFAULT_STORE_ID`, instead of using the Product Builder in the following controller actions: 
- `\Magento\Catalog\Controller\Adminhtml\Product\Duplicate::execute`
- `\Magento\Catalog\Controller\Adminhtml\Product\Save::execute`

The only problem, in my opinion, is that the product entity after loading does not contain all the necessary data. I guess, the bunch of data entries still need to be populated from request.

Thank you!

Jeroenll:
Hi @dmytro-ch,

I can refactor the `\Magento\Catalog\Controller\Adminhtml\Product\Save::execute` te be:
```php
if ($redirectBack === 'duplicate') {
    if ($product->getStoreId() === \Magento\Store\Model\Store::DEFAULT_STORE_ID) {
        $product->unsetData('quantity_and_stock_status');
        $productToDuplicate = $product;
    } else {
        $productToDuplicate = $this->productRepository->get(
            $product->getSku(),
            false,
            \Magento\Store\Model\Store::DEFAULT_STORE_ID,
            true
        );
    }

    $newProduct = $this->productCopier->copy($productToDuplicate);
    $this->messageManager->addSuccessMessage(__('You duplicated the product.'));
}
```

Although i'm not sure what you mean with:
> in my opinion, is that the product entity after loading does not contain all the necessary data. I guess, the bunch of data entries still need to be populated from request.

The product is already saved so all the data we need should be there already when fetched using the repository?

Also, I've found no reference to `\Magento\Catalog\Controller\Adminhtml\Product\Duplicate::execute`. Can we deprecate this class?

dmytro-ch:
@JeroenVanLeusden,
thank you for the update.

The new solution looks correct to me. 

> Although i'm not sure what you mean with:
> 
> > in my opinion, is that the product entity after loading does not contain all the necessary data. I guess, the bunch of data entries still need to be populated from request.
> 
> 
> The product is already saved so all the data we need should be there already when fetched using the repository?

I mean if you compare the results of `$product->getData()` and `$productToDuplicate->getData()` according to your code snippet, you may get the different number of values in resulting array. We just need to make sure the duplicated product does not lose any data.

> Also, I've found no reference to \Magento\Catalog\Controller\Adminhtml\Product\Duplicate::execute.
> Can we deprecate this class?

Sure

Jeroenll:
Hi @dmytro-ch,

I've updated the PR, still using the repository to load the product because it duplicates just fine. Are there some additional tests I can do to make sure it's all good?

sidolov (approved):

magento-engcom-team:
[ENGCOM-5689](https://jira.corp.magento.com/browse/ENGCOM-5689) has been created to process this Pull Request

| Label | Description |

engcom-Delta:
Hi @JeroenVanLeusden during testing PR changes I found issue with product duplicate when different URL Key is set for each store view.
**Manual testing scenario:**

- Create more than one store view:
image
- Create product
- Edit product for each store view and save:
 - - **Default Store View:**
- - - Product Name=simple1 default
- - - URL Key=simple1-def
- - **Second View:**
- - - Product Name=simple1 second
- - - URL Key=simple1-second
- - **Third View:**
- - - Product Name=simple1 third
- - - URL Key=simple1-third
- - **Notice:** **Create Permanent Redirect for old URL** is checked on each store view when URL key is edited
image
- Open product and select Second View
- Click on "Save product & duplicate"

**Result:**
:x: Product Name value is set to **All Store Views**(e.g., "simple1") value for each store view
#23209PRDuplicate
:heavy_check_mark: URL keys is generated correctly using store view value for it
#23209PRURLKey

Could you take a look?

app/code/Magento/Bundle/Controller/Adminhtml/Bundle/Product/Edit/Duplicate.php
```diff
@@ -6,6 +6,9 @@
  */
 namespace Magento\Bundle\Controller\Adminhtml\Bundle\Product\Edit;
 
+/**
+ * @deprecated
```
ihor-sviziev, line 10: Please add description why it became deprecated and add `@see` tag with example what we should use instead

app/code/Magento/Catalog/Block/Adminhtml/Product/Edit.php
```diff
@@ -242,6 +242,7 @@ public function getProductSetId()
 
     /**
      * @return string
+     * @deprecated
```
ihor-sviziev, line 245: Please add description why it became deprecated and add `@see` tag with example what we should use instead

app/code/Magento/Catalog/Controller/Adminhtml/Product/Duplicate.php
```diff
@@ -9,6 +9,9 @@
 use Magento\Backend\App\Action;
 use Magento\Catalog\Controller\Adminhtml\Product;
 
+/**
+ * @deprecated
```
ihor-sviziev, line 13: Please add description why it became deprecated and add `@see` tag with example what we should use instead

app/code/Magento/Downloadable/Controller/Adminhtml/Downloadable/Product/Edit/Duplicate.php
```diff
@@ -6,6 +6,9 @@
  */
 namespace Magento\Downloadable\Controller\Adminhtml\Downloadable\Product\Edit;
 
+/**
+ * @deprecated
```
ihor-sviziev, line 10: Please add description why it became deprecated and add `@see` tag with example what we should use instead

ihor-sviziev (changes requested):
Hi @JeroenVanLeusden,

This PR is really closely related to https://github.com/magento/magento2/pull/24505 and looks like changes from this PR could fix issue that described @engcom-Delta.
Could you collaborate with @webkul-paul to fix both these issues?

sidolov:
@JeroenVanLeusden , I am closing this PR now due to inactivity.

PR #23219, opened 2019-06-12
Fix retrieving remote address when Magento behind reverse proxy

ihor-sviziev:
### Description (*)
Add `X-Forwarded-For` header to list of alternate headers for getting remote address.
This is default header that common used by well known reverse proxies like nginx, apache, varnish and many services.

More info: https://dev98.de/2017/01/02/how-to-add-alternative-http-headers-to-magento-2/

N/A

1. Setup Magento with reverse proxy, nginx for instance
2. Try to get client IP address using following method: `\Magento\Framework\HTTP\PhpEnvironment\RemoteAddress::getRemoteAddress`

**Actual result:**
Retrieved address - IP address of reverse proxy

**Expected result:**
Retrieved address - client's IP address

### DIFF
--- a/app/etc/di.xml
+++ b/app/etc/di.xml
@@ -211,6 +211,13 @@
     <preference for="Magento\Framework\MessageQueue\QueueFactoryInterface" type="Magento\Framework\MessageQueue\QueueFactory" />
     <preference for="Magento\Framework\Search\Request\IndexScopeResolverInterface" type="Magento\Framework\Indexer\ScopeResolver\IndexScopeResolver"/>
     <preference for="Magento\Framework\HTTP\ClientInterface" type="Magento\Framework\HTTP\Client\Curl" />
+    <type name="Magento\Framework\HTTP\PhpEnvironment\RemoteAddress">
+        <arguments>
+            <argument name="alternativeHeaders" xsi:type="array">
+                <item name="x-forwarded-for" xsi:type="string">HTTP_X_FORWARDED_FOR</item>
+            </argument>
+        </arguments>
+    </type>
     <type name="Magento\Framework\Model\ResourceModel\Db\TransactionManager" shared="false" />
     <type name="Magento\Framework\Acl\Data\Cache">
         <arguments>


kassner:
Given that the header can be easily forged, isn't it better to leave this decision for each deployment?

ihor-sviziev:
> Given that the header can be easily forged, isn't it better to leave this decision for each deployment?

As recommended production installation should have Varnish - it's mandatory. Otherwise you will always get IP address of varnish node (or 127.0.0.1 when it installed locally)

kassner:
@ihor-sviziev AFAIK, Varnish is not mandatory, just [recommended](https://devdocs.magento.com/guides/v2.3/performance-best-practices/software.html).

This PR will introduce a security issue to instances that are not running an `optional` software.

ihor-sviziev:
> @ihor-sviziev AFAIK, Varnish is not mandatory, just [recommended](https://devdocs.magento.com/guides/v2.3/performance-best-practices/software.html).
> 
> This PR will introduce a security issue to instances that are not running an `optional` software.

I meant as varnish installation is recommended - this issue will affect a lot of people, so such fix is needed.
Could you contact me in Slack to make sure we have/don't have security issue in my solution? I tried to contact with you, but no response :(

kassner:
@ihor-sviziev the solution is quite simple: don't parse the header unless the admin explicitly enabled this.

You could do that via an admin configuration, or adding those same lines to a di.xml in a module specific to your deployment. The solution might as well be just changing the documentation to make people aware of it.

You also should make sure you can't spoof this even with Varnish (i.e.: the VCL injects the header regardless, and Magento uses is the last IP in the `X-Forwarded-For` header).

ihor-sviziev:
Hi @kassner,
It's still not clear to me what's causing security issue. It think it will be really much more productive to talk in Slack and discuss that. Please ping me in Slack when you'll be available.
Thank you!

FreekVandeursen:
@ihor-sviziev For any customer who is not behind a reverse proxy, the X-Forwarded-For header can easily be set by the customer. And since getRemoteAddress will use it, any functionality that depends on the IP address in Magento could be tricked into returning a false IP address that is entered into the fake header by the customer.

ihor-sviziev:
Thank you @kassner  and @FreekVandeursen. Looks reasonable.

For not I'm closing this PR. Will update it and re-open when will have a time.

PR #23293, opened 2019-06-17
#21361 'Bestsellers collection is using ratingLimit as sql select limit'

yuriichayka:
Adding possibility for change range of bestseller items to be displayed

Drop-down was added with predefined set of values for changing number of items to be displayed

1. ...
2. ...

### DIFF
--- a/app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
+++ b/app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
@@ -7,8 +7,11 @@ declare(strict_types=1);
 
 namespace Magento\Reports\Block\Adminhtml\Bestsellers\Filter;
 
+use Magento\Backend\Block\Template\Context;
 use Magento\Framework\Data\Form\Element\Fieldset;
-use \Magento\Reports\Block\Adminhtml\Filter\Form as ReportsBlockFilterForm;
+use Magento\Framework\Data\FormFactory;
+use Magento\Framework\Registry;
+use Magento\Reports\Block\Adminhtml\Filter\Form as ReportsBlockFilterForm;
 
 /**
  * Sales Adminhtml report bestseller filter form
@@ -19,6 +22,33 @@ use \Magento\Reports\Block\Adminhtml\Filter\Form as ReportsBlockFilterForm;
 class Form extends ReportsBlockFilterForm
 {
     /**
+     * Sorting limits for items displayed
+     *
+     * @var []
+     */
+    private $itemSortLimit;
+
+    /**
+     * Form constructor.
+     *
+     * @param Context $context
+     * @param Registry $registry
+     * @param FormFactory $formFactory
+     * @param array $itemSortLimit
+     * @param array $data
+     */
+    public function __construct(
+        Context $context,
+        Registry $registry,
+        FormFactory $formFactory,
+        array $itemSortLimit,
+        array $data = []
+    ) {
+        $this->itemSortLimit = $itemSortLimit;
+        parent::__construct($context, $registry, $formFactory, $data);
+    }
+
+    /**
      * @inheritdoc
      */
     protected function _prepareForm()
@@ -26,14 +56,14 @@ class Form extends ReportsBlockFilterForm
         parent::_prepareForm();
         /** @var Fieldset $fieldset */
         $fieldset = $this->getForm()->getElement('base_fieldset');
-        if (is_object($fieldset) && $fieldset instanceof Fieldset) {
+        if ($fieldset instanceof Fieldset) {
             $fieldset->addField(
                 'rating_limit',
                 'select',
                 [
-                    'name'    => 'rating_limit',
-                    'label'   => __('Display items'),
-                    'options' => array_combine($i = [5, 10, 20, 50, 100], $i),
+                    'name' => 'rating_limit',
+                    'label' => __('Display items'),
+                    'options' => $this->itemSortLimit
                 ],
                 'to'
             );
--- a/app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
+++ b/app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Reports\Block\Adminhtml\Bestsellers\Filter;
+
+/**
+ * Class Form
+ *
+ * @package Magento\Reports\Block\Adminhtml\Bestsellers\Filter
+ */
+class Form extends \Magento\Reports\Block\Adminhtml\Filter\Form
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function _prepareForm()
+    {
+        parent::_prepareForm();
+        /** @var \Magento\Framework\Data\Form\Element\Fieldset $fieldset */
+        $fieldset = $this->getForm()->getElement('base_fieldset');
+        if (is_object($fieldset) && $fieldset instanceof \Magento\Framework\Data\Form\Element\Fieldset) {
+            $fieldset->addField(
+                'rating_limit',
+                'select',
+                [
+                    'name' => 'rating_limit',
+                    'label' => __('Display items'),
+                    'options' => array_combine($i = [5, 10, 20, 50, 100], $i),
+                ],
+                'to'
+            );
+        }
+        return $this;
+    }
+}
--- a/app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
+++ b/app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
@@ -3,36 +3,41 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+declare(strict_types=1);
 
 namespace Magento\Reports\Block\Adminhtml\Bestsellers\Filter;
 
+use Magento\Framework\Data\Form\Element\Fieldset;
+use \Magento\Reports\Block\Adminhtml\Filter\Form as ReportsBlockFilterForm;
+
 /**
- * Class Form
+ * Sales Adminhtml report bestseller filter form
  *
  * @package Magento\Reports\Block\Adminhtml\Bestsellers\Filter
  */
-class Form extends \Magento\Reports\Block\Adminhtml\Filter\Form
+class Form extends ReportsBlockFilterForm
 {
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     protected function _prepareForm()
     {
         parent::_prepareForm();
-        /** @var \Magento\Framework\Data\Form\Element\Fieldset $fieldset */
+        /** @var Fieldset $fieldset */
         $fieldset = $this->getForm()->getElement('base_fieldset');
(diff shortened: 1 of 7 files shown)


app/code/Magento/Reports/Block/Adminhtml/Sales/Bestsellers/Grid.php
```diff
@@ -40,20 +41,30 @@ public function getResourceCollectionName()
...
             'period',
             [
-                'header' => __('Interval'),
-                'index' => 'period',
-                'sortable' => false,
-                'period_type' => $this->getPeriodType(),
-                'renderer' => \Magento\Reports\Block\Adminhtml\Sales\Grid\Column\Renderer\Date::class,
-                'totals_label' => __('Total'),
-                'html_decorators' => ['nobr'],
+                'header'           => __('Interval'),
```
novikor: Please revert formatting changes. We don`t need to align array key => values

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
```diff
@@ -0,0 +1,43 @@
...
+                    'label'   => __('Display items'),
+                    'options' => array_combine($i = [5, 10, 20, 50, 100], $i),
+                ],
+                'to'
+            );
+        }
+
+        return $this;
+    }
+}
```
novikor: Empty line is missing

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Reports\Block\Adminhtml\Bestsellers\Filter;
+
+use Magento\Framework\Data\Form\Element\Fieldset;
+use \Magento\Reports\Block\Adminhtml\Filter\Form as ReportsBlockFilterForm;
```
novikor: Please avoid usage of leading `\`

app/code/Magento/Sales/Model/ResourceModel/Report/Bestsellers/Collection.php
```diff
@@ -37,25 +49,39 @@ class Collection extends \Magento\Sales\Model\ResourceModel\Report\Collection\Ab
...
     /**
-     * @param \Magento\Framework\Data\Collection\EntityFactory $entityFactory
-     * @param \Psr\Log\LoggerInterface $logger
-     * @param \Magento\Framework\Data\Collection\Db\FetchStrategyInterface $fetchStrategy
-     * @param \Magento\Framework\Event\ManagerInterface $eventManager
-     * @param \Magento\Sales\Model\ResourceModel\Report $resource
-     * @param \Magento\Framework\DB\Adapter\AdapterInterface $connection
+     * @param EntityFactory          $entityFactory
+     * @param LoggerInterface        $logger
+     * @param FetchStrategyInterface $fetchStrategy
```
novikor, line 57: Please revert the phpdoc elements align formatting changes

app/code/Magento/Sales/Model/ResourceModel/Report/Bestsellers/Collection.php
```diff
@@ -91,11 +118,11 @@ protected function _getSelectedColumns()
                 $this->_selectedColumns = $this->getAggregatedColumns();
             } else {
                 $this->_selectedColumns = [
-                    'period' => sprintf('MAX(%s)', $connection->getDateFormatSql('period', '%Y-%m-%d')),
+                    'period'                 => sprintf('MAX(%s)', $connection->getDateFormatSql('period', '%Y-%m-%d')),
```
novikor: Please revert key => value formatting

novikor (changes requested):
Hi, @yuriichayka . Thanks for your contribution.
Please pay attention to my notices, and make sure tests are green.

The review is not finished yet - but we have a scope of work there already,

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * @inheritdoc
+     */
+    protected function _prepareForm()
+    {
+        parent::_prepareForm();
+        /** @var Fieldset $fieldset */
+        $fieldset = $this->getForm()->getElement('base_fieldset');
+        if (is_object($fieldset) && $fieldset instanceof Fieldset) {
```
novikor: I believe `is_object` check is not needed since we already have `instanceof` there

novikor (changes requested):
So, your solution is generally fine, BUT.
But `Magento\Sales\Block\Adminhtml\Report\Filter\Form` is marked as `@api` so we cannot just replace it because of backward compatibility break.

I suggest to act in the following way:
- Leave `\Magento\Reports\Block\Adminhtml\Bestsellers\Filter\Form` as it is
- Copy changes to `\Magento\Sales\Block\Adminhtml\Report\Filter\Form`

So - yes, in that case, the newly created class will not be used anywhere at the moment.
I have marked `\Magento\Sales\Block\Adminhtml\Report\Filter\Form` as deprecated - so it will be replaced in 2.4.

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
```diff
@@ -0,0 +1,44 @@
...
+            $fieldset->addField(
+                'rating_limit',
+                'select',
+                [
+                    'name'    => 'rating_limit',
+                    'label'   => __('Display items'),
+                    'options' => array_combine($i = [5, 10, 20, 50, 100], $i),
```
novikor: That`s fine.
However, it would be to move these hardcoded values into di.xml

novikor (changes requested):
Please check one more point

yuriichayka:
> So, your solution is generally fine, BUT.
> But `Magento\Sales\Block\Adminhtml\Report\Filter\Form` is marked as `@api` so we cannot just replace it because of backward compatibility break.
> 
> I suggest to act in the following way:
> 
>     * Leave `\Magento\Reports\Block\Adminhtml\Bestsellers\Filter\Form` as it is
> 
>     * Copy changes to `\Magento\Sales\Block\Adminhtml\Report\Filter\Form`
> 
> 
> So - yes, in that case, the newly created class will not be used anywhere at the moment.
> I have marked `\Magento\Sales\Block\Adminhtml\Report\Filter\Form` as deprecated - so it will be replaced in 2.4.

Hello, 
Thanks for the review. 

- If I copy changes to '\Magento\Sales\Block\Adminhtml\Report\Filter\Form' - limit for items displaying will appear for some other entities such as Coupons reports, but our aim is to add it only for Bestsellers.
- Maybe extending from _\Magento\Sales\Block\Adminhtml\Report\Filter\Form_ but not from _Magento\Reports\Block\Adminhtml\Filter\Form_(as it is now) can be a solution for backward compatibility here? 
- If not, could you please clarify how can I make the implemented limit functionality being applied only for Bestsellers reports after copying changes to _`\Magento\Sales\Block\Adminhtml\Report\Filter\Form`_

Thanks in advance.

app/code/Magento/Reports/Block/Adminhtml/Sales/Bestsellers/Grid.php
yuriichayka: Done. Associative arrays alignment was removed.

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
yuriichayka: Done. Empty line was added.

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
yuriichayka: fixed

app/code/Magento/Sales/Model/ResourceModel/Report/Bestsellers/Collection.php
yuriichayka: fixed

app/code/Magento/Sales/Model/ResourceModel/Report/Bestsellers/Collection.php
yuriichayka: Done. Associative arrays alignment was removed

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
yuriichayka: Redundant condition was removed

app/code/Magento/Reports/Block/Adminhtml/Bestsellers/Filter/Form.php
yuriichayka: Done. The values array was added to adminhtml/di.xml via argument replacement.

yuriichayka:
Hello @novikor 

Tests for this pull request are failed due to` # `symbol in branch name. 
I`ll recreate this one with correct name, but I need to have answer on my last question regarding backward compatibility. Please check this one https://github.com/magento/magento2/pull/23293#issuecomment-504752705

novikor:
@yuriichayka , - option 2 seems to be applicable: all the arguments provided using DI will be preserved. In case DI `preference` - nothing will break.

Anyway, please leave this line of code as it is: we still need some refactoring for the upcoming major release. https://github.com/magento/magento2/pull/23293/commits/938acb448ef68594ab9e2699a107b4f7b9b5cd75#diff-0faebc6cb763a1c639b40624e2009051R16

@sivaschenko , do you agree with me about the backward compatibility?

sivaschenko:
@yuriichayka @novikor I think the best possible approach to minimizing backwards incompatible changes here is to add `rating_limit` select to form using a plugin (i.e. afterGetForm). 

A condition can be added to the plugin to ensure the filter is added only for Bestsellers reports. In case there is not enough information to identify that - additional argument can be added to a block via layout configuration (please fallback to adding additional argument only if there is no other way to solve it).

Additionally: Plugin is a solution to update Sales block from Report module. In case you'd like to keep the functionality in Sales module - the @novikor proposition looks good to me: add a field to `Magento\Sales\Block\Adminhtml\Report\Filter\Form`, the same condition can be used to ensure the filter is added only when needed

yuriichayka:
> @yuriichayka @novikor I think the best possible approach to minimizing backwards incompatible changes here is to add `rating_limit` select to form using a plugin (i.e. afterGetForm).
> 
> A condition can be added to the plugin to ensure the filter is added only for Bestsellers reports. In case there is not enough information to identify that - additional argument can be added to a block via layout configuration (please fallback to adding additional argument only if there is no other way to solve it).
> 
> Additionally: Plugin is a solution to update Sales block from Report module. In case you'd like to keep the functionality in Sales module - the @novikor proposition looks good to me: add a field to `Magento\Sales\Block\Adminhtml\Report\Filter\Form`, the same condition can be used to ensure the filter is added only when needed

Hello @sivaschenko 

Thanks for the help. 
I've added field to Magento\Sales\Block\Adminhtml\Report\Filter\Form with block layout condition(didn't find other solution)

This pull request will be closed, see a new one with correct branch name - https://github.com/magento/magento2/pull/23461

PR #23304, opened 2019-06-18
magento/magento2#23301: MethodArgumentsSniff works incorrect with arguments passed by reference.

p-bystritsky:

1.magento/magento2#23301: MethodArgumentsSniff works incorrect with arguments passed by reference.

1. Add any changes to \Magento\CatalogImportExport\Model\Import\Product\Option class.
2. Run \Magento\Test\Php\LiveCodeTest::testCodeStyle method.

### DIFF
--- a/dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
+++ b/dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
@@ -417,7 +417,7 @@ class MethodArgumentsSniff implements Sniff
                 );
                 break;
             case 1:
-                if (preg_match('/^\$.*/', $paramDefinitions[0])) {
+                if (preg_match('/^&?\$.*/', $paramDefinitions[0])) {
                     $phpcsFile->addError(
                         'Type is not specified',
                         $paramPointers[$ptr],
@@ -442,7 +442,7 @@ class MethodArgumentsSniff implements Sniff
                 );
                 break;
             default:
-                if (preg_match('/^\$.*/', $paramDefinitions[0])) {
+                if (preg_match('/^&?\$.*/', $paramDefinitions[0])) {
                     $phpcsFile->addError(
                         'Type is not specified',
                         $paramPointers[$ptr],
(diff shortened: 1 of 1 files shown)


dev/tests/static/framework/Magento/Sniffs/Annotation/MethodArgumentsSniff.php
```diff
@@ -681,4 +681,17 @@ private function noneParamsAligned(array $argumentPositions, array $commentPosit
...
     }
+
+    /**
+     * Get argument name starting position in annotation string.
+     *
+     * @param string $paramContent
+     * @param array $paramDefinition
+     * @return int
+     */
+    private function getArgumentPosition(string $paramContent, array $paramDefinition): int
```
phoenix128: I am not 100% this can happen, but looking at the code, seems that params are retrieved from annotations in `\Magento\Sniffs\Annotation\MethodArgumentsSniff::process`. Did you test what happens if a param is missing?
Can `strpos` give a `false` result in this condition? If this happens you could get a strict type validation error since you specified `int` as return type.

Thank you.

Thank you.

engcom-Foxtrot: @phoenix128 thank you for review. You're right about this concern. I've updated the code.

p-bystritsky: @phoenix128 thank you for review. You're right about this concern. I've updated the code.

orlangur (changes requested):
Such syntax is invalid, there is nothing to fix here in sniff. Please fix such occurrences in phpdocs instead.

p-bystritsky:
@orlangur thank you for review. Can you please provide argumentation why do you think this syntax is not correct?

orlangur:
@p-bystritsky nope, it's YOU should provide information why such syntax makes any sense.

Prior to leaving my review note I've checked how PhpStorm generates PhpDoc for such method signature and proposed PhpDoc standard in PHP_FIG.

sidolov:
@p-bystritsky according to the https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types definition ampersand should be used for intersection type. For variables passed by reference, such docblock is invalid. DocBlock should be fixed instead of code sniff. I'm closing this PR as not relevant.

PR #23323, opened 2019-06-19
[23316] Invalid Base URL. Value must be a URL or one of placeholders: {{base_url}}

tnsezer:
Referring issue: https://github.com/magento/magento2/issues/23316

### Preconditions (*)

1. Version 2.3.1

### Steps to reproduce (*)

1. Try to enter a new base_url like "http://bmt.consulting".

### Expected result (*)

1. Change the base URL.

### Actual result (*)
Result: `Invalid Base URL. Value must be a URL or one of placeholders: {{base_url}}`
magento_error_baseurl

1. I modified the value of base_url in MySQL so now it works. However, it seemed like I was still unable to modify anything on the store's web administration page. I incorrectly assumed that it was because of a new style of consulting website address, but finally, I realized that I could add a slash to the end of the URL and it could be accepted. The message `Invalid Base URL` could more clearly why the validation failed. Here, this error occurred simply because I did not have a trailing slash on my base URL.

This issue took a bit of time to resolve. It seems that it was just a missing trailing slash in my base_url. The server seemed to work fine when I had set it via MySQL without a trailing slash. 

My suggestion is to make the error message more clear or to automatically put a trailing slash on website addresses rather than have them fail. The tests at lib/internal/Magento/Framework/Validator/Test/Unit/UrlTest.php don't fail so I don't know why my address without trailing slash gets picked up.


sidolov:
Hi @tnsezer could you please take a look at failed tests?

sidolov:
@tnsezer I have reviewed `\Magento\Config\Model\Config\Backend\Baseurl` class and I believe such cases should be handled by `\Magento\Config\Model\Config\Backend\Baseurl::_validateFullyQualifiedUrl` method and return to the user meaningful message that url should be fully qualified. 

Looks like something went wrong with it, I propose to investigate why we don't receive a specified message instead of url fixing.

sidolov:
@tnsezer could you please take a look at previous comment? I believe we can fix the issue in a different way

tnsezer:
> @tnsezer could you please take a look at previous comment? I believe we can fix the issue in a different way
@sidolov 
You are right but it is useless if we ignore the request. The slash should not be an issue but the method can reject the URL if it doesn't match qualification

sidolov:
@tnsezer please, take a look at failed integration tests

engcom-Foxtrot:
@sidolov updated.

engcom-Foxtrot:
@tnsezer please, sign CLA.

ihor-sviziev:
@magento run all tests

tnsezer:
@magento run all tests

VladimirZaets:
@magento run all tests

app/code/Magento/ConfigurableProduct/Model/Product/VariationHandler.php
```diff
@@ -239,6 +239,10 @@ public function duplicateImagesForVariations($productsData)
 
             foreach ($simpleProductData['media_gallery']['images'] as $imageId => $image) {
                 $image['variation_id'] = $variationId;
+                if (isset($imagesForCopy[$imageId][0])) {
+                    // skip duplicate image for first product
+                    unset($imagesForCopy[$imageId][0]);
+                }
```
ihor-sviziev, line 245: Hi @tnsezer ,
Seems like this change isn't related to this PR and causing functional and integration test failures. Could you revert it?
```suggestion
```

ihor-sviziev (changes requested):

app/code/Magento/ConfigurableProduct/Model/Product/VariationHandler.php
aredridel: Hi! @ihor-sviziev I am not actually a member of the team nor involved in this PR.

ihor-sviziev: @aredridel oh sorry, just mentioned wrong account. I fixed already my comment

engcom-Charlie:
@tnsezer could you please look at https://github.com/magento/magento2/pull/23323#discussion_r505244092?
Thank you.

engcom-Charlie:
@tnsezer please fix failed tests. Otherwise, we can't proceed with your PR.
Thank you.

ihor-sviziev:
@tnsezer  I am closing this PR now due to inactivity.

PR #23351, opened 2019-06-21
Fix some framework coding issues

ghost:
### Description (*)
Used static code analysis to find a range of issues in the Magento Framework code. Mostly missing or incorrect typehints, undefined properties and too many parameters for method calls.

1. N/A

1. As it's all on the framework level not easy to manually test. Easiest way might be to take a look at the code in an IDE and confirm the typehints now resolve to classes where before they didn't.

### DIFF
--- a/dev/tests/integration/framework/Magento/TestFramework/App/MutableScopeConfig.php
+++ b/dev/tests/integration/framework/Magento/TestFramework/App/MutableScopeConfig.php
@@ -23,10 +23,7 @@ class MutableScopeConfig implements MutableScopeConfigInterface
     private $testAppConfig;
 
     /**
-     * @param string $path
-     * @param string $scopeType
-     * @param null $scopeCode
-     * @return bool
+     * @inheritdoc
      */
     public function isSetFlag($path, $scopeType = ScopeConfigInterface::SCOPE_TYPE_DEFAULT, $scopeCode = null)
     {
--- a/lib/internal/Magento/Framework/App/Config.php
+++ b/lib/internal/Magento/Framework/App/Config.php
@@ -50,7 +50,7 @@ class Config implements ScopeConfigInterface
      *
      * @param string $path
      * @param string $scope
-     * @param null|string $scopeCode
+     * @param null|int|string $scopeCode
      * @return mixed
      */
     public function getValue(
@@ -85,7 +85,7 @@ class Config implements ScopeConfigInterface
      *
      * @param string $path
      * @param string $scope
-     * @param null|string $scopeCode
+     * @param null|int|string $scopeCode
      * @return bool
      */
     public function isSetFlag($path, $scope = ScopeConfigInterface::SCOPE_TYPE_DEFAULT, $scopeCode = null)
@@ -95,6 +95,7 @@ class Config implements ScopeConfigInterface
 
     /**
      * Invalidate cache by type
+     *
      * Clean scopeCodeResolver
      *
      * @return void
--- a/lib/internal/Magento/Framework/Cache/Frontend/Decorator/Logger.php
+++ b/lib/internal/Magento/Framework/Cache/Frontend/Decorator/Logger.php
@@ -30,7 +30,7 @@ class Logger extends Bare
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function remove($identifier)
     {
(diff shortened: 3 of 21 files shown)


lib/internal/Magento/Framework/App/Config.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
      *
      * @param string $path
      * @param string $scope
-     * @param null|string $scopeCode
+     * @param null|int|string $scopeCode
```
fooman, line 53:
```
            if (is_numeric($scopeCode) || $scopeCode === null) {
                $scopeCode = $this->scopeCodeResolver->resolve($scope, $scopeCode);
            }
```

lib/internal/Magento/Framework/Cache/Frontend/Decorator/Logger.php
```diff
@@ -44,7 +44,7 @@ public function remove($identifier)
      */
     public function clean($mode = \Zend_Cache::CLEANING_MODE_ALL, array $tags = [])
     {
-        $result = parent::clean($mode, $tags, $mode);
+        $result = parent::clean($mode, $tags);
```
fooman, line 47: public function clean($mode = \Zend_Cache::CLEANING_MODE_ALL, array $tags = [])
    {
        return $this->_getFrontend()->clean($mode, $tags);
    }

fooman:
Okay so looks like there are a range of other things to address that got triggered by my changes...

I am okay addressing those but some of what is to be fixed is not clear at all. For example what is meant by the below?

```
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 25 | ERROR | [x] Missing short description
 26 | ERROR | [x] There must be exactly one blank line before tags
----------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```
phpcbf does not fix anything for me. 

I also tried running  `vendor/bin/phpcs --standard=Magento2` with `magento/magento-coding-standard` against the changed file and that brings up nothing.

orlangur:
@fooman,

> phpcbf does not fix anything for me

Those are buggy sniffs which I already reported. They used `addFixableError` by mistake instead of `addError`.

dev/tests/integration/framework/Magento/TestFramework/App/MutableScopeConfig.php
```diff
@@ -25,7 +25,7 @@ class MutableScopeConfig implements MutableScopeConfigInterface
     /**
      * @param string $path
      * @param string $scopeType
-     * @param null $scopeCode
+     * @param null|int||string $scopeCode
```
orlangur: Why `||`?

orlangur (changes requested):
Thanks @fooman, please provide steps for catching those errors.

Also, class cannot be removed even if it's not used, just deprecate it.

dev/tests/integration/framework/Magento/TestFramework/App/MutableScopeConfig.php
fooman: thanks - fixed

fooman:
Sorry had some issues with git but should all be sorted now. The latest commit fixes the code style issues that were triggered by my touching the files.

The only exception so far is 
```
FILE: ...ar/www/html/lib/internal/Magento/Framework/App/Utility/Files.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
----------------------------------------------------------------------
 618 | WARNING | The use of function md5() is discouraged; use
     |         | improved hash functions (SHA-256, SHA-512 etc.)()
     |         | instead
 771 | WARNING | The use of function md5() is discouraged; use
     |         | improved hash functions (SHA-256, SHA-512 etc.)()
     |         | instead
----------------------------------------------------------------------
```

where I don't blindly want to add an ignore without someone else making a call on this one. The code change to `hash('sha256'` would be simple enough but I am not sure if this would be too disruptive on existing caches, upgrades and what not. On the other side if some one could confirm that the md5 here is not security relevant that be great.

fooman:
> please provide steps for catching those errors.

I have started using phpstan in my personal extensions which very quickly brings up Magento core related issues. The output is quite noisy so have manually been going through small chunks to fix what looks odd. Command would be something like `phpstan analyse  path/to/code -l1`

orlangur:
@fooman,

> The output is quite noisy so have manually been going through small chunks to fix what looks odd. Command would be something like `phpstan analyse path/to/code -l1`

Perfect, such info is enough. Are a lot of false-positives or what kind of noise?

> On the other side if some one could confirm that the `md5` here is not security relevant that be great.

Generally, I don't think this rule should be applied anywhere outside the cryptographic layer. Will check for a particular use case here.

Thanks, will have a look on latest changes tonight.

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-5407](https://jira.corp.magento.com/browse/ENGCOM-5407) has been created to process this Pull Request

| Label | Description |

engcom-Delta:
:heavy_check_mark: QA passed

PR #23369, opened 2019-06-23
#16832 - Reworked query in getAttributeRawValue so that it returns a store specific value even if there is no default value

semajeg:
### Description (*)

Reworked query in getAttributeRawValue so that it returns a store specific value even if there is no default value

The query being generated in this function was querying the default value and then joining onto the store value, so if the default value was not present then it would not return a store value, even if there was one.

The query has been changed so that it left joins on both the default value and the store value, so that it does not prevent the return of a store value if there is no default

1. magento/magento2#16382: Product resource getAttributeRawValue function does not return anything if there is no default value

### Manual testing scenarios

1. Create 2 store views (though 1 will do)
2. Create a product attribute with no default value, and a store view scope (e.g. a  dropdown)
3. On a product, set a value in one of the child store views, and not at the default level
4. Call $resourceModel->getAttributeRawValue(<productId>, <attribute_code>, <store-view-id>);

### Questions or comments

Though the tech guidelines say "1.3.1. All new PHP files MUST have strict type mode enabled by starting with declare(strict_types=1);. All updated PHP files SHOULD have strict type mode enabled. PHP interfaces SHOULD NOT have this declaration.". This was discussed at the mm19uk contribution day with Tom Erskine and we decided that this PR should not do that to this file,since it will widen the scope of the PR too much and probably cause a LOT of knock on effects


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23369) <br/>All committers have signed the CLA.

app/code/Magento/Catalog/Model/ResourceModel/AbstractResource.php
```diff
@@ -466,9 +466,9 @@ protected function _getOrigObject($object)
      * Checks also attribute's store scope:
      * We should insert on duplicate key update values if we unchecked 'STORE VIEW' checkbox in store view.
      *
-     * @param AbstractAttribute $attribute
-     * @param mixed $value New value of the attribute.
-     * @param array &$origData
+     * @param  AbstractAttribute $attribute
```
dmanners: I think the static tests might be failing because of the extra space here and the following lines though I am not too sure.

```
 469 | ERROR   | [x] Method arguments visual alignment must be
     |         |     consistent
```

semajeg: I was getting the same error before the spaces were added, so I added the extra spaces so that they aligned with the return too. Didn't make a difference though.

What's odd is that if I run that static test locally, I don't get the error :( 

(I'll revert the commit anyway since it didn't make a difference)

semajeg: He @dmanners , thanks for looking into that! 

I took another look and thanks to comments on this other PR  (https://github.com/magento/magento2/pull/23304) it led me to find that it was the "&$origdata" causing that error online 469, phpcs must've been reporting the wrong line number (oddly!). I've made some further commits and that error is now cleared along with the other warnings, however one part is still failing: 

```
----------------------------------------------------------------------
 18 | WARNING | Abstract classes MUST NOT be marked as public @api.
----------------------------------------------------------------------
```

I feel like this error is slightly "above my paygrade" so to speak and it might be worth someone Magento side making a ruling on it. The obvious solution is to remove "api" from the class comment, but that's also obviously not a decision to take lightly given how these classes are referred to in the devdocs https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/ (though perhaps I'm also overthinking it :P)

Not sure how we proceed with this one!

Let me know what you think

semajeg: Hey @dmanners 

I've been able to fix the static test by adding: 

phpcs:disable Magento2.Classes.AbstractApi

To the top of Magento\Catalog\Model\ResourceModel\AbstractResource, like in some of the other classes in core Magento (e.g. Magento\Ui\Component\Form\Element\AbstractOptionsField)

This means all tests are passing now :)

semajeg:
Hey @dmanners, got a weird issue here. The "Database Compare Build" is failing now, but when you look at the details, it says it's all passed. Any thoughts?

semajeg:
> Hey @dmanners, got a weird issue here. The "Database Compare Build" is failing now, but when you look at the details, it says it's all passed. Any thoughts?

You can ignore this now @dmanners , all tests are now passing

josefbehr (approved):

magento-engcom-team:
Hi @josefbehr, thank you for the review.
[ENGCOM-5683](https://jira.corp.magento.com/browse/ENGCOM-5683) has been created to process this Pull Request
:eight_spoked_asterisk: @josefbehr, could you please add one of the following labels to the Pull Request?

| Label | Description |

engcom-Delta:
:heavy_check_mark: QA passed

PR #23454, opened 2019-06-27
Resolve _save_after_commit not firing on Guest Checkout

navarr:
### Description (*)
All sorts of commit_after events do not fire on Guest Checkout due to the creation of a transaction level outside of the Abstract Resource models - thereby preventing them from ever reaching a low enough transaction level to call the commit after event hooks

This removes that change and instead creates two very generic resource models solely existing to do the exact same thing while still triggering commit callbacks.

1. magento/magento2#23340: sales_order_item_save_commit_after & sales_order_save_commit_after events do not fire for guest checkout

### DIFF
--- a/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
+++ b/app/code/Magento/Checkout/Model/GuestPaymentInformationManagement.php
@@ -7,6 +7,8 @@ declare(strict_types=1);
 
 namespace Magento\Checkout\Model;
 
+use Magento\Checkout\Model\ResourceModel\GenericCheckoutResource as CheckoutResource;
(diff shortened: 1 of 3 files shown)


josefbehr:
@navarr Thank you for your contribution. Unfortunately, this does not seem like an ideal solution to me and seems to break many tests. Is there no other way to do this?

@magento run all tests

navarr:
I'll have to take another look at it since it looks like the tests are failing during guest checkout.  Perhaps I made a mistake somewhere.

sidolov:
@navarr I'm closing this PR since there is no activity for a long period of time. 
Please, reopen it if you wish to continue!

Issue #7266, opened 2016-11-01
Account signup form is attack vector for spammers, and is active in the wild.


cyuzik:
Using Magento 2.1.2 running on linux with php 5.6.1.

Spammers have been posting to /customer/account/create and similar form pages to send out spam. Our server had thousands of spam messages submitted using this in the past couple of days. The site is pretty much stock running the stock template.

We've done packet captures to the server and the attackers are simply using the last name field for the spam message, and the email address as the recipient.

My proposed fix is to limit the input length for firstname and lastname fields to 15 characters, and disallow any characters except the standard ascii alphabet, lower and uppercase, and the apostrophe. The firstname and lastname fields should not allow a paragraph of text that also contains URLs.

Since in this case it appears that the attackers were using a bot, so the form validation should be done after the post and not in javascript on the page so that form validation can simply be ignored by disabling or ignoring javascript.

Here is a pastebin from one of the packet captures: http://pastebin.com/0WKvF21L

orlangur:
> disallow any characters except the standard ascii alphabet, lower and uppercase, and the apostrophe

Are you serious? :) There are other languages exist besides English, you know.

> The firstname and lastname fields should not allow a paragraph of text that also contains URLs

Agree with this point, looks like slashes and line breaks are not really needed for these fields in any language. But I'm not sure it is worth adding to core as your case is quite narrow (was it ever reported as a problem for Magento 1?).

In vanilla core you can just enable CAPTCHA or implement any other defending techniques relevant for your store on top of it.

cyuzik:
I don't know if this was ever reported in Magento 1. However if spammers found our little low-traffic server, they're definitely going to be looking for others. We have enabled CAPTCHA on the site, and that's stopped the spam, for now. Again, we're running latest patch version of Magento 2.1.2 and were hit by this.

However, we're still back to the point that there is no reasonable length limit on firstname or lastname, and further it allows entire urls in a person's name. Did you look at the pastebin? That's not someone's last name.

VahanDrnoyan:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  v.drnoyan@gmail.com
    Domain magejet.com has exceeded the max emails per hour (203/199 (102%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from o3.sgmail.github.com ([192.254.112.98]:42436)
[...]

PR #23489, opened 2019-06-29
[Fixed] First name and Last name field must not allow to enter URL and Special character's

meetvora69:
### Description (*)

First name and Last name field must not allow to enter URL and Special character's

1. magento/magento2#23487: First name and Last name field must not allow to enter URL and Special character's

1. Go to Admin
2. Customers > All customers > Add New Customer
3. Enter URL details for First name Last name field

### DIFF
--- a/app/code/Magento/Customer/Setup/Patch/Data/UpgradeNameAttributes.php
+++ b/app/code/Magento/Customer/Setup/Patch/Data/UpgradeNameAttributes.php
@@ -1,0 +1,102 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Customer\Setup\Patch\Data;
+
+use Magento\Customer\Setup\CustomerSetupFactory;
+use Magento\Framework\App\ResourceConnection;
+use Magento\Framework\Setup\ModuleDataSetupInterface;
+use Magento\Framework\Setup\Patch\DataPatchInterface;
+use Magento\Framework\Setup\Patch\PatchVersionInterface;
+
+/**
+ * Class UpgradeNameAttributes
+ * @package Magento\Customer\Setup\Patch
+ */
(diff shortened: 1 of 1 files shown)


ihor-sviziev (changes requested):
Hi @meetvora69,

I just reviewed your fix and found one issue - people will not be able to put non English name & surname. For instance you can't put there "Müller", or "Иван" anymore. 
So there should be some blacklist of special symbols.

Could you double check your solution?

ihor-sviziev:
Hi @meetvora69,
Thank you for your contribution. Unfortunately we didn’t get response from you after more than 14 days, so I’m closing this PR.
Feel free to fix issues listed in my last review and reopen PR.

cbichis:
This is the most frequent attack nowdays, we need to have a way to filter out URL's...

ihor-sviziev:
Hi @cbichis,
Original issue was closed as not an issue, more details with explanation why you could find there.
Could you explain what attack do you mean?

cbichis:
The spam bots are registering or signing up to the newsletter and they are using on first name/last name URL's. This would cause the registration confirmation email to include malicious URL's which of course are dangerous for anyone who clicks on them.

ihor-sviziev:
@cbichis oh, I see. Seems like following issue contains all needed info and should be confirmed https://github.com/magento/magento2/issues/28041

cbichis:
Yeah, but thats a bit a diff topic.

I am pointing out here that it's quite hard to understand Magento with the default settings to have such security holes... For the custom made apps I am doing I also strip out not just tags but also URL's from the fields...

ihor-sviziev:
@cbichis, I see. Seems like this is duplicate of https://github.com/magento/magento2/issues/7266 https://github.com/magento/magento2/issues/28041.

Looks like issue is valid, should be fixable by enabling recaptcha and adding some [WAF](https://en.wikipedia.org/wiki/Web_application_firewall).

PR #23530, opened 2019-07-01
Update Helper.php

WJdeBaas:
Remove user-input type-casting.

### Description
Link attributes like "position" should be allowed to be any of the types "integer", "decimal" or "varchar" in custom modules

customization restrictions

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
@@ -17,10 +17,9 @@ use Magento\Catalog\Model\Product\Link\Resolver as LinkResolver;
 use Magento\Catalog\Model\Product\LinkTypeProvider;
 use Magento\Framework\App\ObjectManager;
 use Magento\Catalog\Controller\Adminhtml\Product\Initialization\Helper\AttributeFilter;
+use Magento\Framework\Escaper;
 
 /**
- * Product helper
- *
  * @api
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  * @since 100.0.2
@@ -97,6 +96,11 @@ class Helper
     private $attributeFilter;
 
     /**
+     * @var Escaper
+     */
+    protected $escaper;
+
+    /**
      * Constructor
      *
      * @param \Magento\Framework\App\RequestInterface $request
@@ -105,6 +109,7 @@ class Helper
      * @param ProductLinks $productLinks
      * @param \Magento\Backend\Helper\Js $jsHelper
      * @param \Magento\Framework\Stdlib\DateTime\Filter\Date $dateFilter
+     * @param Escaper $escaper
      * @param CustomOptionFactory|null $customOptionFactory
      * @param ProductLinkFactory|null $productLinkFactory
      * @param ProductRepositoryInterface|null $productRepository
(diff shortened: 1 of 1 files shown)


aleron75:
Hello @WJdeBaas, first of all, thank you for your contribution.

>attributes like "position" should be allowed to be any of the types

please, can you provide a use case for types other than `int`? We are just trying to understand under which circumstances we could need a `varchar` position and how it could be used.

Thank you!

app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
```diff
@@ -119,6 +124,7 @@ public function __construct(
         \Magento\Catalog\Model\Product\Initialization\Helper\ProductLinks $productLinks,
         \Magento\Backend\Helper\Js $jsHelper,
         \Magento\Framework\Stdlib\DateTime\Filter\Date $dateFilter,
+        Escaper $escaper,
```
aleron75, line 127: Let's try to make this change as backward compatible as possible, adding new constructor parameter as **last** and **optional**, with property initialized with object manager if the parameter is not passed.

e.g.

    $this->escaper = $escaper ?: ObjectManager::getInstance()->get(Escaper::class);

WJdeBaas:
Hi Alessandro,

I'll try to explain under which circumstances we could need a varchar
position and how it could be used.

Here's an example: https://www.zandri.nl/catalog/product/view/id/133813

On this page you'll see an exploded view drawing of a scooter carburetor,
where the articles are numbered, but these numbers aren't always typical
integers.
For example: there is 12 en 12-1, and even other drawings can have other
codes like 3a, 3b and 3c for instance.
There are also example where positions are devided with a pipe-symbol (|),
to make it possible to list the same SKU on several different positions in
the drawing.

This functionality has to be migrated to Magento 2, and after some research
and testing I found it's quite easy to reuse and modify parts of the
standard Magento 2 implementations for linked articles like used for
Crosssell and Upsell articles.
But at the moment it's not possible to save divergent values because the
backendcontroller casts the user input, even if the positions' data_type is
declared as a varchar attribute, and the input is saved into the
'catalog_product_link_attribute_varchar' db-table.

In my opinion the variable casting shouldn't be done in the controller, but
in the repository model, where it can check the attributes' type (int,
decimal or varchar), and cast it accordingly just before the data is saved.
I agree with you on the backward compatibility and use the ObjectManager in
case the Escaper class is not passed as a parameter, although the Escaper
class could also be moved the the repository model.

Op do 4 jul. 2019 om 12:35 schreef Alessandro Ronchi <
notifications@github.com>:

aleron75:
Thank you @WJdeBaas for your explanation, it's clear.

My only concern is whether this change should be part of the core or rather should be part of a customized implementation.

Let me check whether we can involve the core team into this discussion.

My best.

sivaschenko (changes requested):
Hi @WJdeBaas thanks for the pull request!

Unfortunately, we can allow only `int` type for link position as this is part of the API: https://github.com/magento/magento2/blob/5588999a108619d09835a2eecbc9579a9e6f27af/app/code/Magento/Catalog/Api/Data/ProductLinkInterface.php#L78

To achieve the desired behaviour you need to extend Magento functionality possibly using an extension attribute for `ProductLinkInterface`

WJdeBaas:
Hi Sergii, Alessandro,

Is it really impossible to apply a minor update to the API to gain more
flexibility  ?
After all, It's "software", not ferroconcrete or titanium ...

    /**
     * Get linked item position
     *
     * @return int|float|string
     */
    public function getPosition();

    /**
     * Set linked item position
     *
     * @param int|float|string $position
     * @return $this
     */
    public function setPosition($position);

Kind regards,
Willem-Jan

Op do 4 jul. 2019 om 15:57 schreef Sergii Ivashchenko <
notifications@github.com>:

sidolov:
@WJdeBaas , I am closing this PR now due to inactivity.

PR #23663, opened 2019-07-11
If customer has already quote address, show as selected

talipkorkmazer:
### Description (*)

When there is no quote address, checkout address page should show default shipping address as selected but when there is a quote address that selected before, checkout address page should show selected address as selected not default shipping address.

assignCustomer function sets quote address object's shipping and billing address with default addresses.

One can produce this issue by selecting different address from default address on checkout address page and continue to review page and return to the checkout address page, then selected address will be default address not the one selected before.


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23663) <br/>All committers have signed the CLA.

PR #23687, opened 2019-07-13
Resolve Warning during search in catalog by multiple custom option values. Warning should not be generated in log

edenduong:
### Description (*)

magento/magento2#23557: Resolve Warning during search in catalog by multiple custom option values. Warning should not be generated in log

The reason because the "$label" should be always "string" to call the _createItem($label, $value) function. 

Solution: Convert array to string by using "implode'

1. magento/magento2#23557: Warning during search in catalog by multiple custom option values. Warning should not be generated in log

1. Create the Category and go to it at the frontend like http://[magento domain]/shirt.html
2. Change URL parameter to filter, add "color=123,11111" (color is the attribute can filter at the navigation and has the options) . Url will be : http://[magento domain]/shirt.html?color=123,11111
3. Press Enter to query filter the color by the value "123,11111" 

Result: No error in Log 
 Warning: strip_tags() expects parameter 1 to be string, array given in /.../magento2/vendor/magento/framework/Filter/StripTags.php on line 48

### DIFF
--- a/app/code/Magento/CatalogSearch/Model/Layer/Filter/Attribute.php
+++ b/app/code/Magento/CatalogSearch/Model/Layer/Filter/Attribute.php
@@ -3,6 +3,8 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+declare(strict_types=1);
+
 namespace Magento\CatalogSearch\Model\Layer\Filter;
 
 use Magento\Catalog\Model\Layer\Filter\AbstractFilter;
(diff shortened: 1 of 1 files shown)


edenduong:
Hi @orlangur : Could you please review my pull request?

swnsma:
I think that such changes need to be done in
`\Magento\Eav\Model\Entity\Attribute\Source\Table::getOptionText`
But in will lead to BC break.

swnsma (approved):

magento-engcom-team:
[ENGCOM-5596](https://jira.corp.magento.com/browse/ENGCOM-5596) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed 

**Before:**
before

**After:**
after

PR #23714, opened 2019-07-15
Add Api to MediaStorage to simplify the interaction with files

diazwatson:
### Description (*)

While working on Adobe Stock Integration development, we noticed that there is not a clear entry point for use Media Storage.

The proposed change pretend to make clear this entry point and therefore make easier for other modules to use and interact with Media Storage module.

1. No behaviour has changed, Media Storage should be working as before.
2. ...


diazwatson:
After discussed the details with @sivaschenko we no longer need to do this change in the core.

PR #23796, opened 2019-07-19
[Fixed] no autofocus in invalid field when collapse is closed

iflam:
### Description (*)

If a collapse is open and a field inside displays an error message (i.e. 'Please enter a valid number in this field', the collapsible is opened.
If there are multiple collapsibles to open, the cursor focuses on the first errored input.

1. magento/magento2#23765: No auto focus in the invalid field if the fieldset collapse is closed

1. Go to Store->Configuration->Catalog->Inventory
2. Open Collapse "Product Stock Options"
3. Fill "ggggggg" into "Maximum Qty Allowed in Shopping Cart"
4. Close collapse "Product Stock Options" Fieldset
5. Click "Save"
6. Repeat 1-5, but also fill "ggggg" into any other text-input fields.

### DIFF
--- a/lib/web/mage/backend/validation.js
+++ b/lib/web/mage/backend/validation.js
@@ -33,6 +33,17 @@
                     $(this.errorList.length && this.errorList[0].element || [])
                         .focus()
                         .trigger('focusin');
+                    $('[data-toggle=dropdown].active').trigger('open.dropdown');
+                    let first = true;
+                    $('.section-config').each(function(){
+                        if($(this).find('input.mage-error').length && !$(this).hasClass("active")){
+                            $(this).find("a[id$='-head']").click();
+                            if(first) {
+                                $(this).find("input.mage-error:text:visible:first").focus();
+                                first = false;
+                            }
+                        }
+                    })
                 } catch (e) {
                     // ignore IE throwing errors when focusing hidden elements
                 }


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23796) <br/>All committers have signed the CLA.

lib/web/mage/backend/validation.js
```diff
@@ -33,6 +33,17 @@
                     $(this.errorList.length && this.errorList[0].element || [])
                         .focus()
                         .trigger('focusin');
+                    $('[data-toggle=dropdown].active').trigger('open.dropdown');
```
VladimirZaets, line 36: Hi @iflam. Thanks for collaboration. 
Here is you try to get an active dropdown, but we can have a case when we have few active dropdowns on the page, so it isn't fully correct.

lib/web/mage/backend/validation.js
```diff
@@ -33,6 +33,17 @@
...
+                    let first = true;
```
VladimirZaets, line 37: Magento support just ES5, `let` variables it is the feature of ES6, so currently we shouldn't use it. Please use `var` instead of `let` or `const`

lib/web/mage/backend/validation.js
```diff
@@ -33,6 +33,17 @@
...
+                    $('.section-config').each(function(){
```
VladimirZaets, line 38: Can you please explain what you try to do here?
For my point of view, all that we should do in this case it's find first element (that already did here `$(this.errorList.length && this.errorList[0].element || [])`) and check is this field inside of collapsible panel or not. If yes, check is collapsible panel is opened and if not - open this panel

VladimirZaets (changes requested):
Hi @iflam. Thanks for collaboration. Please check code review comments.

sidolov:
@iflam , I am closing this PR now due to inactivity.

PR #19143, opened 2018-11-10
Allow to read HTTP/2 response header in curl client.

vovayatsyuk:
This change fixes ability to read response from third-party servers that use HTTP/2 if your server use HTTP/2 too.

### Description
The issue is HTTP/2 sends the following headers `HTTP/2 200`, `HTTP/2 401` and so on. It doesn't have third parameter like HTTP/1.1 has: `HTTP/2 200 OK`

### Fixed Issues
1. magento/magento2#19127: Cannot connect to Magento 2 market place

### Manual testing scenarios
1. Enable HTTP/2 on your server
2. Try to login to marketplace from [System > Web Setup Wizard](https://docs.magento.com/marketplace/user_guide/buyers/install-extension.html). You can use invalid credentials too.
3. You will see a message about invalid response line instead of "Bad Credentials" message.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #23893, opened 2019-07-25
HTTP reason phrase is optional in RFC 7230

ludovic-gasc:
### Description (*)

When we use a HTTP server that uses a version of Tomcat 8.5+ with the HTTP client of Magento, we have this error: **Invalid response line returned from server** in Magento logs.

It's because since Tomcat 8.5, they have removed the reason phrase, as you can read in the [Tomcat changelog](http://tomcat.apache.org/tomcat-8.5-doc/changelog.html): **RFC 7230 states that clients should ignore reason phrases in HTTP/1.1 response messages. Since the reason phrase is optional, Tomcat no longer sends it.**

In [RFC 7230](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html), you can read this: 
`Reason-Phrase  = *<TEXT, excluding CR, LF>`
The * means zero or more.

Concretely, instead to have this HTTP response from Tomcat:

`HTTP/1.1 404 Not Found`

You have:
`HTTP/1.1 404 `

Another bug report from Spring boot project where you can find more explanations: https://github.com/spring-projects/spring-boot/issues/6548

In the Magento source code, I don't see an usage of `$line[2]`.
To my actual understanding, it might be enough to not check if the third element in the array is present.

1. Setup a Spring Boot hello world project or any Java framework that uses Tomcat.
2. Use Magento HTTP client to connect

### DIFF
--- a/lib/internal/Magento/Framework/HTTP/Client/Curl.php
+++ b/lib/internal/Magento/Framework/HTTP/Client/Curl.php
@@ -439,7 +439,7 @@ class Curl implements \Magento\Framework\HTTP\ClientInterface
     {
         if ($this->_headerCount == 0) {
             $line = explode(" ", trim($data), 3);
-            if (count($line) < 2) {
+            if (count($line) < 1) {
                 $this->doError("Invalid response line returned from server: " . $data);
             }
             $this->_responseStatus = (int)$line[1];


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23893) <br/>All committers have signed the CLA.

hostep:
@GMLudo: correct me if I'm wrong, but I think this was already fixed by https://github.com/magento/magento2/pull/19143 & https://github.com/magento/magento2/pull/19239

If we count the parts of `HTTP/1.1 404` we have 2 parts, so this doesn't match with `< 2`, so you shouldn't see the error you are seeing right now.

What version of Magento are you targetting? According to what I'm seeing this issue was fixed in Magento 2.2.8 for the 2.2 release line & Magento 2.3.1 for the 2.3 release line.

ludovic-gasc:
Hi @hostep,

You have fully understand the situation, the HTTP client in Magento is already fixed, an upgrade is enough to fix the problem.

Thanks again for your time to give me the right pieces of information.

Kind regards

PR #23929, opened 2019-07-28
MasterCard validation fix for 58XXXX bins

hidonet:
Current version of Magento not validating MasterCard numbers like 58XXXX


orlangur (changes requested):
@hidonet please remove unrelated changes and add tests for this case.

sidolov:
Hi @hidonet could you please provide proof links that Mastercard card number can have the second digit more than 5? Thank you!

hidonet:
@sidolov I have some numbers which I got from payment gateways. I've checked mastercard's latest range, these numbers not included in their official number ranges. Sorry for that. Closing request.

PR #23930, opened 2019-07-28
Added Troy Card ( Turkey ) validation to current validation patterns

hidonet:
Troy is a card payment system established and operated by Bankalararası Kart Merkezi A.Ş. (Interbank Card Center - “BKM”).


orlangur (changes requested):
I don't think this needs to be a part of core. Why not introduce such rules in a custom module?

miguelbalparda:
I agree with @orlangur, will close for now. Feel free to reopen if anything changes.

grinet:
@orlangur @miguelbalparda Aura is included in validation patterns. I guess Troy has a larger customer base. Troy is a separate brand. Not a sub brand of a bank or big card companies like Visa, Mastercard etc. Troy should be added if the Aura is added to magento.

PR #23980, opened 2019-08-01
#23955: Minicart shows HTML when using file upload custom option - fixed

sudheers-kensium:
### Description (*)

#23955: Minicart shows HTML when using file upload custom option - fixed

1. magento/magento2#23955: Minicart shows HTML when using file upload custom option

1. As an admin, create a product with a custom option with type file upload
2. As a customer, add the product to the cart after having selected a file
3. Open up the mini cart, and expand "See Details" for the recently added product
4. See the custom option label with the filename of the uploaded file

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/template/minicart/item/default.html
+++ b/app/code/Magento/Checkout/view/frontend/web/template/minicart/item/default.html
@@ -45,7 +45,7 @@
                                 <span data-bind="html: option.value.join('<br>')"></span>
                             <!-- /ko -->
                             <!-- ko ifnot: Array.isArray(option.value) -->
-                                <span data-bind="text: option.value"></span>
+                                <span data-bind="html: option.value"></span>
                             <!-- /ko -->
                         </dd>
                         <!-- /ko -->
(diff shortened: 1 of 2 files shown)


app/code/Magento/Checkout/view/frontend/web/template/minicart/item/default.html
```diff
@@ -45,7 +45,7 @@
                                 <span data-bind="html: option.value.join('<br>')"></span>
                             <!-- /ko -->
                             <!-- ko ifnot: Array.isArray(option.value) -->
-                                <span data-bind="text: option.value"></span>
+                                <span data-bind="html: option.value"></span>
```
rleshchenko, line 48: This changes is forbidden due to security reasons. Please contact security team about that

rleshchenko (changes requested):

piotrekkaminski:
@sudheers-kensium  @rleshchenko  From my POV, the problem lies elsewhere - labels should not allow HTML entry in the first place. And the behavior of escaping the output in such a way that it shows HTML instead of rendering it is correct. We are really trying to limit ways to execute XSS from the admin panel to front-end, as well as embed malicious JavaScript in places like this. If you want custom formatting of specific labels, either customize the code or do it using CSS. However for the larger user base, it's safer to not allow this.

sidolov:
@piotrekkaminski thank you for your answer! @sudheers-kensium  I'm closing this PR because it may introduce security issues in the future. Feel free to open the new PR if you find a safe way to resolve the mentioned issue. Thank you for your collaboration!

PR #23988, opened 2019-08-02
fix: escapeUrl("0") should return "0", not ""

### LINKED ISSUE #23987: Escaper#escapeXssInUrl (internally: Escaper#escapeScriptIdentifiers) incorrectly convert un-truthy values to empty string
### Preconditions (*)
1. Any version (up to current HEAD as of July 2nd 2019)

### Steps to reproduce (*)
1. Call Escaper#escapeXssInUrl with value `"0"` (eg: `$this->_escaper->escapeXssInUrl("0");` in block context)

### Expected result (*)
1. return value of `"0"`, ie: unmodified

### Actual result (*)
1. returns an empty string `""`

This appears to be due to lines such as:

        $filteredData = preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $data) ?: '';
        $filteredData = preg_replace(self::$xssFiltrationPattern, ':', $filteredData) ?: '';

Which assume that the only un-truthy value which could be returned by `preg_replace` is an empty string or `false`

These should be updated to something like:

        $filteredData = preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $data);
        if ($filteredData === false || $filteredData === '') return '';
        $filteredData = preg_replace(self::$xssFiltrationPattern, ':', $filteredData);
        if ($filteredData === false) return '';

Arguably, this is not the "real" solution. Theoretically, this happens because of the expectation that `escapeXssInUrl` should only be called on "full" URLs, and is intended to remove URLs such as `"javascript:alert('I am an XSS attack')"` However, `escapeXssInUrl` is used by `escapeUrl`, which is in-turn used by `encodeUrlParam`. I believe this to potentially be a bug in `encodeUrlParam`, which (by its name) seems to be intended to actually do something similar to PHP's builtin `urlencode(...)` function.

I consider this to be a separate issue, however, and this change would both fix the `encodeUrlParam` case for this specific issue, while not negatively-impacting the pathological case of the string `"0"` being passed to `escapeUrl` directly.

wpalmer:
### Description (*)
Fixed: #23987

previously, we assumed that when removing XSS-like script URLs, the only
non-truthy result would be a failure in preg_replace, or an empty
string. A value of "0" would therefor pass through the script-tag
check, but then be converted into the empty string because it is
non-truthy.

instead, we now specifically test for those values which we want to
convert into an empty string, and do so immediately. This allows the
value of "0" to remain intact when passed through a escape methods such
as encodeUrlParam(...) or escapeUrl(...)

1. magento/magento2#23987: Escaper#escapeXssInUrl (internally: Escaper#escapeScriptIdentifiers) incorrectly convert un-truthy values to empty string

1. use of $this->_escaper->escapeUrl("0") should return "0" in block context
2. use of $this->_escaper->encodeUrlParam("0") should return "0" in block context
3. (unit test data has been included)

### Questions or comments
Arguably, this is not the "real" solution. Theoretically, this happens
because of the expectation that escapeXssInUrl should only be called
on "full" URLs, and is intended to remove URLs such as
"javascript:alert('I am an XSS attack')". However, escapeXssInUrl is
used by escapeUrl, which is in-turn used by encodeUrlParam. I believe
this to potentially be a bug in encodeUrlParam, which (by its name)
seems to be intended to actually do something similar to PHP's builtin
urlencode(...) function.

I consider this to be a separate issue, however, and this change would
both fix the encodeUrlParam case for this specific issue, while not
negatively-impacting the pathological case of the string "0" being
passed to escapeUrl directly.

I'm not sure how to check that "all builds are green"

 - [?] All automated tests passed successfully (all builds are green)

### DIFF
--- a/lib/internal/Magento/Framework/Escaper.php
+++ b/lib/internal/Magento/Framework/Escaper.php
@@ -335,8 +335,16 @@ class Escaper
      */
     private function escapeScriptIdentifiers(string $data): string
     {
-        $filteredData = preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $data) ?: '';
-        $filteredData = preg_replace(self::$xssFiltrationPattern, ':', $filteredData) ?: '';
+        $filteredData = preg_replace('/[\x00-\x1F\x7F\xA0]/u', '', $data);
+        if ($filteredData === false || $filteredData === '') {
+            return '';
+        }
+
+        $filteredData = preg_replace(self::$xssFiltrationPattern, ':', $filteredData);
+        if ($filteredData === false) {
+            return '';
+        }
+
         if (preg_match(self::$xssFiltrationPattern, $filteredData)) {
             $filteredData = $this->escapeScriptIdentifiers($filteredData);
         }
--- a/lib/internal/Magento/Framework/Test/Unit/EscaperTest.php
+++ b/lib/internal/Magento/Framework/Test/Unit/EscaperTest.php
@@ -344,6 +344,10 @@ class EscaperTest extends \PHPUnit\Framework\TestCase
     {
         return [
             [
+                '0',
+                '0',
+            ],
+            [
                 'javascript%3Aalert%28String.fromCharCode%280x78%29%2BString.'
                 . 'fromCharCode%280x73%29%2BString.fromCharCode%280x73%29%29',
                 ':alert%28String.fromCharCode%280x78%29%2BString.'


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=23988) <br/>All committers have signed the CLA.

sidolov (approved):

magento-engcom-team:
[ENGCOM-5539](https://jira.corp.magento.com/browse/ENGCOM-5539) has been created to process this Pull Request

| Label | Description |

magento-engcom-team:
@wpalmer thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

wpalmer:
I'm not sure why the "Database Compare Build" is marked as failing. Looking at the build reports, everything seems to be coming back with a label of "success".

engcom-Delta:
:heavy_check_mark: QA passed

wpalmer:
I have noticed a potential change in functionality: we now return `false` in the event of a failure in `preg_replace`, whereas previously we would have returned an empty string. This would seem to be contrary to the intent of the original ternary. While I can't think of any situations which would result in a failure of `preg_replace`, I will update my branch to remove this change, as it is not intended.

wpalmer:
@sidolov, this pull-request is marked as having your approval, but I believe this may warrant a new review, as the contents of the request have changed slightly

sidolov (approved):

magento-engcom-team:
[ENGCOM-5539](https://jira.corp.magento.com/browse/ENGCOM-5539) has been created to process this Pull Request

| Label | Description |

PR #24093, opened 2019-08-09
Move WYSIWYG adapter to adminhtml theme Ui module

sankalpshekhar:
### Description (*)

This PR move the adition of block for WYSIWYG editor from Magento_Theme to Magento_Ui module for following reasons:

1. the block and template file being injected both belonged to Mageto_Ui, so that seemed a proper space for it
2. WYSIWYG adapter being part of base displayed on frontend, leading to exposure of path to wysiwyg editor path, it isn't being used
image

1. ...
2. ...


VladimirZaets (changes requested):
Hi @sankalpshekhar. You have moved the page layout declaration rather than WYSIWYG adapter. It looks like incorrect changes because we shouldn't change the page layout.

sankalpshekhar:
@VladimirZaets The file i have updated isn't page declaration it is just adding a new block to every page by referring to after body start container. Nor this is changing layout, it only moves the block addition to a more relevant area in the module structure.

Can you add more details on what you are referring as change in page layout?

sankalpshekhar:
@VladimirZaets Can you help on whats next steps here?

sankalpshekhar:
@VladimirZaets Changes made a requested

hostep:
Since this PR is slightly related to https://github.com/magento/magento2/issues/21766, can we determine if https://github.com/magento/magento2/issues/21766 needs to be closed as expected behavior or should the inclusion of the block be moved to the `adminhtml` area instead of `base`? If that last one is the case, then maybe it can already been done in this PR?

sankalpshekhar:
@hostep i had the same idea, however @VladimirZaets doesn't seems to agree to this, so i had to modify the change to move it back to base, instead of adminhtml theme.

this is what i had with @VladimirZaets on Slack

> Vladimir Zaets : the main problem with your PR is backward compatibility, because 3rd party developers can create the theme that didn't based on blank or luma. We don't recommend this approach, but it possible. So they can have a case when they not needed this block, but they need UI module, but with your fix they will be automatically render this block, that is not they expected

> **Sankalp Shekhar 8:53 PM**
> This doesn't adds up
> you need atleast to inherit blank theme to create a new theme
> anyways the change i have made is for admin theme not frontend theme, and the UI module is called in that block so if someone needs that UI module then that would be part of that module
> **Vladimir Zaets 11:56 PM**
> No, it also concern me, it was the point that I discussed. We recommend at least  inherit blank theme but it isn't strict rule. Theme developers can create new theme that don't based on any of existing theme.
> **Sankalp Shekhar 6:32 AM**
> Hear you, but the change i made is related to moving the block called into UI module, since the block in the module is called from UI module it make more sence to be in UI module
> **Vladimir Zaets 1:10 PM**
> Not really, the files placed in UI module, but layout is responsible for calling block rendering . Also, this block was placed in base directory, that's mean it was uses in storefront and backend area. Currently you move it to backend area, that's mean that this block willn't rendered on storefront

hostep:
@sankalpshekhar: thanks for the info. And yes indeed, that was my concern as well, that it will break backwards compatibility for people relying on the fact that it was previously available on the frontend.

So I guess we should close #21766 because it's an invalid issue? @VladimirZaets can you confirm? Thanks!

sankalpshekhar:
@hostep I understand that, however the path that this block exposes says _adminhtml_, so i consider that entire wysiwyg should be moved to be more common path to get the #21766 be fixed.

hostep:
Right, I totally understand @sankalpshekhar, you mean the path `mage/adminhtml/wysiwyg/tiny_mce/tinymce4Adapter` contains `adminhtml` and is being used on the frontend.

I would rather have it not being exposed on the frontend, but it's really hard to get feedback from core Magento devs to see if they meant to expose it or not, so we just have to be a bit more patient until somebody finally finds some time to give us the answer 🙂

VladimirZaets (approved):

magento-engcom-team:
[ENGCOM-5755](https://jira.corp.magento.com/browse/ENGCOM-5755) has been created to process this Pull Request

| Label | Description |

VladimirZaets:
@hostep we discussed this question in the slack channel. The main reason why we can't do it is backward compatibility. Someone can use it on frontend and if we will move this file just for backend area it can break there.

hostep:
Ok that's what I expected.

Out of curiosity: if this "problem" was found before 2.3.0 was released, would it have been moved in the `adminhtml` area, or was it actually intended to be exposed on the frontend?

VladimirZaets:
@hostep Really I don't know, it was a long time ago and we have a lot of teams with their own responsibilities. Currently, it looks like useful just for `adminhtml` but what plan was it's nobody knows, maybe it's just mistake.

hostep:
Ok no worries, I've closed https://github.com/magento/magento2/issues/21766 since I agree it might cause backwards incompatibilities. Thanks for the feedback!

engcom-Alfa:
:heavy_check_mark: QA Passed 

**Note:**
The issue [#18548](https://github.com/magento/magento2/issues/18548), that has been fixed in 2.2-develop version is still reproduce on 2.3-develop.
after

PR #24103, opened 2019-08-11
Refactor AdminNotification Render Blocks

DavidLambauer:
### Description

this PR includes 3 new Unit Tests to cover the 3 Block Render Classes of the AdminNotification Module. In addition to the Unit Tests, I refactored the 3 Block Render Classes and generally changed the following:

- Make use of FQCN
- Added strict type hints where ever it can be done easily
- Reduced the Use of magic methods e.g:

```php
#before: 
$object->getIsReady();

#after:
$object->getData('is_ready');
```

- Declared return types where ever it can be done easily
- Added `declare(strict_types=1);` to all files that were changed


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=24103) <br/>All committers have signed the CLA.

torhoehn (approved):

torhoehn (approved):
Thanks for your contribution!

magento-engcom-team:
[ENGCOM-5588](https://jira.corp.magento.com/browse/ENGCOM-5588) has been created to process this Pull Request

| Label | Description |

magento-engcom-team:
@DavidLambauer thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

magento-engcom-team:
[ENGCOM-5588](https://jira.corp.magento.com/browse/ENGCOM-5588) has been created to process this Pull Request

| Label | Description |

magento-engcom-team:
@DavidLambauer thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

engcom-Delta:
Hi @DavidLambauer there is an issue with PR changes: **Mark as Read** and **Remove** links in Notifications grid don't work.
Issue caused by $row->getData('id') because it returns "null".
#24103PR

Could you take a look?

DavidLambauer:
Hi @engcom-Delta, I added the fix. Id was null, you're right. I had to change it to `notification_id`. Please review the change. If there is anything else to do, please let me know.

engcom-Delta:
@torhoehn Could you review new changes?

torhoehn (approved):
Looks good for me now!

magento-engcom-team:
[ENGCOM-5588](https://jira.corp.magento.com/browse/ENGCOM-5588) has been created to process this Pull Request

engcom-Delta:
:heavy_check_mark: QA passed

app/code/Magento/AdminNotification/Block/Grid/Renderer/Actions.php
```diff
@@ -40,16 +45,16 @@ public function __construct(
      * @param   \Magento\Framework\DataObject $row
      * @return  string
      */
-    public function render(\Magento\Framework\DataObject $row)
+    public function render(DataObject $row) : string
```
sidolov: Unfortunately, we are not able to change public method signature according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

app/code/Magento/AdminNotification/Block/Grid/Renderer/Actions.php
```diff
@@ -40,16 +45,16 @@ public function __construct(
...
     {
-        $readDetailsHtml = $row->getUrl() ? '<a class="action-details" target="_blank" href="' .
-            $this->escapeUrl($row->getUrl())
+        $readDetailsHtml = $row->getData('url') ? '<a class="action-details" target="_blank" href="' .
```
sidolov: Why `getData('url')` method call better instead of `getUrl`? I don't see benefits in such usage.

sidolov (changes requested):

app/code/Magento/AdminNotification/Block/Grid/Renderer/Actions.php
DavidLambauer: The [Technical Guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html?itm_source=devdocs&itm_medium=search_page&itm_campaign=federated_search&itm_term=return%20type#1-basic-programming-principles) do describe that:

```
...
1.2. Explicit return types MUST BE declared on functions.

1.3. Type hints for scalar arguments SHOULD be used.

...
```

For me then, it seems to be fine to have a strict type?

app/code/Magento/AdminNotification/Block/Grid/Renderer/Actions.php
DavidLambauer: To declare the key as a string reduces magic getters and makes the code more explicit.

app/code/Magento/AdminNotification/Block/Grid/Renderer/Actions.php
sidolov: @DavidLambauer , yes, it's true for the new code, but for existing code, it's backward incompatible change, because every developer who extended this class and has own render method will get an error because method signature is changed.

app/code/Magento/AdminNotification/Block/Grid/Renderer/Actions.php
sidolov: Methods like 'getUrl' can be defined in the class with specific logic instead of magic method call and by current changes, you may break functionality. In this case, all plugins defined for such methods also will not be called.
It would be better to revert such changes. 
Thank you.

sidolov (changes requested):

engcom-Foxtrot:
@sidolov updated.

sidolov (approved):

sidolov:
@DavidLambauer please, sign CLA to proceed with delivery. Thank you!

magento-engcom-team:
[ENGCOM-5588](https://jira.corp.magento.com/browse/ENGCOM-5588) has been created to process this Pull Request

engcom-Delta:
:heavy_check_mark: QA passed

ihor-sviziev:
Hi @engcom-Foxtrot,
Why this task was closed and re-opened? I believe it was done due to some mistake

ihor-sviziev:
@magento rul all tests

ihor-sviziev:
@magento run all tests

ihor-sviziev:
@magento run Database Compare

ihor-sviziev (approved):
DB compare test failure for sure isn't related to these changes.
Approved

magento-engcom-team:
[ENGCOM-5588](https://jira.corp.magento.com/browse/ENGCOM-5588) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed

Issue #22867, opened 2019-05-13
Magento script and style filtering issue in CMS Blocks and Pages


Siyadev9:
### Preconditions (*)

1. Magento version 2.3.1
2. PHP version 7.2
3. MySQL 5.7

### Steps to reproduce (*)

1. Clean Magento 2.3.1 installation
2. Create a new block 
3. add some HTML and add CSS in <style> tag
4. Save and toggle between Editor and Code using Show/Hide Editor
5. You will see style and script tags are gone

### Expected result (*)

1. This is my input
 ![image](https://user-images.githubusercontent.com/29921007/57622959-01a6ea00-758f-11e9-93f2-5e15fed590a4.png)

### Actual result (*)

1. This is what I get after saving
![image](https://user-images.githubusercontent.com/29921007/57623002-1edbb880-758f-11e9-911e-a483e89d529e.png)

Siyadev9:
Yes I was able to reproduce on a vanilla Magento 2.3.1

krishprakash:
@magento-engcom-team give me 2.3-develop instance

magento-engcom-team:
Hi @krishprakash. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @krishprakash, here is your [Magento instance](https://i-22867-2-3-develop.instances.magento-community.engineering).
Admin access: [https://i-22867-2-3-develop.instances.magento-community.engineering/admin](https://i-22867-2-3-develop.instances.magento-community.engineering/admin)
Login: `admin` Password: `123123q`
Instance will be terminated in up to 3 hours.

krishprakash:
@magento-engcom-team give me 2.2-develop instance

magento-engcom-team:
Hi @krishprakash. Thank you for your request. I'm working on Magento 2.2-develop instance for you

magento-engcom-team:
Hi @krishprakash, here is your [Magento instance](https://i-22867-2-2-develop.instances.magento-community.engineering).
Admin access: [https://i-22867-2-2-develop.instances.magento-community.engineering/admin](https://i-22867-2-2-develop.instances.magento-community.engineering/admin)
Login: `admin` Password: `123123q`
Instance will be terminated in up to 3 hours.

krishprakash:
Hi @Siyadev9.
All css are applied properly.
Can you please check on front end all page/block display properly as applied css.
If any issue in front end then update the issue description.

[...]

PR #24114, opened 2019-08-12
Allow TinyMCE to save content with inline style tags

gwharton:
### Description (*)

Firstly, we need to decide whether we allow `<style>` tags in TinyMCE editors. HTML5.1 and older forbid inline style tags (`<style>` is only valid in `<head>`), although most browsers allow it. HTML5.2 and newer now allow it, so lets get that out of the way and decided first.

Then......

There are two issues that needed to be resolved to allow `<style>` tags within the TinyMCE 4 editor.

### Issue 1
In the `decodeVariables` function of the magentovariable TinyMCE plugin and in the `removeDuplicateAncestorWidgetSpanElement `function of the magentowidget TinyMCE plugin, the content of the TinyMCE editor is passed through a `DOMParser `to convert certain tags. Usually the content parsed through the `DOMParser `is available in the `doc.body.innerHTML` field, and is returned by the plugins once alterations had been made. There is an issue however if the content of the TinyMCE editor starts with a `<style>` tag. The `DOMParser` splits the style elements off and makes them available in `doc.head.innerHTML` and the remainder of the TinyMCE contents are parsed into `doc.body.innerHTML`.

Thus you can see how, in the original function that only processed `doc.body.innerHTML`, that any `<style>` tags were getting stripped out. Note that this only happens if the `<style>` tag is the first item in the editor. `<p>test</p><style>test</style>` would get passed in its entirety to `doc.body.innerHTML` and the bug would not happen, but `<style>test</style><p>test</p>` would get split between `doc.head.innerHTML` and `doc.body.innerHTML` and the <style> tags would be lost.

This PR corrects this by concatenating both `doc.head.innerHTML` and `doc.body.innerHTML` on return from the respective plugin functions.

### Issue 2
The TinyMCE editor enforces strict HTML5 adherence in terms of what tags can go where. Up to HTML5.1 the `<style>` tag is not allowed in the `<body>` section of code. It can only be used in the `<head>` section. This can be confirmed in the source code to TinyMCE 4.9.5 which shows that in the preparation of the HTML5 schema, that `<style>` is not included as a valid tag within a `<body>` tag.

HTML5.2 however DOES now allow the `<style>` section to appear within `<body>`. This can be confirmed [here](https://www.w3.org/TR/html52/document-metadata.html#the-style-element). There are currently no versions of TinyMCE 4 that have been updated to allow it.

This PR adds `<style>` tag as valid to occur inside `<body>` in the TinyMCE config, and thus allows it to be a top level tag within a TinyMCE editor.

1. magento/magento2#22867: Magento script and style filtering issue in CMS Blocks and Pages

1. `<p>test</p><style>test</style>`
2. `<style>test</style><p>test</p>`
3. `<p>test</p>`
4. `<style>test</style>`

### Questions or comments
As suggested in the original description, there might be some contention as to whether to allow inline `<style>` tags in the first place.

### DIFF
--- a/lib/web/mage/adminhtml/wysiwyg/tiny_mce/plugins/magentovariable/editor_plugin.js
+++ b/lib/web/mage/adminhtml/wysiwyg/tiny_mce/plugins/magentovariable/editor_plugin.js
@@ -171,7 +171,8 @@ define([
              * @returns {String}
              */
             decodeVariables: function (content) {
-                var doc = (new DOMParser()).parseFromString(content.replace(/&quot;/g, '&amp;quot;'), 'text/html');
+                var doc = (new DOMParser()).parseFromString(content.replace(/&quot;/g, '&amp;quot;'), 'text/html'),
+                    returnval = '';
 
                 [].forEach.call(doc.querySelectorAll('span.magento-variable'), function (el) {
                     var $el = jQuery(el);
@@ -195,7 +196,12 @@ define([
                     }
                 });
 
-                return doc.body ? doc.body.innerHTML.replace(/&amp;quot;/g, '&quot;') : content;
+                returnval += doc.head.innerHTML ?
+                    doc.head.innerHTML.replace(/&amp;quot;/g, '&quot;') : '';
+                returnval += doc.body.innerHTML ?
+                    doc.body.innerHTML.replace(/&amp;quot;/g, '&quot;') : '';
+
+                return returnval ? returnval : content;
             },
 
             /**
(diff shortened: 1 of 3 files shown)


VladimirZaets (approved):

magento-engcom-team:
[ENGCOM-5607](https://jira.corp.magento.com/browse/ENGCOM-5607) has been created to process this Pull Request

| Label | Description |

gwharton:
@VladimirZaets Cheers for the quick review. Static tests were failing due to coding standards failures, so i've just corrected them.

gwharton:
The exact same problem exists with the script tag which are also being stripped. Would you like me to update this PR to encompass script as well or submit new PR?

engcom-Delta:
:heavy_check_mark: QA passed

PR #24150, opened 2019-08-16
Add .editorconfig file

ihor-sviziev:
### Description (*)
This PR adds .editorconfig file. It will help IDEs automatically format different types of files correctly.

As an example - right now some files in tests have incorrect line endings.

More info: https://editorconfig.org/

N/A

N/A

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,6 @@ indent_size = 4
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


hostep:
Nice, good idea @ihor-sviziev !

I would add an exception for markdown files and set `trim_trailing_whitespace = false`.

Markdown syntax can have lines ending in a double space which means it should render the following line on a new line, if you don't end it with double spaces, it would be rendered on the same line.
Demo:
Screenshot 2019-08-16 at 20 02 55

Not stripping trailing whitespaces in markdown files would prevent these double spaces from getting accidentally deleted.

Not sure if there are markdown files in the repo which make use of these double spaces, but you never know...

ihor-sviziev:
Sure, will update my PR.

Is there any other exceptions that need to be made?

hostep:
Nothing else comes to mind. And searching the web around a bit, I can't find other exceptions which would make sense for Magento code (for example a `Makefile` should have tabs as indentation instead of spaces, but Magento makes no use of a `Makefile` at this point, so this doesn't need to get added).

Would it also make sense to include a static test to enforce the rules from this file?
This one seems to work ok-ish: https://github.com/editorconfig-checker/editorconfig-checker.php
However, if I use the proposed file from this PR, and run that tool, it already reports 241.795 violations in the entire M2 codebase, so cleaning this up would take a lot of time...
Although there are also many false positives it reports, for example this indentation is being reported as incorrect (7 spaces instead of expected 4):
```xml
<tests xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/testSchema.xsd">
...
```
So adding a static test for these rules is probably not a good idea it doesn't seems to be "smart" enough.

ihor-sviziev:
Hi @hostep,
I added exception for `.md` files. About adding static test - I think it's not good idea for now. It's better to wait and get some feedback. That's just first version, there should be more exceptions, it have to be improved for sure.

sidolov (approved):

magento-engcom-team:
[ENGCOM-5675](https://jira.corp.magento.com/browse/ENGCOM-5675) has been created to process this Pull Request

| Label | Description |

hostep:
I have one last remark about this: suppose a developer is already using an `.editorconfig` file in his Magento project with different configurations than this one.
Does that mean that when that developer upgrades to - let's say Magento 2.3.4 - this file will overwrite his file? Because that might be a bit annoying.
Would it make sense to map this file to an `.editorconfig.sample` file in the marshalling section of the `magento2-base` package's `composer.json` file?
Something like:
```
        "map": [
            [
                ".editorconfig",
                ".editorconfig.sample"
            ],
```

Just something which popped in my mind. Not sure if this is important to consider or not?

ihor-sviziev:
In this case it will be useless as it will not be applied to all people.

For case if we’ll have any issues with this version of editorconfig -
someone will report an issue and we will fix it 🙂

PR #24216, opened 2019-08-21
Allow TinyMCE to save content with inline script tags

gwharton:
### Issue 1
Since TinyMCE Version 4.2.0 the ability to have `<script>` tags inside the editor was disabled because of security concerns. See [TinyMCE 4.2.0 Release Notes](https://www.tiny.cloud/docs-4x/changelog/#version420june252015)

It is possible to add it back using the config setting 

`extended_valid_elements : "script[src|async|defer|type|charset]"`

This PR adds this setting.

### Issue 2
TinyMCE enforces that non block elements (of which `<script>` is one) are wrapped within block elements. Given the application of the above fix, `<script>` tags are now allowed, but get automatically wrapped in `<p>` tags.

This behaviour can be disabled by disabling the forced_root_block config setting in TinyMCE. This seems like the desired setting for Magento TinyMCE editors as we are only ever editing code snippets, not complete webpages where root elements `<head><body><html>` etc are present.

Edit : After this PR we were failing an MFTF test which assumed that when entering a category description, it was automatically wrapped with `<p></p>` by TinyMCE on save. Since adding the forced_root_block setting, it was no longer wrapped in `<p>'s` and thus failed. I have updated the test to pass. No other tests seem to expect the presence of `<p>` tags wrapped around any fields.

1. magento/magento2#22867: Magento script and style filtering issue in CMS Blocks and Pages

1. N/A

### Questions or comments
There may need to be an internal discussion at Magento on the resolution to Issue 2 above. Are there any unintended consequences that I am unaware of?

### DIFF
--- a/app/code/Magento/Catalog/Test/Mftf/Section/StorefrontCategoryMainSection.xml
+++ b/app/code/Magento/Catalog/Test/Mftf/Section/StorefrontCategoryMainSection.xml
@@ -20,7 +20,7 @@
         <element name="addToCartProductBySku" type="button" selector="//form[@data-product-sku='{{productSku}}']//button[contains(@class, 'tocart')]" parameterized="true" />
         <element name="SuccessMsg" type="button" selector="div.message-success"/>
         <element name="productCount" type="text" selector="#toolbar-amount"/>
-        <element name="CatalogDescription" type="text" selector="//div[@class='category-description']//p"/>
+        <element name="CatalogDescription" type="text" selector="//div[@class='category-description']"/>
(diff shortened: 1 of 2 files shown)


VladimirZaets:
Hi @gwharton. Thanks for collaboration.
We cannot accept this Pull Request due to security reason. 
It's a bad idea to allow script tag in an editor.

If you have any concerns about this, please reopen this PR and we will discuss it.

gwharton:
No problem. Understood.

Issue #24219, opened 2019-08-21
Add addComplexExceptionMessage method to ManagerInterface


atwixfirster:
### Summary (*)

There are the situations when you should
- log exception 
- display an alternative error message to customer

**Current solution:** use Magento\Framework\Message\ManagerInterface::addExceptionMessage

Minor part of `addExceptionMessage`: we can display to customer a text which does not have HTML or any special tags. For example, "We can not delete a customer". 

But if we want to display `We can't delete a customer` then customer will see `We can&#039;t delete the customer`.

### Proposed solution

Add new method - `Magento\Framework\Message\ManagerInterface`::`addComplexExceptionMessage`

### Examples of usage

*1.* Call: 

```
...
} catch (StateException $e) {
    $this->messageManager->addComplexExceptionMessage(
                'unableDeleteCustomerErrorMessage',
                $e,
                true,
                []
            );
}
....
```

*2.* Declaration in `frontend/di.xml`:

```
    <type name="Magento\Framework\View\Element\Message\MessageConfigurationsPool">
        <arguments>
            <argument name="configurationsMap" xsi:type="array">
                <item name="unableDeleteCustomerMessage" xsi:type="array">
                    <item name="renderer" xsi:type="const">\Magento\Framework\View\Element\Message\Renderer\BlockRenderer::CODE</item>
                    <item name="data" xsi:type="array">
                        <item name="template" xsi:type="string">Magento_Customer::messages/unableDeleteCustomerMessage.phtml</item>
                    </item>
                </item>
            </argument>
        </arguments>
    </type>
```

*3.* Template - `Magento/Customer/view/frontend/templates/messages/unableDeleteCustomerMessage.phtml`

```
<?php
/**
 * Copyright © Magento, Inc. All rights reserved.
 * See COPYING.txt for license details.
 */

/** @var \Magento\Framework\View\Element\Template $block */
?>

<?= $block->escapeHtml(__('We can\'t delete the customer.'));
```

ihor-sviziev:
Based on discussion in related PR - this is not really correct behavior https://github.com/magento/magento2/pull/24220#issuecomment-606065187.
>  Exceptions should not be used for business logic, client code should build the response. The exception itself should not care about its message and moreover manage those messages.

I'm closing this issue as not relevant.

PR #24220, opened 2019-08-21
magento/magento2#24219: Add addExceptionComplexMessage method

atwixfirster:

1. magento/magento2#24219: Add new method to ManagerInterface

1. ...
2. ...

Thank you!


dmytro-ch (changes requested):
@atwixfirster unfortunately it’s not allowed to add new methods to the existing interfaces.
The new interface should be introduced instead. Please check the contributors guide for more details.
Thank you!

atwixfirster:
> The new interface should be introduced instead.

**done**

@dmytro-ch, could you please review a new solution?

Thanks

lib/internal/Magento/Framework/Message/ComplexExceptionManager.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    private $exceptionMessageFactory;
+
+    /**
+     * ComplexExceptionManager constructor.
+     *
+     * @param ExceptionMessageFactoryInterface|null $exceptionMessageFactory
+     */
+    public function __construct(
+        ExceptionMessageFactoryInterface $exceptionMessageFactory = null
```
dmytro-ch:
```suggestion
        ExceptionMessageFactoryInterface $exceptionMessageFactory
```

lib/internal/Magento/Framework/Message/ComplexExceptionManager.php
```diff
@@ -0,0 +1,92 @@
...
+    ) {
+        $this->exceptionMessageFactory = $exceptionMessageFactory ?: ObjectManager::getInstance()
```
dmytro-ch:
```suggestion
        $this->exceptionMessageFactory = $exceptionMessageFactory;
```

lib/internal/Magento/Framework/Message/ComplexExceptionManager.php
```diff
@@ -0,0 +1,92 @@
...
+            ->get(ExceptionMessageLookupFactory::class);
```
dmytro-ch:
```suggestion
```

dmytro-ch (changes requested):
Hi, @atwixfirster, introducing the interface is not recommended unless you need to replace an existing interface.
Please check the Semantic Version Checker report.
Thus I think we may get rid of `Magento\Framework\Message\ComplexExceptionManagerInterface` in our particular case.

lib/internal/Magento/Framework/Message/ComplexExceptionManager.php
atwixfirster: **fixed**

lib/internal/Magento/Framework/Message/ComplexExceptionManager.php
atwixfirster: **fixed**

lib/internal/Magento/Framework/Message/ComplexExceptionManager.php
atwixfirster: **fixed**

Thank you, @dmytro-ch!

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-5786](https://jira.corp.magento.com/browse/ENGCOM-5786) has been created to process this Pull Request

| Label | Description |

sidolov:
Hi @atwixfirster , could you please clarify use cases when you need to collect a few messages in one exception? I'm not sure that collecting messages in one exception is a good practice. For cases when you need to validate few entities and collect error messages it would be better just send all information in response instead of exception throwing. Exceptions should be used in cases when we must stop script execution due to failing logic.

atwixfirster:
> Hi @atwixfirster , could you please clarify use cases when you need to collect a few messages in one exception? I'm not sure that collecting messages in one exception is a good practice. For cases when you need to validate few entities and collect error messages it would be better just send all information in response instead of exception throwing. Exceptions should be used in cases when we must stop script execution due to failing logic.

Hi, @sidolov !

A new function `addExceptionComplexMessage`  has been added regarding to logic of the existing one - `Magento\Framework\Message\Manager`::`addExceptionMessage`:

```
    public function addExceptionMessage(\Exception $exception, $alternativeText = null, $group = null)
    {
        $message = sprintf(
            'Exception message: %s%sTrace: %s',
            $exception->getMessage(),
            "\n",
            Debug::trace(
                $exception->getTrace(),
                true,
                true,
                (bool)getenv('MAGE_DEBUG_SHOW_ARGS')
            )
        );

        $this->logger->critical($message);

        if ($alternativeText) {
            $this->addErrorMessage($alternativeText, $group);
        } else {
            $this->addMessage($this->exceptionMessageFactory->createMessage($exception), $group);
        }

        return $this;
    }
```

Thank you!

atwixfirster:
@sidolov , do you have any updates here?

Thank you! :)

ihor-sviziev:
@magento run all tests

ihor-sviziev:
@magento run WebAPI Tests

ihor-sviziev (dismissed):
Hi @atwixfirster,
In general your code looks good to me.

Questions:
1. how to use it in client code?
2. what is use case for this feature?

And it's not covered with tests yet. Let's leave it after questions will be resolved

atwixfirster:
Hi, @ihor-sviziev !

> 2. what is use case for this feature?

There are the situations when you should
- log exception 
- display an alternative error message to customer

**Current solution:** use Magento\Framework\Message\ManagerInterface::addExceptionMessage

Minor part of `addExceptionMessage`: we can display to customer a text which does not have HTML or any special tags. For example, "We can not delete a customer". 

But if we want to display `We can't delete a customer` then customer will see `We can&#039;t delete the customer`.

> 1. how to use it in client code?

**Examples of usage**

*1.* Call: 

```
...
} catch (StateException $e) {
    $this->messageManager->addComplexExceptionMessage(
                'unableDeleteCustomerErrorMessage',
                $e,
                true,
                []
            );
}
....
```

*2.* Declaration in `frontend/di.xml`:

```
    <type name="Magento\Framework\View\Element\Message\MessageConfigurationsPool">
        <arguments>
            <argument name="configurationsMap" xsi:type="array">
                <item name="unableDeleteCustomerMessage" xsi:type="array">
                    <item name="renderer" xsi:type="const">\Magento\Framework\View\Element\Message\Renderer\BlockRenderer::CODE</item>
                    <item name="data" xsi:type="array">
                        <item name="template" xsi:type="string">Magento_Customer::messages/unableDeleteCustomerMessage.phtml</item>
                    </item>
                </item>
            </argument>
        </arguments>
    </type>
```

*3.* Template - `Magento/Customer/view/frontend/templates/messages/unableDeleteCustomerMessage.phtml`

```
<?php
/**
 */

/** @var \Magento\Framework\View\Element\Template $block */
?>

<?= $block->escapeHtml(__('We can\'t delete the customer.'));
```

ihor-sviziev:
Hi @atwixfirster,
We had discussion with @lbajsarowicz and seems like this change not complies with single responsibility principle. We think that message manager shouldn't care about exceptions but your client code should do that instead. In addition to that seems like original Message Manager also shouldn't have such method, but we can't remove it due to backward compatibility reasons.

@lenaorobei what do you think about deprecating `addExceptionMessage` inside?
https://github.com/magento/magento2/blob/d9ca96bc0b4ef728274ce7d9bf14d90dabafe0f3/lib/internal/Magento/Framework/Message/ManagerInterface.php#L190-L198

lbajsarowicz:
For such cases it's the best approach to introduce separate Interface capable of handling exceptions, that uses Message Manager. That could be `StorefrontExceptionHandlerInterface` that gets the Exception, logs it and adds message to MessageManager.

lenaorobei:
Closing this PR. Exceptions should not be used for business logic, client code should build the response. The exception itself should not care about its message and moreover manage those messages.

PR #24264, opened 2019-08-23
Catch enter on media library

manuelmattar:
### Description (*)
I catch the ENTER event and prevent the page to refresh and add the category

1. magento/magento2#24214:Hitting enter while creation of folder in media browser closes media browser

### DIFF
--- a/lib/web/mage/dropdown_old.js
+++ b/lib/web/mage/dropdown_old.js
@@ -9,7 +9,8 @@ define([
     'use strict';
 
     var ESC_KEY_CODE = '27';
-
+    var ENTER_KEY_CODE = '13';
+    
     $(document)
         .on('click.dropdown', function (event) {
             if (!$(event.target).is('[data-toggle=dropdown].active, ' +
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=24264) <br/>All committers have signed the CLA.

lib/web/mage/dropdown_old.js
```diff
@@ -9,7 +9,8 @@ define([
     'use strict';
 
     var ESC_KEY_CODE = '27';
-
+    var ENTER_KEY_CODE = '13';
```
VladimirZaets, line 12: Hi @manuelmattar. Thanks for collaboration. 
Please use key code mapper (`Ui/view/base/web/js/lib/key-codes.js`).

lib/web/mage/dropdown_old.js
```diff
@@ -26,6 +27,21 @@ define([
             if (event.keyCode == ESC_KEY_CODE) { //eslint-disable-line eqeqeq
                 $('[data-toggle=dropdown].active').trigger('close.dropdown');
             }
+            if (event.keyCode == ENTER_KEY_CODE) { //eslint-disable-line eqeqeq
+                event.preventDefault();
```
VladimirZaets, line 31: I'm not sure that triggered the click on the element is the best solution in this case and also why the press enter button when dropdown open should make accept action?

lib/web/mage/dropdown_old.js
```diff
@@ -26,6 +27,21 @@ define([
...
+                $('.action-accept').click();
+            }
+
+
+
+        })
+        .ready(function() {
+          $(window).keydown(function(event){
```
VladimirZaets, line 39: What this code should do? Looks like you try to prevent action that shouldn't be at all. Usually, when I open some page and press the enter button - nothing happens, because `window` doesn't have any default behavior. If in this case, this is not true maybe we should fix it in another place

VladimirZaets (changes requested):
Hi @manuelmattar. Thanks for collaboration. Please, take a  look at my comments.

slavvka:
#cd19latam

VladimirZaets:
@manuelmattar , I am closing this PR now due to inactivity.
Thank you for collaboration

PR #24273, opened 2019-08-24
Resolve undefined $currentStore in app/code/Magento/Store/App/Action/Plugin/Context.php

edenduong:
### Description (*)

Let see:

image

in app/code/Magento/Store/App/Action/Plugin/Context.php

if `catch (NoSuchEntityException $exception)` , this function is absolutely wrong

Solution: `$this->updateContext($currentStore);` must in `try` block

1. No Testing Scenario. I think it is clear to change the logic.

### DIFF
--- a/app/code/Magento/Store/App/Action/Plugin/Context.php
+++ b/app/code/Magento/Store/App/Action/Plugin/Context.php
@@ -4,6 +4,8 @@
  * See COPYING.txt for license details.
  */
 
+declare(strict_types=1);
+
 namespace Magento\Store\App\Action\Plugin;
 
 use Magento\Framework\App\Action\AbstractAction;
--- a/app/code/Magento/Store/App/Action/Plugin/Context.php
+++ b/app/code/Magento/Store/App/Action/Plugin/Context.php
@@ -94,11 +94,10 @@ class Context
         }
         try {
             $currentStore = $this->storeManager->getStore($storeCode);
+            $this->updateContext($currentStore);
         } catch (NoSuchEntityException $exception) {
             $this->processInvalidStoreRequested($exception);
         }
-
-        $this->updateContext($currentStore);
     }
 
     /**


app/code/Magento/Store/App/Action/Plugin/Context.php
```diff
@@ -94,11 +96,10 @@ public function beforeDispatch(
         }
         try {
             $currentStore = $this->storeManager->getStore($storeCode);
+            $this->updateContext($currentStore);
         } catch (NoSuchEntityException $exception) {
             $this->processInvalidStoreRequested($exception);
         }
-
-        $this->updateContext($currentStore);
```
dmytro-ch, line 101: What is the problem here?
If the `NoSuchEntityException` is thrown, then the `$this->updateContext($currentStore);` is never called.
Because `processInvalidStoreRequested` method throws another exception.

Could you please cover the corresponding case with an automated test?

dmytro-ch (changes requested):

app/code/Magento/Store/App/Action/Plugin/Context.php
edenduong: Hi @dmytro-ch : 
Thank you for your reply. Yes,it may never run if the block go to `NoSuchEntityException` . But why it below the `catch (NoSuchEntityException $exception) ` while it can only run if it run in "Try" block? It should move to the "Try" block. 

Thanks.

dmytro-ch: I don't think that we need to put any other logic to `try` block in this particular case.
The main idea is that exceptions should not be recoverable in a way that it's done here. The logic must be linear as much as possible. Thus the `try` block should contain only the calls that throw an expected exception.

Also, the `$this->processInvalidStoreRequested();` can be called any time even before this try-catch block.
So this change doesn't actually affect anything.

dmytro-ch:
Closing this PR due to the unnecessity of the mentioned improvement.
Feel free to reopen this PR if you wish to provide solid refactoring for undesirable approaches.
Thank you!

PR #24281, opened 2019-08-25
Resolve undefined $product variable in updateItem(...)

edenduong:
### Description (*)

Let check :

image

in : `app/code/Magento/Checkout/Model/Cart.php`

I believe this function absolutely go wrong when the logic runs:
```
 catch (\Magento\Framework\Exception\LocalizedException $e) {
            $this->_checkoutSession->setUseNotice(false);
            $result = $e->getMessage();
        }
```

Solution:

```
        $this->_eventManager->dispatch(
            'checkout_cart_product_update_after',
            ['quote_item' => $result, 'product' => $product]
        );
        $this->_checkoutSession->setLastAddedProductId($productId);
```

**must move to`"try"` block**

1. Test the updateItem function when the catch block runs: (can simulate by throw exception in             `$result = $this->getQuote()->updateItem($itemId, $request, $updatingParams);`
2. Expected Result: No undefined errors.

### DIFF
--- a/app/code/Magento/Checkout/Model/Cart.php
+++ b/app/code/Magento/Checkout/Model/Cart.php
@@ -4,6 +4,8 @@
  * See COPYING.txt for license details.
  */
 
+declare(strict_types=1);
+
 namespace Magento\Checkout\Model;
 
 use Magento\Catalog\Api\ProductRepositoryInterface;
@@ -291,8 +293,8 @@ class Cart extends DataObject implements CartInterface
     /**
      * Get product object based on requested product information
      *
-     * @param   Product|int|string $productInfo
-     * @return  Product
+     * @param Product|int|string $productInfo
+     * @return Product
      * @throws \Magento\Framework\Exception\LocalizedException
      */
     protected function _getProduct($productInfo)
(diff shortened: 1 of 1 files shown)


app/code/Magento/Checkout/Model/Cart.php
```diff
@@ -732,11 +742,6 @@ public function updateItem($itemId, $requestInfo = null, $updatingParams = null)
             throw new \Magento\Framework\Exception\LocalizedException(__($result));
         }
 
-        $this->_eventManager->dispatch(
-            'checkout_cart_product_update_after',
-            ['quote_item' => $result, 'product' => $product]
-        );
-        $this->_checkoutSession->setLastAddedProductId($productId);
```
dmytro-ch, line 739: According to the logic above, if any exception is thrown, then the both `dispatch` and `setLastAddedProductId` are never called. I can not imagine any case which might lead to `Undefined variable` error related $product, $productId variables.

I agree that the `Magento\Checkout\Model\Cart::updateItem` method is unproperly built on exceptions and need to be refactored entirely. Feel free to fix it and cover the changes with an appropriate automated test.

Thank you!

dmytro-ch (changes requested):

PR #24368, opened 2019-08-30
Removed Grunt Config Files From gitignore

sprankhub:
### Description (*)
The following two DevDocs posts nicely show how to use Grunt in order to generate the CSS etc.:

* [Using Grunt for Magento tasks](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/tools/using_grunt.html)
* [Compile LESS using Grunt](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/css-topics/css_debug.html)

When working on a Magento 2 project, it makes sense that anyone working on the project uses the same Grunt configuration, so that the CSS is generated in the exact same way. Hence, it makes sense to include Grunt configuration files into git. The file `/dev/tools/grunt/configs/local-themes.js` has already been removed from `.gitignore` in other pull requests (#18960 / #19350). This pull requests removes the remaining Grunt configuration files from the default `.gitignore`.

1. Create Magento 2 project.
2. Install and configure Grunt as described in the [DevDocs](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/tools/using_grunt.html#grunt_config).
3. Expected behaviour: Grunt configuration files will be added to git.
4. Actual behaviour: Grunt configuration files are not added to git, because they are included in the default `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,11 +28,8 @@ atlassian*
 /lib/internal/flex/varien/.settings
 /node_modules
 /.grunt
-/Gruntfile.js
-/package.json
 /.php_cs
 /.php_cs.cache
-/grunt-config.json
 /pub/media/*.*
 !/pub/media/.htaccess
 /pub/media/attribute/*


torhoehn (dismissed):
Thanks for your contribution!

magento-engcom-team:
[ENGCOM-5740](https://jira.corp.magento.com/browse/ENGCOM-5740) has been created to process this Pull Request

| Label | Description |

engcom-Delta:
:heavy_check_mark: QA passed

VladimirZaets (changes requested):
Hi @sprankhub. Thanks for collaboration. 
This fix is not fully valid. We provide the sample of Gruntfile.js file and package.json, but we don't want to commit it because of users can have their own customization. If you want to add some customization to all members of your team you can just update sample file that will be successfully committed and all team members will have ability yo use it

torhoehn:
@VladimirZaets I think @sprankhub uses the correct way to share the config files. If they follow your suggestion their changes could be gone with the next update of Magento. So his approach will exactly allow to have own customization and even to share it.

hostep:
@VladimirZaets: I agree and this problem could be solved if those lines get removed from the `.gitignore` file which is put in place when running `composer create-project --repository-url=https://repo.magento.com/ magento/project-community-edition`

That file seems to be slightly different then the one in this repo since it is lacking the line `app/etc/config.php` (based on a vanilla 2.3.2 install vs https://github.com/magento/magento2/blob/2.3.2/.gitignore)

So how is that `.gitignore` file changed? Does that file exists in this repo? Or is it somewhere in some non-public repo? Any chance that repo can get opened so people can also contribute to it?

Thanks!

@torhoehn: I think @VladimirZaets means that Magento devs cloning this repo are also using their own version of those files and they don't want to accidentally commit those files, hence why they are ignored.

sprankhub:
I do not really have anything to add besides the points mentioned by @torhoehn and @hostep. I really think this is the right way. If you are working on a client project, it is vital that any dev can checkout the project and directly work on it without having to copy numerous config files from one place to another.

sprankhub:
@VladimirZaets could you have a look at this pull request again given the feedback from @torhoehn, @hostep and me? Thanks!

ihor-sviziev:
@magento run all tests

ihor-sviziev:
Hi @lenaorobei,
Seems like we need to deliver this fix to composer only installations and keep it as is for GIT repo, but not sure how to do that. Seems like some internal tool should be adjusted instead.

Could you help me with processing this PR?

ihor-sviziev:
@sidolov  any thoughts?

sprankhub:
I see "progress needs update". Anything from my side that can be done to push this forward?

ihor-sviziev:
Maybe @gabrieldagama @sivaschenko ?

ihor-sviziev:
maybe @okorshenko ?

sivaschenko:
@ihor-sviziev @sprankhub this pull request requires technical discussion and approval from our side as there are different points of view regarding the .gitignore content that should be considered.

Since the priority on this pull request was set to P4, I don't think we'll be able to be addressed soon.

Is there any workaround that can solve this issue for developers that require that without the change for everyone? (I.e. using composer post install script)

Do you think the priority should be increased?

ihor-sviziev:
Seems like it’s available only for partners

sprankhub:
I still think this change makes sense - especially for client projects. If you would not like to have it in the general repository, it would indeed make sense to implement it somehow in the repository, which creates the meta packages as suggested by @hostep. However, I do not have access to it.

Since adding the respective lines to the `.gitignore` is kind of obvious and straightforward, I do not think that this needs to be added to the docs. As a reference, this is what we usually append to the default `.gitignore` from Magento:

    !/Gruntfile.js
    !/package.json
    !/grunt-config.json

Feel free to close this PR if you do not want to / "cannot" merge.

sivaschenko:
Hi @sprankhub thank you for the pull request! However, considering there are different opinions about the desired content of gitignore file and estimated low priority/severity of this change I have to close this PR.

PR #24455, opened 2019-09-05
Upload fails in image uploader due to unrequired dependency

posttechguy:
### Description (*)

In lib/internal/Magento/Framework/File/Mime.php:97 fails because mime_content_type function does not exist, which php extension fileinfo adds

Thus when the temp uploaded file, i.e., '/tmp/sdfds5we23e', which has no extension, is trying to get the extension of the file via pathinfo function in getFileExtension(), it fails to verify against the validTypes array because the extension does not exist.

1. magento/magento2#24332: [Magento 2.3.2] Upload fails in image uploader

1. Upload image via wysiwyg editor on a page without fileinfo php extension

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "ext-ctype": "*",
         "ext-curl": "*",
         "ext-dom": "*",
+        "ext-fileinfo": "*",
         "ext-gd": "*",
         "ext-hash": "*",
         "ext-iconv": "*",
--- a/lib/internal/Magento/Framework/composer.json
+++ b/lib/internal/Magento/Framework/composer.json
@@ -13,6 +13,7 @@
         "php": "~7.1.3||~7.2.0||~7.3.0",
         "ext-curl": "*",
         "ext-dom": "*",
+        "ext-fileinfo": "*",
         "ext-gd": "*",
         "ext-hash": "*",
         "ext-iconv": "*",


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/magento2?pullRequest=24455) <br/>All committers have signed the CLA.

posttechguy:
@buskamuza I have signed the Contributor License Agreement though it says still pending, my icon is appearing in the PR, (i saw this in some of the documentation), not sure what the issue is

ihor-sviziev:
Hi @posttechguy,
Commit author’s email is not associated with your github account, that’s why it showes that cla is not signed.
Could you add commiter email to your github account and check again?

If you’ll do everything correct - near commit message you’ll have your github account avatar and it will become clickable

composer.json
```diff
@@ -16,6 +16,7 @@
         "ext-ctype": "*",
         "ext-curl": "*",
         "ext-dom": "*",
+        "ext-fileinfo": "*",
```
buskamuza, line 19: Please add the same to `lib/internal/Magento/Framework/composer.json`. Otherwise this extension won't be taken into account during composer-based deployment.

buskamuza (changes requested):
I would suggest to fix this in 2.4 because adding an ext dependency can break deployments w/o this extension (they'll have to update the environment to perform upgrade).
This can be documented for now as a recommended extension to be installed.
The issue should not be critical as it's possible to fix it by installing corresponding extension.

VladimirZaets (changes requested):
Hi @posttechguy. Thanks for collaboration. Please take a look comments in the review.

ihor-sviziev:
@buskamuza 
> I would suggest to fix this in 2.4 because adding an ext dependency can break deployments w/o this extension (they'll have to update the environment to perform upgrade).
> This can be documented for now as a recommended extension to be installed.
> The issue should not be critical as it's possible to fix it by installing corresponding extension.

Actually this dependency we have in 2.3, but it’s not defined anywhere. Otherwise we wouldn’t have this issue.
As for me - it’s ok to add it to 2.3

buskamuza:
> Actually this dependency we have in 2.3, but it’s not defined anywhere.

It's a soft dependency now:
```
        if (function_exists('mime_content_type')) {
            $result = $this->getNativeMimeType($file);
        }
```
Yes, files w/o extensions won't be processed correctly, but files w/ extension will be fine.
With the added ext dependency people w/o it won't be able to install Magento at all, though they can now.

ihor-sviziev:
> > Actually this dependency we have in 2.3, but it’s not defined anywhere.
> 
> It's a soft dependency now:
> 
> ```
>         if (function_exists('mime_content_type')) {
>             $result = $this->getNativeMimeType($file);
>         }
> ```
> 
> Yes, files w/o extensions won't be processed correctly, but files w/ extension will be fine.
> With the added ext dependency people w/o it won't be able to install Magento at all, though they can now.

@buskamuza, ah yes, you’re right, so let’s add it to release line 2.4 due to backward compatibility issues

hostep:
@buskamuza: no it's not a soft dependency. Most of the time in the code it is checked if the function exists, but at one occurrence it's not being checked:
- https://github.com/magento/magento2/blob/2.3.2/app/code/Magento/Cms/Model/Wysiwyg/Images/Storage.php#L333

So either we need to add the dependency to the cms module (or framework?) in 2.3.x, or either that call should also be wrapped in a `function_exists` and an alternative should be used if it doesn't exist.

composer.json
posttechguy: Added `"ext-fileinfo": "*",` to `lib/internal/Magento/Framework/composer.json`

buskamuza:
@hostep , good catch.
This then should be refactored to use the `\Magento\Framework\File\Mime::getMimeType()`, so we don't depend on PHP native functions directly.
Or `Magento_Cms` module should have the dependency on `ext-fileinfo` because it's that module that uses it. This approach is not desirable. That's incorrect usage of PHP functions.

hostep:
Tnx for feedback @buskamuza!

@posttechguy: are you willing to update the PR with the suggested change?

ihor-sviziev (changes requested):
Please review comment from @buskamuza and adjust PR accordingly

ralphgarrison (approved):

ihor-sviziev (changes requested):
Hi @posttechguy,
Could you review following comment https://github.com/magento/magento2/pull/24455#issuecomment-530907778 and adjust your PR accordingly?

posttechguy (commented):
Please review

ihor-sviziev:
Hi @posttechguy,
I can't see changes related to `Magento_Cms` module that @buskamuza requested:
> This then should be refactored to use the `\Magento\Framework\File\Mime::getMimeType()`, so we don't depend on PHP native functions directly.
Or `Magento_Cms` module should have the dependency on `ext-fileinfo` because it's that module that uses it. This approach is not desirable. That's incorrect usage of PHP functions.

Maybe you missed to push them?

ihor-sviziev:
Hi @posttechguy,
Sorry, still can't see requested changes.

ihor-sviziev:
@posttechguy I am closing this PR now due to inactivity.

PR #24508, opened 2019-09-07
fixed issue #24363

ghost:
### Description (*)
I have added a method getDynamicTemplates in app/code/Magento/Elasticsearch6/Model/Client/ElasticSearch.php file to return the dynamic templates array any third party developer who wish to modify it can easily create an after plugin, I have not added any test case for this method if you will found it is the correct way I will surely add that too. 

1. magento/magento2#24363: Elasticsearch mappings cannot be changed

1. create any third party module and try to change the dynamic template array
2. you will face issues because the array is statically bind with the addFieldsMapping method 

### Questions or comments
let me know if I am doing right here, I can create a pool class for that too and can take the array as an argument from di so that anyone can easily add their own template, please suggest

### DIFF
--- a/app/code/Magento/Elasticsearch6/Model/Client/Elasticsearch.php
+++ b/app/code/Magento/Elasticsearch6/Model/Client/Elasticsearch.php
@@ -254,39 +254,7 @@ class Elasticsearch implements ClientInterface
                             'type' => 'text'
                         ],
                     ],
-                    'dynamic_templates' => [
-                        [
-                            'price_mapping' => [
-                                'match' => 'price_*',
-                                'match_mapping_type' => 'string',
-                                'mapping' => [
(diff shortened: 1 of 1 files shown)


sidolov (changes requested):
@ashutoshwebkul , it would be better to extract the configuration from the client class. You could do it in several ways: 
1. Define configuration in the di.xml and inject it via constructor
2. Extract configuration to the new class and use it in the client.

sidolov:
@ashutoshwebkul  I am closing this PR now due to inactivity.

tzyganu:
@sidolov This should be reopened.
I think it's better to NOT have a problem even if the code is not at the ideal standards than to have a problem with the code NOT being at the ideal standards.

PR #24530, opened 2019-09-09
Fix the false directory path comparison failure, which causes incorrect error message for no writable permission.

Hailong:
### Description (*)

It appends a DIRECTORY_SEPARATOR directly to the path without checking whether there is already one or not. This would cause the false comparison failure in some conditions. Please refer to the example in the manual testing scenarios.

None

1. Install Magento 2.3.2 on Linux or Unix machine.
2. Change the owner of Magento root directory to someone else, so that the web application user does not have writable permission.
3. Go to Magento backend, and navigate to Marketing -> Site Map
4. Add a sitemap with Path = '/' and Filename = 'sitemap.xml'
5. Click "Generate" to generate the sitemap.

Would get the following error message:
```    
Path "/var/www/magento_root_path/" cannot be used with directory "/var/www/magento_root_path/"
```
This error message is not meaningful at all, and it's actually caused by the defect in the path comparison. After applying the PR, we would get the following error message:
```
The path "/:///var/www/magento_root_path/" is not writable.
```

### Questions or comments

We might also be interested in why the scheme string in the path is `"/://"` which seems uncommon. It could be a separate issue we could track after this fix.

### DIFF
--- a/lib/internal/Magento/Framework/Filesystem/Test/Unit/Directory/PathValidatorTest.php
+++ b/lib/internal/Magento/Framework/Filesystem/Test/Unit/Directory/PathValidatorTest.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Unit Test for \Magento\Framework\Filesystem\Directory\PathValidator
+ *
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Magento\Framework\Filesystem\Test\Unit\Directory;
+
+use Magento\Framework\Filesystem\Directory\WriteInterface;
+use Magento\Framework\Filesystem\DriverInterface;
+
+class PathValidatorTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * \Magento\Framework\Filesystem\Driver
+     *
+     * @var \PHPUnit_Framework_MockObject_MockObject
+     */
+    protected $driver;
+
+    /**
+     * @var \Magento\Framework\Filesystem\Directory\PathValidator
+     */
+    protected $pathValidator;
+
+    /**
+     * Set up
+     */
+    protected function setUp()
+    {
+        $this->driver = $this->createMock(\Magento\Framework\Filesystem\Driver\File::class);
+        $this->pathValidator = new \Magento\Framework\Filesystem\Directory\PathValidator(
+            $this->driver
+        );
+    }
+
+    /**
+     * Tear down
+     */
+    protected function tearDown()
+    {
+        $this->pathValidator = null;
+    }
+
+    /**
+     * @param string $directoryPath
+     * @param string $path
+     * @param string $scheme
+     * @param bool $absolutePath
+     * @param string $prefix
+     * @dataProvider validateDataProvider
+     */
+    public function testValidate($directoryPath, $path, $scheme, $absolutePath, $prefix)
+    {
+        $this->driver->expects($this->exactly(2))
+            ->method('getRealPathSafety')
+            ->willReturnMap(
+                [
+                    [$directoryPath, $directoryPath],
+                    [null, $prefix . $directoryPath . ltrim($path, '/')],
+                ]
+            );
+
+        $this->assertNull(
+            $this->pathValidator->validate($directoryPath, $path, $scheme, $absolutePath)
+        );
+    }
+
+    /**
+     * @return array
+     */
+    public function validateDataProvider()
+    {
+        return [
+            ['/directory/path/', '/directory/path/', '/', false, '/://'],
+            ['/directory/path/', '/var/.regenerate', null, false, ''],
+        ];
+    }
+}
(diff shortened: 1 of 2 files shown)


DavidLambauer (approved):

ihor-sviziev (changes requested):
Hi @Hailong,

Now it's became more clear what's happened, but as for me this message is not correct:
> The path "/:///var/www/magento_root_path/" is not writable.

I'd like to change it to 
> The path "/var/www/magento_root_path/" is not writable.

Could you fix that?

Also would be great to cover your changes with some tests. Could you do that?

Hailong:
Hi @ihor-sviziev , thanks for your reply. I agree with you the "/://" scheme is also a problem, but I'd like to open a separate issue to track that. As it would need more changes in the file system code, it's not as straight forward as this small fix.

If it sounds good to you, I can open a new issue and take it over.

ihor-sviziev:
Hi @Hailong,

Looks reasonable. Prepared message is not part of this code, let's fix it separately.

As this change in public method - would be great to cover this change with some tests in order not to add regression there later. Will you be able to do that?

Hailong:
Hi @ihor-sviziev , sure let me add the test ASAP.

Hailong:
Hi @ihor-sviziev, I have just added a unit test case to verify this code change.

lib/internal/Magento/Framework/Filesystem/Test/Unit/Directory/PathValidatorTest.php
```diff
@@ -0,0 +1,76 @@
...
+            $this->driver
+        );
+    }
+
+    /**
+     * Tear down
+     */
+    protected function tearDown()
+    {
+        $this->driver = null;
```
ihor-sviziev: Why it’s needed? Never saw that variables are set to null in tear down method

lib/internal/Magento/Framework/Filesystem/Test/Unit/Directory/PathValidatorTest.php
```diff
@@ -0,0 +1,76 @@
...
+            $this->pathValidator->validate($directoryPath, $path, $scheme, $absolutePath)
+        );
+    }
+
+    /**
+     * @return array
+     */
+    public function validateDataProvider()
+    {
+        return [['/directory/path/', '/directory/path/', '/', false, '/://']];
```
ihor-sviziev: Could you add few more cases?

ihor-sviziev (changes requested):
Could you review my comments and failing static tests?

lib/internal/Magento/Framework/Filesystem/Test/Unit/Directory/PathValidatorTest.php
Hailong: I copied over the implementation of of ReadTest and WriteTest, where they clean up the driver mockup pointer in the tearDown method. I guess historically this is for cleaning up the resources allocated in the constructor of mocking target.

While I did some test with latest phpunit, it doesn't initialize the mocking target class at all, so I don't think the clean-up is needed either.

lib/internal/Magento/Framework/Filesystem/Test/Unit/Directory/PathValidatorTest.php
Hailong: Just added one more test case, with sample data that could happen in the real scenario.

BTW, I also opened the issue about the path string issue. https://github.com/magento/magento2/issues/24623

ihor-sviziev:
Hi @Hailong,
Thank you! Now changes looks good, bus static tests are failing. Please review and fix them

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-5852](https://jira.corp.magento.com/browse/ENGCOM-5852) has been created to process this Pull Request

engcom-Alfa:
Hi @Hailong 
During testing we faced with the issue

_**Problem:**_ The path you specified when creating the Sitemap is transferred to the beginning of the url in the error message

**Steps to reproduce:**
1. Create folder in root directory ( "test" for ex.);
2. Go to backend, and navigate to Marketing -> Site Map;
3. Add a sitemap with Path = '/test/' and Filename = 'sitemap.xml';
4. Change permissions for "test" directory (read only);
5. Click "Generate" button to generate the sitemap;

**Actual Result:**
after1

Is this problem related to this PR?

@Hailong Could you take a look, please?

Thanks!

ihor-sviziev:
> Hi @Hailong
> During testing we faced with the issue
> 
> _**Problem:**_ The path you specified when creating the Sitemap is transferred to the beginning of the url in the error message
> 
> **Steps to reproduce:**
> 
> 1. Create folder in root directory ( "test" for ex.);
> 2. Go to backend, and navigate to Marketing -> Site Map;
> 3. Add a sitemap with Path = '/test/' and Filename = 'sitemap.xml';
> 4. Change permissions for "test" directory (read only);
> 5. Click "Generate" button to generate the sitemap;
> 
> **Actual Result:**
> after1
> 
> Is this problem related to this PR?
> 
> @Hailong Could you take a look, please?
> 
> Thanks!

Hi @engcom-Alfa,
We agreed that this message should be fixed separately in https://github.com/magento/magento2/issues/24623. So for now it’s ok. 

Now situation is better than it was earlier and it’s better to deliver one small improvement than not deliver anything

engcom-Alfa:
:heavy_check_mark: QA Passed

**Before:**
before

**After:**
after

PR #24539, opened 2019-09-10
Sort the name of modules before saving them to config.php

reense:
### Description (*)
I've added a simple `ksort()` before all the modules are saved to the `app/etc/config.php` file. With this change, the modules in this file now get sorted alphabetically. 

This currently is a problem because everytime you run `setup:upgrade`, Magento will alter the order of this module list. Git will register this as a change, preventing from updating your store using a simple `git pull`. It's more an annoyance than a real problem.

None.

1. Run `setup:upgrade` on a clean Magento installation. 
2. Open `app/etc/config.php` 
3. All the values of the `modules` key should have been sorted alphabetically.

### DIFF
--- a/setup/src/Magento/Setup/Model/Installer.php
+++ b/setup/src/Magento/Setup/Model/Installer.php
@@ -442,6 +442,7 @@ class Installer
             }
         }
         if (!$dryRun) {
+            ksort($result);
             $this->deploymentConfigWriter->saveConfig([ConfigFilePool::APP_CONFIG => ['modules' => $result]], true);
         }
         return $result;


dmytro-ch:
Hi @reense,
the flow you described is actually correct:

> everytime you run setup:upgrade, Magento will alter the order of this module list. Git will register this as a change, preventing from updating your store using a simple git pull. It's more an annoyance than a real problem.

The `app/etc/config.php` is being updated as a result of `bin/magento setup:upgrade` execution, according to the modules dependency sequence.

There is also `bin/magento: module:config:status` command available for validating the `app/etc/config.php` file. For more details please check the following PR: #22733

Thank you!

hostep:
Just to add something interesting here: [PR 21020](https://github.com/magento/magento2/pull/21020) already added a better way for sorting modules in the `config.php` file. I'm not super convinced of the change, but it seems to work quite well in practice. This change is only available since Magento 2.3.2.

@reense: if you aren't on Magento 2.3.2 yet I can understand your frustration, but since Magento 2.3.2 the changes in the `config.php` file have become a lot more cleaner/manageable.

PR #24609, opened 2019-09-15
#24376  update save/load query building logic, check definition of columns us…

pierzakp:
…ed in conditions and prepare their values properly

### Description (*)

This PR is proposing changes which are casting condition types according to the column definition type during select/update, which will be improving performance especially on int type indexes, currently, in many cases, it is passed as a string, in result index is not used.

1. magento/magento2#24376: Poor performance on sales order update - string to integer

1. turn on query logging, or view in mysql queries being ran
2. update an order
3. look at query
4. entity_id will be int

### DIFF
--- a/app/code/Magento/Sales/Model/ResourceModel/EntityAbstract.php
+++ b/app/code/Magento/Sales/Model/ResourceModel/EntityAbstract.php
@@ -176,7 +176,11 @@ abstract class EntityAbstract extends AbstractDb
      */
     protected function updateObject(\Magento\Framework\Model\AbstractModel $object)
     {
-        $condition = $this->getConnection()->quoteInto($this->getIdFieldName() . '=?', $object->getId());
+        $idFieldValue = $this->getConnection()->prepareColumnValue(
+            $this->_getColumnDefinition($this->getIdFieldName(), $this->getMainTable()),
+            $object->getId()
+        );
+        $condition = $this->getConnection()->quoteInto($this->getIdFieldName() . '=?', $idFieldValue);
         $data = $this->_prepareDataForSave($object);
         unset($data[$this->getIdFieldName()]);
         $this->getConnection()->update($this->getMainTable(), $data, $condition);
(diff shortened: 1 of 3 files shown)


app/code/Magento/Sales/Model/ResourceModel/EntityAbstract.php
```diff
@@ -176,7 +176,11 @@ protected function _afterSave(\Magento\Framework\Model\AbstractModel $object)
      */
     protected function updateObject(\Magento\Framework\Model\AbstractModel $object)
     {
-        $condition = $this->getConnection()->quoteInto($this->getIdFieldName() . '=?', $object->getId());
+        $idFieldValue = $this->getConnection()->prepareColumnValue(
+            (array)$this->_getColumnDefinition($this->getIdFieldName(), $this->getMainTable()),
+            $object->getId()
+        );
+        $condition = $this->getConnection()->quoteInto($this->getIdFieldName() . '=?', $idFieldValue);
```
o-iegorov: quoteInto() will return quoted value in any case, value after prepareColumnValue() will be quoted anyway, that makes this value preparation little bit useless.

lib/internal/Magento/Framework/Model/ResourceModel/AbstractResource.php
```diff
@@ -189,6 +189,17 @@ protected function _prepareDataForTable(DataObject $object, $table)
         return $data;
     }
 
+    /**
+     * @param string $column
+     * @param string $table
+     * @return array
+     */
+    protected function _getColumnDefinition($column, $table)
```
sidolov, line 197: Unfortunately, we are not allowed to add new protected methods to the @api classes according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

sidolov (changes requested):

sidolov:
@pierzakp , I am closing this PR now due to inactivity.

PR #24657, opened 2019-09-19
Removing the $e->getPrevious() Method.

tiagosampaio:
The method `$e->getPrevious()` not always return a value and doesn't seem to make sense in this scenario because it tries to get the previous error.

### Description (*)
After getting the error below:

`Fatal error: Uncaught Error: Call to a member function getMessage() on null in vendor/magento/framework/Crontab/CrontabManager.php:212`

I noticed that the return of the method `$e->getPrevious()` is not being validated before trying to call the `getMessage()` method.

Also, probably it's not right to implement this here since this code is trying to get the message of a previous exception, not the current one.

I didn't find an issue with this description.

It's based on an improvement not in an issue.

### Questions or comments
Does my idea make sense?

### DIFF
--- a/lib/internal/Magento/Framework/Crontab/CrontabManager.php
+++ b/lib/internal/Magento/Framework/Crontab/CrontabManager.php
@@ -209,7 +209,7 @@ class CrontabManager implements CrontabManagerInterface
             $this->shell->execute('echo "' . $content . '" | crontab -');
         } catch (LocalizedException $e) {
             throw new LocalizedException(
-                new Phrase('Error during saving of crontab: %1', [$e->getPrevious()->getMessage()]),
+                new Phrase('Error during saving of crontab: %1', [$e->getMessage()]),
                 $e
             );
         }


lib/internal/Magento/Framework/Crontab/CrontabManager.php
```diff
@@ -209,7 +209,7 @@ private function save($content)
             $this->shell->execute('echo "' . $content . '" | crontab -');
         } catch (LocalizedException $e) {
             throw new LocalizedException(
-                new Phrase('Error during saving of crontab: %1', [$e->getPrevious()->getMessage()]),
+                new Phrase('Error during saving of crontab: %1', [$e->getMessage()]),
```
orlangur, line 212: You can get the idea of what this code does from a failed unit test: https://testing-service.magento-community.engineering/reports/magento/magento2/pull/24657/c2a63ebfba0bbcbfcf7634d0bfe801ca/Unit/allure-report-ce/index.html#suites/d9d312cf829cbf15ca5c954b5cd9bcd5/1969e26d32677399/

However, it seems in some cases `$e->getPrevious()` is not present - please add such logic and write a new unit test to cover it.

orlangur (changes requested):

sidolov:
@tiagosampaio  I am closing this PR now due to inactivity.

PR #24711, opened 2019-09-24
Deprecate result factories in favor of one generic

ihor-sviziev:
#### Issue
In magento we have a mess with result objects in adminhtml controllers. When you want to add new result type, for instance add json result - you have three options

1. _Inject new type with as constructor argument_ 
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Cms/Controller/Adminhtml/Wysiwyg/Images/Upload.php#L39
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Cms/Controller/Adminhtml/Wysiwyg/Images/Upload.php#L65

2. _Inject it via ObjectManger_ (really-really bad, but we could do that):
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Catalog/Controller/Adminhtml/Category/Add.php#L62
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Catalog/Controller/Adminhtml/Category/Add.php#L64

3. _Use **already existing** `resultFactory` object_:
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/lib/internal/Magento/Framework/App/Action/AbstractAction.php#L44
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/AdminNotification/Controller/Adminhtml/System/Message/ListAction.php#L72

#### Examples
In many places we're using 1st option that for different types causes not needed code duplicates, examples: 
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Sales/Controller/Adminhtml/Order.php#L111-L114

https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Checkout/Controller/Onepage.php#L114-L117

https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Customer/Controller/Adminhtml/Index.php#L222-L225

#### Solution 
My PR marks all specific result factories as `@deprecated` in favor of one generic factory that will have the same result, so we'll have only one option how to create result objects. It should decrease amount of mess in controllers.

#### Fixed Issues (if relevant)
N/A

N/A

### Questions or comments
I checked - option 3 works correctly in admin because it replaces result object types:
https://github.com/magento/magento2/blob/20069c4846043d90755d978c799b1e358ef2fe4f/app/code/Magento/Backend/etc/adminhtml/di.xml#L97-L114

### DIFF
--- a/app/code/Magento/Backend/Model/View/Result/RedirectFactory.php
+++ b/app/code/Magento/Backend/Model/View/Result/RedirectFactory.php
@@ -34,7 +34,7 @@ class RedirectFactory extends \Magento\Framework\Controller\Result\RedirectFacto
      */
     public function __construct(
         ObjectManagerInterface $objectManager,
-        $instanceName = \Magento\Backend\Model\View\Result\Redirect::class
+        $instanceName = Redirect::class
     ) {
         $this->objectManager = $objectManager;
         $this->instanceName = $instanceName;
@@ -44,7 +44,7 @@ class RedirectFactory extends \Magento\Framework\Controller\Result\RedirectFacto
      * Create class instance with specified parameters
      *
      * @param array $data
-     * @return \Magento\Backend\Model\View\Result\Redirect
+     * @return Redirect
      */
     public function create(array $data = [])
     {
--- a/lib/internal/Magento/Framework/Controller/Result/JsonFactory.php
+++ b/lib/internal/Magento/Framework/Controller/Result/JsonFactory.php
@@ -6,34 +6,39 @@
 
 namespace Magento\Framework\Controller\Result;
 
+use Magento\Framework\ObjectManagerInterface;
+
 /**
  * Factory class for @see \Magento\Framework\Controller\Result\Json
+ *
+ * @deprecated replaced with more generic ResultFactory
+ * @see \Magento\Framework\Controller\ResultFactory::create
  */
 class JsonFactory
 {
     /**
      * Object Manager instance
      *
-     * @var \Magento\Framework\ObjectManagerInterface
+     * @var ObjectManagerInterface
      */
-    protected $objectManager = null;
+    protected $objectManager;
 
     /**
      * Instance name to create
      *
      * @var string
      */
-    protected $instanceName = null;
+    protected $instanceName;
 
     /**
      * Factory constructor
      *
-     * @param \Magento\Framework\ObjectManagerInterface $objectManager
+     * @param ObjectManagerInterface $objectManager
      * @param string $instanceName
      */
     public function __construct(
-        \Magento\Framework\ObjectManagerInterface $objectManager,
-        $instanceName = \Magento\Framework\Controller\Result\Json::class
+        ObjectManagerInterface $objectManager,
+        $instanceName = Json::class
     ) {
         $this->objectManager = $objectManager;
         $this->instanceName = $instanceName;
@@ -43,7 +48,7 @@ class JsonFactory
      * Create class instance with specified parameters
      *
      * @param array $data
-     * @return \Magento\Framework\Controller\Result\Json
+     * @return Json
      */
     public function create(array $data = [])
     {
--- a/lib/internal/Magento/Framework/Controller/Result/RedirectFactory.php
+++ b/lib/internal/Magento/Framework/Controller/Result/RedirectFactory.php
@@ -9,6 +9,9 @@ use Magento\Framework\ObjectManagerInterface;
 
 /**
  * @api
+ *
+ * @deprecated replaced with more generic ResultFactory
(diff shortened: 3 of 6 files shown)


VladimirZaets (approved):

magento-engcom-team:
[ENGCOM-5951](https://jira.corp.magento.com/browse/ENGCOM-5951) has been created to process this Pull Request

ihor-sviziev:
For now moved to this PR "on hold" till it be discussed with Magento Architects

orlangur:
@ihor-sviziev are we even able to add new deprecations within `2.3.x` release line?

orlangur:
https://testing-service.magento-community.engineering/reports/magento/magento2/pull/24711/7e54f79e68bbc21bc69ccadb41cd037a/SemanticVersionChecker/report-magento2.html LOL :) So naive test implementation. I suppose it is even based on regular expressions.

buskamuza:
Summary of Arch discussion around this topic:
1. the problem with multiple result factories used in one action controller is caused by incorrect object decomposition. With correct implementation (desired state), only one result type is necessary in one action controller
2. Specialized factories provide better interface because they declare returned type. While generic ones don't.

Based on this, we all agreed that **it's better to use specialized factories than generic one**. And from the three described variants, 1st one (factory injected in constructor) should be a guideline.
2nd option is discouraged and should not be used. Maybe somebody implemented it for backward compatibility (though it's unnecessary in action controllers).
3rd option also don't look like a good approach. Probably it was just an "automatic" aggregation of parameters for action controllers with many of them.

ihor-sviziev:
> Summary of Arch discussion around this topic:
> 
> 1. the problem with multiple result factories used in one action controller is caused by incorrect object decomposition. With correct implementation (desired state), only one result type is necessary in one action controller
> 2. Specialized factories provide better interface because they declare returned type. While generic ones don't.
> 
> Based on this, we all agreed that **it's better to use specialized factories than generic one**. And from the three described variants, 1st one (factory injected in constructor) should be a guideline.
> 2nd option is discouraged and should not be used. Maybe somebody implemented it for backward compatibility (though it's unnecessary in action controllers).
> 3rd option also don't look like a good approach. Probably it was just an "automatic" aggregation of parameters for action controllers with many of them.

So in this case should we deprecate generic ResultFactory?

orlangur:
Great @buskamuza!

> With correct implementation (desired state), only one result type is necessary in one action controller

Missed this when we discussed in Slack, really nice catch 👍

ihor-sviziev:
@buskamuza,
> 1. With correct implementation (desired state), only one result type is necessary in one action controller

Actually not, in many places we'll have two types of result - Page & Forward.

**Example**
On frontend we have catalog/product/view action. In case if product isn't enabled on website or such product ID doesn't exists in DB - we need to return 404 page. How to do that - return two types of results - when we have product - return Page result, if we don't have - return forward to "no route" action.

We could apply this example to any entity on frontend, result will be the same.

I see your points, but still don't think that introducing **multiple dependencies** instead of **one** - good idea.

Am I missed something?

> 2. Specialized factories provide better interface because they declare returned type. While generic ones don't.

Sure, this is good point. Maybe we need to introduce new result factory that will have methods "createPageResult", "createRedirectResult"...? Actually currently we have 2 types of results - frontend & backend, they have the same interface

buskamuza:
> Actually not, in many places we'll have two types of result - Page & Forward.

Yes, we have such places.
I looked at a few action controllers and there are cases where:
- a single Forward factory is used
- a single another (Redirect/Page/etc) factory is used
- combination of two different types are used. It's not only Page and Forward. There are cases of using Redirect and Forward. So having a single factory with Page and Forward wouldn't be very useful.

I think this topic is partially subjective. As for me, having two factories injected is a lesser issue than having a single factory with a non-strict interface. So I'd prefer having multiple factories. Especially in action controllers, which are (or should be) small and having one more dependency should not be a big issue. In the same time it would clearly show the dependencies used.
If we would need Page and Forward factories together in all (or almost all) cases, it would definitely make sense for me to combine them in one interface. But because this doesn't look to be the case (though I didn't collect full statistics), I'd rather avoid combining them.

ihor-sviziev:
@buskamuza ok, so another question. If using "pageFactory" is not really good option - maybe we need to mark this function as deprecated with description why and alternatives?

PR #24860, opened 2019-10-04
Fixed #24704 Saving CMS Page Title from REST web API makes content empty

### LINKED ISSUE #24704: Saving CMS Page Title from REST web API makes content empty
I am not sure if this is a bug or feature. I am trying to update the CMS Page title from POSTMAN. It updates title but also makes content and other fields empty.

### Preconditions (*)

1. At least 1 CMS page
2. Web API auth key

### Steps to reproduce (*)

1. Go to the [postman](https://www.getpostman.com/) and use this URL and body content with PUT method
URL: `http://www.example.com/rest/all/V1/cmsPage/ID` (replace ID with your id (integer)
Body: 
```
{"page":
	{
		"title": "My Title"
	}
}
```

### Expected result (*)

1. Update TITLE and keep all other fields same/unchanged
2. OR only update the field provided.

### Actual result (*)

1. Updates title and makes all other fields (not provided in the body) null.

**How do I update title only or content only? I want to update one field only (title at this stage).**

shikhamis11:
Fixed #24704 Saving CMS Page Title from REST web API makes content empty

I am not sure if this is a bug or feature. I am trying to update the CMS Page title from POSTMAN. It updates title but also makes content and other fields empty.

### Preconditions (*)

1. At least 1 CMS page
2. Web API auth key

### Steps to reproduce (*)

1. Go to the [postman](https://www.getpostman.com/) and use this URL and body content with PUT method
URL: `http://www.example.com/rest/all/V1/cmsPage/ID` (replace ID with your id (integer)
Body: 
```
{"page":
	{
		"title": "My Title"
	}
}
```

### Expected result (*)

1. Update TITLE and keep all other fields same/unchanged
2. OR only update the field provided.

### Actual result (*)

1. Updates title and makes all other fields (not provided in the body) null.

**How do I update title only or content only? I want to update one field only (title at this stage).** 


aleron75:
According to [its RFC](https://tools.ietf.org/html/rfc7231#section-4.3.4):

> the PUT method requests that the state of the target resource be created or replaced with the state defined by the representation enclosed in the request message payload

Thus, technically speaking, since your representation only contains the `title`, is't correct that other fields are blanked.

That said, I think that this is not the common use of `PUT` so I'd invoke some help from the engcom team /cc @sidolov @sivaschenko

sivaschenko:
@aleron75 you are right, `PUT` is usually used to replace the information. It should also be able to clean previously set fields (that change could have negative effect for the clients adopted to current API behaviour).

The current implementation appears to be ignoring all the information from the provided CMS Page except the design and I couldn't find any other way to update the existing page via API.

@melnikovi can you please take a look at the proposed changes

sivaschenko:
As there is no response I would assume everybody is happy for this PR to be merged.

@shikhamis11 can you please update your branch and cover the change with a test.

PR #24882, opened 2019-10-06
Added logging to contact form when a LocalizedException get's thrown

BrentRobert:
### Description (*)
When a LocalizedException is being thrown after posting to the contact form, no log entry was being made. This PR fixes this.

1. magento/magento2#23645: Magento 2.3 contact form not sending emails( No log entry)

None

### Questions or comments
None


hws47a (changes requested):
I believe it's assumed that an exception from this block is not critical and can be shown to the customer. 

If there is an error happening which is bad enough to be logged, probably other piece of logic should be changed to throw another type of exception instead of LocalizedException.

PR #24894, opened 2019-10-07
magento/magento2#24780: Checks like `is customer locked` or `scope` s…

kisroman:
…hould happen before the customer will even get a token

1. magento/magento2#24780: Checks like `is customer locked` or `scope` should happen before the customer will even get a token

1. ...
2. ...


app/code/Magento/Customer/Model/ValidateCustomerByShareOption.php
```diff
@@ -0,0 +1,84 @@
...
+     * @return bool
+     */
+    public function execute(int $customerId): bool
+    {
+        $result = true;
+        if ((bool)$this->share->isWebsiteScope() === true) {
+            $websiteId = $this->storeManager->getStore()->getWebsiteId();
+            $customer = $this->customerRepository->getById($customerId);
+
+            if ((int)$customer->getWebsiteId() !== (int)$websiteId
```
naydav, line 71: What about other variants?
Looks like loading of whole customer object only for a couple of fields (website_id, lock_expires) is overhead.

Maybe we can introduce some new "contract" and will use the collection (with `loadData` instead of `load` method) under the hood.

In this case, we will get `CONS` like getting data in a few points/ways but it is very common functionality (will be used very often) and need to take into account performance.

One more way is clearing all of related `tokens` during customer locking.

So, we can communicate with the architecture team and clarify expectations.

app/code/Magento/Customer/Model/ValidateCustomerByShareOption.php
```diff
@@ -0,0 +1,84 @@
...
+                || true === $this->authentication->isLocked($customerId)
+            ) {
+                $result = false;
+            }
+        } else {
+            if (true === $this->authentication->isLocked($customerId)) {
+                $result = false;
```
naydav, line 78: `$result  = true === $this->authentication->isLocked($customerId);`

app/code/Magento/Integration/Model/ResourceModel/Oauth/Token/RequestLog.php
```diff
@@ -27,72 +31,119 @@ class RequestLog extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb im
...
     }
 
     /**
-     * {@inheritdoc}
+     * @inheritdoc
      */
     public function getFailuresCount($userName, $userType)
     {
+        if ($userType === RequestThrottler::USER_TYPE_CUSTOMER && (bool)$this->share->isWebsiteScope() === true) {
+            $websiteId = $this->storeManager->getStore()->getWebsiteId();
```
naydav, line 81: I recommend moving `$websiteId` to input parameters of methods. We should avoid temporal coupling and dependencies on some state of the system.

app/code/Magento/Webapi/composer.json
```diff
@@ -10,11 +10,11 @@
         "magento/module-authorization": "*",
         "magento/module-backend": "*",
         "magento/module-integration": "*",
-        "magento/module-store": "*"
+        "magento/module-store": "*",
+        "magento/module-customer": "*"
```
naydav, line 14: My opinion is `magento/module-webapi` *must not* have the *hard* dependency on `module-customer` module.

`magento/module-webapi` can provide some extension point and `module-customer` can use this extension point.

Example:
GraphQl module provede extension point:
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/GraphQl/Model/Query/ContextFactory.php#L56-L63

CustomerGraphQl module adds own logic via this extension point:
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/CustomerGraphQl/Model/Context/AddUserInfoToContext.php#L17
https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/CustomerGraphQl/etc/graphql/di.xml#L16-L22

dev/tests/api-functional/testsuite/Magento/GraphQl/Customer/ChangeCustomerPasswordTest.php
```diff
@@ -165,7 +165,7 @@ public function testChangePasswordIfNewPasswordIsEmpty()
     /**
      * @magentoApiDataFixture Magento/Customer/_files/customer.php
      * @expectedException \Exception
-     * @expectedExceptionMessage The account is locked.
+     * @expectedExceptionMessage The current customer isn't authorized.
```
naydav, line 168: I am not sure that a message should be changed. It's not obvious.

If my account was locked but I get a message like `The current customer isn't authorized.` then I can spend a lot of time for understanding what is wrong with my token? But the real reason will be another.

naydav (changes requested):

app/code/Magento/Customer/Model/ValidateCustomerByShareOption.php
VladimirZaets, line 71: We use 1 to 1 matching  (customer vs website). We can have a customer with same email on different websites but in DB it's 2 separate records.

According to it, we don't need this check and respectively dependency to customer module

VladimirZaets (changes requested):
Hi @kisroman. 
Thanks for collaboration.

Please take a look at the review comments.

VladimirZaets:
@kisroman, I am closing this PR now due to inactivity.
Thank you for collaboration

PR #24936, opened 2019-10-09
Add REST Api endpoint to Contact module

VitaliyMartyschuk:
### Description (*)

Add REST Api endpoint to Contact module, to allow sending emails from contact page for projects with PWA architecture 

magento/magento2/issues/24934

Create POST request to magento_site_url/rest/V1/contact with data array, params - Name, Email, Telephone, Comment.

### DIFF
--- a/app/code/Magento/Contact/Api/ContactInterface.php
+++ b/app/code/Magento/Contact/Api/ContactInterface.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Magento\Contact\Api;
+
+use Magento\Contact\Api\Data\ContactFormInterface;
+
+interface ContactInterface
+{
+    /**
+     * Post data from contact form.
+     *
+     * @param ContactFormInterface $data
+     * @return string
+     */
+    public function send(ContactFormInterface $data);
+}
--- a/app/code/Magento/Contact/Api/Data/ContactFormInterface.php
+++ b/app/code/Magento/Contact/Api/Data/ContactFormInterface.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Magento\Contact\Api\Data;
+
+interface ContactFormInterface
+{
+    const NAME = 'name';
+    const COMMENT = 'comment';
+    const EMAIL = 'email';
+    const TELEPHONE = 'telephone';
+
+    /**
+     * @return string
+     */
+    public function getName();
+    /**
+     * @param string $name
+     */
+    public function setName($name);
+    /**
+     * @return string
+     */
+    public function getComment();
+    /**
+     * @param string $comment
+     */
+    public function setComment($comment);
+    /**
+     * @return string
+     */
+    public function getEmail();
+    /**
+     * @param string $email
+     */
+    public function setEmail($email);
+    /**
+     * @return string
+     */
+    public function getTelephone();
+    /**
+     * @param string $telephone
+     */
+    public function setTelephone($telephone);
+}
--- a/app/code/Magento/Contact/Model/Contact.php
+++ b/app/code/Magento/Contact/Model/Contact.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Magento\Contact\Model;
+
+use Magento\Contact\Api\ContactInterface;
+use Magento\Contact\Api\Data\ContactFormInterface;
+use Magento\Framework\DataObject;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Serialize\SerializerInterface;
+
+/**
+ * Class Contact
+ * @package Phoenix\Contact\Model
+ */
+class Contact implements ContactInterface
+{
+    /**
+     * @var MailInterface
+     */
+    private $mail;
+    /**
+     * @var SerializerInterface
+     */
+    private $serializer;
+
+    /**
+     * Contact constructor.
+     * @param MailInterface $mail
+     * @param SerializerInterface $serializer
+     */
+    public function __construct(
+        MailInterface $mail,
+        SerializerInterface $serializer
+    ) {
+        $this->mail = $mail;
+        $this->serializer = $serializer;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function send(ContactFormInterface $data)
+    {
+        $result = [];
+        try {
+            $this->mail->send($data->getEmail(), ['data' => new DataObject($data->__toArray())]);
+            $result['message'] = __('Thanks for contacting us with your comments and questions. We\'ll respond to you very soon.');
+        } catch (LocalizedException $e) {
+            $result['error'] = $e->getMessage();
+        } catch (\Exception $e) {
+            $result['error'] = __('An error occurred while processing your form. Please try again later.');
+        }
+
+        return $this->serializer->serialize($result);
+    }
+}
--- a/app/code/Magento/Contact/Model/ContactForm.php
+++ b/app/code/Magento/Contact/Model/ContactForm.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Magento\Contact\Model;
+
+use Magento\Framework\Api\AbstractSimpleObject;
+use Magento\Contact\Api\Data\ContactFormInterface;
+
+class ContactForm extends AbstractSimpleObject implements ContactFormInterface
+{
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return $this->_data[self::NAME];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setName($name)
+    {
+        if (trim($name) === '') {
+            throw new \InvalidArgumentException(__('Enter the Name and try again.'));
+        }
+
+        $this->_data[self::NAME] = $name;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getComment()
+    {
+        return $this->_data[self::COMMENT];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setComment($comment)
+    {
+        if (trim($comment) === '') {
+            throw new \InvalidArgumentException(__('Enter the comment and try again.'));
+        }
+
+        $this->_data[self::COMMENT] = $comment;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getEmail()
+    {
+        return $this->_data[self::EMAIL];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setEmail($email)
+    {
+        if (false === \strpos($email, '@')) {
+            throw new \InvalidArgumentException(__('The email address is invalid. Verify the email address and try again.'));
+        }
+
+        $this->_data[self::EMAIL] = $email;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTelephone()
+    {
+        return $this->_data[self::TELEPHONE];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setTelephone($telephone)
+    {
+        $this->_data[self::TELEPHONE] = $telephone;
+    }
+}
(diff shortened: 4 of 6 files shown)


app/code/Magento/Contact/Api/ContactInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
```
osrecio: Not Contain Copyright sign

app/code/Magento/Contact/Api/ContactInterface.php
```diff
@@ -0,0 +1,16 @@
...
+interface ContactInterface
+{
+    /**
+     * Post data from contact form.
+     *
+     * @param ContactFormInterface $data
+     * @return string
+     */
+    public function send(ContactFormInterface $data);
```
naydav, line 15: Need to choose a more "agnostic " name.
It's service for sending some message (with own structure). From the service perspective, it doesn't matter how this data was got... via HTML form or via some call from a mobile application

naydav, line 15: Please, use type hinting for return value as well

app/code/Magento/Contact/Api/Data/ContactFormInterface.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Magento\Contact\Api\Data;
+
+interface ContactFormInterface
+{
+    const NAME = 'name';
```
naydav, line 7: Why do we need these constant? Looks like some legacy approach

app/code/Magento/Contact/Api/Data/ContactFormInterface.php
```diff
@@ -0,0 +1,44 @@
...
+    const TELEPHONE = 'telephone';
+
+    /**
+     * @return string
+     */
+    public function getName();
+    /**
+     * @param string $name
+     */
+    public function setName($name);
```
naydav, line 19: We considered that DTO should be immutable. It simplifies code, avoids temporal coupling and makes code easier for debug and maintenance.
Now Magento Framework allows creating DTO via constructor (without any setters in class)

app/code/Magento/Contact/Api/Data/ContactFormInterface.php
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public function setEmail($email);
+    /**
+     * @return string
+     */
+    public function getTelephone();
+    /**
+     * @param string $telephone
+     */
+    public function setTelephone($telephone);
```
naydav, line 43: Please, add the extension attributes.
In this case, it could be useful for adding new data fields to the message.

app/code/Magento/Contact/Model/Contact.php
```diff
@@ -0,0 +1,56 @@
+<?php
```
naydav, line 1: Dockblocks

app/code/Magento/Contact/Model/Contact.php
```diff
@@ -0,0 +1,56 @@
...
+        SerializerInterface $serializer
+    ) {
+        $this->mail = $mail;
+        $this->serializer = $serializer;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function send(ContactFormInterface $data)
```
naydav, line 42: Need to consider the possibility to send a list of messages. Does it make sense from a usability perspective?

app/code/Magento/Contact/Model/ContactForm.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        return $this->_data[self::NAME];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setName($name)
+    {
+        if (trim($name) === '') {
```
naydav, line 24: We should avoid logic in DTO (data objects) models.
If you don't have logic in your DTO then you have the possibility to transfer this data between different systems (contexts)

Also, validation depends on different contexts (in one context it could be valid, in another context the same data will not be valid).
We can extract validation logic to separate process(and this process can be extended or changes)
Example:
https://github.com/magento/async-import/blob/2.3-asynchronous-import-module/app/code/Magento/AsynchronousImportSourceDataRetrieving/Model/RetrieveSourceData.php#L86-L89

app/code/Magento/Contact/Model/ContactForm.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+namespace Magento\Contact\Model;
+
+use Magento\Framework\Api\AbstractSimpleObject;
+use Magento\Contact\Api\Data\ContactFormInterface;
+
+class ContactForm extends AbstractSimpleObject implements ContactFormInterface
```
naydav, line 8: Why do we need to extend `AbstractSimpleObject`? It could be more simple to use objects with `__construct`

app/code/Magento/Contact/Api/Data/ContactFormInterface.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    const NAME = 'name';
+    const COMMENT = 'comment';
+    const EMAIL = 'email';
+    const TELEPHONE = 'telephone';
+
+    /**
+     * @return string
+     */
+    public function getName();
```
naydav, line 15: Please, use type hinting for return value

app/code/Magento/Contact/Model/Contact.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $result = [];
+        try {
+            $this->mail->send($data->getEmail(), ['data' => new DataObject($data->__toArray())]);
+            $result['message'] = __('Thanks for contacting us with your comments and questions. We\'ll respond to you very soon.');
```
naydav, line 47: I'm not sure about this message. Looks like it's View layer responsibility.

app/code/Magento/Contact/Model/Contact.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function send(ContactFormInterface $data)
+    {
+        $result = [];
+        try {
+            $this->mail->send($data->getEmail(), ['data' => new DataObject($data->__toArray())]);
```
naydav, line 46: Please don't use NOT contract methods `__toArray`. If somebody provides own implementation of `ContactInterface`, this code will be broken.
We should depend on abstraction but not on the concrete implementation.

Look at `\Magento\Framework\Reflection\DataObjectProcessor::buildOutputDataArray` (or data can be extracted mannualy via getters if mailer don't have extension points in data)

app/code/Magento/Contact/Model/Contact.php
```diff
@@ -0,0 +1,56 @@
...
+            $result['message'] = __('Thanks for contacting us with your comments and questions. We\'ll respond to you very soon.');
+        } catch (LocalizedException $e) {
+            $result['error'] = $e->getMessage();
+        } catch (\Exception $e) {
+            $result['error'] = __('An error occurred while processing your form. Please try again later.');
+        }
+
+        return $this->serializer->serialize($result);
```
naydav, line 54: Looks like we can return `void` because it's a `synchronous operation`

If the operation was successful UI layer will be responsible for success message
If the operation was failed then exception should be thrown

app/code/Magento/Contact/Model/Contact.php
```diff
@@ -0,0 +1,56 @@
...
+    public function send(ContactFormInterface $data)
+    {
+        $result = [];
+        try {
+            $this->mail->send($data->getEmail(), ['data' => new DataObject($data->__toArray())]);
+            $result['message'] = __('Thanks for contacting us with your comments and questions. We\'ll respond to you very soon.');
+        } catch (LocalizedException $e) {
+            $result['error'] = $e->getMessage();
+        } catch (\Exception $e) {
+            $result['error'] = __('An error occurred while processing your form. Please try again later.');
```
naydav, line 51: Generally, it's a bad way to process errors in the same way as the success scenario.
If something went wrong you threw an exception. Your contract should be clear and obvious

For example, now client of the service should work with
```
$result = $service->execute();
if (!empty($result)) {
  // Change HTTP response code, process error message
}
```
But we have a general mechanism for processing Exceptions (in WebApi module)

app/code/Magento/Contact/Api/ContactInterface.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Magento\Contact\Api;
+
+use Magento\Contact\Api\Data\ContactFormInterface;
+
+interface ContactInterface
+{
+    /**
+     * Post data from contact form.
+     *
+     * @param ContactFormInterface $data
```
naydav, line 12: Need to check do we need FQN for SOAP parser?

app/code/Magento/Contact/etc/webapi.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<routes xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Webapi:etc/webapi.xsd">
+    <route url="/V1/contact" method="POST">
+        <service class="Magento\Contact\Api\ContactInterface" method="send" />
```
naydav, line 4: It will be great if PR will be contains API-functional tests.
This kind of tests are absolutely easy (like unit) and doesn't require a large entry threshold

A couple of examples,
https://github.com/magento/inventory/blob/2.3-develop/app/code/Magento/InventoryApi/Test/Api/StockRepository/CreateTest.php#L45

https://github.com/magento/magento2/blob/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Customer/AddSimpleProductToCartTest.php

app/code/Magento/Contact/etc/webapi.xml
```diff
@@ -0,0 +1,9 @@
...
+        <resources>
+            <resource ref="anonymous" />
+        </resources>
+    </route>
+</routes>
```
naydav, line 9: Maybe we also can introduce the GraphQL entry point for these API. It easy to implement but it guarantees that functionality will be used in new approaches (PWA)

naydav (changes requested):
Thank you for contribution. I added a few suggestions for improving the API contract.

app/code/Magento/Contact/Api/ContactInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Magento\Contact\Api;
+
+use Magento\Contact\Api\Data\ContactFormInterface;
+
+interface ContactInterface
```
VladimirZaets, line 7: Please rename the file and interface names to names that will represent what this interface responsible for.
As example "SendMessageInterface"

app/code/Magento/Contact/Api/ContactInterface.php
VladimirZaets, line 15: Regarding that, we will change the name of the interface to name that will represent what this interface responsible for the method can be "agnostic" and according to our practices the name should be "execute"

app/code/Magento/Contact/Api/ContactInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
VladimirZaets, line 1: Please add copyright to all files

app/code/Magento/Contact/Api/Data/ContactFormInterface.php
VladimirZaets, line 7: Please remove constants. They are redundant in this case.

VladimirZaets (changes requested):
Hi @VitaliyMartyschuk. Thanks for collaboration. Please look at the review comments

VladimirZaets:
@VitaliyMartyschuk , I am closing this PR now due to inactivity.
Thank you for collaboration

PR #24954, opened 2019-10-09
[ImportExport] Fixing the Export page that has a large amount of exported files

eduard13:
### Description (*)
This PR adds a grid for Export page, that fixes the issue if your grid has a large amount of data (more than 20k of records)

image

1. magento/magento2#24311: "CData section too big" on Export grid

1. Create a huge amount of csv files into the following folder: `var/export`
2. Try to open the `System / Export`
3. You'll get an error
image

### DIFF
--- a/app/code/Magento/ImportExport/Test/Mftf/ActionGroup/AdminAssertVisiblePagerActionGroup.xml
+++ b/app/code/Magento/ImportExport/Test/Mftf/ActionGroup/AdminAssertVisiblePagerActionGroup.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ /**
+  * Copyright © Magento, Inc. All rights reserved.
+  * See COPYING.txt for license details.
+  */
+-->
+
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
+    <actionGroup name="AdminAssertVisiblePagerActionGroup">
+        <see selector=".admin__data-grid-header .admin__control-support-text" userInput="records found" stepKey="seeRecordsLabel"/>
+        <seeElement selector=".admin__data-grid-header div.admin__data-grid-pager-wrap" stepKey="seeGridPager"/>
+    </actionGroup>
+</actionGroups>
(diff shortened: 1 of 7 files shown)


app/code/Magento/ImportExport/Test/Mftf/ActionGroup/AdminAssertVisiblePagerActionGroup.xml
```diff
@@ -0,0 +1,15 @@
...
+ /**
+  * Copyright © Magento, Inc. All rights reserved.
+  * See COPYING.txt for license details.
+  */
+-->
+
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
+    <actionGroup name="AdminAssertVisiblePagerActionGroup">
+        <see selector=".admin__data-grid-header .admin__control-support-text" userInput="records found" stepKey="seeRecordsLabel"/>
```
rogyar: Please, take a look at the existing elements for admin grids https://github.com/magento/magento2/blob/acf8f9ff281a0fb097ed40b8e58e16abbfec48c9/app/code/Magento/Ui/Test/Mftf/Section/AdminGridControlsSection.xml#L9
They are pretty common and that's a chance we can reuse a selector from there. If not, please, create a separate selector instead of hardcoding it directly in the action group

rogyar (changes requested):

app/code/Magento/ImportExport/Test/Mftf/ActionGroup/AdminAssertVisiblePagerActionGroup.xml
eduard13: Looks like the only available here is the `Per Page` dropdown selector - so we're able to use it here.
Also not sure about the pager selector, if it makes sense to move it in the mentioned by you Section, or leave as currently is.

eduard13:
Hi @rogyar, the Action Group has been adjusted. Please let me know if anything is needed.
Thank you.

app/code/Magento/ImportExport/Ui/DataProvider/ExportFileDataProvider.php
```diff
@@ -67,6 +76,9 @@ public function __construct(
             $meta,
             $data
         );
+
+        $this->fileIO = $fileIO ?: ObjectManager::getInstance()->get(File::class);
+        $this->request = $request;
```
rogyar: Sorry, just spotted one more point. Do we need this $request assignment here? I'm asking because the parent class has the protected request property (https://github.com/magento/magento2/blob/c4eeaa9a827c533e46f35af5cf7e10f8eb4737b3/lib/internal/Magento/Framework/View/Element/UiComponent/DataProvider/DataProvider.php#L100), so we can access it in the child class without the assignment.

rogyar (changes requested):

app/code/Magento/ImportExport/Ui/DataProvider/ExportFileDataProvider.php
eduard13: Hi @rogyar you're right. Just removed it.
Thank you.

rogyar (approved):

engcom-Bravo:
:heavy_check_mark: QA Passed

PR #23468, opened 2019-06-29
[Fixed] - Display double quote in order CSV/XML export on ID field.

niravkrish:
### Preconditions (*)
1. Magento 2.3.2

### Description (*)

Display double quote in order id field in order CSV export time

### Fixed Issues (if relevant)

1. magento/magento2#23465: Display double quote in order id field in order CSV export time

### Manual testing scenarios (*)

1. Login to admin panel
2. Go to sales > order
3. Select Orders and click on export.

### Expected result (*)
1. Double quote's should not be display in CSV field title

### Actual result (*)
image_2019_06_29T05_10_32_105Z

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds are green)

PR #24969, opened 2019-10-10
Issue #23465 display double quote in order id field in order CSV export time 

sashas777:
### Description (*)
 Display double quote in order id field in order CSV export time 

1. magento/magento2#23465 : Display double quote in order id field in order CSV export time

    1. Go to admin panel
    2. Select sales ->Order
    3. Export the order CSV

### Questions or comments
 According to @rogyar  there is another issue  at Magento Commerce edition for archived orders and system data collector modules. (https://github.com/magento/magento2/pull/23479)

After discussion at https://github.com/magento/magento2/pull/23468 @sidolov  and @hostep it should be fixed in Commerce modules instead of Magento_Ui module.

### DIFF
--- a/app/code/Magento/Ui/Model/Export/MetadataProvider.php
+++ b/app/code/Magento/Ui/Model/Export/MetadataProvider.php
@@ -15,6 +15,7 @@ use Magento\Framework\Locale\ResolverInterface;
 use Magento\Framework\Stdlib\DateTime\TimezoneInterface;
 
 /**
+ * Metadata Provider
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 class MetadataProvider
@@ -84,7 +85,7 @@ class MetadataProvider
                 return $childComponent;
             }
         }
-        throw new \Exception('No columns found');
+        throw new \Exception('No columns found'); // @codingStandardsIgnoreLine
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/Ui/Model/Export/MetadataProvider.php
```diff
@@ -84,7 +85,7 @@ protected function getColumnsComponent(UiComponentInterface $component)
                 return $childComponent;
             }
         }
-        throw new \Exception('No columns found');
+        throw new \Exception('No columns found'); // @codingStandardsIgnoreLine
```
aleron75, line 88: Just my curiosity, why is this directive needed?

aleron75 (approved):

engcom-Delta:
:heavy_check_mark: QA passed

app/code/Magento/Ui/Model/Export/MetadataProvider.php
sashas777: @aleron75  without it throws static code check errors.

hostep: More specifically, static code warning is: "Direct throw of generic Exception is discouraged. Use context specific instead."

Source: https://github.com/magento/magento-coding-standard/blob/f062ab4e32d4870c4b91deb2b875eb13bbd9c8f9/Magento2/Sniffs/Exceptions/DirectThrowSniff.php#L20

So it should probably be replaced with some other Exception type? Or is that backwards incompatible?

sashas777: @hostep I didn't check all cases when this function used so I prefer to not change it because it is not related to the current issue. The line for ignore I added just ot pass validation.
I can change it to LocalizedException but again cannot speak if it would be compatible will all use cases of this function.

Issue #24985, opened 2019-10-11
API returns created_at/updated_at in UTC regardless of timezone settings


itmsenior:
### Preconditions (*)

1. Magento 2.3.3
2. Set TimeZone to something that isn't UTC (I.e. GMT Europe/London - currently UTC+1)

### Steps to reproduce (*)
1. Place test order
2. Send GET request for order details to SalesOrderRepository
3. created_at/updated_at will be in UTC

### Expected result (*)

1. created_at/updated_at will use timezone set in Magento admin

### Actual result (*)

1. created_at/updated_at uses UTC timezone

webkul-ratnesh:
@magento I am working on it.

novikor:
Hi, @itwebdev2. Thank you for your collaboration.

The issue is not relevant and should be closed.

The API you referenced is a part of Magento Service Layer and it is not responsible for formatting dates in its output, - it's relevant to the presentation layer logic.
Moreover, even if we assume it is acceptable, it is not really safe to do that because existing integrations do not expect such changes.

The date format and timezone should be handled on the side of a client that uses the API.

Please don't hesitate to contact me if you have any questions regarding the issue.

PR #25002, opened 2019-10-12
updating formattted created_at and updated_at details of order in ord…

webkul-ratnesh:
…er details api #24985

### Description 
This PR contains fix for order details api's created_at and updated_at data in formatted form.

1. magento/magento2#24985: API returns created_at/updated_at in UTC regardless of timezone settings


app/code/Magento/Sales/Model/OrderRepository.php
```diff
@@ -140,6 +157,12 @@ public function get($id)
                     __("The entity that was requested doesn't exist. Verify the entity and try again.")
                 );
             }
+            $date = $this->locale->date($entity->getCreatedAt());
```
novikor, line 160: Hi, @webkul-ratnesh. Thank you for your contribution!
Well, you are right - this solution will solve the problem. However, I don't really think a repository is a place where the date formatting should happen, - it's a part of presentation layer logic, but not the service one.

In this case, I would suggest moving this logic into something kind of after-plugin that applies for `webapi_rest` and `webapi_soap` areas only.

novikor (dismissed):

novikor:
Thank you for your contribution, it is highly appreciated.

Unlikely, I should close the PR because the issue it fixes is not relevant.

PR #25011, opened 2019-10-12
18346

ghost:
### Description (*)

"Visible Section Not Found" message appears if Magento admin session times out. It must show message like you were logged out due to inactivity

1. magento/magento2#18346: "Visible Section Not Found" message appears if Magento admin session times out.

1. Login to admin panel.
2. Wait for session timeout.
3. Reload the page.

### DIFF
--- a/app/code/Magento/Config/Controller/Adminhtml/System/AbstractConfig.php
+++ b/app/code/Magento/Config/Controller/Adminhtml/System/AbstractConfig.php
@@ -7,6 +7,7 @@
 namespace Magento\Config\Controller\Adminhtml\System;
 
 use Magento\Framework\Exception\LocalizedException;
+use Magento\Security\Model\AdminSessionsManager;
 
 /**
  * System Configuration Abstract Controller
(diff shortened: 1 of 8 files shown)


app/design/adminhtml/Magento/backend/web/css/source/actions/_actions-select.less
```diff
@@ -109,7 +109,7 @@
     .action-menu-items {
         left: 0;
         position: absolute;
-        right: 0;
+        right: auto;
```
dmytro-ch, line 112: I guess, this change is not related to the current issue.

app/code/Magento/Config/Controller/Adminhtml/System/AbstractConfig.php
```diff
@@ -31,22 +32,30 @@ abstract class AbstractConfig extends \Magento\Backend\App\AbstractAction
     protected $_configStructure;
 
     /**
-     * @deprecated 100.2.0
+     * @var AdminSessionsManager
+     */
+    protected $_sessionsManager;
```
dmytro-ch, line 37: All the new properties must be declared in the private scope of visibility.
Please also don't use underscores in property names, this is a legacy approach.

app/code/Magento/Config/Controller/Adminhtml/System/AbstractConfig.php
```diff
@@ -31,22 +32,30 @@ abstract class AbstractConfig extends \Magento\Backend\App\AbstractAction
...
     /**
      * @param \Magento\Backend\App\Action\Context $context
      * @param \Magento\Config\Model\Config\Structure $configStructure
+     * @param AdminSessionsManager $sessionsManager
      * @param mixed $sectionChecker - deprecated
      */
     public function __construct(
         \Magento\Backend\App\Action\Context $context,
         \Magento\Config\Model\Config\Structure $configStructure,
+        AdminSessionsManager $sessionsManager,
```
dmytro-ch, line 53: The new constructor parameters should be optional in order to avoid backward compatibility issues.
Thus you won't need to update the classes which inherit this one.

app/code/Magento/Config/Controller/Adminhtml/System/AbstractConfig.php
```diff
@@ -58,6 +67,10 @@ public function __construct(
      */
     public function dispatch(\Magento\Framework\App\RequestInterface $request)
     {
+        if (!$this->_sessionsManager->getCurrentSession()->isLoggedInStatus()) {
+            $this->messageManager->addSuccessMessage(__('You were logged out due to inactivity.'));
```
dmytro-ch, line 71: Please use the default message, which already has translations: "Your current session has been expired."

dmytro-ch (changes requested):
Hi @vishesh-webkul,
could you please fix the code review notices and automated tests failures?

I also have doubts about the current solution.
The actual problem is caused by the incorrect behavior starting from `\Magento\Security\Model\Plugin\AuthSession::aroundProlong` plugin method. I would recommend investigating and fixing the root reason of the bug instead of applying the workarounds.

Thank you!

ghost:
Please close this pull request it has some error by mistake. I will create a new PR.

PR #25046, opened 2019-10-14
Fix incorrect PHPDoc in email transport

FreekVandeursen:
### Description (*)
After changes to email transport and message interface, not all phpdoc was updated

1. Try to send an email
2. Check variable and return types in email transport


FreekVandeursen:
Regarding the failing Semantic Version Checker: The return type has already changed in 2.3.3, so just fixing the PHPDoc doesn't need to be a MAJOR change

VladimirZaets:
Hi @FreekVandeursen. Thanks for collaboration. The bug was not just in doc, this bug was in implementation because two implementations of interface returned different values, currently, it's fixed in 2.3-develop branch and patch provide to 2.3.3 release, so this PR is no more actual.

PR #25113, opened 2019-10-17
Fix warning count parameter must be array or object implements Countable

mrtuvn:
### Description (*)

This is my fix for issue 25112. Not much changes, only add empty check 

1. magento/magento2#25112: Warning: count(): Parameter must be an array or an object that implements Countable

1. First Create New Product
2. Go to System > Export
3. Choose Entity Type: Products

### DIFF
--- a/app/code/Magento/ImportExport/Block/Adminhtml/Export/Filter.php
+++ b/app/code/Magento/ImportExport/Block/Adminhtml/Export/Filter.php
@@ -152,7 +152,7 @@ class Filter extends \Magento\Backend\Block\Widget\Grid\Extended
             }
         }
 
-        if ($size = count($options)) {
+        if (!empty($options)) {
             $arguments = [
                 'name' => $this->getFilterElementName($attribute->getAttributeCode()) . '[]',
                 'id' => $this->getFilterElementId($attribute->getAttributeCode()),
--- a/app/code/Magento/ImportExport/Block/Adminhtml/Export/Filter.php
+++ b/app/code/Magento/ImportExport/Block/Adminhtml/Export/Filter.php
@@ -152,7 +152,7 @@ class Filter extends \Magento\Backend\Block\Widget\Grid\Extended
             }
         }
 
-        if (!empty($options)) {
+        if ($size = count($options)) {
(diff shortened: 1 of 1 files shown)


ihor-sviziev (changes requested):
Hi @mrtuvn,
Your changes looks good, but they're actually fixing not the root of the issue.

It looks like the code that creating `$options` variable expected to return an array all the time.
But if this issue appeared - looks like there is the case when not array was returned, and in this case would be more correct to identify case when not array was returned and return array.

Could you adjust your code according to my feedback?

mrtuvn:
@ihor-sviziev i should add check is_array($options) && !empty($options)

ihor-sviziev:
Hi @mrtuvn,
Idea was to check in which case `$options` could be not an array. As for me - it's an issue that returned not array, when it was expected --> we need to fix code that returns not array, and we don't need to fix code that you touched.

mrtuvn:
@ihor-sviziev yeah i totally agreed with your point.
I have take a look around getOptions and i see something seem not right
1. I can't find define method $attribute->getFilterOptions();
2. $options = $attribute->getSource()->getAllOptions(false); in line 244 near line my touched code have new argument false. Method getAllOptions doesn't have any argument

ihor-sviziev:
> 1\. I can't find define method $attribute->getFilterOptions();
> 2\. $options = $attribute->getSource()->getAllOptions(false); in line 244 near line my touched code have new argument false. Method getAllOptions doesn't have any argument

This is source model for attribute, it could be defined per attribute.

I can recommend you to put breakpoint and use xdebug for undesrtanding what models and what values we're having and why

mrtuvn:
@ihor-sviziev is valid for call $attribute->getSource()->getAllOptions(false)
The origin getAllOptions doesn't have any parameter
Magento\Eav\Model\Entity\Attribute\Source\SourceInterface

I see in someplace in magento used getAllOptions() no parameter and some used getAllOptions(false) or even getAllOptions(false, false)

ihor-sviziev:
@mrtuvn  I am closing this PR now due to inactivity.

PR #25164, opened 2019-10-19
Optimize config data traversal in Store/Config/Placeholder

andrey-legayev:
### Description (*)
Optimizations:
 - get rid of a lot of string concatenation and then explode()
 - replace recursion with iterative algorithm to avoid data copying via parameters
 - utilize builtin array iterator functionality to avoid creation of additional variables
 - read/set data directly to current $data item, skip loops in _getData() and _setData()

My benchmarks for method process() on empty Magento cache have shown reduction in total time spent from 0.291 sec to 0.077 sec (-73%)
Tests were done on PHP 7.2 inside docker, XDebug=Off

N/A

General regression testing

### Questions or comments
N/A

### DIFF
--- a/app/code/Magento/Store/Model/Config/Placeholder.php
+++ b/app/code/Magento/Store/Model/Config/Placeholder.php
@@ -3,6 +3,7 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+
 namespace Magento\Store\Model\Config;
 
 /**
@@ -32,8 +33,8 @@ class Placeholder
      */
     public function __construct(\Magento\Framework\App\RequestInterface $request, $urlPaths, $urlPlaceholder)
     {
-        $this->request = $request;
-        $this->urlPaths = $urlPaths;
+        $this->request        = $request;
+        $this->urlPaths       = $urlPaths;
         $this->urlPlaceholder = $urlPlaceholder;
     }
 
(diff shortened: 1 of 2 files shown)


app/code/Magento/Store/Model/Config/Placeholder.php
```diff
@@ -164,4 +206,26 @@ protected function _setValue(array &$container, $path, $value)
...
+
+    /**
+     * Set array value by path
+     *
+     * @param array &$container
+     * @param array $path
+     * @param string $value
+     * @return void
+     */
+    private function _setCurrentPathValue(array &$container, array $path, $value)
```
dmytro-ch: Please avoid using underscores at the beginning of the private/protected method's name since this is the legacy approach according to the PSR-2 Coding Style Guide:
```
Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.
```

app/code/Magento/Store/Model/Config/Placeholder.php
```diff
@@ -164,4 +206,26 @@ protected function _setValue(array &$container, $path, $value)
         }
         $currentPointer = $value;
     }
+
+    /**
+     * Set array value by path
+     *
+     * @param array &$container
```
dmytro-ch: Could you please fix the static test failure on this line?
I guess the "&" is redundant here.

dmytro-ch (changes requested):

app/code/Magento/Store/Model/Config/Placeholder.php
andrey-legayev: Removed

app/code/Magento/Store/Model/Config/Placeholder.php
andrey-legayev: I've copy-pasted it, now removed. Waiting for new static test results...

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-6133](https://jira.corp.magento.com/browse/ENGCOM-6133) has been created to process this Pull Request

andrey-legayev:
I though a bit more on the algorithm and found a way to get rid of loops inside _getData() and _setData() by using array pointers directly on $data array.
Data read/set directly to current $data element.
Code is complicated (a bit), but it's almost 4 times faster that original algorithm.

Added unit test for empty data.

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-6133](https://jira.corp.magento.com/browse/ENGCOM-6133) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed

According to Blackfire companion:

**Before**
 
before2

**After**

after1

In my case, a reduction in total time spent from 0.443 sec to 0.215 sec (-51%)

andrey-legayev:
Thanks!
Results may be impacted by Blackfire. 
Initially I found this place in code by using Xdebug Profiler.
But then I measured time by adding logging directly into code without profiler (xdebug module was unloaded) to avoid it's impact:
```
$t = microtime(true);
// process()
$t = microtime(true) - $t;
echo $t;
```

PR #25180, opened 2019-10-21
Fix webapi returns variables in JSON using snake case and I need camelCase

### LINKED ISSUE #24681: webapi returns variables in JSON using snake case and I need camelCase
### Precondition (*)
Magento 2.3.x

### Summary (*)
I want to return JSON in camelCase in my custom webapi.xml and interface that I have created.
There is no way to configure in webapi.xml the format that you wish to use.

By default M2 will return data like so: server_transaction_id but my third party JavaScript is expecting serverTransactionId for example.

Secondly, It is also rather annoying to have to create an interface to do this instead of just being able to use json_encode the data. Using json_encode is not the solution as you end up with \ in the format.

See below:

### Examples (*)
```
interface Check3dsVersionInterface
{
    /**
     * @return string
     */
    public function getServerTransactionId();

    /**
     * @param string $serverTransactionId
     * @return $this
     */
    public function setServerTransactionId($serverTransactionId);
```

webapi.xml:
```
<route url="/V1/xx-payments/:cartId/check-3ds-version" method="POST">
        <service class="xx\xx\Api\xxServiceInterface" method="check3dsVersion"/>
        <resources>
            <resource ref="anonymous" />
        </resources>
    </route>
```

### Proposed solution
In webapi.xml, it would be useful to specify the format of the json, for example to return as camelCase or as it is currently which is underscores

jayankaghosh:
### Description (*)
Added objectKeyMap which accepts parameters from di.xml to modify data object keys
We can now modify the keys of any accessor on any class used in the rendering of the REST API, using the following (or similar) snippet in the `di.xml` of the `webapi_rest` directory. 
````
<type name="Magento\Framework\Reflection\DataObjectProcessor">
        <arguments>
            <argument name="objectKeyMap" xsi:type="array">
                <item name="Magento\Cms\Api\Data\PageInterface" xsi:type="array"> <!-- Change the accessors keys of this class -->
                    <item name="page_layout" xsi:type="string">PageLayout</item> <!-- page_layout should be shown as PageLayout in the ouput -->
                </item>
            </argument>
        </arguments>
    </type>
````

1. Fixes magento/magento2#24681 : webapi returns variables in JSON using snake case and I need camelCase

1. Add a `di.xml` inside `webapi_rest` directory of any module and inject object key names to it, similar to the following
````
<type name="Magento\Framework\Reflection\DataObjectProcessor">
        <arguments>
            <argument name="objectKeyMap" xsi:type="array">
                <item name="Magento\Cms\Api\Data\PageInterface" xsi:type="array">
                    <item name="page_layout" xsi:type="string">PageLayout</item>
                </item>
            </argument>
        </arguments>
    </type>
````
2. Call the REST API and observe the new keys reflecting.


ihor-sviziev:
@magento run all tests

ihor-sviziev:
Hi @jayankaghosh,
Thank you for your contribution! I'm really sorry that no one reviewed this PR earlier.

I'll take this PR and will provide my feedback soon.

From the first look - in general changes looks good to me, but some adjustments definitely will be required.

ihor-sviziev:
Oh... I see that the branch from which was created this PR was already removed, so we basically can't accept this PR.

Unfortunately I have to close this PR due to that reason. If you're still interesting in contributing this improvement - please create a new PR with the same changes, I'll try to take it as fast as possible.

tuyennshopstack:
@jayankaghosh could you create a new pull request base on current one.

tuyennn:
@jayankaghosh could you create a new pull request base on current one.

jayankaghosh:
@tuyennn Oh my. I forgot I've even made this PR. Lol.

Sure. I'll create a new PR at the earliest.

jayankaghosh:
@ihor-sviziev @tuyennn Here's the new PR

https://github.com/magento/magento2/pull/31001

PR #25193, opened 2019-10-21
Fixed #23290 /bin/magento config:show returns "Configuration for path: '...' doesn't exist" if config value equals to 0 or empty string 

### LINKED ISSUE #23290: ./bin/magento config:show returns "Configuration for path: '...' doesn't exist" if config value equals to 0 or empty string
### Preconditions (*)
1. Magento 2.4-develop
2. Create module `Vendor_Module`
3. Create `etc/config.xml` in created module `Vendor_Module` with new config path `default/vendor_module/general/value` with value `0` or `''` (empty string)

### Steps to reproduce (*)
1. Open magento cli and run command `./bin/magento config:show vendor_module/general/value`

### Expected result (*)
1. Should return `0` or `''` (empty string)

### Actual result (*)
1. Command returns `Configuration for path: "vendor_module/general/value" doesn't exist`

I think it happens because there `<magento_root>/vendor/magento/module-config/Console/Command/ConfigShowCommand.php:153` magento checks if value is empty and if it actualy is the error message is shown. But empty value does not mean that path does not exist. So we should check if path exist in different method.

shikhamis11:
Fixed #23290 /bin/magento config:show returns "Configuration for path: '...' doesn't exist" if config value equals to 0 or empty string 

### Preconditions (*)
1. Magento 2.3.1
2. Create module `Vendor_Module`
3. Create `etc/config.xml` in created module `Vendor_Module` with new config path `default/vendor_module/general/value` with value `0` or `''` (empty string)

### Steps to reproduce (*)
1. Open magento cli and run command `./bin/magento config:show vendor_module/general/value`

### Expected result (*)
1. Should return `0` or `''` (empty string)

### Actual result (*)
1. Command returns `Configuration for path: "vendor_module/general/value" doesn't exist`

I think it happens because there `<magento_root>/vendor/magento/module-config/Console/Command/ConfigShowCommand.php:153` magento checks if value is empty and if it actualy is the error message is shown. But empty value does not mean that path does not exist. So we should check if path exist in different method.

### DIFF
--- a/app/code/Magento/Config/App/Config/Source/RuntimeConfigSource.php
+++ b/app/code/Magento/Config/App/Config/Source/RuntimeConfigSource.php
@@ -54,13 +54,13 @@ class RuntimeConfigSource implements ConfigSourceInterface
      * Get initial data.
      *
      * @param string $path Format is scope type and scope code separated by slash: e.g. "type/code"
-     * @return array
+     * @return mixed
      * @since 100.1.2
      */
     public function get($path = '')
     {
         $data = new DataObject($this->loadConfig());
-        return $data->getData($path) ?: [];
+        return !is_null($data->getData($path)) ? $data->getData($path): [];
     }
 
     /**
--- a/app/code/Magento/Config/Console/Command/ConfigShowCommand.php
+++ b/app/code/Magento/Config/Console/Command/ConfigShowCommand.php
@@ -150,7 +150,7 @@ class ConfigShowCommand extends Command
             $configPath = $this->pathResolver->resolve($this->inputPath, $this->scope, $this->scopeCode);
             $configValue = $this->configSource->get($configPath);
 
-            if (empty($configValue)) {
+            if (is_array($configValue) && empty($configValue)) {
                 $output->writeln(sprintf(
                     '<error>%s</error>',
                     __('Configuration for path: "%1" doesn\'t exist', $this->inputPath)->render()


app/code/Magento/Config/App/Config/Source/RuntimeConfigSource.php
```diff
@@ -54,13 +54,13 @@ public function __construct(
...
      * @param string $path Format is scope type and scope code separated by slash: e.g. "type/code"
-     * @return array
+     * @return mixed
      * @since 100.1.2
      */
     public function get($path = '')
     {
         $data = new DataObject($this->loadConfig());
-        return $data->getData($path) ?: [];
+        return !is_null($data->getData($path)) ? $data->getData($path): [];
```
swnsma, line 63: Hi @shikhamis11!
Could we add these cases to the unit tests `\Magento\Config\Test\Unit\App\Config\Source\RuntimeConfigSourceTest` ?

swnsma (changes requested):

app/code/Magento/Config/App/Config/Source/RuntimeConfigSource.php
shikhamis11: sure let me check

swnsma:
Could you check failed tests also?

shikhamis11:
> Could you check failed tests also?

yes

VladimirZaets:
@shikhamis11 hi, do you have some updates?

shikhamis11:
hi @VladimirZaets 
yes I am working on it .. I will try to update my changes as soon as possible thanks

app/code/Magento/Config/Console/Command/ConfigShowCommand.php
```diff
@@ -150,7 +150,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $configPath = $this->pathResolver->resolve($this->inputPath, $this->scope, $this->scopeCode);
             $configValue = $this->configSource->get($configPath);
 
-            if (empty($configValue)) {
+            if (is_array($configValue) && empty($configValue)) {
```
swnsma, line 153:
```suggestion
            if ($configValue === null) {
```

app/code/Magento/Config/App/Config/Source/RuntimeConfigSource.php
swnsma, line 63:
```suggestion
        return $data->getData($path) === null ? null : $data->getData($path);
```

shikhamis11: thank you for your review 
also there is  one possibility that config value is actually set `NULL` then in that case result will be error message like `Configuration for path: "test/active/enable" doesn't exist` which will not be correct

PR #25198, opened 2019-10-21
#22084: Remove hard dependency on Magento_Wishlit with Magento_Captcha

viniciusfabri:
### Description (*)
Another Pull Request added a hard dependency between Magento_Captcha and Magento_Wishlist

1. magento/magento2#22084: [2.3.1] Wishlist module shouldn't have a hard dependency on Captcha module

1. Magento 2.3.* with sample data installed
2. Enable captcha on wishlist sharing form
3. Add product to wishlist
4. Try to share wishlist, you should be able to see a captcha and only submit if captcha is correct
5. Remove captcha from wishlist sharing form (but keep module enabled)
6. Try to share wishlist, you should be able to submit normally without captcha
7. Disable Captcha module
8. Try to share wishlist, you should be able to submit normally without captcha

### Questions or comments
Please let me know if I need to update any Tests


viniciusfabri:
I'm checking the failed checks

app/code/Magento/Wishlist/view/frontend/layout/wishlist_index_share.xml
```diff
@@ -9,7 +9,19 @@
     <update handle="customer_account"/>
     <body>
         <referenceContainer name="content">
-            <block class="Magento\Wishlist\Block\Customer\Sharing" name="wishlist.sharing" template="Magento_Wishlist::sharing.phtml" cacheable="false"/>
+            <block class="Magento\Wishlist\Block\Customer\Sharing" name="wishlist.sharing" template="Magento_Wishlist::sharing.phtml" cacheable="false">
+                <block class="Magento\Captcha\Block\Captcha" name="captcha" after="-" cacheable="false" ifconfig="customer/captcha/enable">
```
dmytro-ch, line 13: With this layout update, we still have a soft dependency. So it can not be removed from `app/code/Magento/Wishlist/etc/module.xml`.

dmytro-ch (changes requested):
Hi @viniciusfabri, 
Could you please review the CR comment and check the automated test failures?

app/code/Magento/Wishlist/view/frontend/layout/wishlist_index_share.xml
viniciusfabri: Yeah I agree, while the dependency isn't explicit in module.xml and it will work with Captcha disabled, it's still a dependency

I was looking on how Magento_Customer does it, and it's actually the Magento_Captcha module that adds the block (and then it has a dependency on Magento_Customer)

I didn't had time to deep dive into this issue yet, but do you have any suggestions?

I could move all this logic to the Captcha module but feels weird having Magento_Wishlist as a dependency of Magento_Captcha

Another approach would be an intermediary module Magento_WishlistCaptcha that depends on both and enables Captcha on Wishlist sharing

What do you think?

dmytro-ch:
@magento run all tests

PR #25223, opened 2019-10-22
Braintree - Don't request ClientToken if payment method is not active

andrey-legayev:
### Description (*)
If Braintree module is enabled then it requests ClientToken via Curl request on every shopping cart and checkout page load **regardless active/non-active configuration** of payment method.
Curl request takes 0.3-0.5 sec to respond for servers in N.Virginia which is making Shopping Cart and Checkout pages much slower.

My pull request disables requesting Braintree ClientToken if Braintree payment method is not enabled for current store view.

https://github.com/magento/magento2/issues/25343

1. Test purchases with Braintree payment method.

### Questions or comments
N/A

### DIFF
--- a/app/code/Magento/Braintree/Model/Ui/ConfigProvider.php
+++ b/app/code/Magento/Braintree/Model/Ui/ConfigProvider.php
@@ -67,11 +67,12 @@ class ConfigProvider implements ConfigProviderInterface
     public function getConfig()
     {
         $storeId = $this->session->getStoreId();
+        $isActive = $this->config->isActive($storeId);
         return [
             'payment' => [
                 self::CODE => [
-                    'isActive' => $this->config->isActive($storeId),
-                    'clientToken' => $this->getClientToken(),
+                    'isActive' => $isActive,
+                    'clientToken' => $isActive ? $this->getClientToken() : null,
                     'ccTypesMapper' => $this->config->getCcTypesMapper(),
                     'sdkUrl' => $this->config->getSdkUrl(),
                     'hostedFieldsSdkUrl' => $this->config->getHostedFieldsSdkUrl(),
--- a/app/code/Magento/Braintree/Model/Ui/ConfigProvider.php
+++ b/app/code/Magento/Braintree/Model/Ui/ConfigProvider.php
@@ -3,116 +3,215 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
-namespace Magento\Braintree\Model\Ui;
+declare(strict_types=1);
+
+namespace Magento\Braintree\Test\Unit\Model\Ui;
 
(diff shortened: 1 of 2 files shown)


dmytro-ch (changes requested):
Hello @andrey-legayev,
thank you for your contribution. Due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html#automated-tests) the changes should be covered by tests. Could you please cover your fix by automated tests?
Thank you!

andrey-legayev:
Sorry, I have no time for it now. May be in a week.

dmytro-ch (approved):

engcom-Bravo:
Hello @andrey-legayev. Please cover Your PR with appropriate Auto-Tests. Thank You.

ihor-sviziev:
Hi @engcom-Bravo,
Could you cover these changes by internal team?

These changes really improves performance of shopping cart/checkout pages. I think that's much better to accept this PR even without covering with any tests rather than wait till author will have a time for covering it with tests.

This is really not obvious finding.

engcom-Bravo:
Hi @ihor-sviziev. Ok, I'll find out about it.

andrey-legayev:
Some details about performance impact:
We added debug logging to our production Magento2 to track how much time is spent on Braintree's getClientToken()
The result is: **0.231 sec on average** for 24 hours of logging.

Token is not used in two cases:
1. if payment method is not active - that's obvious, this is why I submitted this pull request
2. on shopping cart - it's triggered via long chain of calls, but all starts from Estimate Shipping Cost block (which doesn't use Braintree's token)

In both cases there are no reason why page should be 0.231 sec slower.

Would be great to avoid token generation on shopping cart as well.
For now I see it doable only via ajax request.

ihor-sviziev:
Hi @andrey-legayev,

I just looked on "clientToken" and found that in Admin client token already taken via AJAX.

https://github.com/magento/magento2/blob/890bf891f86d876c1e2651978ba287958467dd9f/app/code/Magento/Braintree/view/adminhtml/web/js/braintree.js#L136-L144

It runs following controller code:
https://github.com/magento/magento2/blob/0eb8677b0b4e35606032e856cc1ef7c80e68829f/app/code/Magento/Braintree/Controller/Adminhtml/Payment/GetClientToken.php#L59-L75

I think it's good idea to adjust also frontend JS to get it via AJAX, so it will not slow down the page when it's not needed.

But that's should be separate PR.

engcom-Golf:
Hi @andrey-legayev  thank you for Pull Request I will take care of test coverage.

engcom-Golf:
:heavy_check_mark: QA Passed

ihor-sviziev:
Hi @andrey-legayev,
Based on our discussion I reported an issue https://github.com/magento/magento2/issues/25343

PR #25272, opened 2019-10-24
Fixes the problem with page reload when js prompt was closed with enter

krisdante:
### Description (*)

Forms inside prompt.js did not have any submit action, so hitting enter
inside any of the inputs caused submission to the current page - reload of the page.
All the JS logic was lost.
I added a default submit handler (whenever there is a form in the content of the prompt)
and default action that is the same as pressing the OK button on the prompt.
If one does not want this behavior, he can pass an empty selector in options,
and it will disable the onSubmit listener.
This update will prevent faulty submission (by hitting Enter key) on all prompts.

1. magento/magento2#24214: Hitting enter while creation of folder in media browser closes media browser

Go to Content > Blocks > New Block
Use Insert Image option -> Media Browser will open
Use the option to Create Folder
Type in folder name like "foobar" and hit Enter key
Folder shall be created - the dialog shall behave the same way as I would click on OK button

### DIFF
--- a/app/code/Magento/Ui/view/base/web/js/modal/prompt.js
+++ b/app/code/Magento/Ui/view/base/web/js/modal/prompt.js
@@ -73,13 +73,14 @@ define([
          */
         _create: function () {
             var modalContent;
-            
+
             this.options.focus = this.options.promptField;
             this.options.validation = this.options.validation && this.options.validationRules.length;
             this._super();
             modalContent = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
-            if(this.options.formSelector) {
-                modalContentElem.find(this.options.formSelector).on('submit', _.bind(this.submitForm, this));
+
+            if (this.options.formSelector) {
+                modalContent.find(this.options.formSelector).on('submit', _.bind(this.submitForm, this));
             }
             this.modal.find(this.options.modalCloseBtn).off().on('click',  _.bind(this.closeModal, this, false));
 
@@ -126,7 +127,7 @@ define([
 
         /**
          * Handle submit of the form inside prompt
-         * @param event {Event} Event triggered by submit
+         * @param {Event} event
          */
         submitForm: function (event) {
             event.preventDefault();
--- a/app/code/Magento/Ui/view/base/web/js/modal/prompt.js
+++ b/app/code/Magento/Ui/view/base/web/js/modal/prompt.js
@@ -22,6 +22,7 @@ define([
             modalClass: 'prompt',
             promptContentTmpl: promptContentTmpl,
             promptField: '[data-role="promptField"]',
+            formSelector: 'form',
             attributesForm: {},
             attributesField: {},
             value: '',
@@ -71,10 +72,15 @@ define([
          * Create widget.
          */
         _create: function () {
+            var modalContentElem;
+            
             this.options.focus = this.options.promptField;
             this.options.validation = this.options.validation && this.options.validationRules.length;
             this._super();
-            this.modal.find(this.options.modalContent).append(this.getFormTemplate());
+            modalContentElem = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
+            if(this.options.formSelector) {
+                modalContentElem.find(this.options.formSelector).on('submit', _.bind(this.submitForm, this));
+            }
             this.modal.find(this.options.modalCloseBtn).off().on('click',  _.bind(this.closeModal, this, false));
 
             if (this.options.validation) {
@@ -116,6 +122,15 @@ define([
             }));
 
             return formTemplate;
+        },
+
+        /**
+         * Handle submit of the form inside prompt
+         * @param event {Event} Event triggered by submit
+         */
+        submitForm: function (event) {
+            event.preventDefault();
+            this.closeModal(true);
         },
 
         /**
--- a/app/code/Magento/Ui/view/base/web/js/modal/prompt.js
+++ b/app/code/Magento/Ui/view/base/web/js/modal/prompt.js
@@ -72,12 +72,12 @@ define([
          * Create widget.
          */
         _create: function () {
-            var modalContentElem;
+            var modalContent;
             
             this.options.focus = this.options.promptField;
             this.options.validation = this.options.validation && this.options.validationRules.length;
             this._super();
-            modalContentElem = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
+            modalContent = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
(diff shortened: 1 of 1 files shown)


krisdante:
I do not see anything that fails in Static Tests. I do not know why it is marked red.

DrewML:
Shouldn't need any JavaScript for this.

https://www.w3.org/TR/html52/sec-forms.html#implicit-submission

Adding a hidden `<input type="submit">` will prevent `enter` from submitting the form.

https://www.w3.org/TR/html52/sec-forms.html#implicit-submission

image

Alternatively, if this isn't actually being used to submit form data, the `form` element isn't strictly necessary.

krisdante:
@DrewML Yes, we can disable submission with kind of hack with hidden submit but that is for me bad because of worse User Experience (you just want to submit the form with Enter key, not using a mouse. The OK button from the js prompt cannot be even easily accessed by TAB key. 

Input cannot be 'submit' and 'hidden' same time, so the submit button would have to be hidden with CSS, that would be not WCAG pleasent especially that there is another "real" submit.

I could also remove the form - but that is a breaking change, as the template has form now, there is also the whole logic to add additional form attributes via extensions - so we cannot just strip out from tag from the template.

krisdante:
@DrewML Any feedback on above?

app/code/Magento/Ui/view/base/web/js/modal/prompt.js
```diff
@@ -71,10 +72,15 @@ define([
...
         _create: function () {
+            var modalContentElem;
+            
             this.options.focus = this.options.promptField;
             this.options.validation = this.options.validation && this.options.validationRules.length;
             this._super();
-            this.modal.find(this.options.modalContent).append(this.getFormTemplate());
+            modalContentElem = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
+            if(this.options.formSelector) {
+                modalContentElem.find(this.options.formSelector).on('submit', _.bind(this.submitForm, this));
```
krzksz: Base `modal` widget which this `prompt` is extending has this custom handler option that can be provided for closing the modal https://github.com/magento/magento2/blob/e6531ff15903b00c9e46712711572009cf28dbc8/app/code/Magento/Ui/view/base/web/js/modal/modal.js#L136. Maybe we could allow people to pass custom form submition handler just to be on the safe side?

app/code/Magento/Ui/view/base/web/js/modal/prompt.js
```diff
@@ -71,10 +72,15 @@ define([
          * Create widget.
          */
         _create: function () {
+            var modalContentElem;
```
krzksz: I think `modalContent` would be enough for a name here.

krzksz: I think `modalContent` should be enough of a name here.

krzksz (changes requested):
Please also take a look at failing static tests.

app/code/Magento/Ui/view/base/web/js/modal/prompt.js
krisdante: I do not see a benefit - and can be actually confusing. This form is not being submitted by the OK button. Submit here is a side effect. 
I would not add additional possibilities to a "corner case" not predicted by the original functionality of the OK button.

krisdante:
@krzksz Do you know which static test is failing? Because I do not have a clue :( - I do not see this is report. Everything is green.

krzksz:
@krisdante here it is from my local run:
```
jscs app/code/Magento/Ui/view/base/web/js/modal/prompt.js --preset=dev/tests/static/testsuite/Magento/Test/Js/_files/jscs/.jscsrc 
disallowTrailingWhitespace: Illegal trailing whitespace at app/code/Magento/Ui/view/base/web/js/modal/prompt.js :
    74 |        _create: function () {
    75 |            var modalContentElem;
    76 |            
--------^
    77 |            this.options.focus = this.options.promptField;
    78 |            this.options.validation = this.options.validation && this.options.validationRules.length;

requirePaddingNewlinesBeforeKeywords: Keyword `if` should have an empty line above it at app/code/Magento/Ui/view/base/web/js/modal/prompt.js :
    78 |            this.options.validation = this.options.validation && this.options.validationRules.length;
    79 |            this._super();
    80 |            modalContentElem = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
-----------------------------------------------------------------------------------------------------------------^
    81 |            if(this.options.formSelector) {
    82 |                modalContentElem.find(this.options.formSelector).on('submit', _.bind(this.submitForm, this));

requireSpaceAfterKeywords: One space required after "if" keyword at app/code/Magento/Ui/view/base/web/js/modal/prompt.js :
    79 |            this._super();
    80 |            modalContentElem = this.modal.find(this.options.modalContent).append(this.getFormTemplate());
    81 |            if(this.options.formSelector) {
----------------------^
    82 |                modalContentElem.find(this.options.formSelector).on('submit', _.bind(this.submitForm, this));
    83 |            }

jsDoc: Missing hyphen before description at app/code/Magento/Ui/view/base/web/js/modal/prompt.js :
   127 |        /**
   128 |         * Handle submit of the form inside prompt
   129 |         * @param event {Event} Event triggered by submit
-------------------^
   130 |         */
   131 |        submitForm: function (event) {

jsDoc: Missing param type at app/code/Magento/Ui/view/base/web/js/modal/prompt.js :
   127 |        /**
   128 |         * Handle submit of the form inside prompt
   129 |         * @param event {Event} Event triggered by submit
--------------------------^
   130 |         */
   131 |        submitForm: function (event) {


5 code style errors found.
```

krzksz (approved):

magento-engcom-team:
[ENGCOM-6465](https://jira.corp.magento.com/browse/ENGCOM-6465) has been created to process this Pull Request
:eight_spoked_asterisk: @krzksz, could you please add one of the following labels to the Pull Request?

| Label | Description |

engcom-Alfa:
Hi @krisdante .
     Due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html) all code must be covered by tests. Please cover your fix by jasmine test.
Thanks!

engcom-Golf:
I will take care of test coverage

krisdante:
I am working on tests, please leave this to me so I can train it myself

engcom-Golf:
Hi @krisdante any updates with test ? or we can cover it by self ?

VladimirZaets:
@krisdante , I am closing this PR now due to inactivity.
Thank you for collaboration

Issue #25373, opened 2019-10-30
The 'promotion' region of the minicart is never rendered


mattijv:
### Preconditions (*)
1. M2.3.1 or later (and possibly some later M2.2 versions too)

### Steps to reproduce (*)
1. Add a component to the 'promotion' region of the minicart jsLayout

### Expected result (*)
1. Component is rendered

### Actual result (*)
1. Component is not rendered

### Cause
The issue was introduced in https://github.com/magento/magento2/pull/20383 where an if-binding was added to the minicart-widgets div to prevent the div from being rendered if the 'promotion' region does not contain any elements. Unfortunately the if was written badly as the conditional is checking the `length` property of the observable returned from getRegion (always 0) instead of the array contained in the observable. This means the conditional is always false and the element is never rendered.

### Possible fixes
I've create two mutually exclusive PR's that solve the issue in two ways.

https://github.com/magento/magento2/pull/25374 Fixes the if-clause by adding the missing parentheses.

https://github.com/magento/magento2/pull/25375 Is a bit more involved fix as it adds a separate method to the collection component that can be used to check if a region contains elements. It also modifies existing uses of the `getRegion(region)().length` pattern to use the new method.

VladimirZaets:
Hi @mattijv. Thank you for your report. 
The issue has been fixed in magento/magento2#25375 by @mattijv in 2.4-develop branch
Related commit(s):
 - [20d3f40795e8c29aa1a307c55817f557698f2f8a](https://github.com/magento/magento2/commit/20d3f40795e8c29aa1a307c55817f557698f2f8a)
 - [9b6c8d74daf13a1de3124d6999751f158004f2b2](https://github.com/magento/magento2/commit/9b6c8d74daf13a1de3124d6999751f158004f2b2)
 - [ca8c75a303918257d1fca4cc110ad7348b05a72f](https://github.com/magento/magento2/commit/ca8c75a303918257d1fca4cc110ad7348b05a72f)
 - [f9eb4e4c0867eeb8adee181ca6137bfc43cc1fed](https://github.com/magento/magento2/commit/f9eb4e4c0867eeb8adee181ca6137bfc43cc1fed)
 - [109392d8d78f6ada85e13f1523537888a66e8da0](https://github.com/magento/magento2/commit/109392d8d78f6ada85e13f1523537888a66e8da0)

The fix will be available with the upcoming 2.4.0 release.

PR #25374, opened 2019-10-30
Fix 'promotion' region not rendering #25373

mattijv:
The getRegion(name) method returns a knockoutjs observable array which is a function, so the check used in the template will always resolve to false (as function `length` property is the number of arguments a function takes, which for the observable array is 0). Therefore the region is never rendered.

### Description (*)
The fix in this PR follows the current convention used in Magento and adds the missing parentheses to the if condition.

1. https://github.com/magento/magento2/issues/25373: The 'promotion' region of the minicart is never rendered

See related issue

### Questions and comments
This PR is an alternative to https://github.com/magento/magento2/pull/25375. The fix implement in this PR simple and follows the existing code conventions in Magento, but does nothing to prevent a similar mistake from happening in the future.

### DIFF
--- a/app/code/Magento/Checkout/view/frontend/web/template/minicart/content.html
+++ b/app/code/Magento/Checkout/view/frontend/web/template/minicart/content.html
@@ -97,7 +97,7 @@
         </div>
     </div>
 
-    <div id="minicart-widgets" class="minicart-widgets" if="getRegion('promotion').length">
+    <div id="minicart-widgets" class="minicart-widgets" if="getRegion('promotion')().length">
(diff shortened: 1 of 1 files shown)


mattijv:
Apparently there is some latency to the CLA signing being registered? I'm going to try closing and reopening this PR later.

mattijv:
As far as I can tell, all the failing checks are unrelated to the changes in this PR.

mattijv:
As https://github.com/magento/magento2/pull/25375 seems to be the preferred fix for the related issue, I'll close this PR as redundant.

PR #25375, opened 2019-10-30
Fix minicart promotion region not rendering #25373

mattijv:
The minicart promotion region was not rendering due to a wrongly written check for the number of items in the 'promotion' region. The code checked for the length property of the observable returned from the getRegion method instead of the array it contained.

A fast way would have been to add the missing parentheses to the check (and I made antoher PR doing just that), but I felt that the getRegion(region)().length pattern is a bit hard to read and easy to make a mistake with.

### Description (*)
This PR adds a new method to the lib/core/collection.js component of the Magento_Ui module that can be used to check if a region has any elements associated with it. This commit also replaces all occurences of getRegion(region)().length with a call to the new method.

1. https://github.com/magento/magento2/issues/25373: The 'promotion' region of the minicart is never rendered

See related issue.

### Questions or comments
This PR is an alternative to https://github.com/magento/magento2/pull/25374. The fix implement in this PR is more involved, but I feel it improves to quality of Magento code and helps avoid similar mistakes in the future.


mattijv:
Apparently there is some latency to the CLA signing being registered? I'm going to try closing and reopening this PR later.

mattijv:
As far as I can tell, all the remaining failing tests are unrelated to any changes in this PR.

krzksz (approved):
The suffix check was introduced in https://github.com/magento/magento2/commit/a303ce89133fddcc9a2cc2b5d50eac5c76f800af and I agree adjustments to comply with it should be done in a separate PR.

Karlasa (approved):

engcom-Alfa:
Hi @mattijv 
Could you provide more details how to reproduce [this](https://github.com/magento/magento2/issues/25373) issue?

Thanks!

mattijv:
@engcom-Alfa Thank you for taking a look at the issue.

Easiest way to reproduce the problem is to try rendering a component into the 'promotion' region of the minicart. Magento does not use this region for anything itself (as it's designed to be used by extensions), but the issue can easily be reproduced by adding the following into `app/code/Magento/Checkout/view/frontend/layout/default.xml` under the `promotion` item.

```xml
<item name="children" xsi:type="array">
    <item name="test" xsi:type="array">
        <item name="component" xsi:type="string">Magento_Ui/js/form/components/html</item>
        <item name="config" xsi:type="array">
            <item name="content" xsi:type="string">Test</item>
        </item>
    </item>
</item>
```

The expected result is that the text 'Test' is visible in the bottom part of the minicart when it is open. The actual result is that the text 'Test' is NOT visible there.

engcom-Alfa:
@mattijv Thank you for the clarification!

engcom-Alfa:
:heavy_check_mark: QA Passed

sdzhepa:
@magento run all tests

mattijv:
As it seems this PR is tentatively approved as the proper fix, I'll close https://github.com/magento/magento2/pull/25374 as redundant.

VladimirZaets:
B2B tests are failed

engcom-Foxtrot:
@magento run all tests

PR #25381, opened 2019-10-30
Change generated/code from PSR-0 to PSR-4

bentideswell:
### Description (*)
This pull request includes one small change in the \Magento\Framework\Code\Generator\Io class to not change DIRECTORY_SEPARATOR to '\'. This changes the generated/code directory to PSR-4 (currently is PSR-0) and brings it inline with the rest of Magento 2.

This is an issue for certain system when module uses an underscore in the module name. This is allowed in PSR-4 but in not in PSR-0.

MyCompany/Module_SubModule

This becomes:

> generated/code/MyCompany/Module/SubModule

I have found some systems can retrieve generated classes from this directory where as some systems cannot. I have not been able to find the different between the systems to figure out why some can and some can't read generated classes.

### DIFF
--- a/lib/internal/Magento/Framework/Code/Generator/Io.php
+++ b/lib/internal/Magento/Framework/Code/Generator/Io.php
@@ -76,7 +76,7 @@ class Io
      */
     public function generateResultFileName($className)
     {
-        return $this->_generationDirectory . ltrim(str_replace(['\\', '_'], '/', $className), '/') . '.php';
+        return $this->_generationDirectory . ltrim(str_replace('\\', '/', $className), '/') . '.php';
     }
 
     /**


hostep:
Out of curiosity, should the [autoloading section in the `composer.json`](https://github.com/magento/magento2/blob/cc67aa09c01a4f336669c53e7ab286ddcc9dad8e/composer.json#L320-L325) file then also needs to be updated?

bentideswell:
Yes, I think it might.

Do you know why composer.json lists app/code as PSR-0 when this directory seems to follow PSR-4?

Why is PSR-0 used at all?

hostep:
Good question, it was introduced in https://github.com/magento/magento2/commit/97ce2b8e63cc1200c7d2ec4f314d606eab736968#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780 (merge commit: https://github.com/magento/magento2/commit/86d5abe5adf8f452e52e5876c1ba2008aa0e6e30#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780)

But no idea why they chose psr-0 over psr-4 ...

bentideswell:
I've looked into this further and the issue may not be PSR-0 related and may be due to incorrect permissions on the server. As it's not my server (it's a client), I cannot test. I will close this pull request for now and re-open if I find out any more.

PR #25414, opened 2019-10-31
update a way the match is applied

luanalves:
### Description (*)
When elasticsearch is enabled and the search has two strings, the "match" attribute is enabled the search is not matching.
In attachament , the patch with some result

1. The counter that activates the elasticsearch "match_phrase" attribute has been set

1. use sample data with elasticsearch 6.5 enabled
2. perform two-string search
3. The return should be just what you entered


rodrigowebjump:
@luanalves Thanks for your PR.
Your changes will remove the exact phrase match when user search using " (quotes).
For now, I will close your PR. Please review your approach to fix the described issue.

Thanks

PR #25479, opened 2019-11-05
JSON fields support

akaplya:
### Description (*)
Starting MySQL 5.7 supports native JSON data type. https://dev.mysql.com/doc/refman/5.7/en/json.html
This pull request enables the possibility to use JSON fields with Magento declarative schema.

##### db_schema.xml
```
<?xml version="1.0"?>
<!--
/**
 */
-->
<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd">
    <table name="test_table" resource="default" engine="innodb" comment="Data">
        <column xsi:type="bigint" name="id" padding="20" unsigned="true" nullable="false" identity="true" comment="ID"/>
        <column xsi:type="json" name="data" nullable="false" comment="data"/>
            <column name="id"/>
        </constraint>
    </table>
</schema>
```

##### Structure of created table 
```
mysql> show create table test_table\G
*************************** 1. row ***************************
       Table: test_table
Create Table: CREATE TABLE `test_table` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT 'ID',
  `data` json NOT NULL COMMENT 'data',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COMMENT='Data'
1 row in set (0.00 sec)
```

### DIFF
--- a/app/etc/di.xml
+++ b/app/etc/di.xml
@@ -1453,6 +1453,7 @@
                 <item name="primary" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\Primary</item>
                 <item name="foreign" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\Foreign</item>
                 <item name="index" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\Index</item>
+                <item name="json" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Dto\Factories\Json</item>
             </argument>
         </arguments>
     </type>
@@ -1480,6 +1481,7 @@
                 <item name="varchar" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Columns\StringBinary</item>
                 <item name="binary" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Columns\StringBinary</item>
                 <item name="varbinary" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Columns\StringBinary</item>
+                <item name="json" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Columns\Json</item>
                 <item name="index" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Index</item>
                 <item name="unique" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Constraints\Internal</item>
                 <item name="primary" xsi:type="object">\Magento\Framework\Setup\Declaration\Schema\Db\MySQL\Definition\Constraints\Internal</item>
@@ -1593,6 +1595,7 @@
                 <item name="datetime" xsi:type="object">Magento\Framework\Setup\SchemaListenerDefinition\TimestampDefinition</item>
                 <item name="date" xsi:type="object">Magento\Framework\Setup\SchemaListenerDefinition\DateDefinition</item>
                 <item name="boolean" xsi:type="object">Magento\Framework\Setup\SchemaListenerDefinition\BooleanDefinition</item>
+                <item name="json" xsi:type="object">Magento\Framework\Setup\SchemaListenerDefinition\JsonDefinition</item>
             </argument>
         </arguments>
     </type>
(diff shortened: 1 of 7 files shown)


skovalenk:
Looks good!

skovalenk:
Looks good!

akaplya:
SVC failed due to XSD change, but this can not affect existing extensions except the case when someone already declared JSON type locally, which is very unlikely.

ihor-sviziev (changes requested):
Hi @akaplya,

In general your changes looks good to me.

According to mysql [docs](https://dev.mysql.com/doc/refman/5.7/en/json.html):
> As of **MySQL 5.7.8**, MySQL supports a native JSON data type

But in the [magento requirements](https://devdocs.magento.com/guides/v2.3/install-gde/prereq/mysql.html) there is MySQL 5.7.x (it means version 5.7.0-5.7.7 also supported).
> The Magento application requires MySQL 5.6.x. Magento versions 2.1.2 and later are compatible with **MySQL 5.7.x**. Magento is also compatible with MySQL NDB Cluster 7.4.x, MariaDB 10.0, 10.1, 10.2, Percona 5.7 and other binary compatible MySQL technologies.

Introducing this new functionality might brake some functionality for people who using mysql 5.7.0-5.7.7, I mean you're just adding support, but someone will start using it in the core and finally brake it.

I think together with this PR we need to define that requirements are MySQL 5.7.x, but not lower than 5.7.8.

What do you think?

akaplya:
@ihor-sviziev  This PR has introduced a possibility to use the JSON field but did not add any real usages. Actually. This feature more for the community than for core. As an extension developer, you have significantly more freedom and control over the environment. So the developer can make a deciding on usage/not usage JSON fields.
Eventually, this feature will be used inside of core after Magento bumps the version of supported MySQL.

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-6276](https://jira.corp.magento.com/browse/ENGCOM-6276) has been created to process this Pull Request
:eight_spoked_asterisk: @ihor-sviziev, could you please add one of the following labels to the Pull Request?

| Label | Description |

akaplya:
Unfortunately, this feature cannot be covered with integration tests :( because Magento CI/CD, which is still using MySQL 5.6 under the hood. Because Magento still officially supports 5.6.

engcom-Alfa:
:heavy_check_mark: QA Passed 

**Before:**
before

**After:**
after

jonashrem:
This will also break compatibility with MariaDB.

sshymko:
@jonashrem,
Does MariaDB not have the same JSON type as MySQL?

jonashrem:
@sshymko see https://mariadb.com/kb/en/library/incompatibilities-and-feature-differences-between-mariadb-103-and-mysql-57/ there are some differences. 

Depending on how JSON is used, it might still work through.

>
> MariaDB stores JSON as true text, not in binary format as MySQL. MariaDB's JSON functions are much faster than MySQL's so there is no need to store in binary format, which would add complexity when manipulating JSON objects.
> For the same reason, MariaDB's JSON data type is an alias for LONGTEXT. If you want to replicate JSON columns from MySQL to MariaDB, you should store JSON objects in MySQL in a TEXT or LONGTEXT column or use statement based replication. If you are using JSON columns and want to upgrade to MariaDB, you need to either convert them to TEXT or use mysqldump to copy these tables to MariaDB.
> In MySQL, JSON is compared according to json values. In MariaDB JSON strings are normal strings and compared as strings. 
> MariaDB 10.3 does not support MySQL's JSON operators (-> and ->>).

akaplya:
This PR does not introduce any JSON fields into the core product. As a result, it is 100% backward compatible. It presents a possibility to add the JSON field in extension in case a developer is confident enough about the environment where the extension will be deployed.
This PR does not say how to operate with JSON fields, it says how to create a table with JSON fields through the declarative schema.
As a result, it is compatible with MariaDB starting 10.2.7
https://mariadb.com/kb/en/library/json-data-type/

slavvka:
Guys, this PR is currently under the architecture review

slavvka:
The PR is approved by architects

slavvka (approved):

magento-engcom-team:
[ENGCOM-6276](https://jira.corp.magento.com/browse/ENGCOM-6276) has been created to process this Pull Request

PR #25496, opened 2019-11-06
Fixed scheduled indexation with MySQL 8

akaplya:
### Description (*)
Magento indexation by schedule does not work correctly with MySQL 8

https://github.com/magento/architecture/pull/301

MySQL 8 has enabled metadata caching by default. As a result, Magento functionality that expects accurate reading frequently changing values ​​fails.
The most devastating consequence of this change has on scheduled indexation, which tries to read the most recent changelog version from the MySQL metadata.
To resolve the issue a materialized view engine has to rely on real changelog table data instead of MySQL metadata.

### DIFF
--- a/lib/internal/Magento/Framework/Mview/View/Changelog.php
+++ b/lib/internal/Magento/Framework/Mview/View/Changelog.php
@@ -7,7 +7,6 @@
 namespace Magento\Framework\Mview\View;
 
 use Magento\Framework\DB\Adapter\ConnectionException;
-use Magento\Framework\Exception\RuntimeException;
 use Magento\Framework\Phrase;
 
 /**
@@ -170,7 +169,6 @@ class Changelog implements ChangelogInterface
      *
      * @return int
      * @throws ChangelogTableNotExistsException
-     * @throws RuntimeException
      */
     public function getVersion()
     {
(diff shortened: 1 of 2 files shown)


lib/internal/Magento/Framework/Mview/View/Changelog.php
```diff
@@ -170,22 +169,19 @@ public function getList($fromVersionId, $toVersionId)
...
-        $row = $this->connection->fetchRow('SHOW TABLE STATUS LIKE ?', [$changelogTableName]);
-        if (isset($row['Auto_increment'])) {
-            return (int)$row['Auto_increment'] - 1;
-        } else {
-            throw new RuntimeException(
-                new Phrase("Table status for %1 is incorrect. Can`t fetch version id.", [$changelogTableName])
-            );
-        }
+        return (int)$this->connection->fetchOne(
+            $this->connection->select()
```
ihor-sviziev, line 180: Hi @akaplya,
Looks like in case if this table will be empty (for instance new installation) - we’ll have exception, and it will be regression.

Could you update your code and cover this case with some test?

ihor-sviziev (changes requested):

lib/internal/Magento/Framework/Mview/View/Changelog.php
akaplya: Hii @ihor-sviziev if a table will be empty query will return null, which will be converted to 0.

ihor-sviziev: Ah, you’re right 👍

ihor-sviziev (approved):

ihor-sviziev (changes requested):
Hi @akaplya,
Could you fix merge conflict?

ihor-sviziev:
Hi @akaplya,
I just reviewed conflict - looks like this issue was already fixed in https://github.com/magento/magento2/commit/31aebbf5b39b0e93910a72b6bcbf9167f6508292.

Could you double check it?

VladimirZaets:
@ihor-sviziev you are right, it's already fixed.

PR #25520, opened 2019-11-07
Return HTTP response of example import file as CSV

tbajorek:
### Description (*)
We are return content of CSV file here. So without these changes unnecessary file is created in "var" directory and default response type is returned. So file is not downloaded, just displayed in browser. This fix will allow to actually download example file and not use some "hacks" in other modules.
Eg. response type of this link: [https://{BASE_URL}/admin_123/admin/import/download/filename/catalog_product/key/{KEY}/](https://{BASE_URL}/admin_123/admin/import/download/filename/catalog_product/key/{KEY}/) is HTML. So when I click "Download Sample File" link eg. for Entity type = Products, CSV file is displayed in web browser instead of being downloaded.

1. Log in into Magento Admin panel
2. Go to System > Data Transfer > Import
3. Select eg. Entity Type = Products
4. Click the link "Download Sample File"
5. There should be started downloading of CSV example file

 - [X] All automated tests passed successfully (all builds are green)

### DIFF
--- a/app/code/Magento/ImportExport/Controller/Adminhtml/Import/Download.php
+++ b/app/code/Magento/ImportExport/Controller/Adminhtml/Import/Download.php
@@ -101,8 +101,8 @@ class Download extends ImportController
         $resultRaw = $this->resultRawFactory->create();
         $resultRaw->setContents($fileContents);
         $resultRaw->setHeader('Content-Type','text/csv');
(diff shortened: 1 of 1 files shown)


lenaorobei:
Hi @tbajorek! I cannot reproduce the issue on latest `2.4-develop` branch. I went through steps you've described and the actual file download starts for me. 
Closing this PR. Feel free to reopen with the reference to the corresponding issue (if it is still reproducible for you). Thanks.

Issue #14568, opened 2018-04-05
normalizeDate() and convertToMomentFormat() in misc.js is the same


likemusic:
I'm not sure, but seems that normalizeDate() and convertToMomentFormat() in misc.js is the same and one of them should be deleted.

Both of them convert Intl date format to moment.js format. normalizeDate() seems more correct.

### Preconditions

1. 
2. 

### Steps to reproduce

1. 
2. 
3. 

### Expected result

1. 

### Actual result

1. [Screenshot, logs]

krzksz:
Closing as per comment: https://github.com/magento/magento2/pull/25544#issuecomment-558353493

PR #25544, opened 2019-11-09
Remove usage of normalizeDate in favor of convertToMomentFormat

krzksz:
### Description (*)

This change removes the usage of `normalizeDate` in favor of `convertToMomentFormat` because it seems to be used more widely. It is also making its implementation more universal based on `normalizeDate` contents.

1. magento/magento2#14568: normalizeDate() and convertToMomentFormat() in misc.js is the same

I don't see any.

### DIFF
--- a/lib/web/mage/utils/misc.js
+++ b/lib/web/mage/utils/misc.js
@@ -35,12 +35,12 @@ define([
     map = {
         'D': 'DDD',
         'dd': 'DD',
-        'd': 'D',
+        'd': 'DD',
         'EEEE': 'dddd',
         'EEE': 'ddd',
         'e': 'd',
         'yyyy': 'YYYY',
-        'yy': 'YY',
+        'yy': 'YYYY',
         'y': 'YYYY',
         'a': 'A'
     };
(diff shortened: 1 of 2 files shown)


VladimirZaets (changes requested):
Hi @krzksz. Please take a look at failed tests

lib/web/mage/utils/misc.js
```diff
@@ -83,22 +83,6 @@ define([
             owner[target] = _.debounce(fn.bind(owner), limit);
         },
 
-        /**
-         * Converts mage date format to a moment.js format.
-         *
-         * @param {String} mageFormat
-         * @returns {String}
-         */
-        normalizeDate: function (mageFormat) {
```
VladimirZaets, line 92: I think we shouldn't remove this method because someone can already use it and we will break customer implementation

VladimirZaets:
Hi @krzksz. Due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html) all code must be covered by tests. Please cover your fix by automated tests.

krzksz:
Actually, after working with tests I realized that both input and output data is different for both methods and it is not possible to unify them without breaking compatibility. I'm closing this PR and linked issue.

PR #25552, opened 2019-11-11
Fix escape less calc

jasperzeinstra:
### Description (*)
We've escaped the content in the calc, this way the calc in the css file will be `calc(100% + 3px)` instead of `calc(103%)`

1. Hover over root category in main navigation and inspect the height of the after element being added.

### DIFF
--- a/lib/web/css/source/lib/_navigation.less
+++ b/lib/web/css/source/lib/_navigation.less
@@ -338,7 +338,7 @@
                     top: 0;
                     left: 100%;
                     width: 10px;
-                    height: calc(100% + 3px);
+                    height: calc(~'100% + 3px');
                     z-index: 1;
                 }
             }


VladimirZaets (changes requested):
Hi @jasperzeinstra. Thanks for collaboration. According documentation ( https://www.w3schools.com/cssref/func_calc.asp ) of culc function current syntax is correct and should work fine

jasperzeinstra:
@VladimirZaets yes, `calc(100% + 3px)` is fine in CSS, but not in LESS. Because LESS will try to calculate `100% + 3px` and that results in `103%`. You need to wrap this in `~'100% + 3px'`, then LESS will see it as a string and not do the calculation. And then the output will be `calc(100% + 3px)` instead of `calc(103%)`

Here's a screenshot of the currently wrong complied CSS.
Screenshot 2019-11-12 at 19 35 32

You can also see that every where calc is being used in LESS file the `~'…'` syntax is being used.

VladimirZaets:
> @VladimirZaets yes, `calc(100% + 3px)` is fine in CSS, but not in LESS. Because LESS will try to calculate `100% + 3px` and that results in `103%`. You need to wrap this in `~'100% + 3px'`, then LESS will see it as a string and not do the calculation. And then the output will be `calc(100% + 3px)` instead of `calc(103%)`
> 
> Here's a screenshot of the currently wrong complied CSS.
> Screenshot 2019-11-12 at 19 35 32
> 
> You can also see that every where calc is being used in LESS file the `~'…'` syntax is being used.

Ok, thanks for the explanation

VladimirZaets (approved):

magento-engcom-team:
[ENGCOM-6273](https://jira.corp.magento.com/browse/ENGCOM-6273) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed 

**Before:**
before

**After:**
after

PR #25587, opened 2019-11-13
Refactor JavaScript mixins module

krzksz:
### Description (*)
This refactor started as an attempt to fix the fact that mixins were not loaded and applied for bundled modules when using advanced bundling technique. During the process I decided to dive a bit deeper and see if it is possible to improve the way Magento extends RequireJS functionality.

#### Bundling bug
First of all, after a lot of investigation I found that the only way for the plugin to retrieve true modules' paths no matter if they are bundled or not, is to prepare a separate context which excludes any bundles configuration.

#### Leaking of `arguments`
In the previous implementation, functions like `applyMixins`, `load` and overwritten `define` were directly passing down special `arguments` object, which is considered a bad practice as it may leads to performance degradation and memory leaks so I refactored it away.

#### Usage of `require.eval`
Overwritten `define` method used `require.exec` which calls unsafe and poor-performing `eval` to be used.

I understand why this approach was used, the problem is that original RequireJS's `define` is able to append additional dependencies by analysing module's source code which happens [here](https://github.com/requirejs/requirejs/blob/master/require.js#L2090) for example for [bootstrap.js](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Ui/view/base/web/js/lib/knockout/bindings/bootstrap.js).

Indeed, pushing the transformed item to a private `globalDefQueue` doesn't leave much hope for intercepting it in any other way but then I had a breakthrough. Instead of processing dependency names when they are added to the queue, we can do the same when they [get transferred](https://github.com/requirejs/requirejs/blob/master/require.js#L566) from `globalDefQueue` to default context's `defQueue` (which we have direct access to) and [shifted from it](https://github.com/requirejs/requirejs/blob/master/require.js#L1252) by wrapping the `shift` method. All of this happens before any of the dependencies are actually loaded so thre is no performance degradation or delay in requests involved. No more eval, another win.

#### Wrapping global `require` and `define`
Solving the above point made it possible for me to actually get rid of overwriting global `define` function. Additionally, by wrapping default context's `require` method, which always gets the parameters in normalized order and shape I was able to remove a lot of code that recreated original logic in order to retain compatibility with all RequireJS features.

1. magento/magento2#25586: Mixins are not applied for advanced bundled modules
2. magento/baler#23: Mixins are not applied for advanced bundled modules

1. Apply this change.
2. Verify that modules are still properly loaded and executed.
3. Verify that mixins are properly loaded and applied no matter if advanced bundling is used or not.

### Questions or comments

To be honest, I hoped that it would be possible to completely remove the need for plugin itself somehow while retaining it's functionality but I haven't figure out the way yet. That said, I'm very happy with the results of a few weeks worth of evenings and digging into RequireJS internals and the most important goal was achieved.

### DIFF
--- a/lib/web/mage/requirejs/mixins.js
+++ b/lib/web/mage/requirejs/mixins.js
@@ -71,7 +71,7 @@ define('mixins', [
      * from a module name ignoring the fact that it may be bundled.
      *
      * @param {String} name - Name, path or alias of a module.
-     * @param {Object} config - Context's configuartion.
+     * @param {Object} config - Context's configuration.
      * @returns {String}
      */
     function getPath(name, config) {
--- a/lib/web/mage/requirejs/mixins.js
+++ b/lib/web/mage/requirejs/mixins.js
@@ -7,7 +7,25 @@ define('mixins', [
 ], function (module) {
     'use strict';
 
-    var rjsMixins;
+    var contexts = require.s.contexts,
+        defContextName = '_',
+        defContext = contexts[defContextName],
+        unbundledContext = require.s.newContext('$'),
+        defaultConfig = defContext.config,
+        unbundledConfig = {
+            baseUrl: defaultConfig.baseUrl,
+            paths: defaultConfig.paths,
+            shim: defaultConfig.shim,
+            config: defaultConfig.config,
+            map: defaultConfig.map
+        },
+        rjsMixins;
+
+    /**
+     * Prepare a separate context where modules are not assigned to bundles
+     * so we are able to get their true path and corresponding mixins.
+     */
+    unbundledContext.configure(unbundledConfig);
 
     /**
      * Checks if specified string contains
@@ -50,14 +68,14 @@ define('mixins', [
 
     /**
      * Extracts url (without baseUrl prefix)
-     * from a modules' name.
+     * from a module name ignoring the fact that it may be bundled.
      *
      * @param {String} name - Name, path or alias of a module.
-     * @param {Object} config - Contexts' configuartion.
+     * @param {Object} config - Context's configuartion.
      * @returns {String}
      */
     function getPath(name, config) {
-        var url = require.toUrl(name);
+        var url = unbundledContext.require.toUrl(name);
 
         return removeBaseUrl(url, config);
     }
@@ -73,11 +91,11 @@ define('mixins', [
     }
 
     /**
-     * Iterativly calls mixins passing to them
+     * Iteratively calls mixins passing to them
      * current value of a 'target' parameter.
      *
      * @param {*} target - Value to be modified.
-     * @param {...Function} mixins
+     * @param {...Function} mixins - List of mixins to apply.
      * @returns {*} Modified 'target' value.
      */
     function applyMixins(target) {
@@ -94,8 +112,13 @@ define('mixins', [
 
         /**
          * Loads specified module along with its' mixins.
+         * This method is called for each module defined with "mixins!" prefix
+         * in its name that was added by processNames method.
          *
          * @param {String} name - Module to be loaded.
+         * @param {Function} req - Local "require" function to use to load other modules.
+         * @param {Function} onLoad - A function to call with the value for name.
+         * @param {Object} config - RequireJS configuration object.
          */
         load: function (name, req, onLoad, config) {
             var path     = getPath(name, config),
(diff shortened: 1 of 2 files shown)


DrewML:
Exciting stuff! And getting rid of `eval` there gets us one step closer to eventually being able to implement a content security policy effectively.

Traveling at the moment so sending this over to our performance team to give a first round of reviews.

Thanks for the hard work on this!

VladimirZaets:
@magento run all tests

lib/web/mage/requirejs/mixins.js
```diff
@@ -73,51 +91,59 @@ define('mixins', [
...
+                deps = [name].concat(this.getMixins(path));
 
             req(deps, function () {
-                onLoad(applyMixins.apply(null, arguments));
+                var target = arguments[0],
+                    mixins = [],
+                    index = 1;
+
+                for (; index < arguments.length; index++) {
+                    mixins[index] = arguments[index];
```
omiroshnichenko: Array.prototype.slice.call(arguments, 1) could be simpler

DrewML: OP mentions this:

image

For a period of time this was generally a best-practice, as referenced in a [few](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments) [places](https://github.com/davidmarkclements/v8-perf#leaking-and-arrayifying-arguments).

I have no idea if engines have improved optimizations around this in recent years. But I'd bet this isn't something that's going to make or break our performance, so I have no strong preference :)

krzksz: I've just checked it and [it seems](https://jsperf.com/arguments-passing-techniques/1) like passing arguments directly through `apply` is the fastest method of all. I'm going to refactor it.

krzksz: Fixed (reverted).

krzksz:
We are all green 🥳 However, now that I figured out how to write and run frontend unit tests locally, maybe it would be worth it to add some for mixins module?

DrewML:
> now that I figured out how to write and run frontend unit tests locally, maybe it would be worth it to add some for mixins module?

Some smallunit tests would definitely be nice, but I'd hope we already have some e2e tests covering us so I wouldn't say it's critical.

krzksz:
All of my work here is done I think, please post your feedback if any additional changes are needed.

krzksz:
@DrewML I actually added some unit tests, so we are on the safe side 🤗

DrewML (approved):
Fantastic job! Really appreciate the work you put in here.

VladimirZaets (approved):

krzksz:
@magento-engcom-team hey, any chance of moving this ticket forward?

DrewML:
cc @VladimirZaets

engcom-Alfa:
:heavy_check_mark: QA Passed

omiroshnichenko (approved):

omiroshnichenko:
@magento run Unit Tests

ashar01:
Hi, is this fix applied in magento 2.3.4 version? in magento 2.3.4 release notes i cant find it.

adifucan:
Hi @krzksz, will you backport this fix to 2.3-develop? Or I can do it if you do not mind.

krzksz:
I have no idea how to even do it so you have the green light 😁

wt., 31 mar 2020, 18:31 użytkownik Dmytro Poperechnyy <
notifications@github.com> napisał:

adifucan:
@krzksz Sorry, backport means just creating a PR with your fix to 2.3-develop:)
You can do it on your own. Or I'll create a PR to 2.3-develop if you do not mind.

krzksz:
I know, the issue is there is no 2.3-develop anymore. I'll ask other maintainers how to proceed here.

krzksz:
I contacted Magento team and they are already working on a backport to 2.3 🤗

adifucan:
Yeah, I talked to them. I cherry picked your commits to 2.3-develop. I tested it with advanced bundling and with baler js bundling either. 
Modules and loaded and mixins are loaded as well.

gwharton:
Running this PR against 2.3.4 I get javascript errors on page load. Anyone else seen these?

image

krzksz:
@gwharton Can you add a conditional breakpoint for `!names` and see which module is affected? We are using this as a patch in multiple Magento 2.3.x versions with no issues so far.

gwharton:
Yes, turns out to be this module. https://github.com/swissup/module-stickyfill

It is when this file is processed. view/frontend/web/js/stickyfill.js

krzksz:
I verified that it is indeed an issue with modules defined in CommonJS format. I'll prepare PR that addresses this regression shortly.

krzksz:
Hey @gwharton sorry for confusion but I have debugged the issue thoroughly and it is indeed the module's issue. The problem is that it defines itself using 
```javascript
define(Stickyfill);
```
and `Stickyfill` is an object, while according to RequireJS's documentation it should be a function. I'm going to prepare an improvement to handle this scenario anyway, but I don't think it is a blocker anymore.

micwallace:
Hey guys!

Is there any reason why this patch hasn't been included in the 2.3.4 or 2.3.5-p1 releases?

PR #25651, opened 2019-11-19
Add null return hint to RouterInterface docblock

### LINKED ISSUE #29645: [Issue] Add null return hint to RouterInterface docblock
This issue is automatically created based on existing pull request: magento/magento2#25651: Add null return hint to RouterInterface docblock

---------
### Description (*)
In the wild, returns null in almost all implementations:
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/lib/internal/Magento/Framework/App/Router/Base.php#L274
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Cms/Controller/Router.php#L108
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Robots/Controller/Router.php#L60
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/UrlRewrite/Controller/Router.php#L89

The exception seems to be
https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Webapi/Controller/Rest/Router/Route.php#L107 which returns array|false (!?) and must be fixed,
 as well as https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Webapi/Controller/Rest/Router.php#L40 which returns RouterInterface

This fix should resolve issues detected by static analysis

marcusirgens:
### Description (*)
In the wild, returns null in almost all implementations:
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/lib/internal/Magento/Framework/App/Router/Base.php#L274
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Cms/Controller/Router.php#L108
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Robots/Controller/Router.php#L60
- https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/UrlRewrite/Controller/Router.php#L89

The exception seems to be
https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Webapi/Controller/Rest/Router/Route.php#L107 which returns array|false (!?) and must be fixed,
 as well as https://github.com/magento/magento2/blob/c0ab1d9c7a3ee84b1c880ee3fb2d16793a7edb9d/app/code/Magento/Webapi/Controller/Rest/Router.php#L40 which returns RouterInterface

This fix should resolve issues detected by static analysis

### Resolved issues:
1. [x] resolves magento/magento2#29645: Add null return hint to RouterInterface docblock

### DIFF
--- a/lib/internal/Magento/Framework/App/RouterInterface.php
+++ b/lib/internal/Magento/Framework/App/RouterInterface.php
@@ -8,16 +8,17 @@
 namespace Magento\Framework\App;
 
 /**
- * Interface \Magento\Framework\App\RouterInterface
- *
+ * Main application route dispatcher
  */
 interface RouterInterface
 {
     /**
      * Match application action by request
      *
+     * Must either return an action capable of handling the request, or null
+     *
      * @param RequestInterface $request
-     * @return ActionInterface
+     * @return ActionInterface|null
      */
     public function match(RequestInterface $request);
 }


ihor-sviziev (changes requested):
Hi @marcusirgens,
Good finding! Could you fix all places where not null was returned?

ihor-sviziev:
@magento run all tests

engcom-Echo:
@magento run all tests

ihor-sviziev:
Moved to status “to approve” in order to get approval for SVC failure

sidolov:
@magento create issue

sidolov (changes requested):
@marcusirgens , @ihor-sviziev I would prefer to throw an exception in case of route does not match. Right now we can see the different implementation with the return type of `null|false` and all client code should handle them to work correctly. With exception approach, it will be clear how to use it - get an object or exception in case if something went wrong.

ihor-sviziev:
@sidolov I don't agree with such approach. 

In case if you don't have any matched route to the current request - is it exceptional situation? No. something went wrong? Also no. So why we would throw an exception in such case?

In my opinion in such case you have to just return "no results" value, that is "NULL" in programming language.

In addition to what I described above adding throwing an exception should be treated as backward incompatible changes, this is one of the basic interfaces that are working with routing and in case if we'll introduce here any issues - it definitely will be catastrophe. I believe this is a mistake that it's not marked with `@api` tag.

marcusirgens:
@ihor-sviziev 

> In case if you don't have any matched route to the current request - is it exceptional situation? No. something went wrong? Also no. So why we would throw an exception in such case?

Actually, I disagree: I think  `match` implies that this class can handle the request. I expect a result that I can use from this method, not "null". If we are uncertain if the class can handle the request, I think it makes sense to add a method like `canMatch` or something.

However, this would require a major rewrite of almost all classes implementing `RouterInterface`. I think it might be better to deprecate this routing system entirely.

@sidolov, what do you think? Go for the strongly typed version? Require the return type to be `ActionInterface`? Add a type hint? Keep null? If we decide to make a new RouterInterface, we need a lot more code ;)

sidolov:
@ihor-sviziev it's the same as `ProductRepository`, following your logic it should return product object or null in case if the product was not found. If method cannot handle the provided request it is an exceptional case because we cannot correctly work with the response. 

@marcusirgens totally agree with you that such changes are not backward compatible and most likely may be delivered to the next minor release, 2.5 for example. So, if you are interested to do that right, we still have time until the next release :)

marcusirgens:
@sidolov I've opened a discussion in https://github.com/magento/community-features/issues/292 which relates to this issue.

ihor-sviziev:
@marcusirgens @sidolov ok, I don't mind of refactoring this logic at all, but as for me this PR was targeting to fix incorrect return type on **existing implementation**. And as part of it it doesn't make sense to introduce any exceptions, as they weren't there.

gabrieldagama:
The risk was set to`low` due to: the suggested changes shouldn't impact other areas.

ihor-sviziev:
@marcusirgens @sidolov 
As I understood - this PR should be closed and new implementation should be created instead PSR-7 / PSR-15 for HTTP requests?

marcusirgens:
@ihor-sviziev 
> As I understood - this PR should be closed and new implementation should be created instead PSR-7 / PSR-15 for HTTP requests?

That's a major refactor and a big BC break, though so it at least deserves some discussion.

ihor-sviziev:
@marcusirgens sure, I don't mind about that discussion, but not clear should we keep this PR till the discussion will be finished or should we just close it?

marcusirgens:
@ihor-sviziev 
> @marcusirgens sure, I don't mind about that discussion, but not clear should we keep this PR till the discussion will be finished or should we just close it?

Well, the current docblock is still wrong.

ihor-sviziev:
Ok, so moving back to "to approve" status

engcom-Dash:
Hello @marcusirgens,

Thanks for your contributions!

Please resolve the conflict so we can proceed with this PR.

Thanks!

Issue #25268, opened 2019-10-24
$order->getCustomer() returns NULL for registered customer


jakwinkler:
### Preconditions (*)

1. Magento 2.3.2 & 2.4-develop, community edition

### Steps to reproduce (*)

1. Create customer account and make an order
2. Create a module with an observer to `sales_order_save_after` event
```
<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Event/etc/events.xsd">
    <event name="sales_order_save_after">
        <observer name="my-observer" instance="Something\Darkside\Observer\OrderSaveAfter" />
    </event>
</config>
```
3. Try to access Customer object from order object

```
public function execute(Observer $observer): void
    {
         /** @var Order $order */
         $order = $observer->getEvent()->getOrder();
         $customer = $order->getCustomer();
        ....
```

### Expected result (*)

1. Method `getCustomer` should return customer object, as defined in phpdoc block **OR** phpdoc block should be changed to highlight that `NULL` could be returned

### Actual result (*)

1. Method `getCustomer` always returns `NULL`, but it's not highlighted in phpdoc block that it could return null

### Additional info
Method `getCustomer` will return customer only in case if someone added it to Order object, in other cases it doesn't exists. In case if you need to get customer from the Order object you need to do following get customer ID from the order object
```php
$customerId = $order->getCustomerId();

// $customerId could be NULL in case if order was placed as guest
$customer = $customerId ? $this->customerRepository->getById($customerId) : null;
```
Where `$this->customerRepository` is instance of `\Magento\Customer\Api\CustomerRepositoryInterface` that retrieved via constructor

jakwinkler:
Any update on this?

magento-engcom-team:
:white_check_mark: Confirmed by @itsPranith 
 Thank you for verifying the issue. Based on the provided information internal tickets `MC-22188` were created

**Issue Available**: @itsPranith, _You will be automatically unassigned. Contributors/Maintainers can claim this issue to continue. To reclaim and continue work, reassign the ticket to yourself._

mohith-ecommerce:
@qsolutions-pl in this event we are not getting customer object so instead of $order->getCustomer() use like $order->getCustomerFirstname() for customer first name

jakwinkler:
honestly @mohith227 that is not an answer nor a solution, I need customer object to fetch other customer data like:
- customer custom attributes
- customer group

[...]

PR #25728, opened 2019-11-25
Issue 25268

rajeebkumar-sahoo:
resolved issue#25268: order get customer returns null from observer
### Description (*)

This merge request resolves issue#25268 (order get customer returns null from observer)

1. magento/magento2#25268: $order->getCustomer() returns NULL for registered customer

Pre-requisite: 
    i. setup:di:compile after getting the new code
    ii. clean the configuration cache after adding event observer

1. Create one observer for event "sales_order_save_after".
2. Place an order with a registered customer account.
3. Try to access customer object from order object within the observer created above.
4. You will be able to get customer data instead of null after this fix.


kanthy-i95dev (approved):

ihor-sviziev (changes requested):
Unfortunately adding new method to existing functionality will break backward compatibility, more details: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/

Why do you think order model should have method "getCustomer"?

As for me it shoudn't as it's not his responsibility. It should just contain getCustomerId method or something like this, and it is:
https://github.com/magento/magento2/blob/bff0d385cfc6138b5a75423498bf9e9a88061d79/app/code/Magento/Sales/Model/Order.php#L2744-L2747

If you need to have customer - you can get it from customer repository.

I would mark it as not an issue

ihor-sviziev (changes requested):
Hi @rajeebkumar-sahoo,
1. First of all - there is a lot not related changes. Could you force push only needed changes?
2. I don't see any difference in order model, looks like you missed to push them

VladimirZaets:
@rajeebkumar-sahoo , I am closing this PR now due to inactivity.
Thank you for collaboration

PR #25746, opened 2019-11-26
Fix #9581 Add primary key on table catalog_url_rewrite_product_category

### LINKED ISSUE #9581: catalog_url_rewrite_product_category table missing PK
The catalog_url_rewrite_product_category table is missing a PK.  Because of this, it cannot be replicated reliably in Galera based systems (eg. MariaDB, Percona XtradbCluster).  It appears to be the only table without a PK.  This is more of a problem with PXC 5.7 as it recommends enforcing strict mode which will actively break usage of this table.

Magento 2.1.5
Mysql: Percona XtraDBCluster 5.7

### Preconditions
Magento 2.4-develop

### Steps to reproduce
1. Open database table `catalog_url_rewrite_product_category`
2. Check if the table has a Primary Key

### Expected result
The database table `catalog_url_rewrite_product_category` has Primary Key

### Actual result
The database table `catalog_url_rewrite_product_category` doesn't have Primary Key
primary_key

paveq:
### Description (*)

This PR adds primary key on **catalog_url_rewrite_product_category** table. Missing primary key causes issues with Percona Xtradb cluster 5.7: https://www.percona.com/doc/percona-xtradb-cluster/LATEST/features/pxc-strict-mode.html#tables-without-primary-keys

1. magento/magento2#9581: catalog_url_rewrite_product_category table missing PK

1. Run setup:upgrade on existing installation
2. Observe that primary key was added to catalog_url_rewrite_product_category table
3. Optional: if running percona cluster with "wsrep_certify_nonPK" off, setup:upgrade should not throw error

### Questions or comments
While it is possible to run Percona Cluster without primary keys with "wsrep_certify_nonPK" feature, having primary keys to every table is recommended approach. Based on my testing, catalog_url_rewrite is only such table that is missing the primary key.


paveq:
There is potential to have existing duplicate values in **catalog_url_rewrite_product_category**, but this situation seems unintentional / caused by bug. Do you think this PR should address the duplicates, and what to do in case of duplicates?

In existence of duplicates, setup:upgrade fails with SQLSTATE[23000]: Integrity constraint violation

I did investigation on Vaimo's instances, and found out that few of them have duplicates in this table, but it is somewhat rare occurrence. So far I've found out that most of the duplicates have exactly same values for all three columns, which certainly makes the data redundant. Situation might have been caused by some indexing flaw in earlier Magento 2 versions. Conclusion is that we could drop duplicates, and only leave one record to the table.

There is very rare case (only 1 or 2 such instances) where there are duplicate data with different values in the table, still investigating if this is intentional or not. It would be great to have feedback from Core team if this is considered intentional design or not.

ihor-sviziev (changes requested):
@paveq it seems to me good solution, but it's not backward compatible.
As you said it could create issues when we have duplicates in this table, so I believe we needed to add some migration that will remove duplicates.

Not sure if that possible with declarative schema, as data upgrades usually executes after schema changes.

ihor-sviziev:
Hi @paveq,

ihor-sviziev:
@paveq I am closing this PR now due to inactivity.

PR #25774, opened 2019-11-27
[Config] Giving the possibility to have a config dependency based on empty config value

eduard13:
### Description (*)
This PR provides a way to set a dependency to a field based on the empty value.

For example:
You have 2 input config fields, and you want to hide the 2nd one if the 1st one won't be empty.

Our `system.xml` will look like:

```xml
<field id="first_input" translate="label" type="text" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1">
    <label>First Input</label>
</field>
<field id="second_input" translate="label" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
    <label>Second input</label>
    <comment><![CDATA[Dependent by 1st field]]></comment>
    <depends>
        <field id="*/*/first_input"/>
    </depends>
</field>
```

N/A

1. Add to any system config group the provided sample code 
2. Open it in the admin panel
3. Type something in the 1st input
4. The second should be hidden

### DIFF
--- a/app/code/Magento/Config/Model/Config/Structure/Element/Dependency/Field.php
+++ b/app/code/Magento/Config/Model/Config/Structure/Element/Dependency/Field.php
@@ -6,6 +6,8 @@
 namespace Magento\Config\Model\Config\Structure\Element\Dependency;
 
 /**
+ * Field
+ *
  * @api
  * @since 100.0.2
  */
--- a/app/code/Magento/Config/Model/Config/Structure/Element/Dependency/Field.php
+++ b/app/code/Magento/Config/Model/Config/Structure/Element/Dependency/Field.php
@@ -41,7 +41,7 @@ class Field
         if (isset($fieldData['separator'])) {
             $this->_values = explode($fieldData['separator'], $fieldData['value']);
         } else {
-            $this->_values = [$fieldData['value']];
+            $this->_values = [isset($fieldData['value']) ? $fieldData['value'] : ''];
         }
         $fieldId = $fieldPrefix . (isset(
             $fieldData['dependPath']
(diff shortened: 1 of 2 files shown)


eduard13:
Based on the Backward Compatibility Policy, we shouldn't change classes marked as @api, but what do you @paliarush think about having a such feature?
Thank you.

dmytro-ch (changes requested):
Hi @eduard13,
great job, thank you!

> Based on the Backward Compatibility Policy, we shouldn't change classes marked as @api

No, it's not relevant unless we change the signature of public/protected methods or introduce new properties or methods, etc.

The only thing, I guess, the behavior you mentioned in description should apear in case the `negative="1"` attribute is set.
e.g. with the following configuration:
```
<depends>
    <field id="*/*/first_input" negative="1"/>
</depends>
```
the second field should be shown if the first one is empty, and vice versa.
Feel free to contact me for further discussion if necessary, or correct me if I'm wrong.

Could you please also implement the recommendations from the static test report?

Thank you!

eduard13:
Hi @dmytro-ch, thank you for your review. Actually not sure what `negative` does, but we can't have an empty value for our dependency. We'll get `Notice: Undefined index: value in /home/dev/sites/mage23/app/code/Magento/Config/Model/Config/Structure/Element/Dependency/Field.php on line 44`.
Basically the provided change fixes this warning and gives us the possibility to set a dependency on empty value.
Please let me know your thoughts on this matter.

Thank you.

dmytro-ch:
After taking a closer look at the issue and having an additional discussion, we decided to proceed with the current behaviour to keep the logic of `\Magento\Config\Model\Config\Structure\Element\Dependency\Field` as straightforward as possible.

dmytro-ch (approved):

magento-engcom-team:
[ENGCOM-6390](https://jira.corp.magento.com/browse/ENGCOM-6390) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed 

**Before:**
`[2019-12-04 10:05:40] main.CRITICAL: Notice: Undefined index: value in /home/arthur/sites/magento/magento23/app/code/Magento/Config/Model/Config/Structure/Element/Dependency/Field.php on line 44 [] []`

**After:**
after

Issue #25738, opened 2019-11-25
DOMDocument::loadHTML(): Tag date invalid in Entity


nghelo:
### Preconditions (*)

1. Magento 2.3.3 or 2.3-develop
2. Simple Product is created.
3. Customer is created and logged

### Steps to reproduce (*)

1. Add product to cart
2. Place Order
3. Click on the order number
![order](https://user-images.githubusercontent.com/51681435/69724841-6689f200-1125-11ea-8db0-b197d83b8cc0.png)
4. The url is (/sales/order/view/order_id/<order_id>/)
5. See exception.log
### Expected result (*)
No new exception log entries

### Actual result (*)
New Exception Log Entry
![DOM](https://user-images.githubusercontent.com/51681435/69725116-11021500-1126-11ea-9ff2-14e0a5bb176c.png)

### Possible fix
in /module-sales/view/frontend/templates/order/order_date.phtml change "date" tag with "span"

magento-engcom-team:
:white_check_mark: Confirmed by @engcom-Echo 
 Thank you for verifying the issue. Based on the provided information internal tickets `MC-29223` were created

**Issue Available**: @engcom-Echo, _You will be automatically unassigned. Contributors/Maintainers can claim this issue to continue. To reclaim and continue work, reassign the ticket to yourself._

TimQSO:
> in /module-sales/view/frontend/templates/order/order_date.phtml change "date" tag with "span"

I can confirm this solves the exception error.

monotheist:
@GuiltyNL You mean like below. Replace "date" with "/span"

```
<div class="order-date">
    <?= $block->escapeHtml(__('<span class="label">Order Date:</span> %1', '</span>' . $block->formatDate($block->getOrder()->getCreatedAt(), \IntlDateFormatter::LONG) . '</date>'), ['span', 'date']) ?>
</div>

```

TimQSO:
No like this:

```
<?php
/**
 * Copyright © Magento, Inc. All rights reserved.
 * See COPYING.txt for license details.
 */
?>
<div class="order-date">
    <?= $block->escapeHtml(__('<span class="label">Order Date:</span> %1', '<span>' . $block->formatDate($block->getOrder()->getCreatedAt(), \IntlDateFormatter::LONG) . '</span>'), ['span', 'date']) ?>
</div>
```

AndyJAllen:
[...]

PR #25777, opened 2019-11-27
Change tag name

AndreyChorniy:
### Description (*)
This PR fixes the unnecessary issue in exeption.log

1. magento/magento2#25738: DOMDocument::loadHTML(): Tag date invalid in Entity

1. Create a new customer or Login with an existing (user should be logged in)
2. Place order
3. Go to Order view page  (you can delete exeption.log before)
4. Check exeption.log (It should not contain `DOMDocument::loadHTML(): Tag date invalid in Entity`)

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/templates/product/listing.phtml
+++ b/app/code/Magento/Catalog/view/frontend/templates/product/listing.phtml
@@ -16,7 +16,6 @@
  */
 ?>
 <?php
-$start = microtime(true);
 $_productCollection = $block->getLoadedProductCollection();
 $_helper = $this->helper(Magento\Catalog\Helper\Output::class);
 ?>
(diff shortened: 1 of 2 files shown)


rogyar:
With the current implementation, `DOMDocument` will parse a string as HTML4. So yes, the `<date>` tag is only supported starting from HTML5. I believe this solution makes sense as a hotfix until `libxml_use_internal_errors` is implemented globally in the escaper class.

app/code/Magento/Catalog/view/frontend/templates/product/listing.phtml
```diff
@@ -16,7 +16,6 @@
  */
 ?>
 <?php
-$start = microtime(true);
```
rogyar, line 19: Please, remove this change from the current commit. I believe you have already introduced the same change in another PR.

Thank you.

rogyar (changes requested):

AndreyChorniy:
Closed because the new one was created here https://github.com/magento/magento2/pull/25895

PR #25840, opened 2019-11-29
Fix catalog page add configurable product to wishlist

mfickers:
### Description (*)

When changing a configurable product's options on the catalog page, the wishlist item option iterates over every option for every attribute and every product on the page, repeatedly overwriting the saved value and only remembering the last value on the page.
This fix changes the wishlist option update to only evaluate the single option that triggered the event.
The PR also includes some refactoring of the _updateWishlistData method.

1. magento/magento2#25091: Product option not reflated at my wish list when added from listing page

1. Go to a catalog page with several configurable products sharing the same attribute option. For the Magento sample data, this would be for example the women's jackets category.
2. Set the options of any product but the last and add this product to the wishlist.
3. Go to wishlist and click on the just added product
4. The options set on the catalog page should be set on the product detail page

### Questions or comments

Are there situations, where the removed iterations over every option was necessary, that could be negatively affected by this change?

### DIFF
--- a/app/code/Magento/Wishlist/view/frontend/web/js/add-to-wishlist.js
+++ b/app/code/Magento/Wishlist/view/frontend/web/js/add-to-wishlist.js
@@ -60,40 +60,28 @@ define([
          */
         _updateWishlistData: function (event) {
             var dataToAdd = {},
-                isFileUploaded = false,
-                self = this;
+                self = this,
+                element = $(event.target);
 
-            if (event.handleObj.selector == this.options.qtyInfo) { //eslint-disable-line eqeqeq
-                this._updateAddToWishlistButton({});
-                event.stopPropagation();
-
-                return;
-            }
-            $(event.handleObj.selector).each(function (index, element) {
-                if ($(element).is('input[type=text]') ||
-                    $(element).is('input[type=email]') ||
-                    $(element).is('input[type=number]') ||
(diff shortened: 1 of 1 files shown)


krzksz (changes requested):
Hey @mfickers, thanks for the contribution! Please check the functional test that is failing.

app/code/Magento/Wishlist/view/frontend/web/js/add-to-wishlist.js
```diff
@@ -60,40 +60,28 @@ define([
...
+                element.is('input[type=email]') ||
+                element.is('input[type=number]') ||
+                element.is('input[type=hidden]') ||
+                element.is('input[type=checkbox]:checked') ||
+                element.is('input[type=radio]:checked') ||
+                element.is('textarea') ||
+                $('#' + element.id + ' option:selected').length
+            ) {
+                if (element.data('selector') || element.attr('name')) {
+                    dataToAdd = $.extend(dataToAdd, self._getElementData(element));
```
krzksz, line 81: If this `.each` is really unnecessary here then I believe `self` can also be removed in favour of using `this` directly.

krzksz (changes requested):

VladimirZaets:
@mfickers, I am closing this PR now due to inactivity.
Thank you for collaboration

PR #25847, opened 2019-11-29
Remove jQuery Migrate

### LINKED ISSUE #21406: remove jQuery Migrate
### Preconditions (*)
jQuery Migrate should never be used in production as this is just meant to migrate from an older version to the used one (for which the Migrate version is).

https://github.com/magento/magento2/blob/2.3-develop/lib/web/jquery/jquery-migrate.js

See https://github.com/jquery/jquery-migrate/tree/1.x-stable#readme

The same applies to underscore, moment.js and other unnecessary libraries and scripts.

This is a big file and just for debugging purposes.

### Expected result (*)

1. jQuery Migrate should not be loaded
2.

### Actual result (*)

1. jQuery Migrate is loaded
2.

theCapypara:
### Description (*)
Removes jQuery Migrate

1. Fixes magento/magento2#21406 : remove jQuery Migrate

None


theCapypara:
Functional tests are failing.
I don't know for sure if this is, because jquery-migrate is actually required somehow, but it seems only tax rule related tests are failing. Can someone confirm that this is because of the jquery-migrate removal?

krzksz:
Hey @Parakoopa, thanks for the contribution! Could you please follow the scenario of one of the failing tests and check what's the problem yourself?

Additionally, I'm not sure if we are allowed to remove jQuery Migrate at all. There can be 3rd party modules and extensions that would break without it and I'm not sure if we can introduce such change before Magento 2.4. I'll try to discuss this with other maintainers.

krzksz (changes requested):
Please check the tests, internal Magento modules shouldn't rely on jQuery Migrate.

ihor-sviziev:
@krzksz I believe we can remove it as part of Magento 2.4 release

torhoehn (approved):

magento-engcom-team:
[ENGCOM-6386](https://jira.corp.magento.com/browse/ENGCOM-6386) has been created to process this Pull Request

| Label | Description |

krzksz:
This ticket has to wait for `2.4-develop` branch to be introduced so we can change the target of the pull request.

engcom-Bravo:
Please notice, some may say this PR does not fully fix the issue it is intended to fix, i.e. #21406 as there is such a line in it "The same applies to underscore, moment.js and other unnecessary libraries and scripts." But considering the issue's title "Remove jQuery Migrate" and its Expected Result "jQuery Migrate should not be loaded" and the according Actual Result I decided it would be more properly to deem the PR fixes the issue than to deem the opposite.  Everything else with it is OK.

engcom-Bravo:
Sorry, haven't noticed the PR was on hold. So I better put it back there)

ihor-sviziev:
Related to https://github.com/magento/magento2/pull/15531, probably or jquery upgrade will be approved - we should keep updated version of jquery migrate for some time, or document how to fix the issues

hostep:
Folks, what's the status here? Any chance the jquery migrate library is going to get removed any time soon? What is blocking this?

ihor-sviziev:
@sidolov can we accept such change in 2.4.x release line?

ihor-sviziev:
@magento run all tests

sidolov:
Hi @ihor-sviziev , @hostep! Does it make sense to leave jQuery migrate for developer mode only? It will be disabled in production mode but may be helpful for developers, what do you think?

ihor-sviziev:
I think it’s useless, as magento has it for a long time and if someone had
any issues - they definitely already resolved them. I think full removal
should be better, unless migration to latest jquery version

hostep:
I was going to say the exact same thing as @ihor-sviziev 🙂 
I don't know if it makes sense to keep it included for v3 of jquery if that ever happens?
Developers can always install jquery migrate themselves should they see the need for it.

But I'm not a frontend expert, maybe @krzksz or @DanielRuf or @Parakoopa or @kirmorozov or @zetlen or @leoquijano or @jonathanKingston have a much better founded opinion (sorry for the inclusion guys, all of you are involved in discussions around jquery/jquery v3/jquery migrate here in this repo)

DanielRuf:
Well, WordPress 5.5 did the same (removing jQuery Migrate) and many plugins and themes were broken - they even announced it a while back but many big plugin developers seem to have forgotten this and users had issues which were resolved by installing the Enable jQuery Migrate Helper plugin in this case.

So we should clearly communicate this with the Magento partners and solution providers (modules / plugins) with a clear roadmap and schedule when this will be released. Otherwise this would catch many by surprise.

leoquijano:
I agree that any major changes to jQuery dependencies (e. g. jQuery Migrate) should be rolled out carefully. But keep in mind that this has been going on for a long time. My [comment](https://github.com/magento/magento2/issues/13685#issuecomment-399241048) about this dependency blocking newer libraries is already two years old, for example.

A lot of developers (and probably solution partners) are still relying on Magento's default frontend for theming. Having more modern tools would be nice, but if they can't have them, at least jQuery 3 (which is 4 years old) would help them move forward.

DanielRuf:
> I agree that any major changes to jQuery dependencies (e. g. jQuery Migrate) should be rolled out carefully. But keep in mind that this has been going on for a long time. My [comment](https://github.com/magento/magento2/issues/13685#issuecomment-399241048) about this dependency blocking newer libraries is already two years old, for example.
> 
> A lot of developers (and probably solution partners) are still relying on Magento's default frontend for theming. Having more modern tools would be nice, but if they can't have them, at least jQuery 3 (which is 4 years old) would help them move forward.

I totally agree with you. We should just make sure that this is properly communicated and that we cover this in the docs so we can point them at this then (why does my code not work anymore? We've removed jQuery Migrate. Here's some guide / are some tips to resolve the problems). Maybe a warning in the frontend docs at the top. But this is just an idea and my opinion / my two cents.

hostep:
Would it make sense to add a simple configuration option in Magento where you can choose to include jquery migrate yes or no? That way it can be kept backwards compatible whilst also giving people who know what they are doing the option to remove it if they so wish to.

It might be a bit awkward though to put this in a configuration value, since it's very developer oriented. Maybe a better idea would be to provide an option for including/excluding it in the [`view.xml` file](https://github.com/magento/magento2/blob/2.4.0/app/design/frontend/Magento/blank/etc/view.xml) of a theme?

Just an idea ...

DanielRuf:
> Would it make sense to add a simple configuration option in Magento where you can choose to include jquery migrate yes or no? That way it can be kept backwards compatible whilst also giving people who know what they are doing the option to remove it if they so wish to.

That makessense, yes.

> It might be a bit awkward though to put this in the adminhtml configuration, since it's very developer oriented. Maybe a better idea would be to provide an option for including/excluding it in the [`view.xml` file](https://github.com/magento/magento2/blob/2.4.0/app/design/frontend/Magento/blank/etc/view.xml?rgh-link-date=2020-08-20T20%3A53%3A54Z) of a theme?
> 
> Just an idea ...

good question, I'm not yet sure which place is the best option.

ihor-sviziev (changes requested):
Hi @Parakoopa,
I just reviewed failing tests - seems like in some cases w/o jquery migrate magento works incorrectly. Could you double check failing tests and update your changes accordingly?
I believe such fixes might be delivered separately than removing jquery migrate

ihor-sviziev:
Hi @Parakoopa,

Parakoopa:
Sorry, I missed the last comment, I'll check the tests when I find the time.

ihor-sviziev:
@magento run all tests

ihor-sviziev:
Hi @Parakoopa,
I am closing this PR now due to inactivity.

gwharton:
Just incase this ever gets picked up again and worked on.

The current implementation of jquery-validate uses $.event.handle...... calls.

https://github.com/magento/magento2/blob/97722131f66535783cb1ce56c179c1c6f1ca68d5/lib/web/jquery/jquery.validate.js#L1251-L1262

These are undocumented and deprecated.

Jquery.migrate is required for these calls to work under IE11. Whilst IE11 may not be supported anymore, it still does largely work ok (well on my site atleast) so seems a shame to remove jquery-migrate without first addressing the use of $.event.handle in jquery-validate.

Looks like jquery-validate hasn't been updated to any significant degree in a long time!! :(

kanevbg:
jQuery migrate is included for a reason. The best thing is to update the jQuery to latest version and update all core algorithms accordingly, than jQuery migrate wont be needed. I believe that likely it can't prove enough value for Magento to do it. Time passes and jQuery 3 gets required for many plugins and systems based on jQuery, which makes this bigger problem in time as it's introducing difficulty in customizing the frontend with newer software.

jordanvector:
Could we not swap the 

$.event.handle 

 with

$.event.dispatch

and then remove jquery migrate? I have been testing this change in IE, so far so good.  Will update my answer if I run into any issues.

```
                handler: function (e) {
                    var args = arguments;
                    args[0] = $.event.fix(e);
                    args[0].type = fix;
                    return $.event.dispatch.apply(this, args);
                }
            };
            function handler(e) {
                e = $.event.fix(e);
                e.type = fix;
                return $.event.dispatch.call(this, e);
            }
        });
```

ihor-sviziev:
@jordanvector could you create a pr with your suggested changes, it will be much easier to see the test results

jordanvector:
@ihor-sviziev yep no problem, will drop that here once thats ready.

PR #20313, opened 2019-01-15
Fix: URL Rewrites redirect loop [2.3]

ghost:
### Description
Added redirect validate method for url rewrites

### Fixed Issues
1. magento/magento2#20309: URL Rewrites redirect loop

### Manual testing scenarios
1. Go To Admin Url rewrites.
2. Create New Url rewrite with request-path / and target_path /

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #25848, opened 2019-11-29
[FIXES] #20309: URL Rewrites redirect loop

mautz-et-tong:
### Description (*)
This adds a meaningful exception if request path equals target path on url rewrite edit page in admin

1. magento/magento2#20309: URL Rewrites redirect loop

1. Go To Admin -> Marketing -> Url Rewrites.
2. Create a new Url Rewrite with request-path: foo and target_path foo with 301 or 302 redirect.
3. Click Save
4. An error message should appear with "The request path and the target path should differ to prevent an infinite redirecion loop."

### DIFF
--- a/app/code/Magento/UrlRewrite/Controller/Adminhtml/Url/Rewrite/Save.php
+++ b/app/code/Magento/UrlRewrite/Controller/Adminhtml/Url/Rewrite/Save.php
@@ -7,6 +7,7 @@
 
 namespace Magento\UrlRewrite\Controller\Adminhtml\Url\Rewrite;
 
+use Magento\Backend\Model\Session;
 use Magento\Framework\App\Action\HttpPostActionInterface as HttpPostActionInterface;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\UrlRewrite\Model\UrlFinderInterface;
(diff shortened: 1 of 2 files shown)


ihor-sviziev (changes requested):
Hi @mautz-et-tong,
Your changes looks good to me.

Could you cover your changes with some tests?

ihor-sviziev (changes requested):
Actually I remembered that such PR we already had in https://github.com/magento/magento2/pull/20313#pullrequestreview-219384257.


With this solution we have few issues:
1. If such URL rewrite was created _before_ these changes introduced - issue will be persisted. In order to fix that  we have two options (or both?)
- create data patch that will find & delete all such incorrect URL rewrites
- adjust logic to do NOT perform redirect in case of target path equal to request path
2. URL rewrites could be created not only Magento admin, but also through custom modules, so would be better to move this validation into URL rewrite model itself.

Could you adjust your PR?

mautz-et-tong:
Actually I would do everything:
1. create data patch that will find & delete all such incorrect URL rewrites (maybe moving to a working redirect with addition (-unique_hash) is better than a deletion?)
2. adjust logic to do NOT perform redirect in case of target path equal to request path (the PR you mentioned should be fair enough, isn't it?)
3. Create a model based validation to prevent this kind of redirects

ihor-sviziev:
Hi @mautz-et-tong,
Do you have any estimates when you'll be able to update your PR?
Maybe do you have any issues?

ihor-sviziev:
Hi @mautz-et-tong,
I am closing this PR now due to inactivity.

PR #25878, opened 2019-12-02
Function getApiIdentifier can return null when config value is not set.

SimonPrins:
### Description (*)
In this PR I changed the type definition for app/code/Magento/CardinalCommerce/Model/Config.php::getApiIdentifier to allow it to return null values. The returned value can be null when the `three_d_secure/cardinal/api_identifier` config setting is not set. This can cause an exception since, strict types is est to 1 for this file: https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/CardinalCommerce/Model/Config.php#L6

1. None that I could find

1. Make sure three_d_secure/cardinal/api_identifier is not set in the database (I believe it is not set by default).
2. Call app/code/Magento/CardinalCommerce/Model/Config.php::getApiIdentifier
3. You should see an exception.

Note: I encountered the exception when navigating to the onestepcheckout provided by https://www.mageplaza.com/magento-2-one-step-checkout-extension/
I am not sure when the function might be called in a vanilla instance of Magento, however since it will return null when the config value is not set, allowing null in the return type is required and will potentially fix some bugs.

### Questions or comments
Note that there are more functions in the class that also return config values but do not allow null as return type. I haven't checked these, but I suspect they should also be changed in the same way.

### DIFF
--- a/app/code/Magento/CardinalCommerce/Model/Config.php
+++ b/app/code/Magento/CardinalCommerce/Model/Config.php
@@ -54,9 +54,9 @@ class Config
      * GUID used to identify the specific API Key.
      *
      * @param int|null $storeId
-     * @return string
+     * @return string|null
      */
-    public function getApiIdentifier(?int $storeId = null): string
+    public function getApiIdentifier(?int $storeId = null): ?string
     {
         $apiIdentifier = $this->scopeConfig->getValue(
             'three_d_secure/cardinal/api_identifier',


ihor-sviziev (changes requested):
Hi @SimonPrins,
Your changes looks good, however looks like the same issue will be also for all three methods:
https://github.com/magento/magento2/blob/25b50ecd1b5f7dacee5c843f6a9d90dd45edb5a8/app/code/Magento/CardinalCommerce/Model/Config.php#L33-L85

Could you update your code to support null values in all 3 methods?

Also could you cover your changes with some unit test?

ihor-sviziev:
@SimonPrins please sign Adobe CLA, otherwise we will not be able to process your PR.
Thank you!

ihor-sviziev:
@SimonPrins  I am closing this PR now due to inactivity.

PR #25912, opened 2019-12-05
Category filters - Fix notice on incorrect price param

ihor-sviziev:
### Description (*)
See details in #25911 

1. magento/magento2#25911 : Category - Notice on incorrect price filter GET param

1. See details in #25911 

### DIFF
--- a/app/code/Magento/Catalog/Model/Layer/Filter/DataProvider/Price.php
+++ b/app/code/Magento/Catalog/Model/Layer/Filter/DataProvider/Price.php
@@ -310,7 +310,7 @@ class Price
             return false;
         }
         foreach ($filter as $v) {
-            if ($v !== '' && $v !== '0' && (double)$v <= 0 || is_infinite((double)$v)) {
+            if ($v !== '' && $v !== '0' && (!is_numeric($v) || (double)$v <= 0 || is_infinite((double)$v))) {
                 return false;
             }
         }
--- a/app/code/Magento/Catalog/Model/Layer/Filter/DataProvider/Price.php
+++ b/app/code/Magento/Catalog/Model/Layer/Filter/DataProvider/Price.php
@@ -10,6 +10,9 @@ use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\Registry;
 use Magento\Store\Model\ScopeInterface;
 
+/**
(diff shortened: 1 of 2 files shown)


krzksz (approved):

magento-engcom-team:
[ENGCOM-6424](https://jira.corp.magento.com/browse/ENGCOM-6424) has been created to process this Pull Request

engcom-Bravo:
Hello @ihor-sviziev. Thank You for Your contribution. Would You please take a look at the following gif:
category
The system.log also does not contain any main.criticals. So it is that ether the issue this PR is going to fix is not reproduced on 2.4-develop or we miss or misunderstand something when trying to reproduce it. If the true one is the second please help us to reproduce it. Thank You.

engcom-Bravo:
@ihor-sviziev, we have immediately found the cause we were not able to reproduce the issue. In order to have it reproduced "for instance https://example.com/my-category/?price=-100%27[0]" should be replaced with "for instance https://example.com/my-category?price=-100%27[0]" - the "/" before the ? should be removed. I believe in 2.3 version Your reported the issue on it was reproduced with that slash, but as far as I remember there was a commit that brought some changes to category url behavior depending on presence of such a slash in it. So probably that is the reason the issue needs now to have category url be without slash before request part (before ?) in order to get reproduced. So we are proceeding with testing Your PR.

engcom-Bravo:
@ihor-sviziev, everyting works fine, sorry for disturbing You. :heavy_check_mark: QA Passed.

MTheProgrammer:
Hello, this is still an issue in Magento 2.4.2:
```
[2021-08-17 02:41:49] main.CRITICAL: Warning: A non-numeric value encountered in XXXX/vendor/magento/module-catalog-search/Model/Layer/Filter/Price.php on line 194 [] []
```

Type in the url non-numeric value for the price filter:
```https://magento.domain/category.html?price=c-d```

ihor-sviziev:
@magento give me 2.4-develop instance

ihor-sviziev:
Hi @MTheProgrammer,
I can't reproduce the issue on the 2.4-develop issue both on the test instance and on the local machine.
If you are still able to reproduce it - please create a separate issue with steps to reproduce.
Thank you!

PR #22061, opened 2019-03-30
#22047: Magento CRON Job Names are missing in NewRelic: Transactions FIX

PiotrSiejczuk:
### Description (*)

Magento CRON Job Names are missing in NewRelic: Transactions Names - This Pull Request Fixes it as per Detailed Description below

### Fixed Issues (if relevant)

1. magento/magento2#22047: Magento CRON Job Names are missing in NewRelic: "Transaction Names"

### Manual testing scenarios (*)

1. Running Magento 2.3.x
2. Correctly Configured CRON Jobs
3. Schedule Cron Job Execution
4. Wait for any Cron Job Execution
5. See Cron Jobs Executed with proper Transaction Names in NewRelic Transaction traces:

NewRelic-UnknownTraces-ExpectedResult

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #25957, opened 2019-12-09
Add Cron Jobs names to New Relic transactions

lbajsarowicz:
### Description (*)
Feature: Add missing transaction names for Cron Jobs in New Relic reports
image

I've modified private methods, as this is not backward incompatible.

- #22047
- #22061

1. Run `bin/magento cron:run`
2. Enjoy the transaction labels for single cron jobs in NewRelic


app/code/Magento/NewRelicReporting/etc/di.xml
```diff
@@ -41,6 +41,16 @@
...
-        <plugin name="newrelic-describe-commands"  type="Magento\NewRelicReporting\Plugin\CommandPlugin"/>
+        <plugin name="newrelic-describe-commands" type="Magento\NewRelicReporting\Plugin\CommandPlugin"/>
+    </type>
+    <type name="Magento\Framework\Profiler\Driver\Standard\Stat">
+        <plugin name="newrelic-describe-cronjobs" type="Magento\NewRelicReporting\Plugin\StatPlugin"/>
+    </type>
+    <type name="Magento\NewRelicReporting\Plugin\CommandPlugin">
+        <arguments>
+            <argument name="skipCommands" xsi:type="array">
+                <item xsi:type="string" name="cron_run">cron:run</item>
```
ihor-sviziev: Why it should be skipped?

ihor-sviziev (changes requested):
Hi @lbajsarowicz,
In general your changes looks good, but would be really good to have some tests for this feature.
Could you cover your changes with some tests?

app/code/Magento/NewRelicReporting/etc/di.xml
lbajsarowicz: Because we measure what's inside, so testing the container could be misleading.

lbajsarowicz:
That will, actually, take some time.

app/code/Magento/NewRelicReporting/Plugin/CommandPlugin.php
```diff
@@ -46,10 +54,39 @@ public function __construct(
...
+
+    /**
+     * Determines whether the Command is declared to be skipped
+     *
+     * @param Command $command
+     * @return bool
+     */
+    private function isCommandSkipped(Command $command): bool
+    {
+        return in_array($command->getName(), $this->skipCommands);
```
ihor-sviziev: Looks like this code will be executed quite often. In array actually is quite slow, could you replace it to finding in map?
Related https://github.com/magento/magento-coding-standard/issues/143

ihor-sviziev (changes requested):

app/code/Magento/NewRelicReporting/Plugin/CommandPlugin.php
lbajsarowicz: This part is run only for bin/magento commands execution, however - I changed it to avoid in_array

app/code/Magento/NewRelicReporting/Test/Unit/Model/NewRelicWrapperTest.php
```diff
@@ -0,0 +1,197 @@
...
+
+        self::$systemMethodsCalled = [];
+    }
+
+    /**
+     * When NewRelic extension is not loaded, `newrelic_set_appname` should not be called
+     */
+    public function testSetAppNameWhenExtensionNotLoaded()
+    {
+        $this->markTestSkipped('Magento Coding Standard does not allow to mock System methods');
```
ihor-sviziev: Please remove skipping tests, if static tests are complains about this file - let's ignore specific issues, but we don't need to prevent unit tests to be executed

ihor-sviziev (changes requested):
@lbajsarowicz after all your changes  - could you squash all changes to single commit, because 18 commits for this small feature is really too much.

app/code/Magento/NewRelicReporting/Test/Unit/Model/NewRelicWrapperTest.php
lbajsarowicz: Unit Tests were removed as do not introduce any value.

lbajsarowicz:
Confirmed that after all the changes feature works as expected:
image

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-6461](https://jira.corp.magento.com/browse/ENGCOM-6461) has been created to process this Pull Request

engcom-Delta:
:heavy_check_mark: QA passed

PR #25980, opened 2019-12-10
Switch to public method to increase extendability

ghost:
### Description (*)
The tax calculation process does not offer many extension points. Changing the `processItem` method in `\Magento\Tax\Model\TaxCalculation` would allow the ability to affect the calculation on an individual tax item level without requiring to set a preference for the whole class. 

1. no separate issue created

1. no functional changes

### DIFF
--- a/app/code/Magento/Tax/Model/TaxCalculation.php
+++ b/app/code/Magento/Tax/Model/TaxCalculation.php
@@ -275,7 +275,7 @@ class TaxCalculation implements TaxCalculationInterface
      * @param bool $round
      * @return TaxDetailsItemInterface
      */
-    protected function processItem(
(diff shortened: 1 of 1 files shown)


dmytro-ch:
Hi @fooman, thank you for your contribution!
Unfortunately, changing the scope of visibility for the existing methods is not allowed.
Please check the [Contributors Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#documentation-of-backward-incompatible-changes) for more details.

fooman:
@dmytro-ch I believe we are at the start of development for 2.4 so believe this is a good point in time to make such a change. What is the quickest approach to get approval for this?

PR #26046, opened 2019-12-14
Grid export issue  task 25963

konarshankar07:
### Description (*)

This PR will fix the customer_group column data is not reflected in the exported CSV 

1. magento/magento2#25963: Grid Export rendered data is not reflecting in the exported File, Displayed ID instead of Rendered Label

1. Log in to the Magento Admin
2. Click "sales" from the menu on the left
3. Click "order"
4. Click on Export
5. Select CSV/Excel XML

### Expected Result
image


app/code/Magento/Sales/Ui/Component/DataProvider/Order/Document.php
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Class Document
+ */
+class Document extends \Magento\Framework\View\Element\UiComponent\DataProvider\Document
+{
+
+    /**
+     * @var string
+     */
+    private static $customerGroupAttributeCode = 'customer_group';
```
lbajsarowicz, line 22: Use `private const CUSTOMER_GROUP_ATTRIBUTE_CODE` instead.

app/code/Magento/Sales/Ui/Component/DataProvider/Order/Document.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Magento\Sales\Ui\Component\DataProvider\Order;
+
+use Magento\Customer\Api\GroupRepositoryInterface;
+use Magento\Framework\Api\AttributeValueFactory;
+use Magento\Framework\Exception\NoSuchEntityException;
+
+/**
+ * Class Document
+ */
+class Document extends \Magento\Framework\View\Element\UiComponent\DataProvider\Document
```
lbajsarowicz, line 16: Import class with alias.

konarshankar07:
I found that we can reuse the `Magento\Sales\Ui\Component\DataProvider\Document` class instead of creating new `Magento\Sales\Ui\Component\DataProvider\Order\Document` which is introduced in this PR because I think it violates DRY principle. We need to name the column for the `customer_group_id` needs to be same for whole application I found some inconsistency in `sales_order_grid` and `sales_invoice_grid` tables.

novikor:
Hi, @konarshankar07. Thank you for your contribution.
The problem is reproduced globally while exporting any grid, even custom.

Could you please investigate for a more universal solution?
I am going to check it as well.

novikor (dismissed):
I suggest taking a look at `\Magento\Framework\View\Element\UiComponent\DataProvider\SearchResult`.
Also `$document` can be moved to the constructor to create a possibility to provide it via `di.xml` for extensibility purposes.

Moreover, it is OK to modify `\Magento\Framework\View\Element\UiComponent\DataProvider\Document` directly if needed.

konarshankar07:
@novikor ... Thanks for the feedback. I'll try to find the universal solution for this issue

novikor:
@konarshankar07, - If we fail to find the universal solution, the current one is OK as well.
As far as I can see, similar problems for other grids are fixed this way.

Please let me know if you feel some trouble about that and would like to give up :)

novikor:
Hi, @konarshankar07 !
Well, I dug too deep and created my own PR: https://github.com/magento/magento2/pull/26523

It seems that I cannot review this PR anymore since I am an interested person.
I`ll request @swnsma 's review.

novikor:
Hi, @konarshankar07 . Please fix the static tests.

konarshankar07:
Hello @novikor ,
Sorry for the delay. Sure, I'll fix the static test but did you find any universal solution for this issue?
Thanks

lbajsarowicz (changes requested):
I've noticed few things to change

novikor:
> Hello @novikor ,
> Sorry for the delay. Sure, I'll fix the static test but did you find any universal solution for this issue?

Hi. Yes, - it is https://github.com/magento/magento2/pull/26523
So, I afraid current PR is not needed anymore if my one is approved, - despite it is newer than yours.
@lbajsarowicz , what do you think?

PR #26073, opened 2019-12-16
Slightly decrease importance of all cron tasks

fredden:
### Description
Set process priority for scheduled tasks & queue consumers to be less important than website visitors.

When the scheduled processes are run on a separate host to website visitors, this change will have little effect. When the scheduled processes are run on the same host as is used by website visitors, this change makes it less likely that a website would become unavailable when scheduled tasks (such as indexers, queue consumers, or other scheduled tasks) run an intensive process.

Due to the way that queue consumers are initialised, they will end up with a nice level of double this value\*, while standard scheduled tasks (eg, `bulk_cleanup`, `sales_send_order_emails`) will be run with a nice level of this value only. This actually means that queue consumers (such as `exportProcessor`) will be even less important than other scheduled tasks.

> \* The parent process will get this value, and pass the same to its child, which hits the same code-path and then increases its nice level by the same amount again.

### Fixed Issues
1. This does not fix, but may help with: #25634: Magento 2.3.3 Cronjob use too much CPU source

### Manual testing scenarios
1. Run Magento scheduled tasks via cron or by manually running `php bin/magento cron:run`
2. Observe process nice levels with a tool such as `ps` or `htop` on Unix. (I'm not sure what equivalent tools are available on Windows.)

### Questions or comments
I do not know how to write a unit test for this change.

### Contribution checklist


slavvka:
Hey @fredden Thank you for the contribution but such adjustments should be applied on the level of server configuration. Magento code shouldn't control it. I am closing the PR.

PR #26163, opened 2019-12-23
Invalidate indexers in RecurringData

### LINKED ISSUE #19469: 2.3 Customer module Recurring setup script performance problems.
When running `bin/magento setup:upgrade` for a Magento CE 2.3.x installation(or just use Magento Open source), there is an unexpected delay in the recurring setup script execution on the `Magento_Customer` module(every time when you run `bin/magento setup:upgrade`) . This is more pronounced on a large data set (>500K customers).

### References
- MAGETWO-85326
- https://github.com/magento/magento2/commit/dc4bb0eae824588459f1e82b64dae96ab6026092#diff-0ac6816ed3ec11b7a9c59731fae99d4b

### Preconditions (*)
1. Magento CE 2.2.x(or 2.3.x) -> 2.3.x upgraded codebase (pre DB upgrade)
2. A large customer database (>500K records).

### Steps to reproduce (*)
1. After codebase upgrade, proceed to run `bin/magento setup:upgrade`
2. Observe execution delay on process step:
```
Module 'Magento_Customer':
Running data recurring...
```

Repeat these steps and you will notice, since there is a recurring upgrade script, that it runs every time.

### Expected result (*)
1. No recurring data scripts run, or they are or more performant.

### Actual result (*)
1. Recurring data scripts run with each attempt to upgrade the DB.

----
After ending of update you can run again `bin/magento setup:upgrade` and you will meet this problem again.
I am not sure of the need/reason to run a recurring upgrade, but from the reference posted at the top of this issue it's clear the intent to is to handle reindexing on upgrades. This seems unwise and gives room for abusing recurring upgrade scripts with patch-like behavior or long-running processes which can delay deployment times.

Do you have any background regarding the nature of the change?

fredden:
### Description

Calling `$indexer->reindexAll();` can take a long time with a large database.
This code is run during `php bin/magento setup:upgrade`, which is part of a
typical deployment.

This change makes the step significantly quicker, while still achieving the
same result. Minimising downtime during deployments seems sensible.

### Fixed Issues
- Fixes:  #19469
- See also:  #21235

### Manual testing scenarios
1. For best results, create a large number (eg 60k) of customers. This step is not required but will more clearly highlight the difference introduced in this change.
1. Run `php bin/magento setup:upgrade`
1. Observe (lack of) delay when running `Magento_Customer` recurring data setup task.

### Questions or comments
I could not see any existing tests covering this code.


rogyar (changes requested):
Hi, @fredden. Thank you for your collaboration. Please, cover your changes with auto-tests. This step is required for further PR processing. I would recommend an integration test that runs the `setup:upgrade` command and checks the indexer status.

Thank you!

rogyar:
Also, according to the failing auto-tests, we may assume that this change brings a couple of other issues as a side-effect.

ihor-sviziev:
I believe this is related PR https://github.com/magento/magento2/pull/21235 and issue https://github.com/magento/magento2/issues/19469
It wasn’t finished, but had quite long discussion

o-iegorov:
Hi @fredden  
I am from internal team and we are started work on this issue in scope of the internal ticket MC-30539

ihor-sviziev:
@o-iegorov I've assigned you, but probably you wanted to assign to issue #19469

fredden:
@o-iegorov are you saying that the work here will be discarded and a different approach is being investigated in private? Should this pull request be closed therefore?

o-iegorov:
@fredden no, all your commits will be preserved, i plan just to fix failed tests and improve provided integration test

o-iegorov:
@fredden no need to close this PR

slackerzz:
@o-iegorov so the solution to #19469 is to invalidate the index every time we run `setup:upgrade`?

hostep:
@slackerzz: I really hope that's not the case, because that would be very bad solution.

o-iegorov:
@slackerzz  in two words yes, your solution absolutely correct: reindex moved to the patch (first time install, need to create indexer tables), invalidation for recurring (upgrade)

ihor-sviziev:
@o-iegorov,
Could you explain why marking index as invalid is good idea?, it seems to me as a hack. Why it was done in this way before declarative schema?

o-iegorov:
@ihor-sviziev for me seems as hack to reindex in upgrade data script (and then moved to recurring script with declarative schema)

ihor-sviziev:
Yep, so why we need to have remindex or index invalidation during module upgrades?

o-iegorov:
@ihor-sviziev because no warranty that customer attributes still not changed (by other modules for example) during upgrade

ihor-sviziev:
@o-iegorov, if some module changed customer attributes - this module should invalidate customer grid index. Am I wrong?
Or maybe it’s better to add some plugin that will invalidate index when customer attribute was added / changed?

o-iegorov:
@ihor-sviziev yes it's better for sure, but since such behavior was already introduced- in case when this action will be removed (for example from recurring script) some extensions (not magento core) that  based on such behavior mail fail after upgrade. To prevent such situation that may treat as backward incompatible changes better to leave already introduced flow as is

hostep:
I disagree @o-iegorov 

The reindexing in the recurring script got added in Magento 2.3 by accident, I will say it was a bug.
It's not normal to expect the `bin/magento setup:upgrade` command to reindex (or invalidate) an indexer, that's not expected behavior. I execute that command dozens of times a day, I want it to be fast, and I really don't want to see it behave the way it does right now.

If people were relying on this incorrect behavior, we just need to make it clear in the release notes of a new Magento version, that they won't be able to rely on this anymore.

It would be nice to get this finally fixed in Magento 2.4.0, as that major version does allow backwards incompatible changes.

ihor-sviziev:
@o-iegorov i was just writing similar thing while @hostep wrote his comment. 
If we’ll add it to release notes - it would be ok. Also next release 2.4.0 - so it’s time to have backward incompatible changes. I think this is really good example to fix the issue in good way.

o-iegorov:
@hostep even Magento functional test based on such behavior :) Just try to run all function tests on branch with removed app/code/Magento/Customer/Setup/RecurringData.php

ihor-sviziev:
@o-iegorov tests also depends on update by schedule index, it doesn’t mean that all production websites should work like this. tests cloud be updated if needed, it doesn’t mean it should be in this way.

o-iegorov:
@ihor-sviziev i'am totally agree that test need to be updated, that was in may comment https://github.com/magento/magento2/pull/26163#issuecomment-584708154

ihor-sviziev:
@o-iegorov, but still you want to keep index invalidation? I don't think it's really needed.

o-iegorov:
Related internal PR merged, merge commit: https://github.com/magento/magento2/commit/cac512ffe55578d41c33146c6fe03ef6bdf851ee

PR #26303, opened 2020-01-08
Fixed problems in the ResourceModel for the Inbox.

### LINKED ISSUE #29617: [Issue] Fixed problems in the ResourceModel for the Inbox.
This issue is automatically created based on existing pull request: magento/magento2#26303: Fixed problems in the ResourceModel for the Inbox.

---------

### Description (*)

When inserting a notification through `Magento\AdminNotification\Model\Inbox::add` 
with the URL empty(as it by default is), would result in the empty string being inserted into the database and therefore the `IS NULL` where condition would never match the next time around, so calling it twice with the same values and the `$isInternal` parameters set to `NULL` would result in 2 of the same notifications being inserted.
Setting the `$isInternal` parameter to `NULL` is the only way of hitting the duplication check, becuase of `isset($item['internal'])`, but hitting that would error before because `internal` is not a valid column in the `adminnotification_inbox` table.

1. Calling `Magento\AdminNotification\Model\Inbox::add` with `$isInternal` set to `NULL` to hit the duplication check(should crash before fix).
2. Check `adminnotification_inbox` table after inserting a notification with the `$url` parameter empty(should show empty string before fix and `NULL` after fix).

### Questions or comments
This changes the value that will be inputet in the `adminnotification_inbox.url` column when there is not provided any URL, the value will change from an empty string to `NULL`.

wexo-team:
### Description (*)

When inserting a notification through `Magento\AdminNotification\Model\Inbox::add` 
with the URL empty(as it by default is), would result in the empty string being inserted into the database and therefore the `IS NULL` where condition would never match the next time around, so calling it twice with the same values and the `$isInternal` parameters set to `NULL` would result in 2 of the same notifications being inserted.
Setting the `$isInternal` parameter to `NULL` is the only way of hitting the duplication check, becuase of `isset($item['internal'])`, but hitting that would error before because `internal` is not a valid column in the `adminnotification_inbox` table.

1. Calling `Magento\AdminNotification\Model\Inbox::add` with `$isInternal` set to `NULL` to hit the duplication check(should crash before fix).
2. Check `adminnotification_inbox` table after inserting a notification with the `$url` parameter empty(should show empty string before fix and `NULL` after fix).

### Questions or comments
This changes the value that will be inputet in the `adminnotification_inbox.url` column when there is not provided any URL, the value will change from an empty string to `NULL`.

### Resolved issues:
1. [x] resolves magento/magento2#29617: Fixed problems in the ResourceModel for the Inbox.

### DIFF
--- a/app/code/Magento/AdminNotification/Model/ResourceModel/Inbox.php
+++ b/app/code/Magento/AdminNotification/Model/ResourceModel/Inbox.php
@@ -98,16 +98,17 @@ class Inbox extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
 
             if (empty($item['url'])) {
                 $select->where('url IS NULL');
+                unset($item['url']);
             } else {
                 $select->where('url = ?', $item['url']);
             }
 
             if (isset($item['internal'])) {
                 $row = false;
-                unset($item['internal']);
             } else {
                 $row = $connection->fetchRow($select);
             }
+            unset($item['internal']);
 
             if (!$row) {
                 $connection->insert($this->getMainTable(), $item);


sidolov:
@magento create issue

fredden (approved):

engcom-Dash:
Hello @wexo-team,

Thanks for the contributions!

It looks like this branch is out of date with the base branch. Could you please update it so we can proceed with this PR?

Thanks.

Issue #26356, opened 2020-01-12
Create customer's token in swagger UI not working properly


marcoaacoliveira:
When trying to use Swagger UI to login you are blocked by retrieve by email step. I could reproduced that in 2.4, but I'm afraid the older versions has the same problem.

### Preconditions (*)
1. Use Swagger UI to login
2. Magento 2.4

### Steps to reproduce (*)
1. Register a customer in Magento registration form (site);
2. Try to login with that customer through swagger UI (http://yourwebsite.com/swagger);

### Expected result (*)
1. Magento creates a token for the given user;

### Actual result (*)
1. Magento doesn't find the user in the given website (swagger's website is handled as 0, the same as the admin);
2. Magento return an authentication error.

pmclain:
@marcelo-monteiro This is possible by specify in the desired store code as a query parameter in the swagger request as documented https://devdocs.magento.com/guides/v2.3/rest/generate-local.html#generate-a-full-rest-reference-locally

`http://<magento_host>/swagger?store=<store_code>` Specifying the store code of the customer will allow the request to succeed.

PR #26357, opened 2020-01-12
Fixing #26356: Problem with Swagger UI in creating token for customers

marcoaacoliveira:
### Description (*)
I had to change a bit the process of retrieve a customer by email, now if websiteId is 0, it will try through all websites to find the customer.

1. magento/magento2#26356: Create customer's token in swagger UI not working properly

1. Create a user in site by registration form; 
2. Try to login through Swagger UI with that user;

### Questions or comments
I'm not sure if in this case we gonna need new (or even if would be possible) tests. I would mostly appreciate some help from reviewer in that.

### DIFF
--- a/app/code/Magento/Customer/Model/CustomerRegistry.php
+++ b/app/code/Magento/Customer/Model/CustomerRegistry.php
@@ -59,7 +59,8 @@ class CustomerRegistry
         CustomerFactory $customerFactory,
         CustomerSecureFactory $customerSecureFactory,
         StoreManagerInterface $storeManager
-    ) {
+    )
+    {
         $this->customerFactory = $customerFactory;
         $this->customerSecureFactory = $customerSecureFactory;
         $this->storeManager = $storeManager;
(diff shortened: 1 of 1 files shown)


marcoaacoliveira:
@pmclain thanks a lot for reviewing it. There are some tests skipped, I've never seen it before... Do you have any idea what the problem is?

pmclain (changes requested):
@marcoaacoliveira Could you think about exploring how the swagger experience could be modified? Is it possible to add a store code input? This would allow the requests to succeed, since the request is failing because the store code in the request is incorrect.

`https://{{domain}}/rest/{{store_code}}/V1/integration/customer/token` would be needed.

pmclain:
@marcoaacoliveira What I mention above is currently supported. Add the desired store code as a query parameter.

`https://{domain}/swagger?store={store_code}`

pmclain:
Closing the PR as the request works when the store code is passed to swagger, no application changes are required. https://devdocs.magento.com/guides/v2.3/rest/generate-local.html#generate-a-full-rest-reference-locally

marcoaacoliveira:
> @marcoaacoliveira What I mention above is currently supported. Add the desired store code as a query parameter.
> 
> `https://{domain}/swagger?store={store_code}`

Indeed, I was wrong, I didn't know we could do something like that. It isn't a bug at all, is it?

PR #26411, opened 2020-01-15
Move additional dependencies from private getters to constructor - Magento_Catalog

Bartlomiejsz:
### Description (*)

This PR moves additionally introduced dependencies from private getter methods into constructor in Magento_Catalog module.

### DIFF
--- a/app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
+++ b/app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
@@ -3,6 +3,7 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+declare(strict_types=1);
 
 namespace Magento\Catalog\Controller\Adminhtml\Product\Initialization;
 
@@ -132,6 +133,7 @@ class Helper
      * @param FormatInterface|null $localeFormat
      * @param ProductAuthorization|null $productAuthorization
      * @param DateTimeFilter|null $dateTimeFilter
+     * @param LinkResolver|null $linkResolver
      * @SuppressWarnings(PHPMD.ExcessiveParameterList)
      */
     public function __construct(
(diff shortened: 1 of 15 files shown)


app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
```diff
@@ -141,6 +143,7 @@ public function __construct(
         ProductLinks $productLinks,
         Js $jsHelper,
         Date $dateFilter,
+        LinkResolver $linkResolver,
         CustomOptionFactory $customOptionFactory = null,
```
orlangur: Please add all new dependencies in a backward-compatible manner - similar to other added in this class:
1) new dependency is last
2) `null` default value
3) initialized in constructor when it was not passed

orlangur (changes requested):

app/code/Magento/Catalog/Controller/Adminhtml/Product/Initialization/Helper.php
Bartlomiejsz: I was sure that it should be done this way but in few of my previous PRs it was mentioned that for 2.4 release proper way is without ObjectManager, and I had to change it for such like here. So which is correct version?

Bartlomiejsz: @orlangur check https://github.com/magento/magento2/pull/26269 as an example

Bartlomiejsz: Hi @orlangur, also check this one please https://github.com/magento/magento2/commit/4e6f6ae42cc6109a02d35d0363b1cad0bbac66e1 this was commited by Magento core and this is pretty much the same I did here

Bartlomiejsz:
Closing this one to open new PR as draft, since Catalog requires serious amount of modifications

PR #26415, opened 2020-01-16
fix type hinting for getStreet() in AbstractAddress

Pr00xxy:
### Description (*)

Return type hint does not match actual types that can be returned by function.
Is fixed by adding the additional string type

on 2.4-develop branch:
1. Open any competent PHP IDE that has an inspector
2. app/code/Magento/Customer/Model/Address/AbstractAddress.php:219 should warn about return type mismatch with what is declared in the docblock.

with my fix:
1. app/code/Magento/Customer/Model/Address/AbstractAddress.php:219 should not warn about return type mismatch with what is declared in the docblock

### DIFF
--- a/app/code/Magento/Quote/Api/Data/AddressInterface.php
+++ b/app/code/Magento/Quote/Api/Data/AddressInterface.php
@@ -139,7 +139,7 @@ interface AddressInterface extends \Magento\Framework\Api\CustomAttributesDataIn
     /**
      * Get street
      *
-     * @return string[]
+     * @return string|string[]
      */
     public function getStreet();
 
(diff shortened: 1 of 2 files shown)


app/code/Magento/Quote/Api/Data/AddressInterface.php
```diff
@@ -139,7 +139,7 @@ public function setCountryId($countryId);
     /**
      * Get street
      *
-     * @return string[]
+     * @return string|string[]
```
orlangur, line 142: How it can return string?

orlangur (changes requested):

PR #26452, opened 2020-01-18
magento/magento2#: GraphQl. DeletePaymentToken. Remove redundant validation logic. Test coverage.

atwixfirster:
### Description (*)

Pull request:
- removes a redundant validation logic from `\Magento\VaultGraphQl\Model\Resolver\DeletePaymentToken` because `Webonix` will cover a removed validation logic.
- adds a unit test for mentioned class

1. ...
2. ...

Thank you!

### DIFF
--- a/app/code/Magento/VaultGraphQl/Test/Unit/Model/Resolver/DeletePaymentTokenTest.php
+++ b/app/code/Magento/VaultGraphQl/Test/Unit/Model/Resolver/DeletePaymentTokenTest.php
@@ -1,0 +1,230 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\VaultGraphQl\Test\Unit\Model\Resolver;
+
+use Magento\Framework\TestFramework\Unit\Helper\ObjectManager;
+use Magento\Framework\GraphQl\Config\Element\Field;
+use Magento\Framework\GraphQl\Exception\GraphQlNoSuchEntityException;
+use Magento\Framework\GraphQl\Schema\Type\ResolveInfo;
+use Magento\GraphQl\Model\Query\ContextInterface;
+use Magento\GraphQl\Model\Query\ContextExtensionInterface;
+use Magento\Vault\Api\PaymentTokenManagementInterface;
+use Magento\Vault\Api\PaymentTokenRepositoryInterface;
+use Magento\Vault\Api\Data\PaymentTokenInterface;
+use Magento\VaultGraphQl\Model\Resolver\DeletePaymentToken;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Test class for \Magento\VaultGraphQl\Model\Resolver\DeletePaymentToken
+ */
+class DeletePaymentTokenTest extends TestCase
+{
+    /**
+     * Object Manager Instance
+     *
+     * @var ObjectManager
+     */
+    private $objectManager;
+
+    /**
+     * Testable Object
+     *
+     * @var DeletePaymentToken
+     */
+    private $resolver;
+
+    /**
+     * @var ContextInterface|MockObject
+     */
+    private $contextMock;
+
+    /**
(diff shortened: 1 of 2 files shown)


app/code/Magento/VaultGraphQl/Test/Unit/Model/Resolver/DeletePaymentTokenTest.php
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * @var PaymentTokenRepositoryInterface|MockObject
+     */
+    private $paymentTokenRepositoryMock;
+
+    /**
+     * @var PaymentTokenInterface|MockObject
+     */
+    private $paymentTokenMock;
+
```
rogyar, line 70: Please, make sure you don't have duplicated empty lines.

app/code/Magento/VaultGraphQl/Test/Unit/Model/Resolver/DeletePaymentTokenTest.php
```diff
@@ -0,0 +1,232 @@
...
+            ->method('getByPublicHash')
+            ->willReturn($this->paymentTokenMock);
+
+        $this->paymentTokenRepositoryMock
+            ->expects($this->once())
+            ->method('delete')
+            ->with($this->paymentTokenMock)
+            ->willReturn($paymentTokenResult);
+
+        $this->resolver->resolve(
```
rogyar: I would recommend using assertions (`self::assert`) when a testable method is supposed to return something.

rogyar (changes requested):
Hi @atwixfirster. Thank you for providing an additional unit test, that's very useful. However, the provided unit test might not exactly cover the modification provided in this PR.

I would suggest extending the [following](https://github.com/magento/magento2/blob/288dec36c94dc7372ccb540095434cab815607f1/dev/tests/api-functional/testsuite/Magento/GraphQl/Vault/CustomerPaymentTokensTest.php#L20) API-functional test with a simple `testDeletePaymentTokenWithNoPublicHash` test. Since there are similar tests in the mentioned class, it should not be a big deal.

Thank you!

app/code/Magento/VaultGraphQl/Test/Unit/Model/Resolver/DeletePaymentTokenTest.php
atwixfirster: Yep... **fixed**

app/code/Magento/VaultGraphQl/Test/Unit/Model/Resolver/DeletePaymentTokenTest.php
atwixfirster: Good notice!

**implemented**

Thanks @rogyar

atwixfirster:
> Hi @atwixfirster. Thank you for providing an additional unit test, that's very useful. However, the provided unit test might not exactly cover the modification provided in this PR.
> 
> I would suggest extending the [following](https://github.com/magento/magento2/blob/288dec36c94dc7372ccb540095434cab815607f1/dev/tests/api-functional/testsuite/Magento/GraphQl/Vault/CustomerPaymentTokensTest.php#L20) API-functional test with a simple `testDeletePaymentTokenWithNoPublicHash` test. Since there are similar tests in the mentioned class, it should not be a big deal.
> 
> Thank you!

Regarding to @lenaorobei notice, we do not cover framework (`Webonix`) errors in the API-unctional test:



Thank you!

rogyar:
Probably, that makes sense. If we would cover every Webonyx validation case, we could have a really huge test case. 

Anyway, thank you for the fix

rogyar (approved):

magento-engcom-team:
[ENGCOM-6636](https://jira.corp.magento.com/browse/ENGCOM-6636) has been created to process this Pull Request

engcom-Bravo:
:heavy_check_mark:  QA Passed

Hello @atwixfirster 

- The PR adds a new Unit test and the test Passes.
- Also, the redundant validation logic is removed from ```\Magento\VaultGraphQl\Model\Resolver\DeletePaymentToken```

PR #26474, opened 2020-01-21
Adjusted TransportBuilder prepareMessage method to support text/plain…

### LINKED ISSUE #26471: TransportBuilder unable to send emails of Content-Type "text/plain" as of 2.3.3
### Preconditions (*)

1. Magento Community Edition 2.3.3
2. One or more Magento email templates configured as type "text" via an email_templates.xml file.

### Steps to reproduce (*)

1.Attempt to send a template based email of type "text" using Magento\Framework\Mail\Template\TransportBuilder that is of Content-Type "text/plain" using the TransportBuilder class's sendMessage method.

### Expected result (*)

1. An email is dispatched whose "Content-Type" header is set to "text/plain"

### Actual result (*)

1. The email is sent with a "Content-Type" header set to "text/html"

### Additional Information (*)
It appears that as of Magento 2.3.3, it is no longer possible to send "text/plain" emails using the TransportBuilder. This is because, as of 2.3.3, the TransportBuilder's prepareMessage method does not pass the determined template type to the mimePartInterfaceFactory's create method.

### Current method definition as of 2.3.3:

```php
protected function prepareMessage()
    {
        $template = $this->getTemplate();
        $content = $template->processTemplate();
        switch ($template->getType()) {
            case TemplateTypesInterface::TYPE_TEXT:
                $part['type'] = MimeInterface::TYPE_TEXT;
                break;

            case TemplateTypesInterface::TYPE_HTML:
                $part['type'] = MimeInterface::TYPE_HTML;
                break;

            default:
                throw new LocalizedException(
                    new Phrase('Unknown template type')
                );
        }
        $mimePart = $this->mimePartInterfaceFactory->create(['content' => $content]);
        $this->messageData['body'] = $this->mimeMessageInterfaceFactory->create(
            ['parts' => [$mimePart]]
        );

        $this->messageData['subject'] = html_entity_decode(
            (string)$template->getSubject(),
            ENT_QUOTES
        );
        $this->message = $this->emailMessageInterfaceFactory->create($this->messageData);

        return $this;
    }
```
### Proposed change that would resolve issue (only the relevant snippet):

```php
$mimePart = $this->mimePartInterfaceFactory->create(
        	[
        		'content' => $content,
        		'type' => $part['type']
        	]
        );
```

(Pull request incoming)

twoonesixdigital:
### Description (*)

As of 2.3.3, Magento's Magento\Framework\Mail\Template\TransportBuilder class can no longer send emails whose "Content-Type" header is set to "text/plain". This is because the TransportBuilder's prepareMessage method does not pass the determined template type to the mimePartInterfaceFactory's create method. When no "type" key is present in the array passed to the create method, the mimePartInterfaceFactory defaults to "text/html".

This pull request simply modifies the prepareMessage method so that it passes the previously determined template type to the mimePartInterfaceFactory's create method.

1. Fixed magento/magento2#26471: TransportBuilder unable to send emails of Content-Type "text/plain" as of 2.3.3

1. Define an email template of type "text" via an email_templates.xml file
2. Attempt to use the TransportBuilder's sendMessage() method to send an email based on that template:

```php
$transport = $this->TransportBuilder->setTemplateIdentifier('some_template_id')
	->setTemplateOptions(['area' => \Magento\Framework\App\Area::AREA_ADMINHTML, 'store' => \Magento\Store\Model\Store::DEFAULT_STORE_ID])
	->setFrom("email@domain.com")
	->addTo("email@domain.com")
	->getTransport()
	->sendMessage();
```

### DIFF
--- a/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
+++ b/lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
@@ -394,7 +394,12 @@ class TransportBuilder
         }
 
         /** @var \Magento\Framework\Mail\MimePartInterface $mimePart */
-        $mimePart = $this->mimePartInterfaceFactory->create(['content' => $content]);
+        $mimePart = $this->mimePartInterfaceFactory->create(
+        	[
+        		'content' => $content,
+        		'type' => $part['type']
+        	]
+        );
         $this->messageData['encoding'] = $mimePart->getCharset();
         $this->messageData['body'] = $this->mimeMessageInterfaceFactory->create(
             ['parts' => [$mimePart]]


lib/internal/Magento/Framework/Mail/Template/TransportBuilder.php
```diff
@@ -394,7 +394,12 @@ protected function prepareMessage()
         }
 
         /** @var \Magento\Framework\Mail\MimePartInterface $mimePart */
-        $mimePart = $this->mimePartInterfaceFactory->create(['content' => $content]);
+        $mimePart = $this->mimePartInterfaceFactory->create(
+        	[
+        		'content' => $content,
+        		'type' => $part['type']
```
fooman, line 400: I know the variable name is not coming from you but it looks like that $part is never defined as an array prior. Since it only uses type maybe it makes sense to rename `$part['type']` to `$partType`?

twoonesixdigital: I would agree. I suspect usage of the $part array is probably an artifact of an older version of the class. Replacing that with something simpler and less ambiguous would likely reduce confusion.

thiaramus:
The internal Magento team is working on this issue right now.
The fix will be based on commits from the author and delivered according to our internal processes. All original commits made by the author will be preserved.

PR #26475, opened 2020-01-21
Update Store.php

eermolaev:
### Description (*)

Expected result: the \Magento\Eav\Model\Entity\Attribute\Source\Table::getOptionText method returns Store name or array of the Store names by id(s).

Actual result:
1. Notice: Undefined offset: 1 if you have 1 store in your Magento
2. you requested value "1", but method returns array ['value'=> "5", 'label' => 'Store5']
3. you requested value "1", it may returns array ['value'=> "1", 'label' => 'Store1'] accidentally, but the expected result if a store name "Store1".

Logic should be the same as here: \Magento\Eav\Model\Entity\Attribute\Source\Table::getOptionText otherwise we have one of the following issues:

1. I'm trying load the Store name for the customer through the "store_id" proprety
2. The $customer is instance of \Magento\Customer\Model\Backend\Customer. Customer data is loaded.
`$customer->getResource()->getAttribute('store_id')->getFrontend()->getValue($customer)`

### DIFF
--- a/app/code/Magento/Customer/Model/Customer/Attribute/Source/Store.php
+++ b/app/code/Magento/Customer/Model/Customer/Attribute/Source/Store.php
@@ -85,12 +85,18 @@ class Store extends \Magento\Eav\Model\Entity\Attribute\Source\Table
 
         if ($isMultiple) {
             $values = [];
-            foreach ($value as $val) {
-                $values[] = $this->_options[$val];
+            foreach ($options as $item) {
+                if (in_array($item['value'], $value)) {
+                    $values[] = $item['label'];
+                }
             }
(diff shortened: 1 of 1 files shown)


lbajsarowicz (changes requested):
Try to - firstly - cover that change with Unit Tests, then use array_filter and array_map. I believe that would be much better solution than so many `in_array`s (that are considered to be inefficient in PHP)

engcom-Echo:
@eermolaev. Please sign "Adobe  CLA".

engcom-Charlie:
Hi @eermolaev, I'm closing this PR now due to inactivity.

PR #26502, opened 2020-01-22
{ASI} :- Error message to be cached for grid data storage component

konarshankar07:
### Description (*)

This PR is part of the ASI issue in which error message is not cached when user try to apply filter

1. magento/adobe-stock-integration#863: Message disappear when the request with broken credentials is sent second time

Please check https://github.com/magento/adobe-stock-integration/pull/884

### DIFF
--- a/dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
+++ b/dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
@@ -6,9 +6,11 @@
 /*eslint max-nested-callbacks: 0*/
 /*jscs:disable requireCamelCaseOrUpperCaseIdentifiers*/
 define([
+    'jquery',
     'mageUtils',
+    'underscore',
     'Magento_Ui/js/grid/data-storage'
-], function (utils, DataStorage) {
+], function ($, utils, _, DataStorage) {
     'use strict';
 
     describe('Magento_Ui/js/grid/data-storage', function () {
@@ -322,7 +324,12 @@ define([
 
         describe('"updateData" method', function () {
             var model = new DataStorage({
-                dataScope: 'magento'
+                dataScope: 'magento',
+                requestConfig: {
+                    url: 'magento.com',
+                    method: 'GET',
+                    dataType: 'json'
+                }
             });
 
             it('Check for defined ', function () {
@@ -342,6 +349,83 @@ define([
                 }];
 
                 expect(model.updateData(data)).toBeTruthy();
+            });
+        });
+
+        describe('"requestData" method', function () {
+            var model = new DataStorage({
+                dataScope: 'magento'
+            });
+
+            it('Check for defined', function () {
+                expect(model.hasOwnProperty('requestData')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof model.requestData;
+
+                expect(type).toEqual('function');
+            });
+
+            it('Check Ajax request', function () {
+                var params = {
+                    namespace: 'magento',
+                    search: '',
+                    filters: {
+                        store_id: 0
+                    },
+                    sorting: {},
+                    paging: {}
+                },
+                query = utils.copy(params);
+
+                spyOn(model, 'onRequestComplete');
+                spyOn($, 'ajax').and.callFake(function () {
+                    return {
+                        /**
+                         * Success result for ajax request
+                         */
+                        done: function () {
+                            model.onRequestComplete(model, query);
+                        }
+                    };
+                });
+                model.requestData(params);
+                expect($.ajax).toHaveBeenCalled();
+                expect(model.onRequestComplete).toHaveBeenCalled();
+            });
+        });
+
+        describe('"getRequest" method', function () {
+            var model = new DataStorage({
+                dataScope: 'magento'
+            });
+
+            it('Check for defined', function () {
+                expect(model.hasOwnProperty('getRequest')).toBeDefined();
+            });
+
+            it('Check method', function () {
+                var type = typeof model.getRequest;
+
+                expect(type).toEqual('function');
+            });
+
+            it('check "getRequest" has been executed', function () {
+                var params = {
+                    namespace: 'magento',
+                    search: '',
+                    sorting: {},
+                    paging: {}
+                };
+
+                model._requests.push({
+                    ids: ['1'],
+                    params: params,
+                    totalRecords: 1,
+                    errorMessage: ''
+                });
+                expect(model.getRequest(params)).toBeTruthy();
             });
         });
 
@@ -366,6 +450,45 @@ define([
                 };
 
                 expect(model.getRequestData(request)).toBeTruthy();
+            });
+
+            it('check "getByIds" has been executed', function () {
+                var request = {
+                    ids: [1,2,3]
+                };
+
+                spyOn(model, 'getByIds');
+                model.getRequestData(request);
+                expect(model.getByIds).toHaveBeenCalled();
+            });
+
+            it('check "delay" function has been executed', function () {
+                var request = {
+                    ids: [1,2,3],
+                    totalRecords: 3,
+                    errorMessage: ''
+                };
+
+                spyOn(_, 'delay');
+                model.getRequestData(request);
+                expect(_.delay).toHaveBeenCalled();
+            });
+
+            it('check "delay" function has not been executed', function () {
+                var request = {
+                        ids: [1,2,3],
+                        totalRecords: 3,
+                        errorMessage: ''
+                    };
+
+                    model = new DataStorage({
+                        dataScope: 'magento',
+                        cachedRequestDelay: 0
+                    });
+
+                spyOn(_, 'delay');
+                model.getRequestData(request);
+                expect(_.delay).not.toHaveBeenCalled();
             });
         });
 
@@ -531,7 +654,9 @@ define([
                     paging: {}
                 };
 
-                model.wasRequested(params);
+                spyOn(model, 'getRequest').and.callFake(function () {
+                    return false;
+                });
                 expect(model.wasRequested(params)).toBeFalsy();
             });
         });
@@ -558,7 +683,7 @@ define([
                         entity_id: '1'
                     }]
                 },
-params = {
+                params = {
                     namespace: 'magento',
                     search: '',
                     sorting: {},
--- a/dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
+++ b/dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
@@ -62,13 +62,13 @@ define([
             });
         });
 
-        describe('"getByIds"', function() {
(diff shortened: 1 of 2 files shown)


sivaschenko:
@konarshankar07 thanks for opening the pull request! there was a jasmine test created for this method, could you include it with this pull request as well?

konarshankar07:
@sivaschenko Yes, I'm going to add those jasmine test with this PR but I think we should cover other methods with the jasmine test for data-storage so I'm working on it

konarshankar07:
@sivaschenko 
Here is the screenshot with Adobe Stock preview images
image

diazwatson:
Waiting for https://github.com/magento/adobe-stock-integration/pull/884 to be merged in order to merge this one

konarshankar07:
@diazwatson ... Actually its opposite to your comment. We need to merge this PR in order to merge [ASI](https://github.com/magento/adobe-stock-integration/pull/884) PR

diazwatson:
Thanks for clarifying @konarshankar07 
What is left in this PR then?

konarshankar07:
@diazwatson...  Code Review for jasmine test cause I'm very new to the jasmine test so I might have missed any test or needs an improvement 
Thanks

diazwatson:
Maybe @filmaj can help? 😅

konarshankar07:
@diazwatson ... Yes @sivaschenko or @filmaj  both can review :)

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+
+                spyOn(model, 'getRequest');
+                spyOn(model, 'hasScopeChanged').and.callFake(function () {
+                    return false;
+                });
+                model.getData(params);
+                expect(model.getRequest).toHaveBeenCalled();
+            });
+
+            it('Return "getRequestData" method', function () {
```
filmaj: One thing I think we should try to do is provide good test names. This may seem like a silly recommendation, but as a first-time contributor to the project, one of the first things I do to familiarize myself with the project is read the tests! While I understand what the intention behind this test is, it only became clear to me after reading the data-storage `getData` method.
I would suggest naming the test something like "it returns cached request  data if a cached request exists and no refresh option is provided". In this way, the name of the test alone is enough to explain to me some key functionality of this module.

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+
+                spyOn(model, 'getRequestData');
+                spyOn(model, 'getRequest').and.callFake(function () {
+                    return true;
+                });
+                model.getData(params, options);
+                expect(model.getRequestData).toHaveBeenCalled();
+            });
+
+            it('Return "requestData" method', function () {
```
filmaj: Along the same lines as my last comment, let's qualify the test name further. How does it differentiate from the previous test? having multiple tests with the same name is not recommended - that can be quite confusing. Wording matters! Remember that tests are a tool to help us, the developers, be productive and guide us. Spending  the extra few minutes to properly comment, document and name things goes a long way to making the code maintainable long term.

filmaj (commented):
Nicely done on the tests! I left a couple of comments around labeling. Wording is important 😄

konarshankar07:
Thanks @filmaj for reviewing the pull request and will work on the your suggestions

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+            });
+
+            it('check method type', function () {
+                expect(typeof model.getByIds).toEqual('function');
+            });
+
+            it('Check returned value if method called with argument', function () {
+                var ids = [1,2,3];
+
+                expect(model.getByIds(ids)).toBeDefined();
```
sivaschenko: I don't think `toBeDefined` is a specific enough case to be checked separately. I'd focus on testing the concrete return values instead

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+            });
+        });
+
+        describe('"getByIds" method', function () {
+
+            var model = new DataStorage({
+                dataScope: 'magento'
+            });
+
+            it('check for defined', function () {
```
sivaschenko: Do you think `check for defined` and `check method type` are necessary? If the function is not declared the following tests will fail anyway and it wouldn't be hard to identify the problem.

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+                var type = typeof model.initConfig;
+
+                expect(type).toEqual('function');
+            });
+
+            it('Check returned value if method called without arguments', function () {
+                expect(model.initConfig()).toBeDefined();
+            });
+
+            it('Check returned value type if method called without arguments', function () {
```
sivaschenko: I don't think `Check for defined`, `Check method type`,  `Check returned value if method called without arguments` and `Check returned value type if method called without arguments` are necessary

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+                expect(type).toEqual('boolean');
+            });
+
+            it('Return false if "getByIds" has been called', function () {
+                var ids = [1,2,3];
+
+                expect(model.getByIds(ids)).toEqual(false);
+            });
+
+            it('Return array if "getByIds" has been called', function () {
```
sivaschenko: it would be good to test the case when `ids` parameter contains both existing and non-existing ids

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+                    dataScope: 'magento',
+                    data: {
+                        1: {
+                            id_field_name: 'entity_id',
+                            entity_id: '1'
+                        }
+                    }
+                });
+
+                expect(typeof model.getByIds(ids)).toEqual('object');
```
sivaschenko: I'd check for the specific object

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+                var ids = [1,2,3];
+
+                expect(model.getIds(ids)).toBeDefined();
+            });
+
+            it('check returned type if method called with argument', function () {
+                var ids = [1,2,3],
+                    type = typeof model.getIds(ids);
+
+                expect(type).toEqual('object');
```
sivaschenko: I'd check for specific return

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+
+            it('check for defined', function () {
+                expect(model.hasOwnProperty('getIds')).toBeDefined();
+            });
+
+            it('check method type', function () {
+                expect(typeof model.getIds).toEqual('function');
+            });
+
+            it('check returned value if method called with argument', function () {
```
sivaschenko: I don't think it's beneficial to test booth return type and return value - only one check for return value would be enough.

It'd be better to check the case when no arguments are passed to `getIds` method instead.

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -22,7 +25,253 @@ define([
...
+                var ids = [1,2,3];
+
+                expect(model.getIds(ids)).toBeDefined();
```
sivaschenko: `toBeDefined` is not actually a check for value, it would be good to verify the actual returned value

sivaschenko (changes requested):
Hi @konarshankar07 thank you for the pull request! The implementation looks fine, the only comments I have are addressed to the test:

- I don't think that abstract checks like `toBeDefined`, `toBeTruthy`, `expect(type).toEqual` provide enough information to assume that the method is working correctly
- I'd limit the tests to verifying the actual return of the method depending on arguments

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
sivaschenko: I'd agree that test names should be more detailed to differentiate test between each other, however, I wouldn't make them too long. I don't think test results can be accessed without reading the code (by name only).

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
```diff
@@ -401,73 +309,83 @@ define([
...
 
                 spyOn(model, 'removeRequest');
-                spyOn(model, 'getIds').and.callFake(function () {
-                    return ['1','2','3'];
-                });
+                spyOn(model, 'getIds').and.returnValue(ids);
                 model.cacheRequest(data, params);
-                expect(typeof model._requests).toEqual('object');
-                expect(model.getIds).toHaveBeenCalled();
+                expect(model.getRequest(params)).toEqual(request);
```
konarshankar07, line 339: I think here we need to also test `model._requests`

sivaschenko (approved):

magento-engcom-team:
[ENGCOM-6840](https://jira.corp.magento.com/browse/ENGCOM-6840) has been created to process this Pull Request

dev/tests/js/jasmine/tests/app/code/Magento/Ui/base/js/grid/data-storage.test.js
sivaschenko: I think it's better to test it through `getRequest` as `_requests` property is a detail of internal implementation

konarshankar07: Okay, `_requests` array pushed functionality is added in the `cacheRequest` method so I thought that we need to cover in the `cacheRequest` functionality

engcom-Bravo:
:heavy_check_mark: QA Passed

The link to _Configuration - System - Adobe Stock Integration_ doesn't disappear after changing between **per page**
asi26502

PR #26623, opened 2020-02-01
#26622 - Check quote item for parentItem instead of parentItemId

aligent-lturner:
### Description (*)

This PR fixes issue #26622 - a "fixed amount for whole cart" discount is calculated incorrectly when a configurable product is first added to the cart. This change is to use `getParentItem` instead of `getParentItemId` from the quote item when determining if it should be skipped. This is because `parentItem` is set when the quote item is created, but `parentItemId` is only set in its `beforeSave` function.

1. magento/magento2#26622: Fixed cart discount calculated incorrectly when product first added to cart

1. Create fixed amount discount for whole cart of $1 with no conditions/restrictions
2. Navigate to PDP of configurable product
3. Choose configurable product options and add to cart
4. Confirm that discount is correct ($1) in minicart and/or database.

### DIFF
--- a/app/code/Magento/SalesRule/Model/Validator.php
+++ b/app/code/Magento/SalesRule/Model/Validator.php
@@ -383,13 +383,7 @@ class Validator extends \Magento\Framework\Model\AbstractModel
 
                 foreach ($items as $item) {
                     //Skipping child items to avoid double calculations
-                    if ($item->getParentItemId() || $item->getParentItem()) {
-                        continue;
-                    }
-                    if (!$rule->getActions()->validate($item)) {
-                        continue;
-                    }
-                    if (!$this->canApplyDiscount($item)) {
+                    if (!$this->isValidItemForRule($item, $rule)) {
                         continue;
                     }
                     $qty = $this->validatorUtility->getItemQty($item, $rule);
@@ -407,6 +401,32 @@ class Validator extends \Magento\Framework\Model\AbstractModel
         }
 
         return $this;
+    }
+
+    /**
+     * Determine if quote item is valid for a given sales rule
+     * @param AbstractItem $item
+     * @param Rule $rule
+     * @return bool
+     */
+    protected function isValidItemForRule(
+        AbstractItem $item,
+        Rule $rule
+    ) {
+        /** @var AbstractItem $item */
+        if ($item->getParentItemId()) {
+            return false;
+        }
+        if ($item->getParentItem()) {
+            return false;
+        }
+        if (!$rule->getActions()->validate($item)) {
+            return false;
+        }
+        if (!$this->canApplyDiscount($item)) {
+            return false;
+        }
+        return true;
     }
 
     /**
--- a/app/code/Magento/SalesRule/Model/Validator.php
+++ b/app/code/Magento/SalesRule/Model/Validator.php
@@ -405,6 +405,7 @@ class Validator extends \Magento\Framework\Model\AbstractModel
 
     /**
      * Determine if quote item is valid for a given sales rule
+     *
      * @param AbstractItem $item
      * @param Rule $rule
      * @return bool
--- a/app/code/Magento/SalesRule/Model/Validator.php
+++ b/app/code/Magento/SalesRule/Model/Validator.php
@@ -383,7 +383,7 @@ class Validator extends \Magento\Framework\Model\AbstractModel
 
                 foreach ($items as $item) {
                     //Skipping child items to avoid double calculations
-                    if ($item->getParentItem()) {
+                    if ($item->getParentItemId() || $item->getParentItem()) {
                         continue;
                     }
                     if (!$rule->getActions()->validate($item)) {
--- a/app/code/Magento/SalesRule/Model/Validator.php
+++ b/app/code/Magento/SalesRule/Model/Validator.php
@@ -383,7 +383,7 @@ class Validator extends \Magento\Framework\Model\AbstractModel
 
                 foreach ($items as $item) {
                     //Skipping child items to avoid double calculations
-                    if ($item->getParentItemId()) {
+                    if ($item->getParentItem()) {
                         continue;
                     }
                     if (!$rule->getActions()->validate($item)) {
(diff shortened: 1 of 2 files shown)


app/code/Magento/SalesRule/Model/Validator.php
```diff
@@ -383,7 +383,7 @@ public function initTotals($items, Address $address)
 
                 foreach ($items as $item) {
                     //Skipping child items to avoid double calculations
-                    if ($item->getParentItemId()) {
+                    if ($item->getParentItem()) {
```
ihor-sviziev: Hi @aligent-lturner,
What do you think about checking for both options, like this?
```suggestion
                    if ($item->getParentItemId() || $item->getParentItem()) {
```

Could you also cover your changes with some test?

ihor-sviziev (changes requested):

aligent-lturner:
@ihor-sviziev changes made as requested. Modified existing test case in `ValidatorTest` class to handle additional quote items with differing combinations of `parentItemId` and `parentItem`

aligent-lturner:
@ihor-sviziev I had to move the functionality into a separate function to reduce the cyclomatic complexity.
Some functional tests are failing, but they appear to be completely unrelated to my changes.

app/code/Magento/SalesRule/Model/Validator.php
aligent-lturner: Code has been updated to check both cases, and corresponding unit test has been updated to check different combinations.

app/code/Magento/SalesRule/Model/Validator.php
```diff
@@ -409,6 +403,33 @@ public function initTotals($items, Address $address)
...
     }
 
+    /**
+     * Determine if quote item is valid for a given sales rule
+     *
+     * @param AbstractItem $item
+     * @param Rule $rule
+     * @return bool
+     */
+    protected function isValidItemForRule(
```
ihor-sviziev: Could you make it private and defined in one row?

app/code/Magento/SalesRule/Model/Validator.php
```diff
@@ -409,6 +403,33 @@ public function initTotals($items, Address $address)
...
+        AbstractItem $item,
+        Rule $rule
+    ) {
+        /** @var AbstractItem $item */
```
ihor-sviziev: We don’t need this phpdoc block as we have this type defined in method arguments

app/code/Magento/SalesRule/Test/Unit/Model/ValidatorTest.php
```diff
@@ -368,11 +371,21 @@ public function testInitTotalsCanApplyDiscount()
         $validator->expects($this->at(1))->method('isValid')->with($item2)->willReturn(true);
 
         $item1->expects($this->any())->method('getParentItemId')->willReturn(false);
+        $item1->expects($this->any())->method('getParentItem')->willReturn(false);
```
ihor-sviziev: Does it really return false? I believe it usually returns null

app/code/Magento/SalesRule/Test/Unit/Model/ValidatorTest.php
```diff
@@ -368,11 +371,21 @@ public function testInitTotalsCanApplyDiscount()
...
         $item1->expects($this->never())->method('getDiscountCalculationPrice');
         $item1->expects($this->never())->method('getBaseDiscountCalculationPrice');
         $item2->expects($this->any())->method('getParentItemId')->willReturn(false);
+        $item2->expects($this->any())->method('getParentItem')->willReturn(false);
         $item2->expects($this->any())->method('getDiscountCalculationPrice')->willReturn(50);
         $item2->expects($this->once())->method('getBaseDiscountCalculationPrice')->willReturn(50);
+        $item3->expects($this->any())->method('getParentItemId')->willReturn(false);
+        $item3->expects($this->any())->method('getParentItem')->willReturn(true);
```
ihor-sviziev: I think it’s not real case when this method could return true

app/code/Magento/SalesRule/Test/Unit/Model/ValidatorTest.php
```diff
@@ -368,11 +371,21 @@ public function testInitTotalsCanApplyDiscount()
...
+        $item3->expects($this->never())->method('getDiscountCalculationPrice');
+        $item3->expects($this->never())->method('getBaseDiscountCalculationPrice');
+        $item4->expects($this->any())->method('getParentItemId')->willReturn(true);
```
ihor-sviziev: I think it’s not real case when this method could return true

ihor-sviziev (changes requested):

app/code/Magento/SalesRule/Test/Unit/Model/ValidatorTest.php
aligent-lturner: I was following the format used for `getParentItemId` on the line above, which should be returning an int according to the api, but will change to be null, as you're right that this is what it would be returning.

app/code/Magento/SalesRule/Model/Validator.php
aligent-lturner: Will do :+1:

app/code/Magento/SalesRule/Model/Validator.php
aligent-lturner: Ah - thanks - forgot to remove that one.

app/code/Magento/SalesRule/Test/Unit/Model/ValidatorTest.php
aligent-lturner: Will change these too

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-6824](https://jira.corp.magento.com/browse/ENGCOM-6824) has been created to process this Pull Request

engcom-Bravo:
:heavy_check_mark: QA Passed

After switching to PR, the discount amount became 1$ instead of 0.50$
discount_ammount1

Issue #25490, opened 2019-11-06
Magento\Framework\Api\SimpleDataObjectConverter->camelCaseToSnakeCase has no concept of numbers 


antbates91:
### Preconditions (*)

1. Magento 2.3.3

### Steps to reproduce (*)

1. Create a method for an object that contains a number at the end of the method name for example `getMondayWeek1`. This resolves to a db field `monday_week_1`.
2. Call `Magento\Framework\Api\ExtensibleDataObjectConverter`'s `toNestedArray` function passing in the interface data, and the data model.
3. Check the output of array properties after this method has executed

### Expected result (*)

1. `toNestedArray` should return a nested array with the correct object keys `monday_week_1` 

### Actual result (*)

1. `toNestedArray` returns incorrect object keys `monday_week1`

This appears to be caused by `Magento\Framework\Api\SimpleDataObjectConverter`'s `camelCaseToSnakeCase` method which does `preg_replace` without any numbers in the pattern. `return strtolower(preg_replace('/(.)([A-Z])/', "$1_$2", $name));`

`return strtolower(preg_replace('/(.)([A-Z0-9])/', "$1_$2", $name));` would fix the issue

milindsingh:
I am working on this issue #mm20in

milindsingh:
I am closing this issue as it seems the fix may break things in an unexpected manner.

Earlier:
`defaultShipping1` -> `default_shipping1`
Now:
`defaultShipping1` -> `default_shipping_1`

So, If my column is default_shipping1 then it will not store data now.

The recommended snake case is `default_shipping1` which is already followed by Magento.

PR #26631, opened 2020-02-02
Number support added in SimpleDataObjectConverter

milindsingh:
### Related Pull Requests

Not merged for missing test coverage. https://github.com/magento/magento2/pull/25492

1. https://github.com/magento/magento2/issues/25490

1. Create a method for an object that contains a number at the end of the method name for example `getMondayWeek1`. This resolves to a db field `monday_week_1`.
2. Call `Magento\Framework\Api\ExtensibleDataObjectConverter`'s `toNestedArray` function passing in the interface data, and the data model.
3. Check the output of array properties after this method has executed

### Questions or comments

Is it intentionally done to ignore numbers?

### DIFF
--- a/lib/internal/Magento/Framework/Api/Test/Unit/Api/SimpleDataObjectConverterTest.php
+++ b/lib/internal/Magento/Framework/Api/Test/Unit/Api/SimpleDataObjectConverterTest.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\Framework\Api\Test\Unit\Api;
+
+use Magento\Framework\Api\SimpleDataObjectConverter;
+
+/**
+ * Unit test class for \Magento\Framework\Api\SimpleDataObjectConverter
+ */
+class SimpleDataObjectConverterTest extends \PHPUnit\Framework\TestCase
+{
+    public function testCamelCaseToSnakeCase()
+    {
+        $this->assertEquals("default_shipping", SimpleDataObjectConverter::camelCaseToSnakeCase("defaultShipping"));
+    }
+
+    public function testCamelCaseToSnakeCaseWithNumber()
+    {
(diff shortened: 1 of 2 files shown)


lib/internal/Magento/Framework/Api/Test/Unit/Api/SimpleDataObjectConverterTest.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function testSnakeCaseToCamelCase()
+    {
+        $this->assertEquals("defaultShipping", SimpleDataObjectConverter::snakeCaseToCamelCase("default_shipping"));
+    }
+
+    public function testSnakeCaseToCamelCaseWithNumber()
+    {
+        $this->assertEquals("defaultShipping1", SimpleDataObjectConverter::snakeCaseToCamelCase("default_shipping_1"));
```
lbajsarowicz, line 43: Could you add extra test case - for example `defaultShipping123` ?

lbajsarowicz (changes requested):
I feel like there's issue when the number is more than 1 digit.

lib/internal/Magento/Framework/Api/Test/Unit/Api/SimpleDataObjectConverterTest.php
milindsingh: Thanks @lbajsarowicz for pointing out the issue. I have updated the regex and add a few more test cases.

Though I am still worried about a thing!
Earlier:
 `defaultShipping1` -> `default_shipping1`
Now:
 `defaultShipping1` -> `default_shipping_1`

If my column is `default_shipping1` then it will not store data now. I think this may break the 3rd party modules?

milindsingh:
I am closing this PR as it seems the fix may break things in an unexpected manner.

Earlier:
`defaultShipping1` -> `default_shipping1`
Now:
`defaultShipping1` -> `default_shipping_1`

So, If my column is default_shipping1 then it will not store data now.

The recommended snake case is `default_shipping1` which is already followed by Magento.

lbajsarowicz:
Hm. Actually - I really liked your extra test cases - maybe could you reopen Pull Request and extend the Unit Tests with extra cases that you've had before.

Issue #23618, opened 2019-07-08
Invalid address id at check out


onepack:
### Preconditions (*)

1. Magento 2.3.2
2.

### Steps to reproduce (*)

1. Delete a customer and then re-create it with the same email.
2. Go to checkout with a product in the basket.
3. At sending press "next".

### Expected result (*)

1. The next step is supposed to load

### Actual result (*)

1. I get the error:  Unable to save shipping information. Please check input data.
2. Error log: [2019-07-08 10:33:00] main.CRITICAL: Invalid customer-address id 6 {"exception":"[object] (Magento\\Framework\\Exception\\NoSuchEntityException(code: 0): Invalid customer-address id 6 at /home/server/domains/dev3.domain.com/public_html/vendor/magento/module-quote/Model/QuoteAddressValidator.php:79)"} []

3. I removed the address from this customer and re-entered it. Ran the index etc. Nothing.
4. Other accounts are still ok with the checkout.
5. The table quote_address stays empty for this quote.

krishprakash:
@magento give me 2.3-develop instance

magento-engcom-team:
Hi @krishprakash. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @krishprakash, here is your [Magento instance](https://i-23618-2-3-develop.instances.magento-community.engineering).
Admin access: [https://i-23618-2-3-develop.instances.magento-community.engineering/admin](https://i-23618-2-3-develop.instances.magento-community.engineering/admin)
Login: `admin` Password: `123123q`
Instance will be terminated in up to 3 hours.

krishprakash:
Hi @onepack. 
I'm not able to reproduce this issue as provided steps by you.
Please update the reproducible steps and kindly confirm that its reproducible on Vanilla Magento instance.
Thanks.

onepack:
Hi @krishprakash ,

I tried to reproduce the same but was not able to.
The difference now is that as I remember there had been an update via an account import on this account in between.
I will need to check that.

But for now, is there anyone that can explain what is happening to that account and why I get this error on it?

krishprakash:
Please feel free to reopen the issue if you reproduce on venilla instance.
Thanks

ghost:
I have this issue randomly on 2.3.1

Test customer account with several saved addresses from testing the checkout, try to use one to checkout today, "Invalid customer address ID ..."

Sherbieny:
[...]

PR #26637, opened 2020-02-02
MAG-141191-23618: Fixing issue with Invalid customer id

shrma-smit:
### Description (*)

Invalid customer ID issue coming on the checkout while placing the order. This issue comes randomly for the customers who are logged in on the website. This issue is related to the customer quote, the table quote_address stays empty for this quote.

1. magento/magento2#23618: Invalid address id at check out

1. Add a product in the cart and proceed to checkout.
2. Click on Place order.
3. This issue is coming for the logged in customers randomly on the checkout.


app/code/Magento/Quote/Model/QuoteAddressValidator.php
```diff
@@ -127,6 +127,6 @@ public function validate(AddressInterface $addressData)
      */
     public function validateForCart(CartInterface $cart, AddressInterface $address): void
     {
-        $this->doValidate($address, $cart->getCustomerIsGuest() ? null : $cart->getCustomer()->getId());
+        $this->doValidate($address, $cart->getCustomer()->getId() ? $cart->getCustomer()->getId() : null);
```
aleron75: Hello @sumitCueBlocks, thanks for your submission!
In this case, to reduce code redundancy, you can use the `?:` (Elvis) operator, thus the code would become:

`$this->doValidate($address, $cart->getCustomer()->getId() ?: null);`

My best

shrma-smit: Hi @aleron75,

Thank you for your suggestion, I've updated the function code as you've suggested to reduce code redundancy.

Regards
Sumit

aleron75 (dismissed):

magento-engcom-team:
[ENGCOM-7215](https://jira.corp.magento.com/browse/ENGCOM-7215) has been created to process this Pull Request
:eight_spoked_asterisk: @aleron75, could you please add one of the following labels to the Pull Request?

| Label | Description |

magento-engcom-team:
@shrma-smit thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

aleron75:
Hello @shrma-smit, 

due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html), all code must be covered by tests. 

For this specific case, you should cover your fix by automated tests with the scenario which leads to an issue. 
Tests should fail on the mainline and pass with your fix.

To answer the question _"which kind of tests should I write"_, the answer is:
pick the most lightweight (execution time-wise) test type that will provide sufficient coverage.

Since in Magento codebase there is already a unit test for this class, I would recommend adding a test case in `\Magento\Quote\Test\Unit\Model\Quote\Address\ValidatorTest`. 

Thank you again!

engcom-Alfa (changes requested):
https://github.com/magento/magento2/pull/26637#issuecomment-608299232

shrma-smit (commented):
Changes reviewed.

VladimirZaets:
Hi @shrma-smit. Thanks for collaboration. As @aleron75 already mentioned, due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html) all code must be covered by tests. Please cover your fix by automated tests.

j-lloyd-slc:
This seems to be an ongoing problem in 2.3.4 because it assumes that if `quote`.`customer_is_guest` = 1, `quote`.`customer_id` = NULL.

I have found that this is not so, and when it is not, the subsequent validation fails and blocks checkout.

Granted, the root of the problem is that quote records are being populated with contradictory data. (Actually, the root of the problem is that quote records have two columns that can contradict each other at all). But as things are, this dysfunctional validation leaves customers in a state that they can not intuitively recover from.

srenon:
@engcom-Alfa @leron75 using this PR will still cause data inconsistency 

For example there will not be a hyperlink to go directly to the customer record from sales order view
image

image

I think it would be better to set the correct values before submitting  a quote 

vendor/magento/module-quote/Model/QuoteManagement.php

Index: vendor/magento/module-quote/Model/QuoteManagement.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
---  vendor/magento/module-quote/Model/QuoteManagement.php	(date 1589668946000)
+++  vendor/magento/module-quote/Model/QuoteManagement.php	(date 1589668946000)
@@ -494,6 +494,12 @@
      */
     protected function submitQuote(QuoteEntity $quote, $orderData = [])
     {
+        if ($quote->getCustomer()->getId()) {
+            $quote->setCustomerIsGuest(0);
+        } else {
+            $quote->setCustomerIsGuest(1);
+        }
+
         $order = $this->orderFactory->create();
         $this->submitQuoteValidator->validateQuote($quote);
         if (!$quote->getCustomerIsGuest()) {

engcom-Kilo:
@shrma-smit 
Can not reproduce this bug.

Please update descriptions,
add step by step instruction for reproduce.

engcom-Kilo:
Hi @shrma-smit, I'm closing this PR now due to inactivity.

jas8522:
I have reproduced this bug after upgrading from 2.2.9 to 2.4.0. It likely only occurs when there has been some kind of inconstency in the database, so how are you going to expect anyone to just reproduce it normally? Shouldn't the code provided here *still* be applied just in case anyone else experiences such an inconsistency?

jackrevate:
Still happens in 2.4.3-p1 - unreal the amount of bugs with Magento. I just want a basic functioning Ecomm store! 

Doing the below fixed it, but is there anything else that needs to be done?

https://github.com/magento/magento2/pull/26637/files

File code to change:
app/code/Magento/Quote/Model/QuoteAddressValidator.php

Find:
$this->doValidate($address, $cart->getCustomerIsGuest() ? null : $cart->getCustomer()->getId());

Replace with:
$this->doValidate($address, $cart->getCustomer()->getId() ?: null);

alitopaloglu:
This happens when your cart's

- customer_id is not null
- but customer_is_guest=1

so, this should not happen if your quote data is consistent. 

Better to search for any other bug in 3rd parties that causes this inconsistency.

PR #26651, opened 2020-02-02
Fixed php version issue in transalation

shikhamis11:
Fixed php version issue in transalation

**Preconditions (*)**
magento 2.3.3 and above 

**Steps to reproduce (*)**
setup website with multiple language like English and spanish 
and edit product

**Expected result (*)**
product should be able to edit

**Actual result (*)**
error in translation 

**Additional Informations** 
 JSON_HEX_TAG (integer)
    All < and > are converted to \u003C and \u003E. Available as of PHP 5.3.0.
 JSON_INVALID_UTF8_IGNORE (integer)
    Ignore invalid UTF-8 characters. Available as of PHP 7.2.0.

we can refer https://www.php.net/manual/en/json.constants.php

### DIFF
--- a/lib/internal/Magento/Framework/Serialize/Serializer/JsonHexTag.php
+++ b/lib/internal/Magento/Framework/Serialize/Serializer/JsonHexTag.php
@@ -26,7 +26,7 @@ class JsonHexTag extends Json implements SerializerInterface
      */
     public function serialize($data): string
     {
-        $result = json_encode($data, JSON_HEX_TAG);
+        $result = json_encode($data, JSON_INVALID_UTF8_IGNORE);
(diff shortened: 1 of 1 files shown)


lenaorobei:
Hi @shikhamis11!

Im not sure that we can change `JsonHexTag` class because it will be semantically wrong. The description of the class says `Serialize data to JSON with the JSON_HEX_TAG option enabled`.

Another point is that `JSON_INVALID_UTF8_IGNORE` was added in PHP 7.2 and we still need to support 7.1.

Could you please create the issue with detailed steps to reproduce so we can work on the compatible with Magento 2 requirements solution?

PR #26661, opened 2020-02-02
magento/magento2#25619: REST API not same behaviour of searchCriteria…

chirag-d-wagento:
### Description (*)

fixed issue by adding conditional code to treat page number 0 and 1 same in search endpoint

1. magento/magento2#25619: REST API not same behaviour of searchCriteria[currentPage] in different endpoints 

1. Do an API product paginated search query with `/V1/products/searchCriteria[currentPage]=0&searchCriteria[pageSize]=20`
2. Look at returned JSON content.
3. Do an API product GET with `/V1/products/searchCriteria[currentPage]=1&searchCriteria[pageSize]=20`
4. Look at returned JSON content
5. Do an API product paginated search query with 
`/V1/search/?searchCriteria[requestName]=quick_search_container&searchCriteria[filter_groups][0][filters][0][field]=search_term&searchCriteria[filter_groups][0][filters][0][value]=sample&searchCriteria[page_size]=20&searchCriteria[current_page]=0`
6. Look at returned JSON content.
7. Do an API product GET with 
`/V1/search/?searchCriteria[requestName]=quick_search_container&searchCriteria[filter_groups][0][filters][0][field]=search_term&searchCriteria[filter_groups][0][filters][0][value]=sample&searchCriteria[page_size]=20&searchCriteria[current_page]=1`
8. Look at returned JSON content

### DIFF
--- a/lib/internal/Magento/Framework/Search/Search.php
+++ b/lib/internal/Magento/Framework/Search/Search.php
@@ -69,7 +69,7 @@ class Search implements SearchInterface
             }
         }
 
-        $this->requestBuilder->setFrom($searchCriteria->getCurrentPage() * $searchCriteria->getPageSize());
+        $this->requestBuilder->setFrom(($searchCriteria->getCurrentPage() > 0 ? $searchCriteria->getCurrentPage() - 1 : 0) * $searchCriteria->getPageSize());
         $this->requestBuilder->setSize($searchCriteria->getPageSize());
 
         /**


lib/internal/Magento/Framework/Search/Search.php
```diff
@@ -69,7 +69,7 @@ public function search(SearchCriteriaInterface $searchCriteria)
             }
         }
 
-        $this->requestBuilder->setFrom($searchCriteria->getCurrentPage() * $searchCriteria->getPageSize());
+        $this->requestBuilder->setFrom(($searchCriteria->getCurrentPage() > 0 ? $searchCriteria->getCurrentPage() - 1 : 0) * $searchCriteria->getPageSize());
```
rogyar, line 72: Hi @chirag-d-wagento. Thank you for your collaboration.

According to the original issue, I'm not sure that changing the `current page` to the `current page - 1` for all cases except page 0 will not bring additional issues. 
Correct me if I'm wrong, but the problem we need to solve here is getting the same results for `current page = 0` and `current page = 1`. So, from my understanding, we need to decrease the `current page` value only in case if the `current page = 1` but not fo all positive values of `current page`

rogyar (changes requested):

lib/internal/Magento/Framework/Search/Search.php
engcom-Charlie: Hi @rogyar, I debug this PR. If we decrease the current page value only in case if the **current_page** = 1 we will get next behavior (for example we have 3 products) :

- In case we pass **current_page** 0 and _page_size_ 2, we will get products from 0 to 2
- In case we pass **current_page** 1 and _page_size_ 2, we will get also products from 0 to 2
- In case we pass **current_page** 2 and _page_size_ 2, we will get products from 4 to 6(we will not get our last product)

sivaschenko:
Closed due to inactivity

PR #26684, opened 2020-02-03
Move additional dependencies from private getters to constructor - Magento_PageCache

Bartlomiejsz:
### Description (*)

This PR moves additionally introduced dependencies from private getter methods into constructor in Magento_PageCache module.


app/code/Magento/PageCache/Observer/FlushCacheByTags.php
```diff
@@ -21,78 +30,59 @@ class FlushCacheByTags implements ObserverInterface
...
-    {
+    public function __construct(
+        Config $config,
+        Cache $cache,
+        Type $fullPageCache,
+        Resolver $tagResolver
+    ) {
         $this->_config = $config;
         $this->_cache = $cache;
+        $this->fullPageCache = $fullPageCache;
```
rogyar, line 63: Hi @Bartlomiejsz. We can consider this change as backward incompatible. If some other class inherits the current class in some Magento based project, we will have an error upon calling `parent::construct(..)` in a child class because the child class is not supposed to pass `$fullPageCache` and ` $tagResolver` arguments.

What I suggest here is to instantiate the corresponding object if it's not being passed as an argument. See the [following](https://github.com/magento/magento2/blob/f706d81f0ca5a7802ddb02484f6a16775c3106a5/app/code/Magento/Cms/Controller/Adminhtml/Page/PostDataProcessor.php#L55) instantiation as an example.

app/code/Magento/PageCache/Observer/FlushAllCache.php
```diff
@@ -21,48 +28,42 @@ class FlushAllCache implements ObserverInterface
...
-    public function __construct(\Magento\PageCache\Model\Config $config, \Magento\Framework\App\PageCache\Cache $cache)
-    {
+    public function __construct(
+        Config $config,
+        Cache $cache,
+        Type $fullPageCache
+    ) {
         $this->_config = $config;
         $this->_cache = $cache;
+        $this->fullPageCache = $fullPageCache;
```
rogyar, line 52: Please, check my previous comment regarding the backward compatibility

app/code/Magento/PageCache/Observer/FlushAllCache.php
```diff
@@ -1,18 +1,25 @@
 <?php
 /**
- *
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
+declare(strict_types=1);
```
rogyar, line 6: I would not recommend declaring strict types for an object that may already be inherited by another object in some custom implementation. By doing that we can break many existing Magento based systems. Declaring strict_types for new classes is 100% must, but not for the existing classes.

rogyar (changes requested):

app/code/Magento/PageCache/Observer/FlushCacheByTags.php
Bartlomiejsz: Hi @rogyar, I asked similar question in one of my previous PRs, such non-backward compatible changes I'm doing in those PRs are based on info I received in https://github.com/magento/magento2/pull/26269, where I was asked to replace backward-compatible change with one that doesn't use ObjectManager, since we're developing for 2.4 now.
Also such changes are being directly merged to core by Magento core team https://github.com/magento/magento2/commit/4e6f6ae42cc6109a02d35d0363b1cad0bbac66e1

app/code/Magento/PageCache/Observer/FlushAllCache.php
Bartlomiejsz: Hi @rogyar, please check my response to previous comment, it is not clearly described if changes to 2.4-develop should be backward-compatible or not, but I was asked in previous PR to choose clean code over backward-compatibility, so not sure what to do with that

app/code/Magento/PageCache/Observer/FlushCacheByTags.php
rogyar: Hi @Bartlomiejsz. You are right, that's fair since we are editing a non-API class. That's rather a professional habit avoid bringing backward-incompatible changes :)

rogyar:
For QA. Looks like we have a couple of fluky functional tests since changes in the current PR are not related to order creation process somehow.

rogyar (approved):

magento-engcom-team:
[ENGCOM-6828](https://jira.corp.magento.com/browse/ENGCOM-6828) has been created to process this Pull Request

engcom-Delta:
:heavy_check_mark: QA passed

engcom-Charlie:
Failed functional test not related to the changes in this PR test fails in 2.4-develop with MSI modules.
image

Issue #26111, opened 2019-11-29
configurable_options uses disabled products to populate the options


lsv:
### Preconditions (*)

1. magento/module-configurable-product-graph-ql: 100.3.1

### Steps to reproduce (*)

1. Create Configurable product which contains two Color configurations, for example, "Color1" and "Color2". 
2. Disable the "Color2" configuration.
3. Perform query:
``` graphql
{
  products(
    filter: {
      sku: {
        eq: "Configurable"
      }
    }
  ) {
    items {
      id
      attribute_set_id
      name
      sku
      type_id
      categories {
        id
      }
      ... on ConfigurableProduct {
        configurable_options {
          id
          attribute_id
          label
          position
          use_default
          attribute_code
          values {
            value_index
            label
          }
          product_id
        }
        variants {
          product {
            id
            name
            sku
            attribute_set_id
          }
        }
      }
    }
  }
}
```
### Expected result (*)
Disabled option should not be returned.
``` graphql
{
  "data": {
    "products": {
      "items": [
        {
          "id": 4,
          "attribute_set_id": 4,
          "name": "Configurable",
          "sku": "Configurable",
          "type_id": "configurable",
          "categories": [
            {
              "id": 3
            }
          ],
          "configurable_options": [
            {
              "id": 1,
              "attribute_id": "93",
              "label": "Color",
              "position": 0,
              "use_default": false,
              "attribute_code": "color",
              "values": [
                {
                  "value_index": 4,
                  "label": "Color1"
                }
              ],
              "product_id": 4
            }
          ],
          "variants": [
            {
              "product": {
                "id": 5,
                "name": "Configurable-Color1",
                "sku": "Configurable-Color1",
                "attribute_set_id": 4
              }
            }
          ]
        }
      ]
    }
  }
}
```

### Actual result (*)
Disabled "Color2" option is returned in the `configurable_options` section.
``` graphql
{
  "data": {
    "products": {
      "items": [
        {
          "id": 4,
          "attribute_set_id": 4,
          "name": "Configurable",
          "sku": "Configurable",
          "type_id": "configurable",
          "categories": [
            {
              "id": 3
            }
          ],
          "configurable_options": [
            {
              "id": 1,
              "attribute_id": "93",
              "label": "Color",
              "position": 0,
              "use_default": false,
              "attribute_code": "color",
              "values": [
                {
                  "value_index": 4,
                  "label": "Color1"
                },
                {
                  "value_index": 5,
                  "label": "Color2"
                }
[...]

PR #26690, opened 2020-02-04
Added plugin configurable_product_graphql_option to validate disabled…

sedonik:
### Description (*)
This PR provides solution for soling issue https://github.com/magento/magento2/issues/26111
Added product_status option to the values option list.

1. magento/magento2#26111: configurable_options uses disabled products to populate the options #26111

### Preconditions (*)

1. magento/module-configurable-product-graph-ql: 100.3.1

### Steps to reproduce (*)

1. Create Configurable product which contains two Color configurations, for example, "Color1" and "Color2". 
2. Disable the "Color2" configuration.
3. Perform query:
``` graphql
{
  products(
    filter: {
      sku: {
        eq: "Configurable"
      }
    }
  ) {
    items {
      id
      attribute_set_id
      name
      sku
      type_id
      categories {
        id
      }
      ... on ConfigurableProduct {
        configurable_options {
          id
          attribute_id
          label
          position
          use_default
          attribute_code
          values {
            value_index
            label
          }
          product_id
        }
        variants {
          product {
            id
            name
            sku
            attribute_set_id
          }
        }
      }
    }
  }
}
```
### Expected result (*)
Disabled option should not be returned.
``` graphql
{
  "data": {
    "products": {
      "items": [
        {
          "id": 4,
          "attribute_set_id": 4,
          "name": "Configurable",
          "sku": "Configurable",
          "type_id": "configurable",
          "categories": [
            {
              "id": 3
            }
          ],
          "configurable_options": [
            {
              "id": 1,
              "attribute_id": "93",
              "label": "Color",
              "position": 0,
              "use_default": false,
              "attribute_code": "color",
              "values": [
                {
                  "value_index": 4,
                  "label": "Color1"
                }
              ],
              "product_id": 4
            }
          ],
          "variants": [
            {
              "product": {
                "id": 5,
                "name": "Configurable-Color1",
                "sku": "Configurable-Color1",
                "attribute_set_id": 4
              }
            }
          ]
        }
      ]
    }
  }
}
```

### Actual result (*)
Disabled "Color2" option is returned in the `configurable_options` section.
``` graphql
{
  "data": {
    "products": {
      "items": [
        {
          "id": 4,
          "attribute_set_id": 4,
          "name": "Configurable",
          "sku": "Configurable",
          "type_id": "configurable",
          "categories": [
            {
              "id": 3
            }
          ],
          "configurable_options": [
            {
              "id": 1,
              "attribute_id": "93",
              "label": "Color",
              "position": 0,
              "use_default": false,
              "attribute_code": "color",
              "values": [
                {
                  "value_index": 4,
                  "label": "Color1"
                },
                {
                  "value_index": 5,
                  "label": "Color2"
                }
              ],
              "product_id": 4
            }
          ],
          "variants": [
            {
              "product": {
                "id": 5,
                "name": "Configurable-Color1",
                "sku": "Configurable-Color1",
                "attribute_set_id": 4
              }
            }
          ]
        }
      ]
    }
  }
}
```


swnsma:
Hi @sedonik!

Thank you for your pull request.
I think this changes could solve some general problem with options which are disabled.
But in my opinion it will be better to keep solution in scope of GraphQl resolver.
I think the root cause is that we able to receive disabled children for configurable products in:
> \Magento\ConfigurableProductGraphQl\Model\Resolver\ConfigurableVariant::resolve

If dig a little bit deeper we may found that we need to prevent loading of disabled child products in collection:
> \Magento\ConfigurableProductGraphQl\Model\Variant\Collection::getChildProductsByParentId
> \Magento\ConfigurableProductGraphQl\Model\Variant\Collection::fetch
Probably, setup filter will be enough.
What do you think about this?
Could you try to implement this solution?

swnsma (changes requested):
Hi @sedonik!

Thank you for your pull request.
I think this changes could solve some general problem with options which are disabled.
But in my opinion it will be better to keep solution in scope of GraphQl resolver.
I think the root cause is that we able to receive disabled children for configurable products in:

> \Magento\ConfigurableProductGraphQl\Model\Resolver\ConfigurableVariant::resolve

If dig a little bit deeper we may found that we need to prevent loading of disabled child products in collection:

> \Magento\ConfigurableProductGraphQl\Model\Variant\Collection::getChildProductsByParentId
> \Magento\ConfigurableProductGraphQl\Model\Variant\Collection::fetch


Probably, setup filter will be enough.
What do you think about this?
Could you try to implement this solution?

sedonik:
Hi @swnsma ,

I've after some investigation I see that method 

> \Magento\ConfigurableProductGraphQl\Model\Variant\Collection::fetch

gets correct data from Database. The Collection has only one Option Product. Please see the screenshot.

Once it get Product It tryes to get Option Attributes for Product and returns two options as values.

> Magento\ConfigurableProductGraphQl\Model\Options\Collection::fetch()

I don know is it correct logic when options are used as values. 
So I think it needs to take look to this method and maybe add filter to AttributeCollection.
Could you advice?

swnsma:
Hi @sedonik!
Thank you for provided information.
Hm... I think we could try to limit the number of options which we receive from this collection.
Could you try and share results?

sedonik:
Hi @swnsma,

I've added method: getAllowedOptions which detects product status and returns an appropriate option as the value of an enabled product.
But for that, it needs to add 'sku' to options map of Configurable Product Attributes.

Could you advise? Thank you.

app/code/Magento/ConfigurableProductGraphQl/Model/Options/Collection.php
```diff
@@ -123,11 +133,31 @@ private function fetch() : array
...
+     * Returns List of Allowed Options
+     *
+     * @param $attributeData
+     * @return array
+     */
+    protected function getAllowedOptions($attributeData) {
+        $options = [];
+
+        foreach ($attributeData['options'] as $keyOption => $valOption) {
+            $configurableProduct = $this->productRepository->get($valOption['sku']);
```
swnsma: I afraid this approach can kill performance of the endpoint.
Maybe we should add plugin for GraphQL area which will remove disabled options from options select?
Maybe this will be right place to do so: \Magento\ConfigurableProduct\Model\ResourceModel\Attribute\OptionSelectBuilder::getSelect
@paliarush @akaplya we need your advice here.

swnsma (changes requested):

swnsma:
Hi @sedonik,
Can we just try to set limitation in select:
> \Magento\ConfigurableProduct\Model\ResourceModel\Attribute\OptionSelectBuilder::getSelect
and try what will happened?
You can use plugin only for `graphql` application area.

sedonik:
Hi @swnsma,
Ok, I will.

sedonik:
Hi @swnsma ,

I've added "after" plugin "configurable_product_graphql_option_select" for 
Magento\ConfigurableProduct\Model\ResourceModel\Attribute\OptionSelectBuilder::getSelect() method. 

Could you check this solution?

app/code/Magento/ConfigurableProductGraphQl/Model/Plugin/AttributeOptionSelect.php
```diff
@@ -0,0 +1,78 @@
...
+                    'entity_attr_value.value = ' . Status::STATUS_ENABLED
+                ]
+            ),
+            []
+        )->joinInner(
+            ['attribute_status' => $this->attributeResource->getTable('eav_attribute')],
+            implode(
+                ' AND ',
+                [
+                    'attribute_status.attribute_code = ' . self::STATUS_CODE,
```
swnsma: Please use constant \Magento\Catalog\Api\Data\ProductInterface::STATUS

app/code/Magento/ConfigurableProductGraphQl/Model/Plugin/AttributeOptionSelect.php
```diff
@@ -0,0 +1,78 @@
...
+     * @param AbstractAttribute $superAttribute
+     * @return mixed
+     */
+    public function afterGetSelect(
+        OptionSelectBuilder $subject,
+        $select,
+        AbstractAttribute $superAttribute
+    ) {
+        $select->joinInner(
+            ['entity_attr_value' => $superAttribute->getBackendTable()],
```
swnsma, line 49: I am not sure if use backend table of super attribute is right decision.
Maybe it's better to load attribute '**status**' and use it's ID and backend table directly, with only one join?

app/code/Magento/ConfigurableProductGraphQl/Model/Plugin/AttributeOptionSelect.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param OptionSelectBuilder $subject
+     * @param $select
+     * @param AbstractAttribute $superAttribute
+     * @return mixed
+     */
+    public function afterGetSelect(
+        OptionSelectBuilder $subject,
+        $select,
```
swnsma: Please add type hint for **select** and update PHPdoc.

app/code/Magento/ConfigurableProductGraphQl/Model/Plugin/AttributeOptionSelect.php
```diff
@@ -0,0 +1,78 @@
...
+        AbstractAttribute $superAttribute
+    ) {
```
swnsma, line 47: Please add return type hint for **\Magento\Framework\DB\Select** and update PHPdoc.

app/code/Magento/ConfigurableProductGraphQl/Model/Plugin/AttributeOptionSelect.php
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Magento\ConfigurableProductGraphQl\Model\Plugin;
+
+use Magento\Catalog\Model\Product\Attribute\Source\Status;
+use Magento\ConfigurableProduct\Model\ResourceModel\Attribute\OptionSelectBuilder;
+use Magento\ConfigurableProduct\Model\ResourceModel\Product\Type\Configurable\Attribute;
+use Magento\Eav\Model\Entity\Attribute\AbstractAttribute;
+
+/**
+ * Class AttributeOptionSelect
```
swnsma: I am not sure if this line provides any additional information.
Please remove it.

app/code/Magento/ConfigurableProductGraphQl/etc/graphql/di.xml
```diff
@@ -36,4 +36,7 @@
             </argument>
         </arguments>
     </type>
+    <type name="Magento\ConfigurableProduct\Model\ResourceModel\Attribute\OptionSelectBuilder">
+        <plugin name="configurable_product_graphql_option_select" type="Magento\ConfigurableProductGraphQl\Model\Plugin\AttributeOptionSelect" />
```
swnsma: I think **configurable_product_graphql_limit_disabled_options** will be more meaningful name for the plugin.
Also, please add a order for the plugin.

swnsma (changes requested):
Hi @sedonik !
Thank you for your update!
Please check my comments.

Could you also cover your fix with Integration or API functional tests?
It will be really great to have such changes covered with tests.

sedonik:
Hi @swnsma,
Thank you for reviewing. I will.

swnsma:
Hi @sedonik!

I am closing this PR due to inactivity.
If you would like to continue your work with this PR, please feel free to re-open it.

PR #26699, opened 2020-02-04
Fix: missing "null" and "false" return type hints in RouterInterface:…

gsomoza:
### Description (*)
The core code in FrontController expects a "falsy" value to be
returned from routers that don't match a request, but want to
delegate to the next router.

```
foreach ($this->_routerList as $router) {
    try {
        $actionInstance = $router->match($request);
        if ($actionInstance) { // <---- return type must evaluate to "false" for loop to continue
            $result = $this->processRequest();
            // ...
        }
    } catch (\Magento\Framework\Exception\NotFoundException $e) {
        // ...
    }
}
```

Similarly, the code in `Webapi\Rest\Router` expects
specifically a `false` for that same use-case.

However those concrete expectations are not reflected in the
RouterInterface::match return type hint (DocBlock) and are not
documented either. This creates problems with static code analysis like PHPStan.

This pull request:
* Adds the "null" and "false" return type hints to the RouterInterface::match()
DocBlock. This fixes issues with PHPStan and IDE's, and makes the behaviour more
explicit.
* Adds a unit test to FrontControllerTest to make sure specifically returning "null"
works as expected in FrontController.

Not necessary.

### DIFF
--- a/lib/internal/Magento/Framework/App/RouterInterface.php
+++ b/lib/internal/Magento/Framework/App/RouterInterface.php
@@ -14,10 +14,18 @@ namespace Magento\Framework\App;
 interface RouterInterface
 {
     /**
-     * Match application action by request
+     * Match application action by request.
+     *
+     * For web requests, return null if the router doesn't match the request
+     * but other routers might still match. Or to cause a 404, throw a
+     * \Magento\Framework\Exception\NotFoundException.
+     * See \Magento\Framework\App\FrontController.
+     *
+     * For Webapi Rest requests, return FALSE instead of null.
+     * See \Magento\Webapi\Controller\Rest\Router.
      *
      * @param RequestInterface $request
-     * @return ActionInterface
+     * @return ActionInterface|null|false
      */
     public function match(RequestInterface $request);
 }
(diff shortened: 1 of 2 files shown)


gsomoza:
FYI just signed the Adobe CLA @adobe/open-source-office

gsomoza:
I think SemVer Checker's recommendation to bump on MAJOR might not actually be accurate, for the following reasons:

1. We're only **adding** new return types to (A) an interface's DocBlock, that (B) should have had them all along (based on existing behaviour of core code): if anything, (C) it could be considered a small bugfix, because the two major pieces of code that use this interface in the core are in fact technically **misusing** the interface.
2. The changes are 100% backwards compatible with existing implementations of the interface, even for projects using PHPStan or similar.

If this is considered a bug fix then I'd be happy to submit a backport for 2.3.

lenaorobei (changes requested):
Hi @gsomoza. This is a problem of leaky abstraction. The interface should not care about implementation details in classes. 

The solution here should be class refactoring. It will be our step towards strict typing.

gsomoza:
That would be indeed a better fix. Is that refactoring already being worked on or are there plans to start it soon?

lenaorobei:
I believe this issue is not reported yet so I'm going to create a GitHub issue to address it.

lenaorobei:
Closing this PR with reference to the issue https://github.com/magento/magento2/issues/27250

PR #26727, opened 2020-02-06
fix based on issue #26693

shahednasser:
### Description (*)

Changes based on issue #26693 which stated problem when overriding Magento\Customer\Controller\Account\CreatePost. 
Just add after `$customer = $this->accountManagement ->createAccount($customer, $password, $redirectUrl);`
the following:
`$customer->save();` and the customer will be considered confirmed without actually confirming their account. This is caused by setting confirmation key to null without checking whether the email sent is a confirmation email or not in `sendEmailConfirmation` method in Magento\Customer\Model\AccountManagement.
This can be a problem when implementing custom create customer logic and you need to use the customer instance after using the `createAccount` method on `accountManagement`, especially if the customer is saved after. The workaround could be to get a new customer instance, but I think conditionally setting the confirmation key to null in `sendEmailConfirmation` method in account management would be better.

1. magento/magento2#26693: Overriding CreatePost can automatically confirm a customer

1. In Admin, turn on confirmation required through Stores - Configuration - Customers - Customer Configuration - Create New Account Options - Require Emails Confirmation
2. Override Magento\Customer\Controller\Account\CreatePost in a module. Copy everything in execute function as it is, but add after
`$customer = $this->accountManagement ->createAccount($customer, $password, $redirectUrl);`
the following:
`$customer->save()`
3.Open storefront as a guest and create account
4.Customer should not be confirmed until they access the confirmation link they received in their email.

### DIFF
--- a/app/code/Magento/Customer/Model/AccountManagement.php
+++ b/app/code/Magento/Customer/Model/AccountManagement.php
@@ -985,7 +985,9 @@ class AccountManagement implements AccountManagementInterface
                 $templateType = self::NEW_ACCOUNT_EMAIL_REGISTERED_NO_PASSWORD;
             }
             $this->getEmailNotification()->newAccount($customer, $templateType, $redirectUrl, $customer->getStoreId());
-            $customer->setConfirmation(null);
+            if($templateType != self::NEW_ACCOUNT_EMAIL_CONFIRMATION){
+                $customer->setConfirmation(null);
+            }
         } catch (MailException $e) {
             // If we are not able to send a new account email, this should be ignored
             $this->logger->critical($e);


app/code/Magento/Customer/Model/AccountManagement.php
```diff
@@ -985,7 +985,9 @@ protected function sendEmailConfirmation(CustomerInterface $customer, $redirectU
                 $templateType = self::NEW_ACCOUNT_EMAIL_REGISTERED_NO_PASSWORD;
             }
             $this->getEmailNotification()->newAccount($customer, $templateType, $redirectUrl, $customer->getStoreId());
-            $customer->setConfirmation(null);
+            if ($templateType != self::NEW_ACCOUNT_EMAIL_CONFIRMATION) {
```
ihor-sviziev, line 984: Could you replace it to strict comparison?
```suggestion
            if ($templateType !== self::NEW_ACCOUNT_EMAIL_CONFIRMATION) {
```

ihor-sviziev:
@magento run all tests

ihor-sviziev:
@shahednasser could you cover your changes with some kind of tests?

shahednasser:
@ihor-sviziev there's already testGetConfirmationStatus in Magento\Customer\Test\Unit\Model\AccountManagementTest do we need a more specific test for sendEmailConfirmation?

ihor-sviziev:
@shahednasser looks like testGetConfirmationStatus isn't testing this part. I see that method that you've changed used only in `createAccountWithPasswordHash`. I think it's better to cover this case with integration test, similar to
https://github.com/magento/magento2/blob/40a787632f36ca67f8fff8c451e1e8d1bdbaf9f6/dev/tests/integration/testsuite/Magento/Customer/Model/AccountManagement/CreateAccountTest.php#L262-L288

shahednasser:
@ihor-sviziev I have added a test to cover both confirmation required and not required cases.

ihor-sviziev:
@magento run all tests

shahednasser:
@magento run all tests

dev/tests/integration/testsuite/Magento/Customer/Model/AccountManagement/CreateAccountTest.php
```diff
@@ -287,6 +303,56 @@ public function testCreateNewCustomerWithPasswordHash(): void
...
+    public function testCreateNewCustomerConfirmation($expected, $is_confirmation_required): void
+    {
+        $website_id = 1;
+        $this->configWriter->save(
+            AccountConfirmation::XML_PATH_IS_CONFIRM,
+            $is_confirmation_required,
+            ScopeInterface::SCOPE_WEBSITES,
+            $website_id
+        );
+        $this->scopeConfig->clean();
```
ihor-sviziev, line 321: Writing to config inside test itself isn't recommended way, as it have to be cleaned.
Could you use build in feature (that cleans all changed configs automatically) for defining config values and split it to two test cases?
 https://devdocs.magento.com/guides/v2.4/test/integration/annotations/magento-config-fixture.html

shahednasser: Hello @ihor-sviziev ,

I know of the annotation, however, the annotation does not work for website-scope config and as you can see here the function the determines whether confirmation is required or not uses website scope:

https://github.com/magento/magento2/blob/4b75b15706170d487b2615dffa0215af825e614a/app/code/Magento/Customer/Model/AccountConfirmation.php#L54-L65

I've tried the annotation for current_store as well as default, both don't work for website scope. Do you have any other suggestions?

ihor-sviziev: I think value should be set on global scope, similar to this (prefixed with `default/`:
https://github.com/magento/magento2/blob/40a787632f36ca67f8fff8c451e1e8d1bdbaf9f6/dev/tests/integration/testsuite/Magento/Catalog/Block/Product/View/MultiStoreCurrencyTest.php#L34

shahednasser: I have already tried that and it works for SCOPE_TYPE_DEFAULT, however not for SCOPE_WEBSITES which isConfirmationRequired function uses.

ihor-sviziev: Interesting... usually website scope inherits all config values from default scope. Maybe in case of integration tests it works a bit incorrectly.

Ok, in this case let's leave it as is for now, but please fix permissions, variable names issues + define types for arguments in method

ihor-sviziev:
Now changes looks good. Let’s wait for test results

@magento run all tests

magento-engcom-team:
[ENGCOM-7819](https://jira.corp.magento.com/browse/ENGCOM-7819) has been created to process this Pull Request

magento-engcom-team:
@shahednasser thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

engcom-Alfa:
:heavy_check_mark: QA Passed

### Case 1. Creating customer with turn on confirmation required

**Manual testing scenario:**

1. In Admin, turn on confirmation required through **Stores - Configuration - Customers - Customer Configuration - Create New Account Options - Require Emails Confirmation**
2. Override Magento\Customer\Controller\Account\CreatePost in a module. Copy everything in execute function as it is, but add after
`$customer = $this->accountManagement ->createAccount($customer, $password, $redirectUrl);`
the following:
`$this->customerRepository->save($customer);`



3. Open storefront as a guest and create account;

**Before:** :heavy_multiplication_x: 1.Confirmation email is sent but no message is shown about required confirmation
                  :heavy_multiplication_x: Customer is logged in

Screenshot from 2020-07-13 15-16-50

**After:** :heavy_check_mark: A message is shown that the account requires confirmation by the link sent on email.
:heavy_check_mark: Customer is not logged in

Screenshot from 2020-07-13 15-23-55

### Case 2. Creating customer with turn off confirmation required

1. In Admin, turn off confirmation required through **Stores - Configuration - Customers - Customer Configuration - Create New Account Options - Require Emails Confirmation**
2. Override Magento\Customer\Controller\Account\CreatePost in a module. Copy everything in execute function as it is, but add after
`$customer = $this->accountManagement ->createAccount($customer, $password, $redirectUrl);`
the following:
`$this->customerRepository->save($customer);`



3. Open storefront as a guest and create account;

**After:** :heavy_check_mark: The customer was successfully created

Screenshot from 2020-07-13 16-02-19

PR #26754, opened 2020-02-07
#26708 Fix ORDER BY has two similar conditions by removing duplicated entries

jonatanrdsantos:
### Description (*)

Fixed the order by duplication problem.

1. magento/magento2#26708: ORDER BY has two similar conditions

1. We have products in category
2. Go to app/code/Magento/Catalog/Block/Product/ListProduct.php and add:

3. Go to page on category

### DIFF
--- a/app/code/Magento/Catalog/Block/Product/ListProduct.php
+++ b/app/code/Magento/Catalog/Block/Product/ListProduct.php
@@ -475,8 +475,6 @@ class ListProduct extends AbstractProduct implements IdentityInterface
             $layer->setCurrentCategory($origCategory);
         }
 
-        $this->addToolbarBlock($collection);
-
         $this->_eventManager->dispatch(
             'catalog_block_product_list_collection',
             ['collection' => $collection]


lenaorobei (commented):
I think we need to dig deeper here and understand the reason of duplicate values, but not just filtering them. I would suggest to debug and find the place where previous `orderBy` is being added.

joridos:
> I think we need to dig deeper here and understand the reason of duplicate values, but not just filtering them. I would suggest to debug and find the place where previous `orderBy` is being added.

He is called twice because `_beforeToHtml` calls `_getProductCollection` that calls `initializeProductCollection` that finally calls `addToolbarBlock` but in sequence on `_beforeToHtml` we call `addToolbarBlock` again. those changes are introduce in #11473

Back on sort order creation on `Magento\Catalog\Model\ResourceModel\Product\Collection` whe have `addAttributeToSort`  that uses `$this->getSelect()->order('cat_index.position ' . $dir);` to set order position, and not some abstract method like `$this->setOrder();`.

Digging deep I could see that `Magento\Framework\Data\Collection\AbstractDb` has a `_setOrder` that do this kind of validation:
```
unset($this->_orders[$field]);
        // avoid ordering by the same field twice
        if ($unshift) {
            $orders = [$field => $direction];
            foreach ($this->_orders as $key => $dir) {
                $orders[$key] = $dir;
            }
            $this->_orders = $orders;
        } else {
            $this->_orders[$field] = $direction;
        }
```
And because of on `addAttributeToSort` set order direct to the select query every time that `addToolbarBlock` the position get duplicated.

@lenaorobei do u have any suggeston?

lenaorobei:
Wow, that's interesting. Let's try to remove redundant `addToolbarBlock` and see what's happening.

joridos:
> Wow, that's interesting. Let's try to remove redundant `addToolbarBlock` and see what's happening.

I remove the `addToolbarBlock` from `initializeProductCollection` seens to fix the problem.

lenaorobei:
Please address unit test failures https://public-results-storage-prod.magento-testing-service.engineering/reports/magento/magento2/pull/26754/86ced6e099a8499721e97e4a81416337/Unit/allure-report-b2b/index.html#suites/4481bc3204d4175f5b821e4ecab3b89a/958ace50b39a8b84/

lenaorobei:
@joridos are you still up to finishing this PR?

joridos:
@lenaorobei Right now i'm with no free time work with this, sorry :(

lenaorobei:
@joridos I will have to close this PR so somebody else can take this issue. Feel free to reopen at any time if you want to continue working on it.

Issue #25124, opened 2019-10-17
Magento 2.3 Wrong product url for anchor categories for multiple storeviews


juharintanen:
### Preconditions (*)

1. Magento 2.3.2

### Steps to reproduce (*)

1. Regenerate new url rewrite for product

### Expected result (*)

1. All product urls with categories are genereted accordingly by correct store view

### Actual result (*)

1. Product urls that are generated from anchor categories are missing StoreId and are generated wrong.

Underlined $storeId is missing and needs to be added in code.

![image](https://user-images.githubusercontent.com/16240211/67034700-d5386080-f120-11e9-8da0-3556c8b4e6d4.png)

hostep:
This is most likely a duplicate of https://github.com/magento/magento2/issues/11615, but I'm leaving it open, because this problem definitely needs a lot more attention.

We are awaiting for many months now on https://github.com/magento/magento2/pull/20826 finally getting approved...

engcom-Delta:
@juharintanen thank you for your report. This this known issue and PR for it is already created. 
@hostep I mark this issue as duplicate and close it.  You could reopen it if PR won't be merged or issue will appears in the future

Duplicate of [#11615](https://github.com/magento/magento2/issues/11615)

slavvka:
Hi @juharintanen. Thank you for your report. 
The issue has been fixed in magento/magento2#26784 by @hostep in 2.4-develop branch
Related commit(s):
 - [cb685d15b2b590bdb4d99a3d4ebe563037d0aa31](https://github.com/magento/magento2/commit/cb685d15b2b590bdb4d99a3d4ebe563037d0aa31)
 - [acb8642c44d1ecb60ad023c81f6541fa9ebae47d](https://github.com/magento/magento2/commit/acb8642c44d1ecb60ad023c81f6541fa9ebae47d)
 - [97258e1a9ce97ad32971731eceac8fee32f66b74](https://github.com/magento/magento2/commit/97258e1a9ce97ad32971731eceac8fee32f66b74)
 - [6cf26dc19e2f80f0c2da445443d0b48a1974fc77](https://github.com/magento/magento2/commit/6cf26dc19e2f80f0c2da445443d0b48a1974fc77)
 - [0b941218dcc86dcbd518f5400934e995ad1f2e3c](https://github.com/magento/magento2/commit/0b941218dcc86dcbd518f5400934e995ad1f2e3c)
[...]

PR #22212, opened 2019-04-08
Add store id to retrieve localized content from category repository

mirkocesaro:
### Description
This commit solves an issue with the anchor category url generation.
To generate the anchor category url, data is retrieved by category repository without specifying  proper store id. Doing so, the generated url does not use the localized category name into request path.

### Manual testing scenarios
1. We assume we are working with Magento with one website and (at least) two storeviews:
  - store ID: 1 (default)
  - store ID: 2
2. Create an **anchor** category "test" (let say it has id 456) under Default Category with different url key for the storeviews:
  - url_key for store 1 : test-foo
  - url_key for store 2 : test-bar
3. Create a child category of "test" named "test child" and assign one product whose url_key is for exaple: "my-product" (let say id 123)

Expected result:
In the url_rewrite table we exepct to find the following rows:

```
entity_type|entity_id|target_path                             |request_path            |store_id|
-----------|---------|----------------------------------------|------------------------|--------|
product    |123      |catalog/product/view/id/123/category/456|test-foo/my-product.html|1       |
product    |123      |catalog/product/view/id/123/category/456|test-bar/my-product.html|2       |
```

Actual result:

```
entity_type|entity_id|target_path                             |request_path            |store_id|
-----------|---------|----------------------------------------|------------------------|--------|
product    |123      |catalog/product/view/id/123/category/456|test-foo/my-product.html|1       |
product    |123      |catalog/product/view/id/123/category/456|test-foo/my-product.html|2       |

```
as we can see the request_path value for the second row (store_id 2) is not as expected 

### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #20826, opened 2019-01-30
11615: Fix UrlRewriteGenerator for anchor categories

obuchowski:
A commit itself is pretty straightforward. Need to pass storeId param to category repository so resulted category has a correct url path.

### Description (*)
For not default stores adding a category for a product may result in wrong url rewrite for a product in category's anchor parent .
result
### Fixed Issues (if relevant)

1.  [#11615](https://github.com/magento/magento2/issues/11615)

### Manual testing scenarios (*)
1. Enable this 
config
2. Change url key of some anchor category in a specific (NOT DEFAULT) store 
key
3. add a product to a child category from a product form
product
Note that there is a correct result if a product is added from a category form.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #26784, opened 2020-02-09
[Forward Port PR-14344] Fix generating product URL rewrites for anchor categories

hostep:
(cherry picked from commit 63111ed399b2e058efc0bf1c7c5427299cb8c5fc)

### Description (*)
This is a forward port of https://github.com/magento/magento2/pull/14344 to 2.4-develop
This replaces https://github.com/magento/magento2/pull/20826 which is stuck because the author is no longer responding for a few months.
Previous attempts at fixing this issue were done in https://github.com/magento/magento2/pull/7667 & https://github.com/magento/magento2/pull/22212

**FOR THE MAINTAINERS**: if this gets approved, can you please please please backport this to Magento 2.3, because the change is already in Magento 2.2, and by this PR also in 2.4, so it should also be part of 2.3, so we have consistent behavior over all these major Magento versions.

### Related Pull Requests
- https://github.com/magento/magento2/pull/7667
- https://github.com/magento/magento2/pull/14344
- https://github.com/magento/magento2/pull/20826
- https://github.com/magento/magento2/pull/22212

1. https://github.com/magento/magento2/issues/4112: Wrong parent category url_key in URL
2. https://github.com/magento/magento2/issues/11615: URL Rewrites vs multiple storeviews - a never ending battle
3. https://github.com/magento/magento2/issues/11616: URL Rewrites vs multiple storeviews - too many rewrites are being generated
4. https://github.com/magento/magento2/issues/25124: Magento 2.3 Wrong product url for anchor categories for multiple storeviews
5. https://github.com/magento/magento2/issues/26393: Product category url rewrite missing storeview specific url_key
6. Probably some more ...

1. See steps to reproduce in https://github.com/magento/magento2/issues/11615 and https://github.com/magento/magento2/issues/11616


rogyar:
Hi @hostep. It would be awesome to have this fix finally merged. But considering the number of talks/attempts, it will be fair to cover this fix using a functional test. Could I kindly ask you to do that? 

Thank you!

hostep:
I agree @rogyar!

Due to me not having much time free and having very little experience with writing tests, somebody else may pick this up if they want.
Maybe I'll have some more time later this week, we'll see if it happens or not from my part.

hostep:
Thanks for the test @engcom-Golf!

But I'm confused, since it doesn't seem to be testing the generated url rewrites of products belonging to categories of at least 2 levels deep.
Am I missing something obvious here?

Also: why is the parent_id, path & level of those categories with ids 3 & 4 different on a different storeview, that seems like a mistake?

engcom-Echo:
Hi @hostep. Thanks for your comment.

The test checks this code if we use the parameters of $storeId.

> $this->categoryRepository->get($anchorCategoryId, $storeId);

These fixtures are a little scary but in my opinion, we should check the code for different use $storeId - that's exactly what I saw an important change.

hostep:
Ok thanks for the feedback!

In my opinion, we should also test if the generated url_rewrites are correct, as that's the purpose of this PR, see https://github.com/magento/magento2/issues/11615 for what the correct values should be in this case.

Could we add a test for that as well? Thanks!

engcom-Echo:
I will take care of functional test coverage.

hostep:
Thanks for the MFTF test @engcom-Echo!

I think I understand how it works. But just to be clear: this test is failing on the code as it was before this fix, right?

And shouldn't we also test the `url_key_custom_store`? Or is that not necessary? Something like:
```xml
<amOnPage url="{{category.url_key_custom_store}}/{{subCategory.url_key_custom_store}}/{{product.urlKey}}2.html" stepKey="goToProductPage"/>
<see selector="{{StorefrontProductInfoMainSection.productName}}" userInput="{{product.name}}" stepKey="seeProductNameInStoreFront"/>
```
(I'm not sure how the url is build when you should be on a different store in MFTF though)

Thanks!

engcom-Echo:
@magento run all tests

engcom-Echo:
@magento run all tests

engcom-Echo:
@magento run all tests

app/code/Magento/CatalogUrlRewrite/Test/Mftf/ActionGroup/StorefrontAssertProductRewriteUrlSubCategoryActionGroup.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ /**
+  * Copyright © Magento, Inc. All rights reserved.
+  * See COPYING.txt for license details.
+  */
+-->
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
+    <actionGroup name="StorefrontAssertProductRewriteUrlSubCategoryActionGroup">
```
rogyar: Hi @engcom-Echo. Could I ask you to rename the action group in order to follow the general naming rules, please? So we have

`AssertStorefrontProductRewriteUrlSubCategoryActionGroup`

in the result.

Please, check the [technical guidelines](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html) (part 11.3.5).

Thank you.

app/code/Magento/Catalog/Test/Mftf/ActionGroup/ChangeSeoUrlKeyForSubCategoryWithoutRedirectActionGroup.xml
```diff
@@ -0,0 +1,21 @@
...
+<!--
+ /**
+  * Copyright © Magento, Inc. All rights reserved.
+  * See COPYING.txt for license details.
+  */
+-->
+
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
+    <actionGroup name="ChangeSeoUrlKeyForSubCategoryWithoutRedirectActionGroup" extends="ChangeSeoUrlKeyForSubCategoryActionGroup">
```
rogyar: Please, use the area prefix (Admin|Storefront) for this action group. If we have all action groups prefixed correctly, it's much easier to find and reuse an action group in the future.

Thank you!

rogyar (dismissed):

app/code/Magento/CatalogUrlRewrite/Test/Mftf/ActionGroup/StorefrontAssertProductRewriteUrlSubCategoryActionGroup.xml
engcom-Echo: Thank you

engcom-Echo:
Hi @rogyar  could you please approve it? :)

rogyar (approved):
Thank you @engcom-Echo. You rock!

magento-engcom-team:
[ENGCOM-6997](https://jira.corp.magento.com/browse/ENGCOM-6997) has been created to process this Pull Request

dev/tests/integration/testsuite/Magento/CatalogUrlRewrite/Model/Product/AnchorUrlRewriteGeneratorTest.php
```diff
@@ -50,32 +44,49 @@ public function setUp()
...
+     * @throws \Magento\Framework\Exception\LocalizedException
+     * @throws \Magento\Framework\Exception\NoSuchEntityException
      * @magentoDataFixture Magento/CatalogUrlRewrite/_files/product_with_stores.php
      * @magentoDbIsolation disabled
-     *
-     * @return void
+     * @dataProvider getConfigGenerate
      */
-    public function testGenerate(): void
+    public function testGenerate(string $expect): void
```
lbajsarowicz, line 61: For what input should the result change?
I see expectation attribute, but there's not context attribute.

lbajsarowicz:
@mszydlo Today I wanted to forward this change, then realised that @hostep is doing it already for 21 days. Awesome work! I can't wait to get this merged.

engcom-Alfa:
:heavy_check_mark: QA Passed 

**Before:**
before

**After:**
after

slavvka:
@magento run Functional Tests EE

hostep:
Thanks @slavvka, do you know if this is going to be backported to Magento 2.3? Since https://github.com/magento/magento2/pull/14344 was merged in 2.2, we should probably also have this fix on 2.3, no?

hostep:
@slavvka, @chernenm, can you guys let me know if this is being backported to the 2.3.x release?

We have this awkward situation where this bug is fixed in Magento 2.2.10, 2.2.11 and 2.4.0 (I guess?), but the fix is not in 2.3.x ...

peterjaap:
@hostep seems like at least a part of this PR is present in 2.3.6; https://github.com/magento/magento2/blob/2.3.6/app/code/Magento/CatalogUrlRewrite/Model/Product/AnchorUrlRewriteGenerator.php#L70

hostep:
Ah nice find @peterjaap !

It was added in [MC-32810: [Magento Cloud] - Incorrect URL Rewrites for non-default websites after upgrade to 2.3.4](https://github.com/magento/magento2/commit/82c91868a2d47d9c052149da1908c5ce1903587b) (what a silly description, this has been a bug for many years)

However, it seems to consist of two changes and only one of them is done in 2.4-develop not the other.

Fix one: [2.3.6](https://github.com/magento/magento2/blob/2.3.6/app/code/Magento/CatalogUrlRewrite/Model/CategoryProductUrlPathGenerator.php#L47) vs [2.4-develop](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/CatalogUrlRewrite/Model/CategoryProductUrlPathGenerator.php#L46-L48) => missing in 2.4
Fix two: [2.3.6](https://github.com/magento/magento2/blob/2.3.6/app/code/Magento/CatalogUrlRewrite/Model/Product/AnchorUrlRewriteGenerator.php#L70) vs [2.4-develop](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/CatalogUrlRewrite/Model/Product/AnchorUrlRewriteGenerator.php#L70) => both in 2.3 and 2.4

So this is again a weird thing and maybe somebody forgot to do something here.

@slavvka, @chernenm, @sdzhepa: can we figure this out please? Was that first change in `MC-32810` forgotten to be included in `2.4-develop`, or was this left out intentionally for some reason?

PR #26835, opened 2020-02-12
Unit test for Magento\Downloadable\Model\Sample\DeleteHandler

karyna-t:
### Description (*)
Unit test for Magento\Downloadable\Model\Sample\DeleteHandler

### DIFF
--- a/app/code/Magento/Downloadable/Test/Unit/Model/Sample/DeleteHandlerTest.php
+++ b/app/code/Magento/Downloadable/Test/Unit/Model/Sample/DeleteHandlerTest.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+declare(strict_types=1);
+
+namespace Magento\Downloadable\Test\Unit\Model\Sample;
+
+use Magento\Catalog\Api\Data\ProductInterface;
+use Magento\Catalog\Model\Product;
+use Magento\Downloadable\Api\Data\SampleInterface;
+use Magento\Downloadable\Api\SampleRepositoryInterface as SampleRepository;
+use Magento\Downloadable\Model\Product\Type;
+use Magento\Downloadable\Model\Sample\DeleteHandler;
+use Magento\Framework\TestFramework\Unit\Helper\ObjectManager as ObjectManagerHelper;
+use PHPUnit\Framework\MockObject\MockObject;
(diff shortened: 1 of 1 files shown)


app/code/Magento/Downloadable/Test/Unit/Model/Sample/DeleteHandlerTest.php
```diff
@@ -0,0 +1,100 @@
...
+use Magento\Framework\TestFramework\Unit\Helper\ObjectManager as ObjectManagerHelper;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Unit Test for \Magento\Downloadable\Model\Sample\DeleteHandler
+ */
+class DeleteHandlerTest extends TestCase
+{
+    const STUB_PRODUCT_TYPE = 'simple';
```
rogyar, line 26: Just a recommendation for the future to declare constant scope as `private`. It will reduce the chance of "temptation" to make another dependency on the current test in another test.

app/code/Magento/Downloadable/Test/Unit/Model/Sample/DeleteHandlerTest.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->deleteHandler = (new ObjectManagerHelper($this))->getObject(
+            DeleteHandler::class,
+            ['sampleRepository' => $this->sampleRepositoryMock]
+        );
+    }
+
+    /**
+     * Test case when provided Product has type Downloadable.
+     */
+    public function testExecuteWithDownloadableProduct()
```
rogyar, line 65: Another advice is to use return type hinting if we declare `strict types` in the future. So we have

```suggestion
    public function testExecuteWithDownloadableProduct(): void
```

rogyar (approved):

magento-engcom-team:
[ENGCOM-6887](https://jira.corp.magento.com/browse/ENGCOM-6887) has been created to process this Pull Request

app/code/Magento/Downloadable/Test/Unit/Model/Sample/DeleteHandlerTest.php
karyna-t: Thanks for the note, will keep it in mind!

engcom-Bravo:
:heavy_check_mark: QA Passed

The test is added and passes 
deleatehendler

PR #21796, opened 2019-03-16
issue #21193

MaksymGrom:
Fixed merging process in configs between global and scoped (frontend, adminhtml etc) areas.

### Description
Fixed merging process in configs between general and specific (frontend, adminhtml etc) areas. 
Magento has a lot classes with array node in constructor used as services pull.
Usage array_replace did replacing nodes on first level of constructor attributes, frontend or adminhtml di totally rewrite services pull node.
Developer want's to have possibility to register different services for different scopes (frontend, adminhtml) without loosing general config.

### Fixed Issue

1. magento/magento2#21193: Types not merged for etc/di.xml etc/adminhtml/di.xml
2. magento/magento2#20891: Fix issue with di.xml config merging for classes array argument in case of different sources and the class name difference in leading slash.

### Manual testing scenarios (*)

1. Create new module or use existing
2. Add in etc/adminhtml/di.xml
`

    <type name="Magento\Framework\View\Element\UiComponent\DataProvider\CollectionFactory">
        <arguments>
            <argument name="collections" xsi:type="array">
                <item name="cms_page_listing_data_source" xsi:type="string">Magento\Cms\Model\ResourceModel\Page\Grid\Collection</item>
                <item name="custom_listing_data_source_adminhtml" xsi:type="string">Magento\Sales\Model\ResourceModel\Order\Creditmemo\Order\Grid\Collection</item>
            </argument>
        </arguments>
    </type>
`
3. Open Sales -> Order page in admin
4. After fix, grid is opened correct.
Before fix, "Not registered handle sales_order_grid_data_source" error was instead grid

### Contribution checklist
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #26851, opened 2020-02-12
MC-21193: Renamed the class CollectionFactory to CollectionPool

AleksLi:
Renamed the class CollectionFactory to CollectionPool everywhere

### Description (*)

I found the reason for this. The class name `\Magento\Framework\View\Element\UiComponent\DataProvider\CollectionFactory` created new class every time and not merged all the collections set before. I had to rename this class to CollectionPool to fix this issue. 

1. magento/magento2[#21193](https://github.com/magento/magento2/issues/21193): Types not merged for etc/di.xml etc/adminhtml/di.xml

- Do everything as described in the issue.

### DIFF
--- a/app/code/Magento/AsynchronousOperations/etc/di.xml
+++ b/app/code/Magento/AsynchronousOperations/etc/di.xml
@@ -59,7 +59,7 @@
             </argument>
         </arguments>
     </type>
-    <type name="Magento\Framework\View\Element\UiComponent\DataProvider\CollectionFactory">
+    <type name="Magento\Framework\View\Element\UiComponent\DataProvider\CollectionPool">
         <arguments>
             <argument name="collections" xsi:type="array">
                 <item name="bulk_listing_data_source" xsi:type="string">Magento\AsynchronousOperations\Ui\Component\DataProvider\SearchResult</item>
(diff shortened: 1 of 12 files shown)


ihor-sviziev (changes requested):
@AleksLi your change is not backward compatible and makes fails for many tests, so it make impossible to accept such changes without real need.

It seems to me that renaming class is just workaround for the issue. Could you debug deeper or explain if you already did why this issue not appearing if you rename class and starting reproducing with old name?

ihor-sviziev:
@AleksLi I see that there were another trials to fix the issue https://github.com/magento/magento2/pull/20891 https://github.com/magento/magento2/pull/21796 but they were not merged due to backward compatibility issues. Maybe it will help with debug

AleksLi:
The Factory itself is the type of classes that generates once per instance. So I tried to debug the issue and find any other explanation why the configuration in other di.xml files were not merged but it seems like the file name is the key problem here. 
Magento thinks that this is the FACTORY class and setting the latest instance of it without any updates of $collections variable. only the last one. 

But if you think there is another solution for this I will dig dipper to find it.

ihor-sviziev:
@AleksLi I just looked on the object manager classes - i didn't found any special cases for factories.

It seems to me as issue with merging constructor arguments. Don't you think so?

AleksLi:
The reason is not in the file name. I found the exact line where the config of Global DI.xml would be overridden by the local di.xml file.

`lib/internal/Magento/Framework/ObjectManager/Config/Config.php:248` _mergeConfiguration method. Replacing the old configuration with the new one.

PR #26892, opened 2020-02-16
Search query throwing error in all case. #26520.

novikor:
### Description (*)

Execute full reindex at the very end of the Magento installation process.

1. magento/magento2#26520: Search query throwing error in all case.

See #26520 
### Questions or comments
See https://magentocommeng.slack.com/archives/CBSL1DF8B/p1581588239024600


rogyar:
Hi @novikor. Thank you for the fix. The issue is really annoying. 

I usually handle this issue in the similar way. However, I didn't have a chance to check what particular index causing the issue. So, If we will figure out what index causes the problem we may reindex only one index. In that case, we will have a little bit more optimal solution from the platform installation performance perspective.

novikor:
Hi, @rogyar.
We have discussed this issue in the last Group Code review meeting. It is caused by `catalogsearch_fulltest` index tables missing. I cannot see any performance problems there: the full reindex takes not more than 1 second with empty database. Moreover, it seems that despite it not described anywhere, the reindex is a required step of Magento installation.

rogyar:
Hi @novikor. I see your point. It makes sense. Since most of the index tables are not being populated automatically after the platform installation, we might have a similar problem with another index in the future. So, it's a good idea to perform the full reindex.

novikor:
@magento run all tests

PR #26920, opened 2020-02-19
Fix checkout password field visibility incorrectly when user email was not registered

mrtuvn:
### Description (*)

Add additional check in javascript for visibility password field

1. magento/magento2#26903: Guest visitor e-mail in checkout kicks off password field visibility

1. Add product to cart as guest
2. Add non registered email to user email
3. Go to homepage or cart page
4. Switch back to checkout one more

### Questions or comments

If this approach fine i can add some test on this case
Maybe i'm missing some cases. You guys feel free  please correct me. Very appreciate


app/code/Magento/Checkout/view/frontend/web/js/view/form/element/email.js
```diff
@@ -192,7 +192,20 @@ define([
...
         resolveInitialPasswordVisibility: function () {
-            if (checkoutData.getInputFieldEmailValue() !== '' && checkoutData.getCheckedEmailValue() === '') {
+            // If customer already logged in so we will hide password field
+            if (checkoutData.getInputFieldEmailValue() !== '' &&
+                customer.isLoggedIn()) {
+                return false;
+            }
+
+            // In this case user put email but not logged in yet password field should not show
+            if (checkoutData.getInputFieldEmailValue() !== '' &&
```
omiroshnichenko, line 202: Are you sure that we need to hide the password field when the user puts an email to the email field and the user is not logged in?

speedupmate: this offered fix will not solve the issue and probably creates more issues . if customer is logged in theres never a e-mail field to fill anyway so the first condition is also bit odd 

if field is filled (like on page load from cache) it needs to be validated against registration status all over again . Customer could go back register on another window etc and reload the page and your check would fail. 

Also try to dive in to the other issue referenced that created the issue , it probably needs to cover the issue reported there as well.

PR #26934, opened 2020-02-19
25038 Only set collection to toolbar once in ListProduct block

wigman:
Fix for: https://github.com/magento/magento2/issues/25038

When using ElasticSearch, pagination does not work in some cases if pagination is set multiple times (in the toolbar).
The catalog does not show results when clicking past the first page of results in catalog.

Only setting the $collection once fixes the issue.

### DIFF
--- a/app/code/Magento/Catalog/Block/Product/ListProduct.php
+++ b/app/code/Magento/Catalog/Block/Product/ListProduct.php
@@ -516,7 +516,9 @@ class ListProduct extends AbstractProduct implements IdentityInterface
             $toolbar->setModes($modes);
         }
         // set collection to toolbar and apply sort
-        $toolbar->setCollection($collection);
+        if (!$toolbar->getCollection()) {
+            $toolbar->setCollection($collection);
+        }
         $this->setChild('toolbar', $toolbar);
     }
 }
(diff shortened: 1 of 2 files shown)


app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -516,7 +517,9 @@ private function configureToolbar(Toolbar $toolbar, Collection $collection)
             $toolbar->setModes($modes);
         }
         // set collection to toolbar and apply sort
-        $toolbar->setCollection($collection);
+        if (!$toolbar->getCollection()) {
```
ihor-sviziev, line 520: I think it's not really logical to solve issues with initializing toolbar object in the client code. In case if someone use it in custom module - issue will appear again. I think it should be moved inside Toolbar class as internal logic.

I'm not totally sure what is the best solution would be there, but not setting collection not looks logically here.

Could you describe with more details how issue actually appears?

ihor-sviziev (changes requested):

ihor-sviziev (changes requested):
@wigman could you check my review above?

ihor-sviziev:
@wigman  I am closing this PR now due to inactivity.

Issue #26541, opened 2020-01-27
Order status pending payment not changing to processing after invoicing


shyam100379:
Order status is not getting updated from pending payment to processing once invoice is generated. When a customer choose payment option and click order, he/she will be redirected to payment gateway making that order as pending payment. Somehow the order is not fulfilled and order stays as pending payment.

### Preconditions (*)
1. Magento 2.3.2
2. Single store
3. Payment method which makes redirect to that corresponding payment gateway once order is clicked.

### Steps to reproduce (*)
1. Customer comes to checkout page and fills all needed information. 
2. Once clicking order, Magento sets pending payment as order status. 
3. Somehow customer didnt completed order, store user contact them personally and arrange cod or something, the invoice is generated.
4. Pending payment status not changing to processing or complete even after invoice and shipment is generated.

### Expected result (*)
1. Pending payment status need to change to processing once invoice is generated.
2. Processing turns complete once shipment is generated.

### Actual result (*)
1. Order status stays as pending payment even after shipment is generated

shyam100379:
Hi didnt anybody checked the mentioned issue. Its seems their is a bug in the system, order status from pending payment not getting updated even if that order is marked as complete. Please advise.

sedonik:
Hi

Cause I didn't reproduce moment when after Payment Getaway it lefts status "pending_payment'. I've only changed status in DB. Here is a screenshot where should be set (in item #26541).

![pending_payment](https://user-images.githubusercontent.com/5873715/74853150-2f9f3300-5346-11ea-88fa-f7412b0f9362.jpg)

aqdennis:
Facing the same problem in versions 2.3.2.  
We use Mollie Payments as our PSP and after successful payment status remains on Pending, hence neither we nor the client will receive a confirmation e-mail....
I've tried the fix as described in https://github.com/magento/magento2/issues/18148, but without result.

woutersamaey:
I also have a store facing this issue. The invoice is created + marked as paid. The total due amount drops to 0, but status stays "pending payment".

My case is with a custom built payment module that extends the now deprecated \Magento\Payment\Model\Method\AbstractMethod

Perhaps this is related somehow to the deprecation of this model?

ihor-sviziev:
Hi @woutersamaey,
This issue duplicates https://github.com/magento/magento2/issues/25659 that was already fixed.
I'm closing this issue as duplicate

PR #25876, opened 2019-12-02
Advance the order state to processing when a capture notification is received

azambon:
### Description (*)
When the registerCaptureNotificationCommand is executed, if the order is in state "new" or "pending payment", then its state is advanced to "processing", as it was up to Magento 2.2.5.
This should fix issues with orders remaining in "pending payment" state even though they were correctly paid. See issue #25659
Previously this command would set the state to "processing" only if the order did not already have a state set. This PR changes that check to do the same for "new" and "pending payment" orders too.
The unit test covering the command has been updated to include the new test cases.

### Fixed Issues (if relevant)
magento/magento2#25659: Paypal Payments Pro IPN keeping payments marked as Pending Payment

### Manual testing scenarios (*)
Prerequisite: a payment method must be installed and enabled that initially creates orders in state "new" or "pending payment" and then uses the "registerCaptureNotification" command to indicate when the payment is captured.

1. Place an order with the aforementioned payment method
2. When the payment is complete and the payment method has issued the registerCaptureNotification command, check the state of the generated order. Without this PR the state will still be "new" or "pending payment". With this PR the state becomes "processing"

PR #26935, opened 2020-02-19
Set STATE_PROCESSING when order is in STATE_PENDING_PAYMENT

sedonik:
### Description (*)

Added one more case when order is checked of status. It validates is status "pending_payment" then set "processing" status to order.

1. magento/magento2#26541: Order status pending payment not changing to processing after invoicing #26541

1. Cause I didn't reproduce moment when after Payment Getaway it lefts status "pending_payment'. I've only changed status in DB. Here is a screenshot where should be set (in item #26541). 
2. Create an invoice and submit.

### DIFF
--- a/app/code/Magento/Sales/Model/ResourceModel/Order/Handler/State.php
+++ b/app/code/Magento/Sales/Model/ResourceModel/Order/Handler/State.php
@@ -24,7 +24,9 @@ class State
     public function check(Order $order)
     {
         $currentState = $order->getState();
-        if ($currentState == Order::STATE_NEW && $order->getIsInProcess()) {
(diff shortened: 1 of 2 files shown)


ihor-sviziev:
Hi @sedonik,
The described issue was already fixed in https://github.com/magento/magento2/pull/25876.

I'm closing this PR as there is no issue anymore.

tuyennn:
@ihor-sviziev I think @sedonik is trying to create an invoice in Admin-end while the order is till pending_payment and expect this should be `processing` after the invoice created. I don't think Magento supported this unless the order is treated as offline payment such as has state: `new and status: `pending` for example.

PR #26972, opened 2020-02-21
Bugfix: Changed saving sorting order in RefundOrder class

### LINKED ISSUE #26975: Tests are failing when they are running at 00:00
### Summary

- Created a new PR: https://github.com/magento/magento2/pull/26972
- PR didn't include any DB changes, so these test should not fail. 
- Tests are failing due to them running and comparing results between 2 different dates: https://public-results-storage-prod.magento-testing-service.engineering/reports/magento/magento2/pull/26972/4b011829bf88bf9a4456d5585652a609/DatabaseCompare/ee-skip-sd-db-compare-report.html

### Examples

image

### Proposed solution

- Don't use current date when adding items to the DB.

dverkade:
### Description

Bugfix: Changed saving sorting order. For Magento Commerce a comment is set on the order when refunding to store credit. This comment will not be saved and the store credit refund will not be shown on the order page, because it's not saved as the order was already saved before saving the credit note.

### Version
Magento Commerce 2.3.5
Fixes: https://github.com/magento/magento2/issues/26975

### DIFF
--- a/app/code/Magento/Sales/Model/RefundOrder.php
+++ b/app/code/Magento/Sales/Model/RefundOrder.php
@@ -160,8 +160,8 @@ class RefundOrder implements RefundOrderInterface
                 : $this->config->getStateDefaultStatus($orderState);
             $order->setStatus($status);
 
-            $order = $this->orderRepository->save($order);
             $creditmemo = $this->creditmemoRepository->save($creditmemo);
+            $order = $this->orderRepository->save($order);
             $connection->commit();
         } catch (\Exception $e) {
             $this->logger->critical($e);
--- a/app/code/Magento/Sales/Model/RefundOrder.php
+++ b/app/code/Magento/Sales/Model/RefundOrder.php
@@ -19,7 +19,8 @@ use Magento\Sales\Model\Order\Validation\RefundOrderInterface as RefundOrderVali
 use Psr\Log\LoggerInterface;
 
 /**
- * Class RefundOrder
+ * Class is used for creating a credit memo and refunding it.
+ *
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 class RefundOrder implements RefundOrderInterface


pmclain (dismissed):
@dverkade Thank you for the contribution. Would you please cover the change with an integration test validating this behavior?

dverkade:
@pmclain I do not have the time to create an integration test. This needs some serious research from my side in order for me to create one. I would suggest accepting this PR since it solved the issue and creating another ticket for creating integration test for the whole RefundOrder class since there is not any test for that class.

dverkade:
@pmclain I've been thinking about this a little bit further. Creating an integration test is not feasible, since you'll need to have a event observing the creditmemo_save_after event, then changing something in the order and this will not get saved.

engcom-Kilo:
Working with MFTF test.

engcom-Kilo:
@dverkade 
Please update description, describe step by step for check this Bug. 

I am checked refund for Store credit with Magento B2B (latest version), 
and seems this works correctly. 
Comment is saved and showed on Credit memo view page.
Thanks.

dverkade:
@engcom-Kilo this has nothing to do with the B2B functionality. That's not mentioned anywhere by me. It's a commerce issue: 

Bugfix: Changed saving sorting order. For Magento Commerce a comment is set on the order when refunding to store credit. This comment will not be saved and the store credit refund will not be shown on the order page, because it's not saved as the order was already saved before saving the credit note.

So try to create a creditmemo with the refundorder class, set a comment and store credit refund. That data will not be shown on the order because the order was already saved before the creditmemo has been saved.

engcom-Kilo:
@dverkade 

> So try to create a creditmemo with the refundorder class, set a comment and store credit refund.

You mean using _refundorder class_ in custom extension(module)?

dverkade:
Hi @engcom-kilo,

Yes.

engcom-Kilo:
@dverkade
Please update Pull Request Description. 
Write Magento version number.
Better will be to describe step by step for reproduce bug.

Please provide example of your code where happened bug.

Thanks.

dverkade:
@engcom-Kilo in the `save` function in the file `Magento\CustomerBalance\Model\Creditmemo\Balance` it's doing all kind of things with the order class. See the following code snippet:

```
            $order = $creditmemo->getOrder();
            $order->setBsCustomerBalTotalRefunded(
                $order->getBsCustomerBalTotalRefunded() + $creditmemo->getBsCustomerBalTotalRefunded()
            );
            $order->setCustomerBalTotalRefunded(
                $order->getCustomerBalTotalRefunded() + $creditmemo->getCustomerBalTotalRefunded()
            );
            $order->setBaseCustomerBalanceRefunded(
                $order->getBaseCustomerBalanceRefunded() + $creditmemo->getBaseCustomerBalanceRefunded()
            );
            $customerBalanceRefunded = $creditmemo->getCustomerBalanceRefunded();
            $order->setCustomerBalanceRefunded(
                $order->getCustomerBalanceRefunded() + $customerBalanceRefunded
            );
            $status = $order->getConfig()->getStateDefaultStatus($order->getState());
            $comment = __(
                'We refunded %1 to Store Credit',
                $order->getBaseCurrency()->formatTxt($customerBalanceRefunded)
            );
            $order->addCommentToStatusHistory($comment, $status, false);
```

In the RefundOrder class the order is saved before the Order before the CreditMemo causing the changes the CustomerBalance does not to be saved.

sidolov:
@magento create issue

sidolov (dismissed):

magento-engcom-team:
[ENGCOM-8048](https://jira.corp.magento.com/browse/ENGCOM-8048) has been created to process this Pull Request

engcom-Charlie (changes requested):
@dverkade please cover your change by autotest. Could you provide a manual scenario to reproduce the issue?
Thank you.

dverkade:
Hi @engcom-Charlie,

The progress on this PR is appalling. There's no manual scenario as far as I'm aware. The code sample speaks for itself. Regarding the tests: I do not have the time to create an integration test. This needs some serious research from my side in order for me to create one. I would suggest accepting this PR since it solved the issue and creating another ticket for creating integration test for the whole RefundOrder class since there is not any test for that class.

It's up to @magento to add test to the whole class, since there are none anyway. 

Again, this needs to be accepted and merged. All the work on my side to research the issue, fix it and create the PR is all done.

engcom-Charlie:
@dverkade We are trying to understand what the problem is, but unfortunately we are not able to yet.

Peek 2020-09-22 13-48

dverkade:
Hi @engcom-Charlie,

Unfortunately, there are different ways to create a credit memo. In Magento there's a `RefundOrder` class which refunds an order. The "Create Credit Memo" functionality on the frontend does not use that class, so the issue is not reproducible on the frontend of Magento. 

If you use this class to refund an order and have Commerce edition with customer balance enabled it will not work correctly. As shown in the code snippet the `RefundOrder` class saves the Order before the credit memo is saved. Any observers, such as the `Magento\CustomerBalance\Model\Creditmemo\Balance` class which make changes on the afterSave or beforeSave events to the order will not be saved, because the order is already saved. By changing the order it will make sure that the changes in the order will be saved as well which are set by the executed observers on the order object. 

This behaviour is already mentioned in previous comments from me. I'm trying to understand what's so hard do termine why this 1 code line change takes over 6 months to process and @magento still has not a clue what is being fixed here.

engcom-Charlie:
@dverkade we have WebAPI test for it `\Magento\Sales\Service\V1\RefundOrderTest`. Could you extend existed test or create a new one to catch the issue?
Thank you.

dverkade:
Hi @engcom-Charlie,

The WebAPI test `\Magento\Sales\Service\V1\RefundOrderTest` you're referring to is not testing the `RefundOrder` class. It does not get instantiated within that test. 

As mentioned numerous times now. Magento does not have test code coverage on this class and I'll not be creating it. The complete scope of the issue and resolution are within this ticket and PR. It can be processed accordingly. If Magento feels that additional stuff should be done, they are free to do some. The time I've invested in this issue to research the cause and the solution is tremendous and it's up to Magento to invest their time into fixing anything that they seem fit and add test coverage which is not there in the first place. I do not have anything to add to this discussion.

engcom-Charlie:
@dverkade it should not instantiate within `\Magento\Sales\Service\V1\RefundOrderTest` but this test makes API calls that will be executed by service class. 
image

dverkade:
As mentioned numerous times now. Magento does not have test code coverage on this class and I'll not be creating it. The complete scope of the issue and resolution are within this ticket and PR.

gabrieldagama:
Hi @dverkade, sorry for the delay in process this, and thanks for the contribution!

We can help you with the automated tests, no worries. But more important than that we need to understand the problem, I read carefully the comments on this and I understand the problem that you are solving, but I don't agree with the solution for it.

Yes, you are right, `CustomerBalance/Model/Creditmemo/Balance` is called from the `CustomerBalance/Observer/CreditmemoSaveAfterObserver` and it does make some changes on the `Order` object and those changes may not be saved, but, I don't think that inverting the saving order is the correct solution.

When a client subscribes to an event where the subject is the `Creditmemo` it should not be expected that the `Order` gets saved, unless that is explicit inside the `Creditmemo`save function, to fix this issue we should actually save the order on `CustomerBalance/Model/Creditmemo/Balance`, or, create the plugin to a process like `RefundOrder` where is expected that the`Order` will be saved.

Basically what I'm saying is that if `$this->creditmemoRepository->save($creditmemo);` is called anywhere else, the`Order` won't be saved, so that observer also won't work on those places, and having this dependency becomes wrong, so if you need to change`Order` information that should be saved within the observer, or subscribe to another event that the order save is actually in the scope of that method.

I hope you understand my thoughts, please let me know what you think.

Thanks.

dverkade:
@gabrieldagama I see your point of view but do not totally agree. If every observer needs to save the order on it's own it will create a lot of overhead doing multiple save actions. This has a performance impact. 

Please note that this issue only occurs when you use the RefundOrder class. If you do this in the Magento backend everything works fine. The Magento Backend already does a `$order->save()`after the observers have been executed. So the behaviour is different and in all other places it already does a `$order->save()` after saving the credit memo. 

I've raised this issue with Magento Commerce Support as well.

engcom-Charlie:
Hello @dverkade.
We have discussed this issue internally and decided to close this PR in favor of magento/partners-magento2ee#392.

dverkade:
@engcom-Charlie I do not have access to that PR so can't verify that this is the correct approach.

PR #26999, opened 2020-02-25
Fixed URL Rewrite addition/removal on product website add/remove

### LINKED ISSUE #24184: Product URL Rewrites are not removed when product removed from website
### Preconditions (*)
1. 2.4-develop

### Steps to reproduce (*)
1. Deploy 2.4-develop
2. Create 2nd Website, 2nd Store, 2nd Store View
3. Create New Cateogory
4. Create New Product in New Category
5. Add product to Main Website and 2nd Website
6. Check URL table contains entries for Product, Category/Product for each store in each website
7. Remove Product from 2nd Website

### Expected result (*)
6. Check URL table contains entries for Product, Category/Product for only the stores in the first website

### Actual result (*)
1. URL Rewrites remain for all stores in all websites

gwharton:
### Related Pull Requests
https://github.com/magento/inventory/pull/2887

### Description (*)
Moved scope of ProductToWebsiteChangeObserver to base instead of adminhtml

as this functionality has been moved to a consumer. With the observer set
to adminhtml, this observer was not being called properly on changing
the websites for a particular product in the consumer.

Corrected ProductProcessUrlRewriteSavingObserver to correctly add and remove
rewrite URLs when products are added and removed from websites using the
Product Edit Admin Page.

Corrceted ProductToWebsiteChangeObserver to correctly add and remove
rewrite URLs when products are added and removed using the Products
Grid Mass Action function.

Extended unit tests to cover both Observers.

1. Closes magento/magento2#24184: Product URL Rewrites are not removed when product removed from website
This also fixes the issue with URL rewrites not being generated correctly using the mass action remove/add from website.

1. See issue


gwharton:
@magento run all tests

ihor-sviziev:
@magento run Database Compare

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -32,20 +40,38 @@ class ProductProcessUrlRewriteSavingObserver implements ObserverInterface
...
      * @param UrlPersistInterface $urlPersist
      * @param ProductUrlPathGenerator|null $productUrlPathGenerator
+     * @param StoreManagerInterface|null $storeManager
+     * @param StoreWebsiteRelationInterface|null $storeWebsiteRelation
      */
     public function __construct(
         ProductUrlRewriteGenerator $productUrlRewriteGenerator,
         UrlPersistInterface $urlPersist,
-        ProductUrlPathGenerator $productUrlPathGenerator = null
+        ProductUrlPathGenerator $productUrlPathGenerator = null,
```
ihor-sviziev: This class isn’t marked with `@api` tag, could you make all constructor arguments not optional?

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -65,10 +91,24 @@ public function execute(\Magento\Framework\Event\Observer $observer)
             || $product->getIsChangedWebsites()
             || $product->dataHasChangedFor('visibility')
         ) {
-            if ($product->isVisibleInSiteVisibility()) {
+            if ($product->getVisibility() != Visibility::VISIBILITY_NOT_VISIBLE) {
```
ihor-sviziev: Can we use strict comparison here?

app/code/Magento/CatalogUrlRewrite/Test/Unit/Observer/ProductProcessUrlRewriteSavingObserverTest.php
```diff
@@ -8,38 +8,51 @@
...
+ * old URLs) are called the correct number of times.
+ *
  * @SuppressWarnings(PHPMD.TooManyFields)
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
  */
 class ProductProcessUrlRewriteSavingObserverTest extends \PHPUnit\Framework\TestCase
 {
     /**
-     * @var \Magento\UrlRewrite\Model\UrlPersistInterface|\PHPUnit_Framework_MockObject_MockObject
+     * @var UrlPersistInterface|\PHPUnit_Framework_MockObject_MockObject
```
ihor-sviziev: Could you replace `\PHPUnit_Framework_MockObject_MockObject` to `\PHPUnit\Framework\MockObject\MockObject` in all places ?

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -65,10 +91,24 @@ public function execute(\Magento\Framework\Event\Observer $observer)
...
                 $product->unsUrlPath();
                 $product->setUrlPath($this->productUrlPathGenerator->getUrlPath($product));
                 $this->urlPersist->replace($this->productUrlRewriteGenerator->generate($product));
+
+                //Remove any rewrite URLs for websites the product is not in
+                foreach ($this->storeManager->getWebsites() as $website) {
+                    $websiteId = $website->getWebsiteId();
+                    if (!in_array($websiteId, $product->getWebsiteIds())) {
+                        foreach ($this->storeWebsiteRelation->getStoreByWebsiteId($websiteId) as $storeId) {
+                            $this->urlPersist->deleteByData([
```
ihor-sviziev: Doing a lot of queries in a loop doesn’t look good enough solution. Can we use execute single query?

app/code/Magento/CatalogUrlRewrite/Observer/ProductToWebsiteChangeObserver.php
```diff
@@ -77,24 +77,29 @@ public function __construct(
...
-                                UrlRewrite::ENTITY_ID => $product->getId(),
-                                UrlRewrite::ENTITY_TYPE => ProductUrlRewriteGenerator::ENTITY_TYPE,
-                                UrlRewrite::STORE_ID => $storeId
-                            ]);
-                        }
+            // Remove the URLs from websites this product no longer belongs to
+            if ($observer->getEvent()->getActionType() == "remove" && $observer->getEvent()->getWebsiteIds()) {
+                foreach ($observer->getEvent()->getWebsiteIds() as $webId) {
+                    foreach ($this->storeWebsiteRelation->getStoreByWebsiteId($webId) as $storeId) {
+                        $this->urlPersist->deleteByData([
```
ihor-sviziev: The same, probably we could do more efficient query here

app/code/Magento/CatalogUrlRewrite/Test/Unit/Observer/ProductProcessUrlRewriteSavingObserverTest.php
```diff
@@ -49,42 +62,87 @@ class ProductProcessUrlRewriteSavingObserverTest extends \PHPUnit\Framework\Test
...
     /**
-     * @var \Magento\CatalogUrlRewrite\Observer\ProductProcessUrlRewriteSavingObserver
+     * @var ProductProcessUrlRewriteSavingObserver
      */
     protected $model;
 
+    /**
+     * @var StoreManagerInterface|\PHPUnit_Framework_MockObject_MockObject
+     */
+    protected $storeManager;
```
ihor-sviziev: Please make all properties private

app/code/Magento/Catalog/Model/Product/Action.php
```diff
@@ -169,6 +169,14 @@ public function updateWebsites($productIds, $websiteIds, $type)
             $categoryIndexer->reindexList(array_unique($productIds));
         }
 
-        $this->_eventManager->dispatch('catalog_product_to_website_change', ['products' => $productIds]);
+        //Dispatch event to update Rewrite URLs for new/removed websites
+        $this->_eventManager->dispatch(
```
ihor-sviziev: Why we need to update event? Can’t we use after plugin instead?

ihor-sviziev (dismissed):
Hi @gwharton,
Thank you so much for the fix! From first view your fix looks good, but url rewrites quite complex thing and usually there are no issues with single unit, but issues appears when few components works together. For such cases unit tests much less effective than integration tests, as there you’re having real objects, real db, etc and you could do some queries to db and compare with what you expect.
Could you review my comments and cover your changes with integration tests?

gwharton:
I have added a new function to the class Magento/UrlRewrite/Model/Storage/DbStorage.php.

Previously this class only provided the ability to remove a single url rewrite at a time, which lead to lengthy and nested foreach loops, with many many database transactions to process for a single operation where multiple url rewrites needed removing.

I added a new function which allows you to remove multiple url rewrite's based on arrays of entity_ids and websites_ids, allowing you to remove multiple items from multiple websites in a single operation.

Question, Would it be worth adding this new deleteEntitiesFromStores function to the UrlPersistInterface so it can sit alongside the existing deleteByData function already in that interface, and provided by DbStorage. It could be useful elsewhere.

app/code/Magento/UrlRewrite/Model/Storage/DbStorage.php
```diff
@@ -375,4 +375,33 @@ public function deleteByData(array $data)
...
+    /**
+     * Function deleteEntitiesFromStores
+     *
+     * Deletes multiple URL Rewrites from database
+     *
+     * @param array $store_ids
+     * @param array $entity_ids
+     * @param int $entity_type
+     */
+    public function deleteEntitiesFromStores($store_ids, $entity_ids, $entity_type)
```
ihor-sviziev: Could you extract this method to separate "service" class? and use something like this:
```
$this->deleteEntitiesFromStores->execute($store_ids, $entity_ids, $entity_type)
```

In addition to that please use `$storeIds`, `$entityIds`, `$entityType` instead, define argument types and return type.

ihor-sviziev (changes requested):

app/code/Magento/CatalogUrlRewrite/Plugin/ProductProcessUrlRewriteRemovingPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+            if (!empty($this->productUrlRewriteGenerator->generate($product))) {
+                if ($product->isVisibleInSiteVisibility()) {
+                    $this->urlPersist->replace($this->productUrlRewriteGenerator->generate($product));
+                }
+            }
+        }
+
+        $storeIdsToRemove = [];
+        // Remove the URLs from websites this product no longer belongs to
+        if ($type == "remove" && $websiteIds && $productIds) {
```
ihor-sviziev: Could you use strict comparison here?

app/code/Magento/UrlRewrite/Model/Storage/DbStorage.php
gwharton: Done. I have just moved into own class and renamed. I haven't created any interfaces/API structures. Is that what you were after?

gwharton:
Apologies for the large number of additions of `<group value="urlRewrite"/>` in the MFTF tests. Group allocation was inconsistent amongst UrlRewrite and CatalogUrlRewrite modules, so I took the opportunity to tag them all in a consistent group to aid running of UrlRewrite related MFTF tests.

gwharton:
Let me know if you want me to cleanup commit history and squash.

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -6,56 +6,108 @@
...
+ *
+ * Observer to update the Rewrite URLs for a product.
+ * This observer is triggered on the save function when making changes
+ * to the products website on the Product Edit page.
  */
 class ProductProcessUrlRewriteSavingObserver implements ObserverInterface
 {
     /**
-     * @var ProductUrlRewriteGenerator
+     * @var ProductUrlRewriteGenerator $productUrlRewriteGenerator
```
ihor-sviziev: Why did you added variable name here?It is needed only for inline blocks.

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -65,11 +117,46 @@ public function execute(\Magento\Framework\Event\Observer $observer)
...
             }
+
+            $storeIdsToRemove = [];
+            if ($this->productScopeRewriteGenerator->isGlobalScope($product->getStoreId())) {
+                //Remove any rewrite URLs for websites the product is not in, or is not visible in. Global Scope.
+                foreach ($this->storeManager->getWebsites() as $website) {
+                    $websiteId = $website->getWebsiteId();
+                    foreach ($this->storeWebsiteRelation->getStoreByWebsiteId($websiteId) as $storeid) {
+                        //Load the product for the store we are processing so we can see if it is visible
+                        $storeProduct = $this->productRepository->getById(
```
ihor-sviziev: This looks as really heavy operation to load whole product object with all attributes, extension attributes, etc in this case. Can we get this info by doing some query?

I didn't ran any performance tests, but I think it will significantly slow down product save on multi store websites

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -65,11 +117,46 @@ public function execute(\Magento\Framework\Event\Observer $observer)
...
+                            $product->getId(),
+                            false,
+                            $storeid,
+                            true
+                        );
+                        if (!$storeProduct->isVisibleInSiteVisibility() ||
+                            !in_array($websiteId, $product->getWebsiteIds())) {
```
ihor-sviziev: In array in loop is not a good thing from performance. I know that loading product will be much more noticable, but anyway. We can create map outsize of the loop, like this `$productWebsiteMap = array_flip($product->getWebsiteIds());`
and in the loop check `isset($productWebsiteMap[$websiteId])`. It will be really faster

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -65,11 +117,46 @@ public function execute(\Magento\Framework\Event\Observer $observer)
...
+                            $storeIdsToRemove[] = $storeid;
+                        };
+                    }
+                }
+            } else {
+                //Only remove rewrite for current scope
+                if (!$product->isVisibleInSiteVisibility() ||
+                    !in_array($product->getStoreId(), $product->getStoreIds())) {
```
ihor-sviziev: the same for stores, as for websites.

ihor-sviziev (changes requested):
Looks like this is the last iteration :)

Let's squash all changes once everything will be ready

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
gwharton: I was never sure whether it was required or not. There were some entries where they were there, and some where they weren't. I was just making them consistent. I see now I went the wrong way on this one :). I'll remove them to make it ahem.... consistent with the rest of Magento :)

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
gwharton: @ihor-sviziev What do you think to something like this

```
$storeProduct = $productCollection->addAttributeToFilter('entity_id', $product->getId())
                        ->addAttributeToSelect(['visibility','entity_id'])
                        ->setPage(1,1)
                        ->setStoreId($storeId)
                        ->getFirstItem();
```
That gives a slimmed down product

image

I couldn't think of a better way of doing it?

ihor-sviziev: I think for this specific case this solution would be much much better

gwharton: Hmm, slight issue with the above. The Observer that is processing the URL Rewrites is called from the afterSave function in the Product model. This afterSave function is mid database transaction. The data retrieved by the productCollection seems to come from the database itself, and doesn't include the current active transaction set that is yet to be committed. Therefore if the product is being saved due to a change in visibility, the observer cannot see the changed values in the data it gets back from the collection. Any advice......

gwharton: In the current commit the product is obtained with the productRepository with the reload flag set to true, which seems to retrieve the correct info.

gwharton: OK, perhaps it isn't related to database transactions. I've asked on Slack for help  https://magentocommeng.slack.com/archives/C4YS78WE6/p1583489502082300

gwharton: Right, I found a much more pleasing way to get the visibilities for all products in all websites in a single call.

`$storeVisibilities = $this->productCollection->getAllAttributeValues(ProductInterface::VISIBILITY);`

This returns an array of ['productId']['storeId']['visibility']

Commit coming shortly

app/code/Magento/CatalogUrlRewrite/Observer/ProductProcessUrlRewriteSavingObserver.php
```diff
@@ -32,30 +43,62 @@ class ProductProcessUrlRewriteSavingObserver implements ObserverInterface
...
      */
     public function __construct(
         ProductUrlRewriteGenerator $productUrlRewriteGenerator,
         UrlPersistInterface $urlPersist,
-        ProductUrlPathGenerator $productUrlPathGenerator = null
+        ProductUrlPathGenerator $productUrlPathGenerator,
+        StoreManagerInterface $storeManager,
+        ProductScopeRewriteGenerator $productScopeRewriteGenerator,
+        DeleteEntitiesFromStores $deleteEntitiesFromStores,
+        Collection $productCollection
```
ihor-sviziev: Product collection contains state, so for collection you need to use factory in constructor and create collection object only when needed. I think this might cause test failures

gwharton: Yes, I'm learning. That explains why when repeatedly getting product data using product collection, it was not showing updates that were being saved with repository. Thanks. I have updated the code.

ihor-sviziev:
@magento run Functional Tests B2B, Functional Tests CE, Functional Tests EE, Integration Tests

gwharton:
@ihor-sviziev I don't know how to fix the integration test failure. The Allure report doesn't give any details on what went wrong.

ihor-sviziev:
@slavvka @lenaorobei could you give any info why integration test fails, but no failures in results?

lenaorobei:
Have to rerun in order to see build. Going to update branch.

lenaorobei:
@magento run all tests

lenaorobei:
```
</span><span class="timestamp"><b>11:32:02</b> </span>[Integration-1-ce] ...................................................Error in fixture: "\/var\/www\/html\/dev\/tests\/integration\/testsuite\/..\/..\/..\/..\/app\/code\/Magento\/InventoryGroupedProductIndexer\/Test\/_files\/custom_stock_with_eu_website_grouped_products.php".
<span class="timestamp"><b>11:32:02</b> </span>[Integration-1-ce]  The product was unable to be saved. Please try again.
```

gwharton:
ok, i wasnt running my local tests with MSI installed. I'll take a look. Thanks @lenaorobei

gwharton:
OK, i've setup my dev environment with MSI included and unfortunately all tests that reference the custom_stock_with_eu_website_grouped_products.php fixture file pass.

Not sure where else to go now.

ihor-sviziev:
@engcom-Charlie @engcom-Golf could you help with failing integration test?

engcom-Echo:
Will try to find issues

engcom-Echo:
@magento run all tests

engcom-Echo:
Failed functional, static tests not related to the changes in this PR

gwharton:
The integration test that used to fail now passes.
The functional tests seem to pass if you re-run them enough times.
The static test failiure is nothing to do with my code.

I guess this is what you get by running the tests against the PR merged with 2.4-develop as opposed to running against the forked repo as if someone manages to commit something bad to 2.4-develop, everyones tests fail.

ihor-sviziev:
@engcom-Echo thank you so much!

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-7162](https://jira.corp.magento.com/browse/ENGCOM-7162) has been created to process this Pull Request

engcom-Bravo:
:heavy_check_mark: QA Passed

The record for the second website disappears from the ```catalog_url_rewrite_product_category``` table after the product is unassigned
PR_26999

engcom-Echo:
Failed functional, static tests not related to the changes in this PR

naydav:
@gwharton 
Hi Graham,
We run Performance build with `medium` profile on `2.4-develop` branch and we discovered huge degradation during Product saving in Admin panel (approximately 70% of degradation).

```
Server Side EE API
Create product with extensible data objects	+77.7 %	[+205 ms]
Create product	+160.8 %	[+201 ms]
Server Side EE Add to Wishlist

Server Side EE Admin Create Order
Save Shipment	+7.4 %	[+27 ms]
Add Products	+5.7 %	[+18 ms]

Server Side EE Admin Create Product
New Configurable Product Save	+85.4 %	[+698 ms]
Server Side EE Admin Customer Management

Server Side EE Admin Edit Product
Edit Simple Product Save	+13 %	[+47 ms]
Edit Configurable Product Save
```
Thus we had to revert changes.

We need to find a more appreciative solution or make refact of current.
Thanks

gwharton:
Understood. We did do some work on optimising it, but obviously not enough.

PR #27034, opened 2020-02-26
FIX: wrong img size attributes

GrimLink:
### Description (*)

Since commit https://github.com/magento/magento2/commit/da7a64f2585ac534762235821db727f7af59f808#diff-eed94ca7983b385a07b8fb5ee2dbc20d the html attributes on the `image_with_borders.phtml` have changed from;
- width to max-width
- height to max-height.

This is off course not valid html.
And will not work.

### Fixes

This will use the right html width and height tags and improve rendering if used correctly.

_No testing required._


app/code/Magento/Catalog/view/frontend/templates/product/image_with_borders.phtml
```diff
@@ -13,7 +13,7 @@
         <img class="<?= $block->escapeHtmlAttr($block->getClass()) ?>"
             <?= $block->escapeHtmlAttr($block->getCustomAttributes()) ?>
             src="<?= $block->escapeUrl($block->getImageUrl()) ?>"
-            max-width="<?= $block->escapeHtmlAttr($block->getWidth()) ?>"
-            max-height="<?= $block->escapeHtmlAttr($block->getHeight()) ?>"
+            width="<?= $block->escapeHtmlAttr($block->getWidth()) ?>"
```
ihor-sviziev, line 16: Looks like there html attributes hadn’t any effect, so we could just remove them. From another side - width and height might be used by custom extensions that processes images somehow, so maybe it’s better to replace them to data-width and data-height.
What do you think?

ihor-sviziev (changes requested):

app/code/Magento/Catalog/view/frontend/templates/product/image_with_borders.phtml
GrimLink: @ihor-sviziev No the images should have the html width and height tag.
This allows the browsers to render the page better.
As it already know what for image size to expect.

ihor-sviziev:
Hi @GrimLink,

Right now we have pull request that includes the same changes and it's already approved https://github.com/magento/magento2/pull/27033.

I'm closing this Pull Request as duplicate of https://github.com/magento/magento2/pull/27033

PR #27077, opened 2020-02-28
Update FlushCacheByTags to use after plugins instead of around plugins

### LINKED ISSUE #29558: [Issue] Update FlushCacheByTags to use after plugins instead of around plugins
This issue is automatically created based on existing pull request: magento/magento2#27077: Update FlushCacheByTags to use after plugins instead of around plugins

---------

### Description (*)

This PR contains a few small quality of life improvements to the FlushCacheByTags class:

1. Conversion of around plugins to after plugins.  
These original plugins were created when after plugins did not contain the arguments to the original call.  Now that they do, we can move to after plugins and not affect the call stack
2. Micro-optimization in cleanCacheByTags. 
Calling array_unique in a loop, _for the same variable_ isn't great.  This change now caches it in a local variable.
3. Updating doc-blocks. 
Newer documentation standards request the exclusion of parameters and return types that are duplicative in nature.  I've also updated the small descriptions to read better 
### Preconditions (*)
1. Magento 2.4-develop

### Steps to reproduce (*)
* Perform step debugging in Plugin before/after changes and compare that tags are the same during the following scenarios:
  * Product Creation
  * Production Deletion

### Expected result (*)
The tags are the same

### Actual result (*)
The tags are differant

navarr:
### Description (*)

This PR contains a few small quality of life improvements to the FlushCacheByTags class:

1. Conversion of around plugins to after plugins.  
These original plugins were created when after plugins did not contain the arguments to the original call.  Now that they do, we can move to after plugins and not affect the call stack
2. Micro-optimization in cleanCacheByTags. 
Calling array_unique in a loop, _for the same variable_ isn't great.  This change now caches it in a local variable.
3. Updating doc-blocks. 
Newer documentation standards request the exclusion of parameters and return types that are duplicative in nature.  I've also updated the small descriptions to read better 

### Manual testing scenarios
* Perform step debugging in Plugin before/after changes and compare that tags are the same during the following scenarios:
  * Product Creation
  * Production Deletion

### Resolved issues:
1. [x] resolves magento/magento2#29558: Update FlushCacheByTags to use after plugins instead of around plugins


navarr:
Failed test: `testCodeStyle` is a false-positive.  Code style is correct according to https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html#functions-methods

lbajsarowicz:

@navarr Unfortunately you need to provide DocBlock. Add them for now, and in the meantime I'm going to ask Lena about the Static Tests.

@lenaorobei I thought this changed some time ago and if there are strict types - there's no need of adding doc blocks?

lbajsarowicz (changes requested):
Unfortunately, you need to add the DocBlocks.

navarr:
@lbajsarowicz Is that because the rule is "methods must have parameters" or b/c the rule is "tests can't fail"

If its the latter, we can fix the static test thats improperly failing. If its the former, its easy enough to do.

lbajsarowicz:
Moved to manual approval, as the failing test is false negative.

drew7721:
@lbajsarowicz This is not ok to merge. It's fine to change the _aroundSave_ in to an _afterSave_ but not the delete. The delete makes operations **before and after** the `proceed` statement. So unless that is tested and 100% sure it will not break anything nor will it leave useless data in the Redis cache then this can NOT be merged in. Also, why is this removing docs from the constructor?

drew7721:
@lbajsarowicz Also, the `clean` method in the `Magento\Framework\Cache\FrontendInterface` is expecting an **array** type. What will happen when null is passed?? Seriously?!

This will introduce more bugs than it fixes. I agree it should be an afterSave but the rest of the changes are 👎 .

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
```diff
@@ -95,18 +85,18 @@ public function aroundDelete(AbstractResource $subject, \Closure $proceed, Abstr
...
         if (!$tags) {
             return;
         }
+        $uniqueTags = null;
         foreach ($this->cacheList as $cacheType) {
             if ($this->cacheState->isEnabled($cacheType)) {
                 $this->cachePool->get($cacheType)->clean(
                     \Zend_Cache::CLEANING_MODE_MATCHING_ANY_TAG,
-                    \array_unique($tags)
+                    $uniqueTags = $uniqueTags ?? \array_unique($tags)
```
drew7721, line 114: Why? Is this really needed?

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
```diff
@@ -37,12 +37,6 @@ class FlushCacheByTags
      */
     private $tagResolver;
 
-    /**
```
drew7721: Please don't remove code blocks. If you want add the long class declarations back instead.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
```diff
@@ -56,36 +50,32 @@ public function __construct(
...
      */
-    public function aroundDelete(AbstractResource $subject, \Closure $proceed, AbstractModel $object): AbstractResource
-    {
+    public function afterDelete(
+        AbstractResource $subject,
+        AbstractResource $result,
+        AbstractModel $object
+    ): AbstractResource {
         $tags = $this->tagResolver->getTags($object);
-        $result = $proceed($object);
```
drew7721, line 88: The tags are retrieved before deleting the Resource, not after. But we need to Purge the cache only after the delete. If the delete fails, it should rollback the transaction and not commit, hence not purge the cache. This needs to stay an around.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
```diff
@@ -56,36 +50,32 @@ public function __construct(
...
         $this->cleanCacheByTags($tags);
 
         return $result;
     }
 
     /**
-     * Clean cache on delete object
+     * Clean cache when object is deleted
      *
-     * @param AbstractResource $subject
```
drew7721, line 79: Do not delete the params.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
```diff
@@ -56,36 +50,32 @@ public function __construct(
     }
 
     /**
-     * Clean cache on save object
+     * Clean cache when object is saved
      *
-     * @param AbstractResource $subject
-     * @param \Closure $proceed
```
drew7721, line 62: Do not change the doc block

drew7721 (changes requested):

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
navarr: A rollback will throw an Exception, right?  In which case the line after $proceed is not called.  This same scenario occurs with the after; an exception is thrown and the after is not called.

I failed to find any instance in which the tags changed after an object was deleted.  This would be the only reason to call getTags before deletion.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
navarr: Previously the code was performing array_unique in a loop (slow!) Caching the value locally makes sense.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
navarr: Re-reviewing this, this was an erroneous removal as the type for `$cacheList` disappears.  I'll return this block.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
navarr: Why?

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
navarr: The parameters are redundant and duplicative.

Code style guide says that they're unnecessary and should not be here:

1. https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html#general-principles

> Include all necessary information without duplication.

2. https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html#functions-methods

> The declaration of all arguments (if any) using the @param tag, unless the argument type is indicated in the method signature.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
drew7721: I agree, so let's do it once before and pass the variable to the for each. Write easy to read code and easy to debug code. Debugging that will be a pain in the future.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
drew7721: I'm not sure if the entityid to crate the tag will still be available on the object once it's deleted. Need to debug and test that to make sure the tag will be created the same way once the object is deleted. The object might be a product, saleable item, category, cms page, etc... This needs to account for all these types.

lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php
drew7721: Try to run phpdoc on this and see. 

The whole reason for the docblock is to make life easier and code more readable and to make note of what the method does. That link in the standards above is pointing you to the [docblock description](https://docs.phpdoc.org/latest/references/phpdoc/basic-syntax.html#what-is-a-docblock). It's important to keep the doc blocks there to continue support for all IDEs. Not everyone is developing on the same IDE.

Maybe other can weight in on this. But IMO it's not good practice to start removing doc blocks just because PHP now supports strict types.

I would actually change the params on multiple lines instead to respect the PSR-2 line length recommendation instead.

drew7721: The whole reason for the docblock is to make life easier and code more readable and to make note of what the method does. That link in the standards above is pointing you to the [docblock description](https://docs.phpdoc.org/latest/references/phpdoc/basic-syntax.html#what-is-a-docblock). It's important to keep the doc blocks there to continue support for all IDEs. Not everyone is developing on the same IDE.

Maybe other can weight in on this. But IMO it's not good practice to start removing doc blocks just because PHP now supports strict types.

I would actually change the params on multiple lines instead to respect the PSR-2 line length recommendation instead.

navarr:
@magento rerun all tests

navarr:
@magento rerun

lbajsarowicz:
@magento run all tests

sidolov:
@magento create issue

sidolov:
@navarr could you please sign Adobe CLA?

navarr:
@sidolov I have previously signed it, is there a new one?  Usually a bot posts a link but that's not the case here

sidolov:
@navarr looks like something went wrong, I have closed and re-opened PR and check is passed now.

sidolov:
@magento run all tests

sidolov (approved):

magento-engcom-team:
[ENGCOM-8079](https://jira.corp.magento.com/browse/ENGCOM-8079) has been created to process this Pull Request

| Label | Description |

lbajsarowicz (approved):
✔️  All green from my side.

magento-engcom-team:
Hi @lbajsarowicz, thank you for the review.
[ENGCOM-8079](https://jira.corp.magento.com/browse/ENGCOM-8079) has been created to process this Pull Request

engcom-Bravo:
:heavy_check_mark: QA Passed

### Manual testing scenario
1. Configure xDebug
2. Open the file _lib/internal/Magento/Framework/App/Cache/FlushCacheByTags.php_
3. Put debug on the following lines and **Start listening...**
- 73 `$this->cleanCacheByTags($tags);`
- 93 `$this->cleanCacheByTags($tags);`
4. Go to Magento admin and Add a new product. Click **Save**
5. When xDebug stops, check the `$tags` array
tags1
6. Go to the next step, and verify that the Product has been successfully saved
7. **Delete** the previously created product from Magento admin
8. When xDebug stops, check the `$tags` array
tags2
9. Assert that the `$tags` after saving are the same as tags after deleting

AR: The `$tags` array is the same

PR #27143, opened 2020-03-03
Fix issue with incorrect images custom attributes output in catalog module

arendarenko:
### Description (*)

Commit https://github.com/magento/magento2/commit/f40f24c3ff971fe49480a470868f443b6061d349 which was merged since `Magento 2.3.3-p1` caused a regression issue with custom attributes on category pages because method `getCustomAttributes` can contains content like `data-attr-value="some-attr-content" data-attr2-value="some-attr-content"` so it's a bad idea to escape it.

My changes reverts those escapes.

### DIFF
--- a/app/code/Magento/Catalog/view/frontend/templates/product/image.phtml
+++ b/app/code/Magento/Catalog/view/frontend/templates/product/image.phtml
@@ -10,7 +10,7 @@
 ?>
 
 <img class="photo image <?= $escaper->escapeHtmlAttr($block->getClass()) ?>"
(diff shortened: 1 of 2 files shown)


ihor-sviziev (changes requested):
Hi @arendarenko,
Your changes looks good to me.
Could you cover your changes with integration tests in order not to make regression here in future?

ihor-sviziev:
Hi @arendarenko,
I'm closing this PR because described issue was already fixed in https://github.com/magento/magento2/pull/26959 (in little bit different way), and this PR already merged.

Thank you for collaboration!

PR #27159, opened 2020-03-04
Fix phpdoc

norgeindian:
This function either returns an array or null.

1. ...
2. ...

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Item.php
+++ b/app/code/Magento/Sales/Model/Order/Item.php
@@ -491,7 +491,7 @@ class Item extends AbstractModel implements OrderItemInterface
      * If code is null return all options
      *
      * @param string $code
-     * @return array | null
+     * @return array|null
      */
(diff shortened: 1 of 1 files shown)


app/code/Magento/Sales/Model/Order/Item.php
```diff
@@ -491,7 +491,7 @@ public function getProductOptions()
      * If code is null return all options
      *
      * @param string $code
-     * @return array
+     * @return array | null
```
drew7721: No spaces. `@return array|null`

drew7721 (changes requested):

lenaorobei:
Hi @norgeindian. Adding PHPDoc annotation looks like hiding of actual issue. The code needs to be refactored in order to use strict typing. So let's avoid adding more confusion to the code.

Closing this PR because the fix should be complex refactoring. Feel free to reopen if you are up to working on it. Thanks.

PR #27165, opened 2020-03-04
[feature] Display category filter item in layered navigation based on the system configuration from admin area

vasilii-b:
### Description (*)
This PR aims to add an easy way to hide the category filter item from the layered navigation on the storefront, without help from the developer's side.

## The reason behind the PR
As a Magento 2 Frontend Developer working on projects daily, I receive often requests to not show the categories filter in the layered navigation on the category pages. To achieve this, there are 2 options:

1. Install a 3rd-party module that does the job;
This is not the best option if the module adds a bunch of other features/functionalities to the store that are not used - this becomes trash for the project.
2. Write a custom solution to achieve the result;
See below details for this.

## Why custom changes aren't the best solution
You might say that this could be achieved by having a plugin (interception) for `Magento\Catalog\Model\Layer\FilterList::getFilters` and you'd be right.

https://github.com/magento/magento2/blob/5a9fcab6a35b6b39e10f47ca1ea22dfca4b794a4/app/code/Magento/Catalog/Model/Layer/FilterList.php#L64-L81

There are 2 ways I see for such an approach and their downsides, IMHO:

1. #### `afterGetFilters` plugin
The `getFilters` method returns an array of available filters. Which means the `afterGetFilters` plugin should look like this (not necessarely exactly the same):
```php
public function afterGetFilters(\Magento\Catalog\Model\Layer\FilterList $subject, array $result) {
    if (array_key_exists(0, $result)) { // check whenever the key we wanna remove exists. It might be so that other plugin (3rd party) already did it
        unset($result[0]);
    }

    return $result;
}
```

Downsides here:
1. We introduce a (one more) plugin to the system
2. We remove the category filter item by the key (0) from the `$result`. Which, in my opinion, is not the cleanest way to do that.
3. Last, but not least - the category filter is already initialized and added to the result. Which means the time and resources were consumed already. You may say that I'm pedant - OK.

2. #### `aroundGetFilters` plugin
I won't provide the example here, because this means we're overriding the whole method and do the same stuff as in the original one, except the adding of the category filter.  [Around plugins are generally a thing to avoid](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/plugins.html?itm_source=devdocs&itm_medium=search_page&itm_campaign=federated_search&itm_term=interception%20around#around-methods).

## This PR's solution

Add a new system configuration field under _Stores -> Configuration -> Catalog -> Catalog -> Layered Navigation_.

System configuration: choose whenever to show or not the category filter item in the layered navigation

### Further actions
Once this PR's concept/idea is approved, I'll proceed with covering it with MFTF tests.

### Related Pull Requests
N/A

N/A

**Preconditions**
Magento 2 with sample data OR any other Magento 2 instance with:
- a category on the storefront that has layered navigation;
- no 3rd-party module or other custom solution to remove the categories from the layered navigation;
- categories filter is visible on the storefront;

With the changes added by this PR, please do so:
1. Open the system configuration page: _Stores -> Configuration -> Catalog -> Catalog -> Layered Navigation_;
2. Change the value of "Display Category" to "No"
3. Flush the caches
4. Open the category page with layered navigation on the storefront
5. See that categories are not present in the layered navigation

### Questions or comments

Would be nice to know your thoughts on the changes and possible downsides to the introduced changes

### DIFF
--- a/app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
+++ b/app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Catalog\Model\Config;
+
+use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Store\Model\ScopeInterface;
+use Magento\Store\Model\StoreManagerInterface;
+
+/**
+ * Config for category in the layered navigation
+ */
+class LayerCategoryConfig
+{
+    private const XML_PATH_CATALOG_LAYERED_NAVIGATION_DISPLAY_CATEGORY = 'catalog/layered_navigation/display_category';
+
+    /**
+     * @var ScopeConfigInterface
+     */
+    private $scopeConfig;
+
+    /**
+     * @var StoreManagerInterface
+     */
+    private $storeManager;
+
+    /**
+     * LayerCategoryConfig constructor
+     *
+     * @param ScopeConfigInterface $scopeConfig
+     * @param StoreManagerInterface $storeManager
+     */
+    public function __construct(
+        ScopeConfigInterface $scopeConfig,
+        StoreManagerInterface $storeManager
+    ) {
+        $this->scopeConfig = $scopeConfig;
+        $this->storeManager = $storeManager;
+    }
+
+    /**
+     * Check if category filter item should be added in the layered navigation
+     *
+     * @param string $scopeType
+     * @param null|int|string $scopeCode
+     *
+     * @return bool
+     */
+    public function isCategoryFilterVisibleInLayerNavigation(
+        $scopeType = ScopeInterface::SCOPE_STORES,
+        $scopeCode = null
+    ): bool {
+        if (!$scopeCode) {
+            $scopeCode = $this->getStoreId();
+        }
+
+        return $this->scopeConfig->isSetFlag(
+            static::XML_PATH_CATALOG_LAYERED_NAVIGATION_DISPLAY_CATEGORY,
+            $scopeType,
+            $scopeCode
+        );
+    }
+
+    /**
+     * Get the current store ID
+     *
+     * @return int|null
+     */
+    protected function getStoreId(): ?int
+    {
+        try {
+            return (int) $this->storeManager->getStore()->getId();
+        } catch (NoSuchEntityException $e) {
+            return null;
+        }
+    }
+}
--- a/app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
+++ b/app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
@@ -69,14 +69,12 @@ class LayerCategoryConfig
     /**
      * Get the current store ID
      *
-     * @return int|null
+     * @return int
+     *
+     * @throws NoSuchEntityException
      */
-    private function getStoreId(): ?int
+    private function getStoreId(): int
     {
-        try {
-            return (int) $this->storeManager->getStore()->getId();
-        } catch (NoSuchEntityException $e) {
-            return null;
-        }
+        return (int) $this->storeManager->getStore()->getId();
     }
 }
--- a/app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
+++ b/app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
@@ -71,7 +71,7 @@ class LayerCategoryConfig
      *
      * @return int|null
      */
-    protected function getStoreId(): ?int
+    private function getStoreId(): ?int
     {
         try {
             return (int) $this->storeManager->getStore()->getId();
--- a/app/code/Magento/Catalog/Model/Layer/FilterList.php
+++ b/app/code/Magento/Catalog/Model/Layer/FilterList.php
@@ -7,6 +7,9 @@ declare(strict_types=1);
 
 namespace Magento\Catalog\Model\Layer;
 
+use Magento\Catalog\Model\Config\LayerCategoryConfig;
+use Magento\Framework\App\ObjectManager;
+
 /**
  * Layer navigation filters
  */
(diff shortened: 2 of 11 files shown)


app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Get the current store ID
+     *
+     * @return int|null
+     */
+    private function getStoreId(): ?int
+    {
+        try {
+            return (int) $this->storeManager->getStore()->getId();
+        } catch (NoSuchEntityException $e) {
```
rogyar: Could you shed some light on cases where we may have `NoSuchEntityException` during getting the store ID, please? 

Thank you.

app/code/Magento/Catalog/Model/Layer/FilterList.php
```diff
@@ -44,18 +47,27 @@ class FilterList
...
     public function __construct(
         \Magento\Framework\ObjectManagerInterface $objectManager,
         FilterableAttributeListInterface $filterableAttributes,
-        array $filters = []
+        array $filters = [],
+        LayerCategoryConfig $layerCategoryConfig = null
     ) {
         $this->objectManager = $objectManager;
         $this->filterableAttributes = $filterableAttributes;
+        $this->layerCategoryConfig = $layerCategoryConfig ?:
```
rogyar: Since we are bringing changes to the minor release, it's not necessary to invoke the object manager here. You may remove it.

app/code/Magento/Catalog/etc/adminhtml/system.xml
```diff
@@ -138,6 +138,12 @@
                     <hide_in_single_store_mode>1</hide_in_single_store_mode>
                 </field>
             </group>
+            <group id="layered_navigation">
+                <field id="display_category" translate="label" type="select" sortOrder="15" showInDefault="1" showInWebsite="1" showInStore="1" canRestore="1">
+                    <label>Display Category</label>
```
rogyar: I would clarify the title of the current option and name it as "Display Category Filter", what do you think?

rogyar (changes requested):
Hi @vasilii-b. Thanks a lot for the great feature. Please, check my comments below. Also, I would kindly ask you to cover your feature with MFTF test. 
The MFTF test will create preconditions (category, product) for showing layered navigation on the storefront. Then enable/disable the newly introduced configuration option using CLI command. And finally, assert the category filter presence on the storefront. 

Please, let me know if you need any help.

app/code/Magento/Catalog/Model/Config/LayerCategoryConfig.php
vasilii-b: Since the `getStore()` throws the exception if the store is not found, I've considered to handle the exception to prevent errors on the storefront.

app/code/Magento/Catalog/Test/Mftf/ActionGroup/StorefrontChangeCategoryFilterVisibilityInLayeredNavigationActionGroup.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+-->
+
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
```
rogyar: Please, use 1 action group per file. Basically, you need to split this file into two separate files

rogyar: In MFTF for verifying some state we use term "Assert". So, I would suggest renaming the action groups correspondingly, so we have something like the following:

- `AssertStorefrontLayeredNavigationCategoryFilterVisibleActionGroup`
- `AssertStorefrontLayeredNavigationCategoryFilterNotVisibleActionGroup`

rogyar (changes requested):

app/code/Magento/Catalog/Test/Mftf/ActionGroup/StorefrontChangeCategoryFilterVisibilityInLayeredNavigationActionGroup.xml
vasilii-b: Thank you, @rogyar! I'll add the changes accordingly.

vasilii-b:
Hi @rogyar,
I've adjusted the PR changes according to your review. Please let me know whenever the changes are good to go. I'd like to squash all the changes into a single commit before merging to the target branch.
Thank you!

rogyar:
@magento run all tests

rogyar:
The failing tests are not related to the current PR since we have the same failing tests for other PRs

rogyar (approved):

magento-engcom-team:
[ENGCOM-7087](https://jira.corp.magento.com/browse/ENGCOM-7087) has been created to process this Pull Request

engcom-Bravo:
:heavy_check_mark: QA Passed

When the **Display Category Filter** is set to **No**, the _CATEGORY_ filter disappears from the layered navigation
See the screenshot made when the filter is **Enabled**
7087_enabled

and when the filter is **Disabled**
7087_disabled


One small note, which might be wors of attention in the future, or in the scope of another bug.

In Admin, the **Display Category Filter** it the third field in _Stores - Configuration - Catalog - Catalog - Layered Navigation_. And when the upper, second, field **Price Navigation Step Calculation** is set to **Manual**, for example, another two fields appear, **Default Price Navigation Step** and **Maximum Number of Price Intervals**. These two fields appear lower than the **Display Category Filter**, and the **Display Category Filter** remains on the third position.
I think it should appear in the last position, after the **Default Price Navigation Step** and **Maximum Number of Price Intervals**
display_cat_filter_field

vasilii-b:
Hi @engcom-Bravo,
Thanks for giving this PR a round of QA!

I've adjusted the system configuration field so it appears in the right order ✅ 


Thanks!

engcom-Echo:

Issue #26882, opened 2020-02-14
2.3.4 Email Template New Pickup Order or add New Template order items are missing in Mail


stradd:
**Preconditions (*)**
magento 2.3.4 

I installed magento 2.3.4 with sample data.
In the customer order mail order items are missing

**Steps to reproduce(*)**

Stores> Configuration> Sales > Sales Emails> New Order Confirmation select Template: New Pickup Order
Sales > Order: Select a order, goto Information, Send Mail

In the mail that Magento sends, the section with the order items is not transferred.
The same applies if you create a new template and select a template New Order or New Order (Magento Luma).

Only if you select
Stores> Configuration> Sales Emails> New Order Confirmation Template: New Order (default)
the order items will be transferred

**Actual result (*)**
In the customer order mail order items are missing

**Expected result (*)**
In the customer order mail order items should be there

magento-engcom-team:
:white_check_mark: Confirmed by @shikhamis11 
 Thank you for verifying the issue. Based on the provided information internal tickets `MC-31544` were created

**Issue Available**: @shikhamis11, _You will be automatically unassigned. Contributors/Maintainers can claim this issue to continue. To reclaim and continue work, reassign the ticket to yourself._

ramesh-durairajaram:
Hi @shikhamis11 , I don't get this new pickup template in `new order confirmation` option. How do we get this new pickup template

fongd:
I believe this issue also affects 2.3.3-p1 as well.

verheesj:
@magento give me 2.4-develop instance

magento-engcom-team:
Hi @verheesj. Thank you for your request. I'm working on Magento 2.4-develop instance for you

magento-engcom-team:
Hi @verheesj, here is your [Magento instance](https://i-26882-2-4-develop.instances.magento-community.engineering).
Admin access: [https://i-26882-2-4-develop.instances.magento-community.engineering/admin_8d71](https://i-26882-2-4-develop.instances.magento-community.engineering/admin_8d71)
Login: `9254450c` Password: `9cc5942a15ef`
Instance will be terminated in up to 3 hours.

achleitner:
How can this be solved? All my order emails are empty. Anyone has a fix? Thanks

singhvivek64:
$magento give me 2.4-develop instance

puntable:
I can confirm. It seems every 2.3.4 are missing order_items_grid in emails, if they are not using default.

[...]

PR #27209, opened 2020-03-09
Fix sales_email_order_items rendering

BrentRobert:
### Description (*)

This pull request fixes the rendering of the sales order items in emails. Due to strict mode the order variable gets stripped and is not available anymore.

1. magento/magento2#26882: 2.3.4 Email Template New Pickup Order or add New Template order items are missing in Mail

1. Test with the default new order email.
2. Test with a new default template created in the backend.
3. Both scenario's should now render the order items.

### DIFF
--- a/app/code/Magento/Sales/Block/Order/Email/Items.php
+++ b/app/code/Magento/Sales/Block/Order/Email/Items.php
@@ -9,7 +9,12 @@
  *
  * @author     Magento Core Team <core@magentocommerce.com>
  */
+
 namespace Magento\Sales\Block\Order\Email;
+
+use Magento\Framework\View\Element\Template;
+use Magento\Sales\Api\Data\OrderInterface;
+use Magento\Sales\Api\OrderRepositoryInterface;
 
 /**
  * @api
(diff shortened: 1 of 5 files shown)


gwharton:
Feels very much like this fix is being applied at the wrong level. There is something more fundamental wrong here as the same problem is affecting shipment, invoice and credit memo emails too.

Look at the difference between legacy mode and normal mode for template processing. legacy mode works, normal mode doesn't.

BrentRobert:
@gwharton this PR is supposed to fix the 'normal' mode for order items. Passing objects is just not allowed any more in normal mode as all passed variables need to be scalar for security reasons. Therefore I pass the ID of the order and retrieve the order.

This PR however does only fix the order items, not those of invoices etc..

gwharton:
Hi Brent, oh yes. I see that now. My apologies. I've taken a look through the commits made back in Sept/Oct where the scalar variables only mode was introduced, quite extensive changes i see.

gwharton:
Is the intention to extend this to include the others? or do we need another bug raising?

gwharton:
Seems in the OrderSender.php we are still setting various objects in the transportobject which will no longer be accepted, e.g $store, $billing (is that object or array), $order. Shouldnt these be removed if they are no longer accepted by the template processor?

The variables section at the top of the email templates also needs work.

Are these entries still valid????




Seems like a perfect opportunity to make the data in the OrderSender match the variables available in the template.

gwharton:
Note, the following files need looking at too

vendor/magento/module-sales/Model/Order/Email/Sender/CreditmemoCommentSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/CreditmemoSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/InvoiceCommentSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/InvoiceSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/OrderCommentSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/OrderSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/ShipmentCommentSender.php
vendor/magento/module-sales/Model/Order/Email/Sender/ShipmentSender.php
vendor/magento/module-sales/Model/Order/Creditmemo/Sender/EmailSender.php
vendor/magento/module-sales/Model/Order/Invoice/Sender/EmailSender.php
vendor/magento/module-sales/Model/Order/Shipment/Sender/EmailSender.php

PR #27230, opened 2020-03-10
Add the ability for plugin overrides.

lukepaoloni:
### Description (*)

Adds the ability for plugin overrides on two shipping address fieldsets. This is important as if you'd like to display a city in the shipping address fieldset on checkout/cart, you're unable to do so easily without creating a preference. This isn't ideal and there is no reason these methods shouldn't be public.

### DIFF
--- a/app/code/Magento/Checkout/Block/Cart/LayoutProcessor.php
+++ b/app/code/Magento/Checkout/Block/Cart/LayoutProcessor.php
@@ -59,7 +59,7 @@ class LayoutProcessor implements \Magento\Checkout\Block\Checkout\LayoutProcesso
      * @return bool
      * @codeCoverageIgnore
      */
-    protected function isCityActive()
+    public function isCityActive()
     {
         return false;
     }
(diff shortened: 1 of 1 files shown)


dmytro-ch (changes requested):
Hello, @lukepaoloni, thank you for your contribution!

Unfortunately, we are not allowed to change the scope of visibility for methods and properties according to the [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development#adding-parameters-in-public-methods).

Feel free to introduce a new implementation of the methods that you wish to make public, in the scope of a separate service class. Than proxy the `isCityActive` and `isStateActive` calls to the newly created methods.

Please also keep in mind, due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html#automated-tests) the changes should be covered by automated tests.

Thank you!

dmytro-ch:
Hello @lukepaoloni, I'm closing this PR now due to inactivity.
Feel free to reopen and update if you wish to continue.
Thank you!

PR #27243, opened 2020-03-11
Whitelist Generation of tables and columns that are allowed to be edited by the declaration installer Log Errors

### LINKED ISSUE #27241: Whitelist Generation of tables and columns that are allowed to be edited by the declaration installer Log Errors
### Description (*)
 Whitelist Generation of tables and columns that are allowed to be edited by the declaration installer doesn't log any errors. The **db_whitelist_schema.json** file is not created automatically when running the `bin/magento setup:db-declaration:generate-whitelist --module-name=Vendor_Modulename` command because there’s an error in the etc/db_schema.xml file of the module.

### Expected behavior (*)
It should log or throw the error in the console 
### Benefits
1. The developer will know what is the error in its db_schema.xml file of the module.
2. The developer can fix the error easily.

### Additional information
Steps to produce:
1. Create a module **Vendor/Module**
2. Create a **db_schema.xml** file with an error
3. Run Command `bin/magento setup:db-declaration:generate-whitelist --module-name=Vendor_Modulename`
4. The command will be executed successfully **no error was logged or given in console**.

immuhammad:
Whitelist Generation of tables and columns that are allowed to be edited by the declaration installer doesn't log any errors. The db_whitelist_schema.json file is not created automatically when running the bin/magento setup:db-declaration:generate-whitelist --module-name=Vendor_Modulename command because there’s an error in the etc/db_schema.xml file of the module.

### Description (*)
Changed the file `lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php` and added a try catch statment around **getDeclarationConfig()** function to log the errors in `var/log/schema.log` 

1. magento/magento2#<27241>: [Whitelist Generation of tables and columns that are allowed to be edited by the declaration installer Log Errors](https://github.com/magento/magento2/issues/27241) 

1. Create a module **Vendor/Module**
2. Create a db_schema.xml file with an error
3. Run Command `bin/magento setup:db-declaration:generate-whitelist --module-name=Vendor_Modulename`
4. The command will be executed successfully error will be logged in **var/log/schema.log**

### DIFF
--- a/lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
+++ b/lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
@@ -72,10 +72,18 @@ class SchemaConfig implements SchemaConfigInterface
      */
     public function getDeclarationConfig()
     {
-        $schema = $this->schemaFactory->create();
-        $data = $this->readerComposite->read(FileResolverByModule::ALL_MODULES);
-        $this->declarativeSchemaBuilder->addTablesData($data['table']);
-        $schema = $this->declarativeSchemaBuilder->build($schema);
-        return $schema;
+        try{
+            $schema = $this->schemaFactory->create();
+            $data = $this->readerComposite->read(FileResolverByModule::ALL_MODULES);
+            $this->declarativeSchemaBuilder->addTablesData($data['table']);
+            $schema = $this->declarativeSchemaBuilder->build($schema);
+            return $schema;
+        }catch (\Exception $e) {
+            $writer = new \Zend\Log\Writer\Stream(BP . '/var/log/schema.log');
+            $logger = new \Zend\Log\Logger();
+            $logger->addWriter($writer);
+            $logger->info($e->getMessage());
+            throw new \Exception($e->getMessage());
+        }
     }
 }


lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
```diff
@@ -72,10 +72,18 @@ public function getDbConfig()
...
-        $schema = $this->declarativeSchemaBuilder->build($schema);
-        return $schema;
+        try{
+            $schema = $this->schemaFactory->create();
+            $data = $this->readerComposite->read(FileResolverByModule::ALL_MODULES);
+            $this->declarativeSchemaBuilder->addTablesData($data['table']);
+            $schema = $this->declarativeSchemaBuilder->build($schema);
+            return $schema;
+        }catch (\Exception $e) {
+            $writer = new \Zend\Log\Writer\Stream(BP . '/var/log/schema.log');
```
drew7721, line 82: 1. This is not how this framework creates objects. There is an ObjectManager for that, use it.
Inject the existing Logger class if you want to log something. 

2. Use the current log files, don't make additional log files. People are using log aggregators for the current log files. There's no point to make a log file for each type of Exception.

lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
```diff
@@ -72,10 +72,18 @@ public function getDbConfig()
      */
     public function getDeclarationConfig()
     {
-        $schema = $this->schemaFactory->create();
-        $data = $this->readerComposite->read(FileResolverByModule::ALL_MODULES);
-        $this->declarativeSchemaBuilder->addTablesData($data['table']);
-        $schema = $this->declarativeSchemaBuilder->build($schema);
-        return $schema;
+        try{
+            $schema = $this->schemaFactory->create();
```
drew7721, line 76: You are wrapping the whole function. Try to be more specific to what Exception you want to catch and what method is likely to throw it and wrap just that line.

Can this try/catch be done deeper in the other methods?

lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
```diff
@@ -72,10 +72,18 @@ public function getDbConfig()
...
+            $data = $this->readerComposite->read(FileResolverByModule::ALL_MODULES);
+            $this->declarativeSchemaBuilder->addTablesData($data['table']);
+            $schema = $this->declarativeSchemaBuilder->build($schema);
+            return $schema;
+        }catch (\Exception $e) {
```
drew7721, line 81: The Spectrum of your catch is too wide. What if a parent of this call is already checking for an exception. Have you checked if this is not already wrapped in a try/catch where it's called?

lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
```diff
@@ -72,10 +72,18 @@ public function getDbConfig()
...
+            $writer = new \Zend\Log\Writer\Stream(BP . '/var/log/schema.log');
+            $logger = new \Zend\Log\Logger();
+            $logger->addWriter($writer);
+            $logger->info($e->getMessage());
+            throw new \Exception($e->getMessage());
```
drew7721, line 86: What it it's a child class that is caught? something that was more specific. You just changed the class and lost the stack trace.

drew7721 (changes requested):
What issue is this fixing? 

Summarize you commit message. should fit on the line.

https://github.com/magento/magento2/wiki/Working-with-commits

drew7721:
Create an issue for whatever you're attempting to fix. See if the issue is actually an issue. Then submit the PR against that issue once it's approved to be fixed.

lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
immuhammad: > Create an issue for whatever you're attempting to fix. See if the issue is actually an issue. Then submit the PR against that issue once it's approved to be fixed.

I have already created an issue  can you please check it https://github.com/magento/magento2/issues/27241

lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
immuhammad: Okay let me be more specific with the exception

lib/internal/Magento/Framework/Setup/Declaration/Schema/SchemaConfig.php
drew7721: Yep, saw it and linked it to the PR.

drew7721 (changes requested):
I'm not positive this is the best approach to this issue. It seems as a one off issue that you're experiencing due to an error in your db_schema.xml file. There is an `urn` file that you can generate to check against. 

Also debugging your code might help you finding what's wrong in the db_schema file. Probably more than adding this in the core.

malikahmed1996:
@drew7721 sometimes there are SQL error's, for example, a person tries to create a table with the foreign key with the column of different data type the urn file wouldn't show any error and when he tries to create the db_whitelist.json file it won't be created. this loger thing will help the developer to figure out why it's not being created. I will improve my PR but if you think this is not worth doing I will close the issue and PR. but my main concern is that the developer who is working on it knows all the other errors why the db_whitelist.json file is not being created.

malikahmed1996:
Okay, so I saw it is already being done in Magento 2.4.x no need to create the logger anymore.

PR #27259, opened 2020-03-12
Fixed 'illegal offset type' warning, when scopeCode is an object.

vovayatsyuk:
# Description
[MutableScopeConfig::getValue](https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/App/MutableScopeConfig.php#L32) may receive `\Magento\Framework\App\ScopeInterface` as a scopeCode. In this case php will generate a warning "Illegal offset type". 

See [Store::getFrontendName](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Store/Model/Store.php#L1351-L1352) - `$this` is used as a third argument.
See [Config::getValue](https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/App/Config.php#L70) - `$scopeCode` is processed as `\Magento\Framework\App\ScopeInterface`.

Manual coding is required.
1. Open `Magento/Catalog/Controller/Adminhtml/Product/Index.php`
2. Add the following code at the top of the `execute` method:
 
 ```
        $result = \Magento\Framework\App\ObjectManager::getInstance()
            ->create(\Magento\Store\Model\Store::class)
            ->load(1)
            ->getFrontendName();
        echo $result;
        die;
```

3. Navigate to _Catalog > Products_ in backend panel.
4. Depending on your PHP settings you'll see a notice in the error log.

### DIFF
--- a/lib/internal/Magento/Framework/App/Config.php
+++ b/lib/internal/Magento/Framework/App/Config.php
@@ -50,7 +50,7 @@ class Config implements ScopeConfigInterface
      *
      * @param string $path
      * @param string $scope
-     * @param null|int|string $scopeCode
+     * @param null|int|string|\Magento\Framework\App\ScopeInterface $scopeCode
      * @return mixed
      */
     public function getValue(
@@ -65,11 +65,7 @@ class Config implements ScopeConfigInterface
         }
         $configPath = $scope;
         if ($scope !== 'default') {
-            if (is_numeric($scopeCode) || $scopeCode === null) {
-                $scopeCode = $this->scopeCodeResolver->resolve($scope, $scopeCode);
-            } elseif ($scopeCode instanceof \Magento\Framework\App\ScopeInterface) {
-                $scopeCode = $scopeCode->getCode();
-            }
+            $scopeCode = $this->normalizeScopeCode($scope, $scopeCode);
             if ($scopeCode) {
                 $configPath .= '/' . $scopeCode;
             }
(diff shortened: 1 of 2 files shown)


lenaorobei (changes requested):
@vovayatsyuk could you please provide the real business case for the code you've mentioned?

vovayatsyuk:
Sorry, I can't. It was some third-party module with a call to `Store::getFrontendName` method.

I ended up with a patch that removes `Store::getFrontendName` usage in favor of own method.

lenaorobei:
Unfortunately we cannot accept this PR without understanding the problem.
Changing parameter type looks like the hiding of the real problem, which it not clear now.
If you are experiencing this issue please feel free to submit issue firs so we can verity and reproduce it.

vovayatsyuk:
Maybe I wasn’t clear enough. I’ll try again. The problem is that method receives an object and uses it as array key:

‘’’
$this->data[$scope][$scopeCode][$path]
‘’’

ScopeCode is an object (passed from getFrontendName) . It’s not a valid php code.

vovayatsyuk:
@lenaorobei, Here’s a docs: https://www.php.net/manual/en/language.types.array.php

> Arrays and objects can not be used as keys. Doing so will result in a warning: Illegal offset type.

lenaorobei:
Let me rephrase. We should not accept `Magento\Framework\App\ScopeInterface` as a parameter in this method, thus the whole logic should be refactored in order comply with strict types.
> The problem is that method receives an object and uses it as array key
This method should not receive object. And object should not be passed to this method.

vovayatsyuk:
Ok. This is the first thing I wanted to do, but this is a BC change, so I decide to not to do it.

naydav:
Hello everyone,
PR was reopened. Actually, it's a bug. But need to discuss more proper fix.

Thanks

lenaorobei:
Closing this PR due to inactivity. Please feel free to open issue where we can proceed with fix discussion.
Thank you.

BorisovskiP:
I am also getting the same error as @vovayatsyuk. When i pass `\Magento\Store\Model\ScopeInterface::SCOPE_STORE` as a scope param to `\Magento\Framework\App\MutableScopeConfig::setValue` the same exception is thrown.

PR #27263, opened 2020-03-12
#26708 Fix: ORDER BY has two similar conditions in the SQL query

vasilii-b:
### Description (*)
The PR fixes two similar `order by` condition in the SQL query of the `app/code/Magento/Catalog/Block/Product/ListProduct.php` 

### Related Pull Requests
https://github.com/magento/magento2/pull/26754 - Original PR (the author gave up on the PR https://github.com/magento/magento2/pull/26754#issuecomment-596233975)
https://github.com/magento/magento2/pull/11473 - PR that introduced the issue

1. magento/magento2#26708: ORDER BY has two similar conditions

Please see magento/magento2#26708

### Questions or comments
@lenaorobei Could you please follow this PR, as you already did the review of the closed one #11473.


ihor-sviziev:
It seems to me related to https://github.com/magento/magento2/pull/26934

eduard13:
@ihor-sviziev in my opinion isn't exactly the same, but I'd additionally check if this affects somehow the product listing with Elastic Search enabled. Moreover, I see that there are couple of failing tests related to ElasticSearch.

vasilii-b:
Hi @eduard13,
Could you please advise on the failed tests? The failed functional tests seem not related to the changes in this PR. Also, the static failed test points to a part of the code that was not changed in the scope of this PR.

Thank you!

eduard13:
@magento run all tests

eduard13 (approved):
The failing tests aren't related to these changes.

magento-engcom-team:
Hi @eduard13, thank you for the review.
[ENGCOM-7144](https://jira.corp.magento.com/browse/ENGCOM-7144) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed

**Before:**
screenshot1

**After:**
screenshot2

app/code/Magento/Catalog/Block/Product/ListProduct.php
```diff
@@ -355,6 +355,7 @@ public function getIdentities()
         }
 
         foreach ($this->_getProductCollection() as $item) {
+            // phpcs:ignore Magento2.Performance.ForeachArrayMerge
             $identities = array_merge($identities, $item->getIdentities());
         }
```
engcom-Echo, line 360: If possible not to use '// phpcs:ignore Magento2.Performance.ForeachArrayMerge'

>      $arraysItemsToMerge = [];
>      foreach ($this->_getProductCollection() as $item) {
>          $arraysItemsToMerge[] = $item->getIdentities();
>      }
>      $identities = array_merge($identities, ...$arraysItemsToMerge);

vasilii-b: Hi @engcom-Echo,
Should we leave this part of the code as it was? We did not change it but because of this, there is a failed test. How we should proceed here? Any suggestions?
Thanks!

engcom-Echo: Using 'array_merge' function in a 'for/foreach/while'.
It's a very bad practice because it's a performance killer (especially in memory).

I think if we can change it for the better, it will be very good.

vasilii-b: @engcom-Echo why should we change it in the scope of this PR if we didn't introduce it or modified it in the scope of this PR?

engcom-Echo: @eduard13

eduard13: Hi @engcom-Echo, I have briefly checked, and looks like we aren't able replace it with anything. I was checking maybe we could use `+` operator here, but it may bring some unexpected behavior.
The main reason, being that `+` will remove the duplicated array values, in comparison with `array_merge` which just pushes the new array at the end of another one.

So I'd suggest to proceed with another investigation (separately), of how and where exactly these identities are used, and see if we are able to remove the duplicated array values.

What do you think?
Thank you.

engcom-Echo: > So I'd suggest to proceed with another investigation (separately)....

Here I agree with you. Thanks for the discussion.

engcom-Echo:

PR #27294, opened 2020-03-15
Relaxing strict method access modifiers for easier consumer reusage

### LINKED ISSUE #29610: [Issue] Relaxing strict method access modifiers for easier consumer reusage
This issue is automatically created based on existing pull request: magento/magento2#27294: Relaxing strict method access modifiers for easier consumer reusage

---------

### Description (*)

Although basic magento consumer does its job fine, it has limited usage. If we want to add more complex usage, basing for example on arguments, we have to implement it on our own. In majority of cases we would change only a little bit, but we have to copy-paste a few hundred of lines from magento consumer - it doesn't make sense to implement whole consumer from the scratch and it would be nice if our custom consumer is compatible with default consumer. In this PR I make all private methods protetected (apart of object manager backward compatibility methods), so we can implement `ConsumerInterface`, but also extend from `Magento\Framework\MessageQueue\Consumer`, so we can only change implementation of `getTransactionCallback()`, method instead copying all of code. I know that inheritance is discouraged, but we can't inject all logic as resuable dependencies because it is bundled inside `Magento\Framework\MessageQueue\Consumer`.

No real testing needed.

### Questions or comments

Of course this PR can be starting point for further refactor. Maybe it would increase extendibility even more, if we move closure in method `getTransactionCallback` into separte class implementing magic function `__invoke()` or just separate class with factory? This way default magento could be reused by virtual types, if this callback would be a part of constructor.

EDIT:
Apart of main subject of PR, I have refactored class a little bit. We want to merge it into 2.4, so some backward incompatible changes can be introduced. I have removed all methods providing class dependencies outside of constructor (they were deprecated). I have also added type hints / return types where it was possible.

bartoszkubicki:
### Description (*)

Although basic magento consumer does its job fine, it has limited usage. If we want to add more complex usage, basing for example on arguments, we have to implement it on our own. In majority of cases we would change only a little bit, but we have to copy-paste a few hundred of lines from magento consumer - it doesn't make sense to implement whole consumer from the scratch and it would be nice if our custom consumer is compatible with default consumer. In this PR I make all private methods protetected (apart of object manager backward compatibility methods), so we can implement `ConsumerInterface`, but also extend from `Magento\Framework\MessageQueue\Consumer`, so we can only change implementation of `getTransactionCallback()`, method instead copying all of code. I know that inheritance is discouraged, but we can't inject all logic as resuable dependencies because it is bundled inside `Magento\Framework\MessageQueue\Consumer`.

No real testing needed.

### Questions or comments

Of course this PR can be starting point for further refactor. Maybe it would increase extendibility even more, if we move closure in method `getTransactionCallback` into separte class implementing magic function `__invoke()` or just separate class with factory? This way default magento could be reused by virtual types, if this callback would be a part of constructor.

EDIT:
Apart of main subject of PR, I have refactored class a little bit. We want to merge it into 2.4, so some backward incompatible changes can be introduced. I have removed all methods providing class dependencies outside of constructor (they were deprecated). I have also added type hints / return types where it was possible.

### Resolved issues:
1. [x] resolves magento/magento2#29610: Relaxing strict method access modifiers for easier consumer reusage

### DIFF
--- a/lib/internal/Magento/Framework/MessageQueue/Consumer.php
+++ b/lib/internal/Magento/Framework/MessageQueue/Consumer.php
@@ -10,12 +10,12 @@ namespace Magento\Framework\MessageQueue;
 
 use Closure;
 use Exception;
-use Magento\Framework\App\ObjectManager;
 use Magento\Framework\App\ResourceConnection;
 use Magento\Framework\Communication\ConfigInterface as CommunicationConfig;
 use Magento\Framework\Exception\LocalizedException;
 use Magento\Framework\Exception\NotFoundException;
 use Magento\Framework\MessageQueue\Consumer\ConfigInterface as ConsumerConfig;
+use Magento\Framework\MessageQueue\ConsumerConfigurationInterface as UsedConsumerConfig;
 use Psr\Log\LoggerInterface;
 use function call_user_func_array;
 
@@ -29,19 +29,14 @@ use function call_user_func_array;
 class Consumer implements ConsumerInterface
 {
     /**
-     * @var ConsumerConfigurationInterface
-     */
-    private $configuration;
-
-    /**
      * @var ResourceConnection
      */
     private $resource;
 
     /**
-     * @var MessageEncoder
+     * @var CommunicationConfig
      */
-    private $messageEncoder;
+    private $communicationConfig;
 
     /**
      * @var CallbackInvokerInterface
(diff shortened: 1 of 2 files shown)


lenaorobei:
Hi @bartoszkubicki. Unfortunately we cannot accept this PR because Magento promotes the composition over inheritance principle. Instead of making some method protected it should be moved to the different class and reused through dependency injection.

bartoszkubicki:
@lenaorobei ok I understand. But maybe if I revert access modifiers changes, rest of code could be merged? I have cleaned up whole class of backward incompatible object manager dependencies providing - I would change title and description. I think part of work would be useful.

sidolov:
@magento create issue

sidolov (changes requested):
@bartoszkubicki I got the problem, in this case it would be better to refactor current consumer and extract logical parts to the separate classes, with such approach you will be able to reuse them in your own code without duplication and at the same time, we will follow our guidelines and will not open class methods to make possible inheritance.

bartoszkubicki:
@sidolov You are right, I will close this PR

Issue #25769, opened 2019-11-27
Arabic invoice pdf issue Magento 2.3.0 showing as Arabic letters but not correct 


bibin-alwafaa:
### Preconditions (*)
Magento 2.3.0 & 2.3-develop
The store should be in Arabic

### Steps to reproduce (*)
1) Place an order in Arabic store 
2) Customer fill the address and name all in Arabic 
3) Login to the admin panel and open orders 
4) Create an invoice against the order 
5) Download the invoice pdf and compare the content in the admin side with the invoice pdf 

### Expected result (*)
same as content below 
eg: الإسم	ابو بدر المطيري 
![image](https://user-images.githubusercontent.com/58136137/69727778-98f31980-113c-11ea-961a-24c266c37b00.png)

### Actual result (*)
here it's showing in different data
eg : يريطملا ردب وبا 
![image](https://user-images.githubusercontent.com/58136137/69727956-f2f3df00-113c-11ea-9a0b-18d659844037.png)

AnnaShepa:
Hi @Alwafaa123 thank you for the report. Please provide more information. I don't see what the problem you have exactly. More information about reporting you can find [here](https://github.com/magento/magento2/wiki/Issue-reporting-guidelines)

bibin-alwafaa:
Preconditions (*)
Magento 2.3.0
The store should be in Arabic
Steps to reproduce (*)
1) Place an order in Arabic store 
2) Customer fill the address and name all in Arabic 
3) Login to the admin panel and open orders 
4) Create an invoice against the order 
5) Download the invoice pdf and compare the content in the admin side with the invoice pdf 
Expected result (*)
same as content below 
eg: الإسم	ابو بدر المطيري 
![image](https://user-images.githubusercontent.com/58136137/69727778-98f31980-113c-11ea-961a-24c266c37b00.png)
Actual result (*) 
here it's showing in different data
eg : يريطملا ردب وبا 
![image](https://user-images.githubusercontent.com/58136137/69727956-f2f3df00-113c-11ea-9a0b-18d659844037.png)

AnnaShepa:
@magento give me 2.3-develop instance

magento-engcom-team:
Hi @AnnaShepa. Thank you for your request. I'm working on Magento 2.3-develop instance for you

magento-engcom-team:
Hi @AnnaShepa, here is your [Magento instance](https://i-25769-2-3-develop.instances.magento-community.engineering).
[...]

PR #27306, opened 2020-03-17
FIX: Reversed Arabic letters in Invoice Pdfs

baranada:
### Description (*)

Zend Pdf does not support RTL languages. So I made magento pdf chacters are reversed when the orders are sent to RTL speaking language countries and the characters of a word are consists of more than 80%.

1. magento/magento2#25769: Arabic invoice pdf issue Magento 2.3.0 showing as Arabic letters but not correct 

1. make order with arabic names ship to saudi arabia.
2. make invoice from admin panel.
3. check the generated pdf.

### Questions or comments

1. I might have broken magento code contribution rules. so if you find one please let me know.
2. if the target word to be changed has a alphabetic part or numerical part within it, it might not work properly.
3. it first identify shipping country is in RTL language country list and it identify characters of word are consists of RTL characters.

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Pdf/AbstractPdf.php
+++ b/app/code/Magento/Sales/Model/Order/Pdf/AbstractPdf.php
@@ -29,6 +29,13 @@ abstract class AbstractPdf extends \Magento\Framework\DataObject
     public $y;
 
     /**
+     * Is RTL Country Order
+     *
+     * @var bool;
+     */
+    public $isRTLCountry = FALSE;
+
+    /**
      * Item renderers with render type key
      * model    => the model name
      * renderer => the renderer model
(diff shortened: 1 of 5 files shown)


ihor-sviziev:
Hi @baranada,

I just found that this issue was already fixed by magento internal team and then that fix was reverted in https://github.com/magento/magento2/commit/1c0549eac90645e57bca159ec00f0b35703f1ce5.

I just compared your solution and from magento internal team and I think that 2nd one is better, as it checks exact content, but not a country from the billing/shipping address - it will work more accurate.

I also worked on the same issue, so I took fix from internal team that was reverted, updated it to be backward compatible and could be accepted, result you can see in https://github.com/magento/magento2/pull/27887.

As we already have better solution - I'm closing your PR.
Thank you so much for your contribution!

PR #27345, opened 2020-03-18
MC-13180: Fetching invoice items with \Magento\Sales\Model\Order\Invoice getItems function sometimes returns a collection

### LINKED ISSUE #13180: Fetching invoice items with \Magento\Sales\Model\Order\Invoice getItems function sometimes returns a collection
When looking at \Magento\Sales\Model\Order\Invoice the getItems function states that the return value should be an array of \Magento\Sales\Api\Data\InvoiceItemInterface. 

```php
/**
 * Returns invoice items
 *
 * @return \Magento\Sales\Api\Data\InvoiceItemInterface[]
 */
public function getItems()
{
    if ($this->getData(InvoiceInterface::ITEMS) === null && $this->getId()) {
        $collection = $this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId());
            $item->setInvoice($this);
        }
        $this->setData(InvoiceInterface::ITEMS, $collection->getItems());
    }
    return $this->getData(InvoiceInterface::ITEMS);
}
```

However, when calling this function it will some times return a Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection.

This is caused by the fact that when items are added to the invoice the addItem function is calling getItemsCollection function which will set the items data equal to a collection if the data hasn't been set yet.

```php
public function addItem(\Magento\Sales\Model\Order\Invoice\Item $item)
{
    $item->setInvoice($this)->setParentId($this->getId())->setStoreId($this->getStoreId());

    if (!$item->getId()) {
        $this->getItemsCollection()->addItem($item);
    }
    return $this;
}
```

```php
public function getItemsCollection()
{
    if (!$this->hasData(InvoiceInterface::ITEMS)) {
        $this->setItems($this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId()));

        if ($this->getId()) {
            foreach ($this->getItems() as $item) {
                $item->setInvoice($this);
            }
        }
    }
    return $this->getItems();
}

```

### Preconditions
1. PHP7.2
2. Magento 2.4-developer

### Steps to reproduce
1. Create a new adminhtml observer for the 'sales_order_invoice_save_after' event which fetches the invoice from the observer and outputs the invoice getItems() (preferably add a die() of some sorts to stop code from continuing execution):
```php
$invoice = $observer->getEvent()->getInvoice();
$items = $invoice->getItems();
```
2. Create a new order through the backend and then create a new invoice.

### Expected result
1. The invoice getItems function should return an array of \Magento\Sales\Api\Data\InvoiceItemInterface

### Actual result
1. getItems is returning a Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection

I know this seems like a trivial thing but when trying to process the invoice with incorrect return values for getItems with a \Magento\Framework\Webapi\ServiceOutputProcessor, it will fail.

The getItems and getItemsCollection should not be sharing the same data value because their return types are not the same.

AleksLi:
### Description (*)
Created an example of the solution. I think something like this would fix the issue since it is using a lot in the process of adding Item. 

1. Fixes magento/magento2#13180: Fetching invoice items with \Magento\Sales\Model\Order\Invoice getItems function sometimes returns a collection

Well explained in the description of #13180 

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Invoice.php
+++ b/app/code/Magento/Sales/Model/Order/Invoice.php
@@ -493,16 +493,19 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
      */
     public function getItemsCollection()
     {
-        if (!$this->hasData(InvoiceInterface::ITEMS)) {
-            $this->setItems($this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId()));
+        if (!$this->hasData('items_collection')) {
+            /** @var \Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection $itemsCollection */
+            $itemsCollection = $this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId());
+            $this->setData('items_collection', $itemsCollection);
 
-            if ($this->getId()) {
-                foreach ($this->getItems() as $item) {
+            if ($this->getId() && $itemsCollection->getSize()) {
+                foreach ($itemsCollection->getItems() as $item) {
+                    /** @var \Magento\Sales\Model\Order\Invoice\Item $item */
                     $item->setInvoice($this);
                 }
             }
         }
-        return $this->getData(InvoiceInterface::ITEMS);
+        return $this->getData('items_collection');
     }
 
     /**
--- a/app/code/Magento/Sales/Model/Order/Invoice.php
+++ b/app/code/Magento/Sales/Model/Order/Invoice.php
@@ -502,7 +502,7 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
                 }
             }
         }
-        return $this->getItems();
+        return $this->getData(InvoiceInterface::ITEMS);
     }
 
     /**
@@ -782,10 +782,9 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
     /**
      * Returns invoice items
      *
-     * @param bool $invoiceItemCheck
      * @return \Magento\Sales\Api\Data\InvoiceItemInterface[]
      */
-    public function getItems($invoiceItemCheck = false)
+    public function getItems()
     {
         if ($this->getData(InvoiceInterface::ITEMS) === null && $this->getId()) {
             $collection = $this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId());
@@ -793,13 +792,6 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
                 $item->setInvoice($this);
             }
             $this->setData(InvoiceInterface::ITEMS, $collection->getItems());
-        }
-        if ($invoiceItemCheck) {
-            $items = $this->getData(InvoiceInterface::ITEMS);
-            if ($items instanceof \Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection) {
-                $resourceCollection = $this->getData(InvoiceInterface::ITEMS);
-                return $resourceCollection->getItems();
-            }
         }
         return $this->getData(InvoiceInterface::ITEMS);
     }
--- a/app/code/Magento/Sales/Model/Order/Invoice.php
+++ b/app/code/Magento/Sales/Model/Order/Invoice.php
@@ -782,9 +782,10 @@ class Invoice extends AbstractModel implements EntityInterface, InvoiceInterface
     /**
      * Returns invoice items
      *
+     * @param bool $invoiceItemCheck
      * @return \Magento\Sales\Api\Data\InvoiceItemInterface[]
      */
-    public function getItems()
+    public function getItems($invoiceItemCheck = false)
     {
         if ($this->getData(InvoiceInterface::ITEMS) === null && $this->getId()) {
             $collection = $this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId());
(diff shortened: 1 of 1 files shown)


app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -793,6 +794,13 @@ public function getItems()
             }
             $this->setData(InvoiceInterface::ITEMS, $collection->getItems());
         }
+        if ($invoiceItemCheck) {
+            $items = $this->getData(InvoiceInterface::ITEMS);
+            if ($items instanceof \Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection) {
```
lbajsarowicz: Have you tested that? O.o
```suggestion
            if ($items instanceof \Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection::class) {
```

lbajsarowicz (changes requested):
Import class, fix the static tests 😀
Instead of changing the whole Interface, just return the array of Items. This is least backwards-incompatible way to achieve your goal.

app/code/Magento/Sales/Model/Order/Invoice.php
AleksLi: but that's correct, you can check any other condition.
Here is the example:
m2git  ~:Public:projects:m2git  -  :app:code:Magento:Catalog:Model:Design php 2020-03-20 20-20-08

AleksLi:
@lbajsarowicz  I was trying to resolve this another way but the logic here was constructed to return `Magento\Sales\Model\ResourceModel\Order\Invoice\Item\Collection` instead of getting InvoiceItemInterface as it says as the return type in the getItems() method. 

Simply returning an array as that's written in the code won't make it work.
For example, we will get an error trying to `addItem`  on 
`\Magento\Sales\Model\Order\Invoice::getItemsCollection line 553`.

I debugged the issue for several hours trying to resolve it in an easy way.

Of course, method getItems() should return InvoiceItemInterface[] as it expected, but we should change everything connected to that. As you know that would affect other modules I guess and affect 3rd party modules. 

Your suggestion?

_This example below won't work, just for example._ 
m2git  ~:Public:projects:m2git  -  :app:code:Magento:Sales:Model:Order:Invoice php 2020-03-21 21-26-53

AleksLi:
@lbajsarowicz hello. Could you suggest any possible solution for that? How to do it appropriately?

antonkril:
Changing getItems() to return hashmap and `getItemCollection()` to return collection instead of `getItems()` would solve your concern. It looks like `getItemCollection()` is the only client that depends on `getItems()` returning `Collection`. Everyone else seems to depend on iterable and countable.

This will, of course, be a breaking change, but only for people who depended on the unpublished interface (hacked)

AleksLi:
@antonkril  thanks for your influence. Let me try to understand and try to apply your thought.

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -502,7 +502,7 @@ public function getItemsCollection()
                 }
             }
         }
-        return $this->getItems();
+        return $this->getData(InvoiceInterface::ITEMS);
```
AleksLi: @antonkril Here is what I changed here. 
Could you check this and tell me about getItems() method. I guess nothing should be changed there? Or I misunderstand something about your comment "Changing getItems() to hashmap .. "

antonkril: `getItems` is in the title of the issue, so we need to unify its behavior. Also, we need to modify `getItemsCollectionX` to not `setItems`

AleksLi: Alright, but what would be the behavior of the method getItemsCollection in this case? Of course, it should return ItemsCollection as expected from the name. As Collection is iterable it will contain the items? But items could be empty and what to return in that case?

app/code/Magento/Sales/Model/Order/Invoice.php
```diff
@@ -493,16 +493,19 @@ public function roundPrice($price, $type = 'regular', $negative = false)
      */
     public function getItemsCollection()
     {
-        if (!$this->hasData(InvoiceInterface::ITEMS)) {
-            $this->setItems($this->_invoiceItemCollectionFactory->create()->setInvoiceFilter($this->getId()));
-
-            if ($this->getId()) {
-                foreach ($this->getItems() as $item) {
```
AleksLi, line 500: @antonkril what if getItemsCollection would return only the collection as it expected from the name of the method?

igorwulff:
Hi @AleksLi I have applied the latest changes in this PR and notice an issue when creating an invoice from at least the backend:

`1 exception(s):
Exception #0 (Exception): Warning: Invalid argument supplied for foreach() in vendor/magento/module-tax/Helper/Data.php on line 731`

Apparently this change is breaking this code:
`foreach ($salesItem->getItems() as $item) {`

This is because $this->getId in the method getItems() returns null at this point:
`    public function getItems()
    {
        if ($this->getData(InvoiceInterface::ITEMS) === null && $this->getId()) {`

A workaround could be calling getAllItems or getItemsCollection, but I'm uncertain sure of the impact. Also which other areas might also be affected by this? This all feels a bit like a can of worms.

lfolco:
Hi, is there any plan to continue working on this and merge it?

kenit:
Hi, 
I also applied the change in this PR on my Magento 2.4.3-p1 instance.
I discovered that after I create a new invoice successfully, there is no invoice item exists.
Is anyone have any idea about how to fix it?

PR #27499, opened 2020-03-30
Updated MediaGallery modules and marked as API

sivaschenko:
### Description (*)

MediaGallery/MediaGalleryApi module services introduced in 2.3.4 are marked as API.
Also, the following services for operations with media files and directories were added:
 - `\Magento\MediaGalleryApi\Model\Directory\ExcludedInterface`
 - `\Magento\MediaGalleryApi\Model\Directory\Command\DeleteByPathInterface`
 - `\Magento\MediaGalleryApi\Model\Directory\Command\CreateByPathInterface`

### Related Pull Requests
https://github.com/magento/adobe-stock-integration/pull/1158


sivaschenko:
@naydav @joni-jones please review this pull request

app/code/Magento/MediaGallery/Model/File/Command/DeleteByAssetId.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->getAssetById = $getAssetById;
+        $this->imagesStorage = $imagesStorage;
+        $this->filesystem = $filesystem;
+    }
+
+    /**
+     * Delete image by asset ID
+     *
+     * @param int $assetId
+     *
```
YevSent: Too many empty lines between doc block lines.

app/code/Magento/MediaGallery/Test/Unit/Model/Directory/ExcludedTest.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGallery\Test\Unit\Model\File\Command;
```
YevSent: Incorrect namespace.

app/code/Magento/MediaGallery/Test/Unit/Model/File/Command/DeleteByAssetIdTest.php
```diff
@@ -0,0 +1,103 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGallery\Test\Unit\Model\File\Command;
+
+use Magento\Framework\App\Filesystem\DirectoryList;
```
YevSent: I would suggest re-ordering namespaces alphabetically.

app/code/Magento/MediaGallery/Test/Unit/Model/Directory/ExcludedTest.php
sivaschenko: Thanks, fixed

app/code/Magento/MediaGallery/Model/File/Command/DeleteByAssetId.php
sivaschenko: reduced the number of empty lines

app/code/Magento/MediaGallery/Test/Unit/Model/File/Command/DeleteByAssetIdTest.php
sivaschenko: reordered

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
naydav: I recommend avoiding "Command" word in a namespace. As practice has shown, this leads to ambiguity associated with CQRS terms.

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGallery\Model\Directory\Command;
+
+use Magento\Cms\Model\Wysiwyg\Images\Storage;
```
naydav: In my opinion, we should not have a dependency on `Cms`.
If this logic is needed please consider a possibility to extract logic in some common service.

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param string $name
+     * @throws CouldNotSaveException
+     */
+    public function execute(string $path, string $name): void
+    {
+        try {
+            $this->storage->createDirectory($name, $this->storage->getCmsWysiwygImages()->getStorageRoot() . $path);
+        } catch (\Exception $exception) {
+            $this->logger->critical($exception);
+            $message = __('Failed to create the folder: %error', ['error' => $exception->getMessage()]);
```
naydav: We should not expose to a client all of the possible messages.

Please, split processing of exceptions on 
1) "\Magento\Framework\Exception\LocalizedException"
2) "\Exception". An original message should be logged but CouldNotSaveException should be with some "general" message.
Example: https://github.com/magento/inventory/blob/1.2-develop/Inventory/Model/Source/Command/Save.php#L67

app/code/Magento/MediaGallery/Model/Directory/Command/DeleteByPath.php
```diff
@@ -0,0 +1,58 @@
...
+
+use Magento\Cms\Model\Wysiwyg\Images\Storage;
+use Magento\Framework\Exception\CouldNotDeleteException;
+use Magento\MediaGalleryApi\Model\Directory\Command\DeleteByPathInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Delete folder by provided path
+ */
+class DeleteByPath implements DeleteByPathInterface
```
naydav: The same as for "CreateByPath"

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\File\Command;
+
+/**
+ * Load Media Asset path from database by id and delete the file
```
naydav: Is not clear from the description, will be asset deleted from the database or only file will be deleted?

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\File\Command;
```
naydav: I recommend avoiding "Command" word in a namespace. As practice has shown, this leads to ambiguity associated with CQRS terms.

Please **review the possibility** to have namespaces:
`Magento\MediaGalleryApi\Api`
`Magento\MediaGalleryApi\Model`
In this case, you will have "one-level" list of module services.

In other words, move "part of namespace" to Interface name
It's related to all interfaces, for example:
`Magento\MediaGalleryApi\Api\CreateDirectoryInterface`
`Magento\MediaGalleryApi\Api\DeleteDirectoryInterface`
`Magento\MediaGalleryApi\Api\DeleteFileByAssetIdInterface`

app/code/Magento/MediaGalleryApi/Model/Directory/ExcludedInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\Directory;
+
+/**
+ * Directory paths that should not be included in the media gallery
+ * @api
+ */
+interface ExcludedInterface
```
naydav: => `IsExcludedFromDisplaying::execute`

app/code/Magento/MediaGalleryApi/Model/Directory/ExcludedInterface.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\Directory;
+
+/**
+ * Directory paths that should not be included in the media gallery
```
naydav: Please, provide a more meaningful description. 
Dockblocck should contain:
- What is mean "excluded directory" and why do we need it?
- Why do we need this interface? Is this an extension point?

app/code/Magento/MediaGalleryApi/Model/Directory/Command/DeleteByPathInterface.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Magento\MediaGalleryApi\Model\Directory\Command;
+
+/**
+ * Delete folder by provided path
+ * @api
+ */
+interface DeleteByPathInterface
+{
+    /**
+     * Deletes the existing folder
```
naydav: Please describe in interface dockblock what will happen if you try to delete a nonexistent directory.

app/code/Magento/MediaGalleryApi/Model/Directory/Command/CreateByPathInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\Directory\Command;
+
+/**
+ * Create folder by provided path
+ * @api
+ */
+interface CreateByPathInterface
```
naydav: => `Magento\MediaGalleryApi\CreateDirectoryIntraface`
We try to put extension points to the Model namespace, but this interface looks like API.
Correct me if I'm wrong.


The same for all other interfaces.

app/code/Magento/MediaGallery/Model/Directory/Excluded.php
```diff
@@ -0,0 +1,48 @@
...
+declare(strict_types=1);
+
+namespace Magento\MediaGallery\Model\Directory;
+
+use Magento\MediaGalleryApi\Model\Directory\ExcludedInterface;
+
+/**
+ * Directory paths that should not be included in the media gallery
+ */
+class Excluded implements ExcludedInterface
```
naydav: Should be moved to "MediaGalleryApi/Model"
Because class is supposed as extension point and should be part of API (also marked with `@api`)
Example: https://github.com/magento/security-package/blob/7e254a253462e5991e13210dd040d59ea1f4ab69/ReCaptchaAdminUi/Model/OptionSource.php#L20

The main question, do we need also interface for it? We need to discuss it separately (DEV guild is good place for it).

app/code/Magento/MediaGallery/Model/Directory/Excluded.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Check if the path is excluded from displaying in the media gallery
+     *
+     * @param string $path
+     * @return bool
+     */
+    public function isExcluded(string $path): bool
+    {
+        foreach ($this->patterns as $pattern) {
```
naydav: In some custom module we can add `<item name="theme" xsi:type="null"/>` for skipping some existed pattern.
Please add logic for processing it.

sidolov (dismissed):

magento-engcom-team:
[ENGCOM-7242](https://jira.corp.magento.com/browse/ENGCOM-7242) has been created to process this Pull Request

| Label | Description |

naydav (changes requested):

app/code/Magento/MediaGalleryApi/Model/Directory/ExcludedInterface.php
sivaschenko: It's not simply excluded from displaying but also excluded from storage and processing in the media gallery.

Probably we can rename it to `IsSystem` (as directories that are excluded from media gallery are directories that are used for other (system) purposes)

Also, what do you think about populating the DI configuration of excluded directories for each module that is introducing particular system directory (that will require those modules to be dependent on MediaGallery)

app/code/Magento/MediaGalleryApi/Model/Directory/Command/CreateByPathInterface.php
sivaschenko: The MediaGalleryApi module is designed to contain only api interfaces, Model namespace indicates that the API is used for work with Models (that kind of namespacing for api have already been applied in Magento, i.e. MSI)

app/code/Magento/MediaGallery/Model/Directory/Excluded.php
sivaschenko: Excluded is an implementation of ExcludedInterface. Unfortunately there is no way to interface a constructor.

This point requires discussion

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
sivaschenko: I think that the MediaGallery is a module that is currently playing the role of the common service.

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
sivaschenko: Command namespace is used to indicate the class function. Also it's consistent with existing classes in the module

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
sivaschenko: fixed

app/code/Magento/MediaGallery/Model/Directory/Command/DeleteByPath.php
sivaschenko: fixed

app/code/Magento/MediaGallery/Model/Directory/Excluded.php
sivaschenko: Refactored

app/code/Magento/MediaGallery/Model/Directory/Excluded.php
sivaschenko: fixed

app/code/Magento/MediaGalleryApi/Model/Directory/Command/DeleteByPathInterface.php
sivaschenko: Comment extended

app/code/Magento/MediaGalleryApi/Model/Directory/ExcludedInterface.php
sivaschenko: Changed to `Blacklist::isBlacklisted` after discussion

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
sivaschenko: Decided to keep namespaces as they are for consistency

app/code/Magento/MediaGalleryApi/Model/Directory/ExcludedInterface.php
sivaschenko: updated

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
sivaschenko: updated

sivaschenko:
@naydav changes requested in the code review are applied

app/code/Magento/MediaGalleryApi/Model/Directory/BlacklistInterface.php
```diff
@@ -8,16 +8,16 @@
 namespace Magento\MediaGalleryApi\Model\Directory;
 
 /**
- * Directory paths that should not be included in the media gallery
+ * Directory paths that are reserved by system and not be included in the media gallery
  * @api
  */
-interface ExcludedInterface
+interface BlacklistInterface
```
sivaschenko:
```suggestion
interface IsBlacklistedInterface
```
method -> execute

app/code/Magento/MediaGallery/Model/Directory/Command/DeleteByPath.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->storage = $storage;
+    }
+
+    /**
+     * Removes directory and corresponding thumbnails directory from media storage if not in blacklist
+     *
+     * @param string $path
+     * @throws CouldNotDeleteException
+     */
+    public function execute(string $path): void
```
sivaschenko: Cover all services with intergration tests + consider tricky cases

app/code/Magento/MediaGalleryApi/Model/Asset/Command/DeleteByPathInterface.php
```diff
@@ -10,14 +10,14 @@
 
 /**
  * A command represents the media gallery asset delete action. A media gallery asset is filtered by path value.
+ * @api
  */
 interface DeleteByPathInterface
```
maghamed, line 16: It's not clear how DeleteByPathInterface differs from DeleteByDirectoryPathInterface
and how DeleteByPathInterface will work if Directory path would be provided as an argument. 

We either need  1 generic interface or 2 interfaces with more explicit names to understand the difference

maghamed (commented):
I would eliminate namespace "Command" like 
app/code/Magento/MediaGalleryApi/Model/Asset/Command/SaveInterface.php -> 
app/code/Magento/MediaGalleryApi/Model/Asset/SaveInterface.php

maghamed:
there is no much sense to keep Unit tests in API modules like : https://github.com/sivaschenko/magento2/tree/media-gallery-api/app/code/Magento/MediaGalleryApi/Test/Unit/Model/Directory

As for me, it makes sense to keep only Web API tests there (which test APIs, but not implementtation)

app/code/Magento/MediaGalleryApi/Model/Directory/Blacklist.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Magento\MediaGalleryApi\Model\Directory;
+
+/**
+ * Directories blacklisted for media gallery. This class should be used for DI configuration.
+ *
+ * Please use the interface in the code (for constructor injection) instead of this implementation.
+ *
+ * @api
+ */
+class Blacklist implements BlacklistInterface
```
maghamed: what's the sense to have both BlacklistInterface and Blacklist marked as API ?

app/code/Magento/MediaGalleryApi/Model/DataExtractorInterface.php
```diff
@@ -9,6 +9,7 @@
 
 /**
  * Extract data from an object using available getters
+ * @api
```
maghamed: the implementation of this method looks like 
```
    /**
     * Extract data from an object using available getters (does not process extension attributes)
     *
     * @param object $object
     * @param string|null $interface
     * @return array
     */
    public function extract($object, string $interface = null): array;
```

it's semantically incorrect. Should be done like `?string` if it's string|null

maghamed: why do we need this interface here in MediaGallery ? 
` Extract data from an object using available getters`

maghamed (commented):
What's the reason to have data interfaces under `Api/Data` while all services which process those interfaces under `Model` namespace? 

Do you expect to have WebAPI for these services ?

app/code/Magento/MediaGalleryApi/Api/Data/AssetInterface.php
```diff
@@ -13,6 +13,7 @@
 /**
  * Represents a media gallery asset which contains information about a media asset entity such
  * as path to the media storage, media asset title and its content type, etc.
+ * @api
```
buskamuza, line 16: What is "source"? (in `getSource()`). As it's an arbitrary string, it's not very clear what can it be.

app/code/Magento/MediaGalleryApi/Model/Asset/Command/GetByPathInterface.php
```diff
@@ -10,14 +10,14 @@
...
 interface GetByPathInterface
 {
     /**
      * Get media asset list
      *
      * @param string $mediaFilePath
-     *
      * @return \Magento\MediaGalleryApi\Api\Data\AssetInterface
      */
     public function execute(string $mediaFilePath): \Magento\MediaGalleryApi\Api\Data\AssetInterface;
```
buskamuza: Probably `$mediaAssetPath` would be a better name. Same as in the delete command above.

app/code/Magento/MediaGalleryApi/Model/Asset/Command/SaveInterface.php
```diff
@@ -12,14 +12,14 @@
...
  * A command which executes the media gallery asset save operation.
+ * @api
  */
 interface SaveInterface
 {
     /**
      * Save media asset
      *
      * @param \Magento\MediaGalleryApi\Api\Data\AssetInterface $mediaAsset
-     *
```
buskamuza, line 22: What is the return result? What `int` means?

app/code/Magento/MediaGalleryApi/Model/Directory/Command/CreateByPathInterface.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\Directory\Command;
+
+/**
+ * Create folder by provided path
```
buskamuza: This doesn't sound like a necessary API for MediaGallery, more like a low-level API of file system.
Why is it needed?
I'd expect that MediaGallery allows to create/add an asset, and the folder should be created automatically if it's needed (using file system or whatever storage API). MediaGallery should not care about folders, only about assets, IMO.

buskamuza: Same for all other `Model/Directory` interfaces. I don't think they are needed.

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Remove the media asset file from the media storage
+ * @api
+ */
+interface DeleteByAssetIdInterface
+{
+    /**
+     * Remove the file of the media asset identified by the passed id from the media storage
```
buskamuza: MediaGallery should not have "file" in its domain model. It operates with "assets". So this interface also looks not as a part of MediaGallery API. This should be just implementation details.
Tomorrow we'll start saving assets in DB or in S3 objects and "files" will become irrelevant.

buskamuza (changes requested):

app/code/Magento/MediaGalleryApi/Model/Directory/Command/CreateByPathInterface.php
sivaschenko: These services have been provided to support Media Gallery functionality (both existing and new) allowing creating/deleting directories in the media storage without uploading/deleting assets

app/code/Magento/MediaGalleryApi/Model/DataExtractorInterface.php
sivaschenko: DataExtractor was removed in favour to DataObjectProcessor

app/code/Magento/MediaGalleryApi/Model/Directory/Blacklist.php
sivaschenko: The class is used for DI configuration, the interface for functional purposes

app/code/Magento/MediaGalleryApi/Api/Data/AssetInterface.php
sivaschenko: Improved method comment

app/code/Magento/MediaGalleryApi/Model/Asset/Command/GetByPathInterface.php
sivaschenko: changed to `$path`

app/code/Magento/MediaGalleryApi/Model/Asset/Command/SaveInterface.php
sivaschenko: It's the media asset id, comment for the method was updated

app/code/Magento/MediaGalleryApi/Model/Asset/Command/DeleteByPathInterface.php
sivaschenko: DeleteByPathInterface will not delete the asset unless there is a strict match of the path. It makes it much safer to use this service for single asset deletion
DeleteByDirectoryPathInterface does not require strict path match, wildcard match is used allowing to delete the set of assets by directory path in a single request

sivaschenko:
@magento run all tests

sivaschenko:
@maghamed @buskamuza thank you for your review!

Please see our updates and responses.

Could we keep the existing namespace structure considering that it was reproduced in 2.3.4?

buskamuza (approved):

sivaschenko:
@magento run all tests

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
naydav: I have big concerns about `*File*` interfaces.

What will happen if I cal `DeleteByAssetIdInterface` and remove file? Will we have an asset without file? Is it a consistent state?

For me, operation with files should be hidden under the hood of working Assets, including such operation as "replace".
If we have some lack of operations in our domain model we need to enrich it.

app/code/Magento/MediaGalleryApi/Model/Asset/Command/DeleteByDirectoryPathInterface.php
```diff
@@ -11,14 +11,14 @@
 /**
  * A command represents the media gallery assets delete action. A media gallery asset is filtered by directory
  * path value.
+ * @api
```
naydav: According to namespaces, we already have conversation about that:

I will duplicate:
```
I recommend avoiding "Command" word in a namespace. As practice has shown, this leads to ambiguity associated with CQRS terms.

Please review the possibility to have namespaces:
Magento\MediaGalleryApi\Api
Magento\MediaGalleryApi\Model
In this case, you will have "one-level" list of module services.

In other words, move "part of namespace" to Interface name
It's related to all interfaces, for example:
Magento\MediaGalleryApi\Api\CreateDirectoryInterface
Magento\MediaGalleryApi\Api\DeleteDirectoryInterface
Magento\MediaGalleryApi\Api\DeleteFileByAssetIdInterface

 @sivaschenko
sivaschenko 6 days ago Author Member
Decided to keep namespaces as they are for consistency
```

In my opinion, we should not do wrong, just because we once did it wrong.
These names are contrary to our conventions.

For Magento2 it will be new module... new module with bad name conventions.
If we need to keep BI, we need to mark old interfaces as deprecated in favor of new with proper names.

app/code/Magento/MediaGalleryApi/Model/Directory/Command/CreateByPathInterface.php
```diff
@@ -0,0 +1,24 @@
...
+interface CreateByPathInterface
+{
+    /**
+     * Create new directory by provided path
+     *
+     * @param string $path
+     * @param string $name
+     * @throws \Magento\Framework\Exception\CouldNotSaveException
+     */
+    public function execute(string $path, string $name): void;
```
naydav: Why do we need two parameters instead of one? In general, a directory name is just final part of directory path.
Even in tests, we follow this approach
https://github.com/magento/magento2/pull/27499/files?file-filters%5B%5D=.php#diff-1d9775d2fc9075b05e8166c777672c52R58

app/code/Magento/MediaGalleryApi/Model/Directory/IsBlacklistedInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryApi\Model\Directory;
+
+/**
+ * Directory paths that are reserved by system and not be included in the media gallery
+ * @api
+ */
+interface IsBlacklistedInterface
```
naydav: Why do we need this interface? We introduced separate config as an extension point for marking directories as a blacklist.

app/code/Magento/MediaGalleryApi/Model/Keyword/Command/SaveAssetKeywordsInterface.php
```diff
@@ -9,6 +9,7 @@
 
 /**
  * A command represents the media gallery asset keywords save operation.
+ * @api
  */
 interface SaveAssetKeywordsInterface
```
naydav, line 15: Should be bulk operation. For following BC, old interface should be marked as deprecated in flavor on new with bulk.
It's related all cases.

dev/tests/integration/testsuite/Magento/MediaGallery/Model/Directory/Command/CreateByPathTest.php
```diff
@@ -0,0 +1,95 @@
...
+{
+    /**
+     * Test directory name
+     */
+    private CONST TEST_DIRECTORY_NAME = 'testCreateDirectory';
+
+    /**
+     * Absolute path to the media direcrory
+     */
+    private static $_mediaPath;
```
naydav: We did not use `underscore` in variable/property names.

dev/tests/integration/testsuite/Magento/MediaGallery/Model/Directory/Command/CreateByPathTest.php
```diff
@@ -0,0 +1,95 @@
...
+        self::$_mediaPath = Bootstrap::getObjectManager()->get(Filesystem::class)
+            ->getDirectoryRead(DirectoryList::MEDIA)->getAbsolutePath();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function setUp()
+    {
+        $this->createByPath = Bootstrap::getObjectManager()->create(CreateByPathInterface::class);
```
naydav: We don't need new instance please use `get` for object retrieving.

dev/tests/integration/testsuite/Magento/MediaGallery/Model/Directory/Command/CreateByPathTest.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        $fullPath = self::$_mediaPath . self::TEST_DIRECTORY_NAME;
+        $this->createByPath->execute('', self::TEST_DIRECTORY_NAME);
+        $this->assertFileExists($fullPath);
+    }
+
+    /**
+     * @return void
+     * @throws \Magento\Framework\Exception\CouldNotSaveException
+     * @expectedException \Magento\Framework\Exception\CouldNotSaveException
```
naydav: I'm not sure that os proper exception for this case

dev/tests/integration/testsuite/Magento/MediaGallery/_files/media_asset.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
naydav, line 1: Each fixture file **must have** a rollback file.

dev/tests/integration/testsuite/Magento/MediaGallery/Model/File/Command/DeleteByAssertIdTest.php
```diff
@@ -0,0 +1,99 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ *
+ */
+namespace Magento\MediaGallery\Model\Directory\Command;
```
naydav: There are only 3 integration tests. All interfaces **must be** covered by tests (by integration/API integration/WebAPI but NOT unit). 
We were agreement about MFTF tests (especially if we fon't have UI), but it's doesn't mean that we don't need any tests. It "must have" part of DoD.

dev/tests/integration/testsuite/Magento/MediaGallery/Model/File/Command/DeleteByAssertIdTest.php
```diff
@@ -0,0 +1,99 @@
...
+
+use Magento\Framework\App\Filesystem\DirectoryList;
+use Magento\Framework\Filesystem;
+use Magento\MediaGallery\Model\File\Command\DeleteByAssetId;
+use Magento\MediaGalleryApi\Model\File\Command\DeleteByAssetIdInterface;
+use Magento\TestFramework\Helper\Bootstrap;
+use Magento\MediaGalleryApi\Model\Asset\Command\GetByIdInterface;
+
+/**
+ * Test methods of class DeleteByPath
```
naydav: Check all dock blocks.

app/code/Magento/MediaGallery/etc/directory.xsd
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
naydav: Should be in `MediaGalleryApi`

app/code/Magento/MediaGallery/Model/Asset/Command/GetByPath.php
```diff
@@ -59,28 +59,26 @@ public function __construct(
...
-     * @param string $mediaFilePath
+     * @param string $path
      *
      * @return AssetInterface
      * @throws IntegrationException
      */
-    public function execute(string $mediaFilePath): AssetInterface
+    public function execute(string $path): AssetInterface
     {
         try {
```
naydav, line 72: In general, if model operates with MySQL queries, it should be in `ResourceModel` namespace

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
```diff
@@ -68,7 +68,10 @@ public function execute(AssetInterface $mediaAsset): int
             $connection = $this->resourceConnection->getConnection();
             $tableName = $this->resourceConnection->getTableName(self::TABLE_MEDIA_GALLERY_ASSET);
 
-            $connection->insertOnDuplicate($tableName, $this->extractor->extract($mediaAsset, AssetInterface::class));
+            $connection->insertOnDuplicate(
+                $tableName,
+                array_filter($this->objectProcessor->buildOutputDataArray($mediaAsset, AssetInterface::class))
```
naydav: We be lead to MySQL error if `extension_attributes` will be not empty.
Pay attention, `array_filter` will not allow inserting values like '0' or an empty string.

app/code/Magento/MediaGalleryApi/Api/Data/AssetInterface.php
```diff
@@ -38,7 +39,7 @@ public function getPath(): string;
     public function getTitle(): ?string;
 
     /**
-     * Get source of the file
+     * Get the name of the channel/stock/integration file was retrieved from. null if not identified.
```
naydav, line 42: Looks like
`\Magento\MediaGalleryApi\Api\Data\AssetInterface::setExtensionAttributes` should be marked as deprecated

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
```diff
@@ -68,7 +68,10 @@ public function execute(AssetInterface $mediaAsset): int
             $connection = $this->resourceConnection->getConnection();
             $tableName = $this->resourceConnection->getTableName(self::TABLE_MEDIA_GALLERY_ASSET);
 
-            $connection->insertOnDuplicate($tableName, $this->extractor->extract($mediaAsset, AssetInterface::class));
+            $connection->insertOnDuplicate(
```
naydav: We already talk that all "persisting" interfaces should work with bulk but not with singular entities.

app/code/Magento/MediaGallery/Model/Directory/Command/CreateByPath.php
naydav: The media gallery should work without `Cms`.
> please consider a possibility to extract logic in some common service

app/code/Magento/MediaGallery/Model/Directory/IsBlacklisted.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Magento\MediaGallery\Model\Directory;
+
+use Magento\MediaGalleryApi\Model\Directory\IsBlacklistedInterface;
+
+/**
+ * Check if the path is blacklisted for media gallery. Directory path may be blacklisted if it's reserved by the system
+ */
+class IsBlacklisted implements IsBlacklistedInterface
+{
+    const XML_PATH_BLACKLIST_PATTERNS = 'blacklist/patterns';
```
naydav: It's should be part of Config object which should encapsulate all details related to the reading of configuration.
Also, it's should be `private` const

app/code/Magento/MediaGallery/etc/directory.xsd
```diff
@@ -0,0 +1,38 @@
...
+
+    <xs:complexType name="blacklistType">
+        <xs:sequence>
+            <xs:element type="patternsType" name="patterns" maxOccurs="unbounded" minOccurs="0"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="patternsType">
+        <xs:sequence>
+            <xs:element type="patternType" name="pattern" maxOccurs="unbounded" minOccurs="0"/>
```
naydav: Please add a description (`<xs:documentation>...</xs:documentation>`) that the pattern should use PCRE regex syntax.

app/code/Magento/MediaGallery/etc/directory.xsd
```diff
@@ -0,0 +1,38 @@
...
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="patternType">
+        <xs:simpleContent>
+            <xs:extension base="xs:string">
+                <xs:attribute type="xs:string" name="name" use="required"/>
+                <xs:attribute type="xs:string" name="example" use="optional"/>
```
naydav: Why do we need this directive?

app/code/Magento/MediaGallery/Model/Directory/Config/Converter.php
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param \DOMDocument $source
+     * @return array
+     */
+    public function convert($source): array
+    {
+        $result = [];
+
+        if (!$source instanceof \DOMDocument) {
+            return $result;
```
naydav: Please move to the signature of the method (type hinting)

app/code/Magento/MediaGallery/Model/File/Command/DeleteByAssetId.php
```diff
@@ -0,0 +1,72 @@
...
+    public function execute(int $assetId): void
+    {
+        $mediaFilePath = $this->getAssetById->execute($assetId)->getPath();
+        $mediaDirectory = $this->filesystem->getDirectoryRead(DirectoryList::MEDIA);
+
+        if (!$mediaDirectory->isFile($mediaFilePath)) {
+            throw new LocalizedException(__('File "%1" does not exist in media directory.', $mediaFilePath));
+        }
+
+        $this->imagesStorage->deleteFile($mediaDirectory->getAbsolutePath() . $mediaFilePath);
```
naydav: Please make sure that we covered case which restricts the possibility to remove any file on a server.

app/code/Magento/MediaGallery/Test/Unit/Model/Asset/Command/SaveTest.php
```diff
@@ -7,16 +7,17 @@
 
 namespace Magento\MediaGallery\Test\Unit\Model\Asset\Command;
 
+use Magento\Eav\Helper\Data;
```
naydav: It looks like redundant dependency.

app/code/Magento/MediaGallery/Test/Unit/Model/Asset/Command/SaveTest.php
```diff
@@ -112,7 +113,7 @@ protected function setUp(): void
             Save::class,
             [
                 'resourceConnection' => $this->resourceConnectionMock,
-                'extractor'          => $this->extractorMock,
+                'objectProcessor'          => $this->objectProcessor,
```
naydav: Please fix all style in all tests

app/code/Magento/MediaGallery/Model/Directory/Config/Reader.php
```diff
@@ -0,0 +1,70 @@
...
+        '/config/patterns' => 'patterns',
+        '/config/patterns/pattern' => 'name',
+    ];
+
+    /**
+     * XML Configuration file name
+     */
+    private const XML_FILE_NAME = 'directory.xml';
+
+    /**
```
naydav: Looks like we don't need override construct due to we provide values in DI
https://github.com/magento/magento2/pull/27499/files#diff-3ddae2bd1c7d9c63b0ce61ab103f3ac5R37

naydav (changes requested):
Also, all logic related to **Catalog** should be moved to separate module.

app/code/Magento/MediaGallery/etc/directory.xsd
VladimirZaets: Why not? it gives more visibility on what case this regexp covered. We use this approach in a similar situation in magento

app/code/Magento/MediaGallery/Test/Unit/Model/Asset/Command/SaveTest.php
sivaschenko: fixed

dev/tests/integration/testsuite/Magento/MediaGallery/Model/Directory/Command/CreateByPathTest.php
sivaschenko: fixed

dev/tests/integration/testsuite/Magento/MediaGallery/_files/media_asset.php
sivaschenko: fixed

dev/tests/integration/testsuite/Magento/MediaGallery/Model/Directory/Command/CreateByPathTest.php
sivaschenko: What is the proper exception?

app/code/Magento/MediaGallery/etc/directory.xsd
sivaschenko: moved

app/code/Magento/MediaGalleryApi/Api/Data/AssetInterface.php
sivaschenko: We would like to allow extension attributes for the asset

app/code/Magento/MediaGallery/Model/Directory/IsBlacklisted.php
sivaschenko: moved

app/code/Magento/MediaGallery/etc/directory.xsd
sivaschenko: added

app/code/Magento/MediaGallery/etc/directory.xsd
sivaschenko: removed

app/code/Magento/MediaGallery/Model/Directory/Config/Reader.php
sivaschenko: Class removed in favour to virtual type

app/code/Magento/MediaGallery/Test/Unit/Model/Asset/Command/SaveTest.php
sivaschenko: removed

app/code/Magento/MediaGallery/Model/Directory/Config/Converter.php
sivaschenko: That will conflict with the `\Magento\Framework\Config\ConverterInterface::convert` contract

app/code/Magento/MediaGalleryApi/Model/Directory/IsBlacklistedInterface.php
sivaschenko: This service is required for identifying if the given directory is blacklisted

app/code/Magento/MediaGalleryApi/Model/Keyword/Command/SaveAssetKeywordsInterface.php
sivaschenko: It is bulk

app/code/Magento/MediaGalleryApi/Model/File/Command/DeleteByAssetIdInterface.php
sivaschenko: This interface was removed

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
sivaschenko: What is the approach for returning the identifier for the saved assets in this case?

app/code/Magento/MediaGalleryApi/Model/Asset/Command/DeleteByPathInterface.php
sivaschenko: Interfaces were merged

app/code/Magento/MediaGalleryApi/Model/Directory/Blacklist.php
sivaschenko: The Blacklist class removed in favour to configuration

app/code/Magento/MediaGalleryApi/Model/Asset/Command/DeleteByDirectoryPathInterface.php
sivaschenko: Done

app/code/Magento/MediaGalleryApi/Model/Keyword/Command/SaveAssetKeywordsInterface.php
sivaschenko: made it bulk bulk

dev/tests/integration/testsuite/Magento/MediaGallery/Model/Directory/Command/CreateByPathTest.php
sivaschenko: Done

app/code/Magento/MediaGallery/Model/Asset/Command/GetByPath.php
sivaschenko: Applied

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
sivaschenko: Done

app/code/Magento/MediaGalleryApi/Model/Directory/Command/CreateByPathInterface.php
sivaschenko: done

app/code/Magento/MediaGallery/Model/File/Command/DeleteByAssetId.php
sivaschenko: This service was removed

dev/tests/integration/testsuite/Magento/MediaGallery/Model/File/Command/DeleteByAssertIdTest.php
sivaschenko: All actual api have been covered by integration tests

dev/tests/integration/testsuite/Magento/MediaGallery/Model/File/Command/DeleteByAssertIdTest.php
sivaschenko: Checked

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
sivaschenko: array_filter was replaced with custom filtering

sivaschenko:
@naydav pull request is ready for review

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
```diff
@@ -68,12 +70,39 @@ public function execute(AssetInterface $mediaAsset): int
...
+     * Filter data to get flat array without null values
+     *
+     * @param array $data
+     * @return array
+     */
+    private function filterData(array $data): array
+    {
+        $filteredData = [];
+        foreach ($data as $key => $value) {
+            if ($value === null) {
```
VladimirZaets: Merge if conditions

app/code/Magento/MediaGallery/Model/AssetKeywords.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function getExtensionAttributes(): AssetKeywordsExtensionInterface
+    {
+        return $this->_getExtensionAttributes();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function setExtensionAttributes(AssetKeywordsExtensionInterface $extensionAttributes): void
```
VladimirZaets: Remove

app/code/Magento/MediaGallery/Model/Directory/Config.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Magento\MediaGallery\Model\Directory;
+
+use Magento\Framework\Config\DataInterface;
+
+/**
+ * Media gallery directory config
+ */
+class Config
+{
+    private const XML_PATH_BLACKLIST_PATTERNS = 'blacklist/patterns';
```
VladimirZaets: Create interface getBlacklistPatternsInterface in model folder

VladimirZaets: remove get

app/code/Magento/MediaGallery/Model/Directory/IsBlacklisted.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+namespace Magento\MediaGallery\Model\Directory;
+
+use Magento\MediaGalleryApi\Api\IsPathBlacklistedInterface;
+
+/**
+ * Check if the path is blacklisted for media gallery. Directory path may be blacklisted if it's reserved by the system
+ */
+class IsBlacklisted implements IsPathBlacklistedInterface
```
VladimirZaets, line 16: Look to UI module, how it uses there

app/code/Magento/MediaGallery/Model/Keyword.php
```diff
@@ -18,7 +18,6 @@
 class Keyword extends AbstractExtensibleModel implements KeywordInterface
 {
     private const ID = 'id';
-
     private const KEYWORD = 'keyword';
```
VladimirZaets: Remove extends AbstractExtensibleModel.
Constructor and getters can be generated by phpshtorm

app/code/Magento/MediaGallery/Model/ResourceModel/DeleteAssetsByPaths.php
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function execute(array $paths): void
+    {
+        $failedPaths = [];
+
+        foreach ($paths as $path) {
+            try {
+                $this->validateDirectoryPath($path);
```
VladimirZaets, line 57: Move validation step to separate class in model namespace. And create an interface for it in model namespace

app/code/Magento/MediaGallery/Model/ResourceModel/DeleteAssetsByPaths.php
```diff
@@ -0,0 +1,102 @@
...
+     * Delete assets from database based on the first part (beginning) of the path
+     *
+     * @param string $path
+     */
+    private function deleteAssetsByDirectoryPath(string $path): void
+    {
+        /** @var AdapterInterface $connection */
+        $connection = $this->resourceConnection->getConnection();
+        $tableName = $this->resourceConnection->getTableName(self::TABLE_MEDIA_GALLERY_ASSET);
+        $connection->delete($tableName, [self::MEDIA_GALLERY_ASSET_PATH . ' LIKE ?' => $path . '%']);
```
VladimirZaets, line 87: Check on  SQL injection

app/code/Magento/MediaGallery/Model/ResourceModel/GetAssetsByIds.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function execute(array $ids): array
+    {
+        $assets = [];
+        try {
+            foreach ($this->getAssetsData($ids) as $assetData) {
+                $assets[] = $this->assetFactory->create(['data' => $assetData]);
```
VladimirZaets: Should be refactored after removing the abstract extensible model from asset model class

app/code/Magento/MediaGallery/Model/ResourceModel/Keyword/GetAssetsKeywords.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Load keywords data
+     *
+     * @param array $assetIds
+     * @return array
+     * @throws \Zend_Db_Statement_Exception
+     */
+    private function getKeywordsData(array $assetIds): array
```
VladimirZaets, line 85: Create DTO (Link between asset id and keywords) for result

app/code/Magento/MediaGallery/Model/ResourceModel/Keyword/SaveAssetLinks.php
```diff
@@ -5,18 +5,17 @@
...
 use Magento\Framework\App\ResourceConnection;
 use Magento\Framework\DB\Adapter\AdapterInterface;
 use Magento\Framework\DB\Adapter\Pdo\Mysql;
 use Magento\Framework\Exception\CouldNotSaveException;
+use Magento\MediaGalleryApi\Api\Data\KeywordInterface;
 use Psr\Log\LoggerInterface;
 
 /**
- * Class SaveAssetLinks
+ * Save links between asset and keyword to media_gallery_asset_keyword table
```
VladimirZaets, line 18: Rename the class, create a link DTO

app/code/Magento/MediaGallery/Model/ResourceModel/Keyword/SaveAssetLinks.php
```diff
@@ -35,8 +34,6 @@ class SaveAssetLinks
     private $logger;
 
     /**
-     * SaveAssetLinks constructor.
-     *
      * @param ResourceConnection $resourceConnection
      * @param LoggerInterface $logger
      */
```
VladimirZaets, line 39: Check keys in DB

VladimirZaets: Use insert multiple on a duplicate (see MSI)

app/code/Magento/MediaGallery/Model/ResourceModel/SaveAssets.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        $connection = $this->resourceConnection->getConnection();
+        $tableName = $this->resourceConnection->getTableName(self::TABLE_MEDIA_GALLERY_ASSET);
+
+        $failedAssets = [];
+        foreach ($assets as $asset) {
+            try {
+                $connection->insertOnDuplicate(
+                    $tableName,
+                    $this->filterData($this->objectProcessor->buildOutputDataArray($asset, AssetInterface::class))
```
VladimirZaets: Use AssetInterface directly instead of ` $this->filterData($this->objectProcessor->buildOutputDataArray`

app/code/Magento/MediaGalleryApi/Api/IsPathBlacklistedInterface.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
```
VladimirZaets, line 6: Look how it uses in media gallery UI module

app/code/Magento/MediaGalleryApi/Api/IsPathBlacklistedInterface.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Magento\MediaGalleryApi\Api;
+
+/**
+ * Directory paths that are reserved by system and not be included in the media gallery
```
VladimirZaets: Clarify descriptions

app/code/Magento/MediaGalleryApi/etc/directory.xsd
```diff
@@ -0,0 +1,52 @@
...
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element type="patternsType" name="patterns" maxOccurs="unbounded" minOccurs="0"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="patternsType">
+        <xs:annotation>
+            <xs:documentation>
+                List of directory paths regexp patterns
```
VladimirZaets: rename regexp to RegExp

app/code/Magento/MediaGalleryCatalog/Plugin/Product/Gallery/Processor.php
```diff
@@ -3,25 +3,24 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
-
```
VladimirZaets: Rename plugin according responsibility

app/code/Magento/MediaGalleryCatalog/etc/di.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
+    <type name="Magento\Catalog\Model\Product\Gallery\Processor">
+        <plugin name="media_gallery_image_remove_metadata" type="Magento\MediaGallery\Plugin\Product\Gallery\Processor"
```
VladimirZaets: Investigate is plugins sequence is important

app/code/Magento/MediaGalleryCatalog/etc/directory.xml
```diff
@@ -0,0 +1,14 @@
...
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_MediaGalleryApi:etc/directory.xsd">
+    <blacklist>
+        <patterns>
+            <pattern name="catalog">/^catalog\/product/</pattern>
```
VladimirZaets, line 11: Investigate do we need to add regexp to  category

sivaschenko: no, we don't

app/code/Magento/MediaGalleryCatalog/etc/di.xml
sivaschenko: Why do you think it is important?

app/code/Magento/MediaGalleryApi/Api/IsPathBlacklistedInterface.php
sivaschenko: looked

app/code/Magento/MediaGallery/Model/ResourceModel/SaveAssets.php
sivaschenko: The idea is to correct the `buildOutputDataArray` to meet our needs. Also, this is a private method.

app/code/Magento/MediaGallery/Model/ResourceModel/DeleteAssetsByPaths.php
sivaschenko: The dynamic value is passed by binding.

app/code/Magento/MediaGallery/Model/Directory/IsBlacklisted.php
sivaschenko: looked

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
sivaschenko: Why?

app/code/Magento/MediaGallery/Model/AssetKeywords.php
sivaschenko: Why?

app/code/Magento/MediaGallery/Model/Directory/Config.php
sivaschenko: Why?

app/code/Magento/MediaGallery/Model/Keyword.php
sivaschenko: Why?

app/code/Magento/MediaGallery/Model/ResourceModel/GetAssetsByIds.php
sivaschenko: see questions above

app/code/Magento/MediaGalleryCatalog/Plugin/Product/Gallery/Processor.php
sivaschenko: please provide a desired namespace and name

app/code/Magento/MediaGalleryApi/Api/IsPathBlacklistedInterface.php
sivaschenko: What is not clear?

app/code/Magento/MediaGallery/Model/ResourceModel/Keyword/SaveAssetLinks.php
sivaschenko: This is an internal class, not an API service

app/code/Magento/MediaGallery/Model/ResourceModel/Keyword/GetAssetsKeywords.php
sivaschenko: I don't understand the concern. The public method is returning the `\Magento\MediaGalleryApi\Api\Data\AssetKeywordsInterface[]`

app/code/Magento/MediaGallery/Model/ResourceModel/DeleteAssetsByPaths.php
sivaschenko: That is a check if the path is empty. I don't think extracting the class/interface for this purpose is necessary.

app/code/Magento/MediaGalleryApi/etc/directory.xsd
sivaschenko: renamed

sivaschenko:
@VladimirZaets thanks for the review. All comments were either addressed or the code was updated, please see my questions.

app/code/Magento/MediaGallery/Model/AssetKeywords.php
sivaschenko: Remove only if webapi is not affected

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
sivaschenko: Save hardcoded list of fields instead of extracting and filtering data

app/code/Magento/MediaGallery/Model/Directory/Config.php
sivaschenko: get is removed and interface was created, but interface has to be moved to MediaGalleryApi\Model\BlacklistPatternsConfigInterface::get

app/code/Magento/MediaGalleryApi/Api/IsPathBlacklistedInterface.php
sivaschenko: Check if the path is blacklisted for media gallery. Directory path may be blacklisted if it's reserved by the system

app/code/Magento/MediaGalleryCatalog/etc/di.xml
sivaschenko: Remove sortorder and disabled attributes

app/code/Magento/MediaGalleryCatalog/Plugin/Product/Gallery/Processor.php
sivaschenko: `Magneto\MediaGallery\Plugin\RemoveAssetAfterRemoveImage`

app/code/Magento/MediaGallery/Model/Asset/Command/Save.php
```diff
@@ -72,37 +64,24 @@ public function execute(AssetInterface $mediaAsset): int
...
+                    'id' => $mediaAsset->getId(),
+                    'path' => $mediaAsset->getPath(),
+                    'title' => $mediaAsset->getTitle(),
+                    'source' => $mediaAsset->getSource(),
+                    'content_type' => $mediaAsset->getContentType(),
+                    'width' => $mediaAsset->getWidth(),
+                    'height' => $mediaAsset->getHeight(),
+                    'size' => $mediaAsset->getSize(),
+                    'created_at' => $mediaAsset->getCreatedAt(),
+                    'updated_at' => $mediaAsset->getUpdatedAt(),
```
sivaschenko: Update at and created at should not be saved to the database if their values are `null`

app/code/Magento/MediaGallery/Model/Directory/Config.php
```diff
@@ -8,11 +8,12 @@
 namespace Magento\MediaGallery\Model\Directory;
 
 use Magento\Framework\Config\DataInterface;
+use Magento\MediaGalleryApi\Model\BlacklistPatternsConfigInterface;
 
 /**
  * Media gallery directory config
  */
-class Config
+class Config implements BlacklistPatternsConfigInterface
```
sivaschenko: I think the class name should also be changed according to interface change

app/code/Magento/MediaGallery/Model/ResourceModel/SaveAssets.php
```diff
@@ -66,7 +58,18 @@ public function execute(array $assets): void
...
+                        'id' => $asset->getId(),
+                        'path' => $asset->getPath(),
+                        'title' => $asset->getTitle(),
+                        'source' => $asset->getSource(),
+                        'content_type' => $asset->getContentType(),
+                        'width' => $asset->getWidth(),
+                        'height' => $asset->getHeight(),
+                        'size' => $asset->getSize(),
+                        'created_at' => $asset->getCreatedAt(),
+                        'updated_at' => $asset->getUpdatedAt(),
```
sivaschenko: Update at and created at should not be saved to the database if their values are null

lenaorobei:
Performance build: https://bamboo.corp.magento.com/browse/MCCE24-MCCE24400-1

sidolov:
@magento run all tests

lenaorobei:
Performance build:: https://bamboo.corp.magento.com/browse/MCCE24-MCCE24404-1

sidolov:
@magento run all tests

sivaschenko:
PAT is green

dev/tests/integration/testsuite/Magento/MediaGallery/_files/media_asset_rollback.php
```diff
@@ -0,0 +1,20 @@
...
+use Magento\TestFramework\Helper\Bootstrap;
+
+$objectManager = Bootstrap::getObjectManager();
+
+/** @var DeleteByPathInterface $mediaSave */
+$mediaAssetDelete = $objectManager->get(DeleteByPathInterface::class);
+
+try {
+    $mediaAssetDelete->execute('testDirectory/path.jpg');
+} catch (\Exception $exception) {
```
coderimus, line 18: @sivaschenko https://github.com/magento/adobe-stock-integration/pull/1188/files/9041038e8c3232997f94744031813009a6a7635a#r410128808

sivaschenko:
@magento run all tests

sivaschenko:
The pull request was tested using basic Media Gallery functionality in Adobe Stock Integration

PR #27502, opened 2020-03-30
Fix bug 25227: Issue with Product Review when disabled from admin

tna274:
### Description (*)
Fix bug 25227: Issue with Product Review when disabled from admin

1. magento/magento2#25227: Issue with Product Review when disabled from admin

1. Open Product page from frontend and go to add review section 
2. Now login into Admin.
3. Go to opened product page on frontend
4. Disable Product Review from admin. Navigate to **Store > Configuration** then **Catalog > Catalog** now disable **Product Reviews**
5. Now submiting review (without refresh product detail page)


lenaorobei (changes requested):
Please have a look at `Magento\Review\Observer\PredispatchReviewObserver`.
This behaviour should also be covered with automated tests.

app/code/Magento/Review/Controller/Product/Post.php
```diff
@@ -31,6 +31,13 @@ public function execute()
             return $resultRedirect;
         }
 
+        $isReviewEnable = $this->_objectManager->get(\Magento\Review\Helper\Data::class)->isEnableReview();
```
Stepa4man: Please don't use Object manager, use constructor dependency injection instead.
Also please don't extend existing helpers with new functionality. If you want to decompose class, use services instead.
In this particular controller, I'm sure you can use just private class.

app/code/Magento/Review/Helper/Data.php
```diff
@@ -67,6 +70,19 @@ public function getDetailHtml($origDetail)
         return nl2br($this->filter->truncate($this->_escaper->escapeHtml($origDetail), ['length' => 50]));
     }
 
+    /**
+     * Return an indicator of whether or not review is enable
+     *
+     * @return bool
+     */
+    public function isEnableReview()
```
Stepa4man: Please avoid extending existing helpers and creating new ones.

app/code/Magento/Review/Controller/Product/Post.php
Stepa4man: Also please add `declare(strict_types=1);` right after copyright.

Stepa4man (changes requested):
Hi @tna274 
Thanks for your contribution

First of all, I want to mention that I don't think it's an issue.
If merchant disables review he has to expect that somebody who is in the progress of submitting a review wants to be able to finish it. 
I think interrupting customers during this action is bad user experience.
In case if the form would be opened for a long time - form_key will be invalidated.

Second, please spend some time and familiarize yourself with [Technical guideline](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html#1-basic-programming-principles) and [Definition of done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html)
Update your PR according to these documents.

Thanks in advance.

tna274:
Hi @Stepa4man 
I have updated my pull request. I think that it is an issue, Magento has already tried to handle this case, you can see file Magento\Review\Observer\PredispatchReviewObserver for more detail. But the action in that observer has not stopped the save action.

Hi @lenaorobei 
I have checked class Magento\Review\Observer\PredispatchReviewObserver, it has already been covered by automation test.

lenaorobei:
@tna274 I meant that logic you've provided already present in this observer. I believe we need to check fired events and investigate why this logic was not executed.

tna274:
@lenaorobei I think that observer did not work because function setRedirect did not block process in the execute function. So even it calls PredispatchReviewObserver, the saving action will still be executed.
In addition, plugin before function usually use for change parameter but not change the logic. At my point of view, that observer should be deprecated.

mrtuvn (changes requested):
Can you change file permissions back to nornal

mrtuvn (commented):
Should we avoid inheritance action here ?  @lenaorobei

lenaorobei:
Unfortunately I need to close this PR.
Please see the reason here https://github.com/magento/magento2/issues/25227#issuecomment-606275553

PR #27541, opened 2020-04-01
Fix (#27505): Prevent custom attributes being mapped twice

### LINKED ISSUE #27505: Shipping Method Estimator not working with custom address attributes
### Preconditions (*)

1. Magento 2.3.4 with Sample data
2. A module that declares a custom attribute for as per instructions from dev docs (https://devdocs.magento.com/guides/v2.3/howdoi/checkout/checkout_new_field.html#add)

### Steps to reproduce (*)

1. Install a module that adds a custom address attribute (either follow the [dev docs](https://devdocs.magento.com/guides/v2.3/howdoi/checkout/checkout_new_field.html#add) to create your own or install the module below)
    1. https://github.com/SamJUK/m2-custom-attribute-bug
2. Enable and Redeploy site
3. Add item to cart and proceed to shipping step of checkout
4. Fill out shipping information (custom attribute + address data)
5. Select a shipping rate
6. Navigate back to your cart
7. Trigger a estimate shipping event
    1. Either try to use shipping estimator on the cart with a new postcode
    2. Or continue back to checkout and change postcode

### Expected result (*)

1. Pull through shipping rates for the new postcode

### Actual result (*)

1. https://www.youtube.com/watch?v=cy3gQqQZ0fs
2. Estimate shipping methods request returns a 400 bad request with the following error message 
`{"message":"Error occurred during \"custom_attributes\" processing. A custom attribute is specified with a missing attribute code. Verify the code and try again."}`

Due to strange nesting within the custom_attribute property. Originally while its working the address payload is of the following format 
```json
custom_attributes: [{
    "attribute_code": "custom_field", 
    "value": "123" 
}]
``` 

Where was when we are getting 400 errors the custom attributes looks like so
```json
custom_attributes: [{
    "attribute_code": "0", 
    "value": {
        "attribute_code": "custom_field", 
        "value": "123" 
    }
}]
```

Currently using the following Hotfix to allow for checkout
https://gist.github.com/SamJUK/479ca3f108b8091fbf512695f97363fb

### Additional information
https://github.com/magento/magento2/issues/27505#issuecomment-627565714

SamJUK:
### Description (*)

Check if the custom attributes data has already been mapped, in the `address-converter` js component to prevent it being remapped into an invalid format.

Currently after you set the shipping address at checkout, the `address-converter` component maps the custom_attributes data into a new format, then this later gets saved to local storage. Then when this data is refetched from local storage it tries to map it again breaking the format.

1. Fixes magento/magento2#27505: Shipping Method Estimator not working with custom address attributes

- Create a custom address attribute
- Add item to cart and proceed to shipping step of checkout
- Fill out shipping information (custom attribute + address data)
- Select a shipping rate
- Navigate back to your cart
- Trigger a estimate shipping event
    - Either try to use shipping estimator on the cart with a new postcode
    - Or continue back to checkout and change postcode


app/code/Magento/Checkout/view/frontend/web/js/model/address-converter.js
```diff
@@ -64,6 +64,12 @@ define([
                 addressData['custom_attributes'] = _.map(
                     addressData['custom_attributes'],
                     function (value, key) {
+                        var hasAlreadyBeenMapped = value instanceof Object && value.hasOwnProperty('attribute_code') && value.hasOwnProperty('value');
```
VladimirZaets, line 67: Hi @SamJUK. Thanks for collaboration. 
I think in this case we should move the check to the upper level. Firstly we should filter the values that already mapped and after that map only not mapped value without any check.

Also, due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html) all code must be covered by tests. Can you please cover it by the functional test?

VladimirZaets (changes requested):
Hi @SamJUK. Thanks for collaboration. Please take a look at my review comment

engcom-Charlie:
Hi @SamJUK, I'm closing this PR now due to inactivity.

PR #27620, opened 2020-04-05
MAGENTO-PWA#2190: Fixed dependent GraphQL schema according to best practices

andrewbess:
### Description (*)

This PR fixes GraphQL schema for merging carts according to Magento best practices

1. magento/pwa-studio#2190: Call mergeCarts to merge guest and customer carts within customer login

### DIFF
--- a/app/code/Magento/QuoteGraphQl/etc/schema.graphqls
+++ b/app/code/Magento/QuoteGraphQl/etc/schema.graphqls
@@ -20,7 +20,7 @@ type Mutation {
     setPaymentMethodOnCart(input: SetPaymentMethodOnCartInput): SetPaymentMethodOnCartOutput @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\SetPaymentMethodOnCart")
     setGuestEmailOnCart(input: SetGuestEmailOnCartInput): SetGuestEmailOnCartOutput @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\SetGuestEmailOnCart")
     setPaymentMethodAndPlaceOrder(input: SetPaymentMethodAndPlaceOrderInput): PlaceOrderOutput @deprecated(reason: "Should use setPaymentMethodOnCart and placeOrder mutations in single request.") @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\SetPaymentAndPlaceOrder")
-    mergeCarts(source_cart_id: String!, destination_cart_id: String!): Cart! @doc(description:"Merges the source cart into the destination cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\MergeCarts")
+    mergeCarts(input: MergeCartsInput): Cart! @doc(description:"Merges the source cart into the destination cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\MergeCarts")
     placeOrder(input: PlaceOrderInput): PlaceOrderOutput @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\PlaceOrder")
 }
 
@@ -128,6 +128,11 @@ input ShippingMethodInput {
 input SetPaymentMethodAndPlaceOrderInput {
     cart_id: String!
     payment_method: PaymentMethodInput!
+}
+
+input MergeCartsInput {
+    source_cart_id: String!
+    destination_cart_id: String!
 }
 
 input PlaceOrderInput {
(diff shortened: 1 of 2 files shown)


app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -20,7 +20,7 @@ type Mutation {
     setPaymentMethodOnCart(input: SetPaymentMethodOnCartInput): SetPaymentMethodOnCartOutput @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\SetPaymentMethodOnCart")
     setGuestEmailOnCart(input: SetGuestEmailOnCartInput): SetGuestEmailOnCartOutput @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\SetGuestEmailOnCart")
     setPaymentMethodAndPlaceOrder(input: SetPaymentMethodAndPlaceOrderInput): PlaceOrderOutput @deprecated(reason: "Should use setPaymentMethodOnCart and placeOrder mutations in single request.") @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\SetPaymentAndPlaceOrder")
-    mergeCarts(source_cart_id: String!, destination_cart_id: String!): Cart! @doc(description:"Merges the source cart into the destination cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\MergeCarts")
+    mergeCarts(input: MergeCartsInput): Cart! @doc(description:"Merges the source cart into the destination cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\MergeCarts")
```
rogyar, line 23: I'm not sure about the backward compatibility of this change. 
@lenaorobei, could you confirm, please, that we may perform such changes? 

Thank you

lenaorobei: The implementation of `mergeCarts` was added based on approved proposal from the architecture repo https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/shared-cart.md

Need to understand how this will help PWA devs.

Addind @cpartica @paliarush

lenaorobei: @andrewbess could you please explain the isse with using `mergeCarts(source_cart_id: String!, destination_cart_id: String!)`?

rogyar: I believe the main concern here is extensibility. If we use a separate type for input (as we do in most cases), we may extend this input (add fields, for example) in a separate module. If we have the parameters hardcoded as they are now, the extensibility is compromised. 

But the main question here is the backward compatibility. If some merchant already uses the existing `mergeCarts` mutation somehow, he will be surprised after the upgrade.

andrewbess: Hello @rogyar 
The GraphQL for merging carts looks like incorrect. It doesn't use Magento [best practices](https://devdocs.magento.com/guides/v2.3/graphql/develop/resolvers.html#mutation-syntax).
I don't understand why it has been changed from the correct code to incorrect.
Please take a look at commits below.
Correct commit (IMHO)
magento/graphql-ce#950
Incorrect commit that changes previous (IMHO)
magento/magento2@f19f727
I also want to note that this is a relatively new feature.
It appeared only in version 2.3.4

andrewbess: Hello @lenaorobei 
Thank you for your question.
This mutation isn't written according to [Magento best practices](https://devdocs.magento.com/guides/v2.3/graphql/develop/resolvers.html#mutation-syntax).
Currently, I am working on [PWA-studio#2190](https://github.com/magento/pwa-studio/issues/2190).
So, I think it would be better to change it now and don't alter PWA after changing this mutation in the future.

lenaorobei: @rogyar regarding your concern about extensibility - both options should work - adding new field and parameter.

@andrewbess there was an issue on the DevDocs page that is already addressed https://github.com/magento/devdocs/pull/7001
This recommendation was outdated.

Please refer to the extensibility approach in the architecture repo https://github.com/magento/architecture/blob/master/design-documents/graph-ql/extensibility.md

rogyar: Hi @lenaorobei. Thank you for clarification. We used to use containers for input arguments for merging purposes. So containers with the same name from different modules were supposed to be merged. If you can confirm that the raw input parameters approach (with no containers) has merging capabilities, then we good to go.

Thank you!

lenaorobei: @rogyar yes, mutations/queries with the same name are supposed to be merged. This is the way of adding new parameters.

I would really appreciate if someone can double check that.

lenaorobei:
Closing this PR.
Please see https://github.com/magento/magento2/pull/27620#discussion_r404337029

PR #27647, opened 2020-04-08
Add Inline Form Key Updater to Page Cache | Fix Issue #13746

### LINKED ISSUE #13746: Add to Cart Form wrong Form Key in FPC
When the page initially loads the form*key that is placed on the page is cached between independent requests. This is not an issue if the entire page loads before the "Add to Cart" button is pressed, however if the Add to Cart is clicked before the Javascript has fully initialized (and therefore the form is actually posted via a standard POST HTTP request) then the form*key that is sent in the form data does not match the users session and the product is not added to the cart.

### Preconditions
1. Magento CE 2.2.2 with sample data installed
1. Full Page Cache enabled 
1. Redirect to cart on "Add to Cart" set to yes in configuration (Sales>Checkout>Shopping Cart)

### Steps to reproduce
1. Open a browser window and navigate to a product page.
1. View the page source and search for the form*key that is part of product_addtocart*form form (form id)
1. Note the form_key down
1. Open an incognito window and navigate to the same product page
1. View the page source again and search for the form_key field

### Expected result
1. The two form_key values should be different

### Actual result
1. The two form_key values are the same (meaning the second one is wrong as it will not match the cookie)

### Additional information
- detailed explanation https://github.com/magento/magento2/issues/13746#issuecomment-368062429
- confirmed for `2.4-develop` https://github.com/magento/magento2/issues/13746#issuecomment-570216197

nicolasddev:
### Description (*)

This fix ensures that form keys are updated on frontend pages as soon as the form_key input field is loaded in order to prevent delays resulting in the "Invalid Form Key. Please refresh the page" error. 

A script tag with a unique ID is being added and executed inline directly after each form_key input ensuring its value is updated as soon as possible.

An internally-developed extension using the same approach is running on one of our customer's production server preventing this issue from occurring. The issue was more frequently encountered on long category pages due to the fact that the HTML had to be downloaded in multiple packets.

**This fix is provided by [MageMontreal](https://magemontreal.com), a [Magento Business Solution Partner](https://partners.magento.com/portal/details/partner/id/2232).**

### Fixed Issue

- fixes magento/magento2#13746: Add to Cart Form wrong Form Key in FPC

1. Create a long category page with several visible products
2. Make sure FPC is enabled and visit the page once with a browser
3. Using another browser or incognito, visit the same page which is now cached
4. Try adding a product to cart without waiting for the page to load

### Comments

Before applying this fix, when a long page or a page requiring several resources was opened, the form keys on the page were not updated immediately as the below script hadn't been executed yet.

**Script:** Magento/PageCache/view/frontend/web/js/page-cache.js

As this fix makes sure that the form keys are each updated inline as soon as the content has been downloaded by the browser, the issue doesn't occur anymore.

**Note:** Uses pure JavaScript to ensure that the scripts are executed without having to wait for RequireJS or any dependencies.

### Functional Requirement
Form key input fields must be integrated to the form through the Magento\Framework\View\Element\FormKey block which is best practice, in order to benefit from this fix/improvement.


ihor-sviziev:
Hi @nicolasddev,
First of all thank you so much for your contribution and I'm really sorry that this PR was not processed for long time. Now I'm going to review it and will try to move it forward.

@magento run all tests

app/code/Magento/PageCache/Plugin/Framework/View/Element/InlineFormKeyUpdater.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+ 
+namespace Magento\PageCache\Plugin\Framework\View\Element;
+
+
+class InlineFormKeyUpdater
+{
+    public function afterToHtml(\Magento\Framework\View\Element\FormKey $subject, $result)
+    {
+        return $result.$this->getInlineJavaScript();
```
ihor-sviziev, line 15: Seems like this shouldn't happen in case if FPC is disabled. Am I correct?

app/code/Magento/PageCache/view/frontend/templates/inline_form_key_provider.phtml
```diff
@@ -0,0 +1,21 @@
...
+    
+    var value = "; " + document.cookie;
+    var parts = value.split("; " + 'form_key' + "=");
+    
+    if (parts.length == 2) {
+        return parts.pop().split(";").shift();
+    }
+    
+})();
+</script>
```
ihor-sviziev, line 21: 1. As Magento introduced CSP - we need to use something like this:
https://github.com/magento/magento2/blob/89d3cda99975c0db2967165311184be84d26e9b4/app/code/Magento/Theme/view/frontend/templates/page/js/require_js.phtml#L15

2. Seems like case with missing form_key cookie isn't covered here
https://github.com/magento/magento2/blob/9544fb243d5848a497d4ea7b88e08609376ac39e/app/code/Magento/PageCache/view/frontend/web/js/page-cache.js#L120-L123

3. I think we need to remove following code as we basically replacing it with inline implementation https://github.com/magento/magento2/blob/9544fb243d5848a497d4ea7b88e08609376ac39e/app/code/Magento/PageCache/view/frontend/web/js/page-cache.js#L300-L303

app/code/Magento/PageCache/Plugin/Framework/View/Element/InlineFormKeyUpdater.php
```diff
@@ -0,0 +1,25 @@
...
+    }
+    
+    private function getInlineJavaScript()
+    {
+        $uniqueId = uniqid();
+        
+        return '<script id="'.$uniqueId.'">document.getElementById("'.$uniqueId.'").previousSibling.value = formKey;</script>';
```
ihor-sviziev, line 22: As Magento introduced CSP, we need to replace this code with something like this:
https://github.com/magento/magento2/blob/89d3cda99975c0db2967165311184be84d26e9b4/app/code/Magento/Theme/view/frontend/templates/page/js/require_js.phtml#L15

ihor-sviziev (changes requested):

gabrieldagama:
The risk was set to`medium` due to: the suggested changes may affect other areas/modules.

sidolov:
@nicolasddev  I am closing this PR now due to inactivity.

Issue #27667, opened 2020-04-09
When programatically creating a CMS block, it cannot be assigned to "All store views"


erfanimani:
### Preconditions (*)

1. Magento 2.3.4

### Steps to reproduce (*)

1. In a patch setup script, run the following:

        $block = $this->blockFactory->create();
        $block
            ->setTitle('Foo')
            ->setIdentifier('foo')
            ->setContent('<p>Hello world</p>')
            ->setStoreId(0);

        $this->blockRepository->save($block);

### Expected result (*)

1. CMS block is assigned to "All store views" 
2. in `cms_block_store` table, the block ID is assigned to store_id 0

### Actual result (*)

1. CMS block is assigned to a single store view (I'm not exactly sure how it pick a single store view from the context of a setup script, but it seems random, or it picks the first store view ID).

erfanimani:
Also reproduced on 2.4-develop, see this line here: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Cms/Model/BlockRepository.php#L114

Creating a PR would be easy as it involves changing the `empty` check to a null check, but I'm not sure what consequences this has, or why empty was chosen. Should I go ahead and check if tests pass?

sudheers-kensium:
Hi @erfanimani , I have used below code to create CMS block. 

$block = $blockFactory->create();
 $block
     ->setTitle('Foo')
     ->setIdentifier('foo')
     ->setContent('<p>Hello world</p>')
     ->setStoreId(0)->save();

erfanimani:
@sudheers-kensium ah thanks, yes that works. However, we're discouraged form using the save method directly on models and instead encouraged to use repositories (interfaces if possible).

sudheers-kensium:
@erfanimani I am closing this issue. 
Feel free to reopen if you have any issue.

erfanimani:
Thanks @sudheers-kensium , I still think this is an issue. Using save on the model directly is bad practice and should not be used. Repository should have the same behaviour as calling save directly, therefore I still think this is a bug.

amarroni:
Hi @erfanimani seen to be not a issue. Please try way. As you can see store_id is waiting for array.
$data = [
'title' => "test",
'identifier' => "zzzzzzzzzz",
'is_active' => "1",
'store_id' =>[0],
'content' => "test"];

$block = $this->blockFactory->create()->addData($data);
$this->blockRepository->save($block);

erfanimani:
[...]

PR #27695, opened 2020-04-10
Addressed issue #27667: improved CMS Block/Page setStoreId/getStoreId method signature

erfanimani:
### Description (*)

Small change: Improved method signature.

1. magento/magento2#27667. Although the issue itself was caused due to improper usage of the method, this PR improves the wrong method signature.

Why does it need to expect an array? See following place where the store ID array is processed and cast to an array (same for Page store entity manager save handler):

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Cms/Model/ResourceModel/Block/Relation/Store/SaveHandler.php#L54

And read handler:

https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Cms/Model/ResourceModel/Block.php#L235

(fetchCol returns array)

### DIFF
--- a/app/code/Magento/Cms/Model/Block.php
+++ b/app/code/Magento/Cms/Model/Block.php
@@ -12,8 +12,8 @@ use Magento\Framework\Model\AbstractModel;
 /**
  * CMS block model
  *
- * @method Block setStoreId(int $storeId)
- * @method int getStoreId()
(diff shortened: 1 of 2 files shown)


sidolov:
@erfanimani I'm closing this PR since the issue mentioned in the description is closed. Thank you for the collaboration!

PR #27888, opened 2020-04-17
Fix 24666 disabled modules still included less files

mrtuvn:
### Description (*)

This PR bring the patch for fix issue modules disabled still include less files. It's should not included in output final style files. Previously styles still included in output final css styles

Use case 1: Disable module example Company_CustomCatalog in app/etc/config.php. But assume accidental add module styles in theme folder in
app/design/frontend/theme/theme_name/Company_CustomCatalog/web/css/source/_module.less
Actually after run deploy output css magento still included  Company_CustomCatalog styles in file styles-m.css in pub folder

### Related Pull Requests

Keep continue works by @Echron after long inactivity in PR 25183

### Fixed Issues (if relevant)

1. Fixes magento/magento2#24666 : Static content is deploying for disabled modules

### Manual testing scenarios (*)

1. Magento 2.3.4+
2. blank or luma theme
3. Disable a couple module use less styles: Ex: Magento_Review, Magento_LoginAsCustomerFrontendUi
4. Run deploy static content and clear cache

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds are green)

PR #27606, opened 2020-04-05
Test coverage for PR #27589 (E-mail templates)

lbajsarowicz:
### Description (*)
Additional Integration Tests coverage for PR provided by @ptylek 

### Fixed Issues (if relevant)
1. magento/magento2#27589

### Related issues
1. magento/magento2#27607

### Manual testing scenarios (*)

1. ...
2. ...

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds are green)

PR #27892, opened 2020-04-18
Fixes unstable email integration tests

hostep:
### Description (*)
This is re-adding the fixes added in https://github.com/magento/magento2/pull/25296

Following changes made the email integration tests unstable again:
- [MC-31752](https://github.com/magento/magento2/commit/811097da79b9bcd65c2b8e260715aa77f56beaaf) reverted some changes done by https://github.com/magento/magento2/pull/25296 for no good reason
- [MC-19918](https://github.com/magento/magento2/commit/efd77d3e7faf175c48bae45d3b1c237a00b37043) reverted some changes done by https://github.com/magento/magento2/pull/25296 for no good reason

### Related Pull Requests
Should fix integration tests which now fail in https://github.com/magento/magento2/pull/27888

- None

See https://github.com/magento/magento2/pull/25296

### Questions or comments
Can we do something to avoid Magento core devs for re-introducing this incorrect way of testing email content?

### DIFF
--- a/dev/tests/integration/testsuite/Magento/Newsletter/Model/SubscriberTest.php
+++ b/dev/tests/integration/testsuite/Magento/Newsletter/Model/SubscriberTest.php
@@ -89,7 +89,7 @@ class SubscriberTest extends TestCase
         $this->assertEquals($subscriber, $subscriber->unsubscribe());
         $this->assertContains(
             'You have been unsubscribed from the newsletter.',
-            $this->transportBuilder->getSentMessage()->getRawMessage()
+            $this->transportBuilder->getSentMessage()->getBody()->getParts()[0]->getRawContent()
         );
         $this->assertEquals(Subscriber::STATUS_UNSUBSCRIBED, $subscriber->getSubscriberStatus());
         // Subscribe and verify
@@ -97,7 +97,7 @@ class SubscriberTest extends TestCase
         $this->assertEquals(Subscriber::STATUS_SUBSCRIBED, $subscriber->getSubscriberStatus());
         $this->assertContains(
             'You have been successfully subscribed to our newsletter.',
-            $this->transportBuilder->getSentMessage()->getRawMessage()
+            $this->transportBuilder->getSentMessage()->getBody()->getParts()[0]->getRawContent()
         );
     }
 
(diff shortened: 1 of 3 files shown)


lbajsarowicz:
I've addressed that in:
https://github.com/magento/magento2/pull/27606
Using DOM structure instead of `assertContains` (which is deprecated, btw).

There's also ticket to introduce DOM assertions:
https://github.com/magento/magento2/issues/27607

hostep:
Aha, thanks for the info @lbajsarowicz!

Let me try to update this PR to use the same DOM assertions then.
That will then also settle ticket https://github.com/magento/magento2/issues/27607 I assume?

hostep:
@lbajsarowicz: updated PR

Before we fix all the occurrences (I count about 16 more classes using the `getRawContent` method), here are a couple of questions first:
- Can you explain the deprecation of the `assertContains` method a bit, I can't seem to find information about that, what should it be replaced with?
- I've written a generic `assertEmailContains` method, which can probably be re-used in all classes. Is that one ok? Where can we put this method so we don't need to copy it in all the test files?

Thanks!

lbajsarowicz:
@hostep assertContains won't work with string anymore

https://phpunit.readthedocs.io/en/9.0/assertions.html

It's replacement is assertStringContainsString

hostep:
I came by that article indeed, and tested `assertStringContainsString`, but `assertStringContainsString` doesn't exists yet in PHPUnit 6.5 which Magento is using ...
I've seen you working on PHPUnit 8 compatibility, but that's not finished yet I think?

lbajsarowicz:
Could you change this to `assertStringContainsString`? We will add your change once the PHPUnit 8 is finally introduced (team of people are working on this right now)

hostep:
Hmm, I'd rather solve this as fast as possible, because the unstable email integration tests cause failures in this very good PR: https://github.com/magento/magento2/pull/27888

If you want me to revert back to my old changes and merge that first. And later handle this better way of handling it, that's also fine by me.

lbajsarowicz:
Yeah, so choose the MVP, and later on we'll do it the right way : )

hostep:
Ok, I've reverted to my initial changes.

I've pushed my other changes to a different branch, just in case if somebody want to get some inspiration from them: https://github.com/hostep/magento2/compare/c163118...first-part-to-fix-issue-27607 when trying to fix #27607

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-7467](https://jira.corp.magento.com/browse/ENGCOM-7467) has been created to process this Pull Request

engcom-Alfa:
:heavy_check_mark: QA Passed

PR #27914, opened 2020-04-20
[GraphQl] Single mutation for adding different types of products to the shopping cart

### LINKED ISSUE #28524: Single mutation for Add to Cart that supports all product types
Right now we have 7 mutations for adding different types of products to the cart. When we started working on wishlist functionality it turned out that we will have to create 14 more mutations (adding to wishlist and moving from wishlist to cart). It leads to overcomplicated flow and does not help to improve developers' experience.

**Expected behavior**
A single mutation is used to add all types of products to the cart.
Supported Product Types

Simple
Configurable
Virtual
Downloadable
Bundled
Gift Card
Grouped

**Acceptance Criteria**
All product types can be added to the cart with a single mutation.
User-friendly errors are returned in case the add to cart operation is unsuccessful.
All existing Add to Cart tests must pass for the new mutation.

**Benefits**
Less code complexity

**Additional information**
Approved Schema

- https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/AddProductsToCart.graphqls
- https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md

rogyar:
### Description (*)
This PR introduces a new approach that allows adding different types of products using the same mutation. 

More details will be provided soon ...

[Solution Architecture](https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md)

### Related Pull Requests
https://github.com/magento/partners-magento2ee/pull/260
https://github.com/magento/partners-magento2-infrastructure/pull/17

1. [Single mutation for Add to Cart that supports all product types](https://github.com/magento/magento2/issues/28524)

TBP

### Questions or comments
The data providers for different product types were decoupled as separate modules according to the suggestion from @mslabko. 

- This PR does not include resolvers for ID_V2 of different options. This functionality will be provided in a separate PR
- This PR does not include a possibility to use Parent SKU + SKU pair for adding items to cart (only via Parent SKU + product options in the same manner as the standard storefront works). The possibility to use Parent SKU + SKU may be provided in a separate PR (still in discussion)


app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -22,6 +22,32 @@ type Mutation {
     setPaymentMethodAndPlaceOrder(input: SetPaymentMethodAndPlaceOrderInput): PlaceOrderOutput @deprecated(reason: "Should use setPaymentMethodOnCart and placeOrder mutations in single request.") @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\SetPaymentAndPlaceOrder")
     mergeCarts(source_cart_id: String!, destination_cart_id: String!): Cart! @doc(description:"Merges the source cart into the destination cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\MergeCarts")
     placeOrder(input: PlaceOrderInput): PlaceOrderOutput @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\PlaceOrder")
+
+    addProductsToCart(cart_id: String!, cart_items: [CartItemInput!]!): AddProductsToCartOutput @doc(description:"Add products to the cart. Provide single interface for all product types ") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\AddProductsToCart")
+}
+
+type CheckoutUserInputError @doc(description:"An error encountered while adding an item the the cart."){
```
paliarush: We need to make sure this type definition is not duplicated in re-order schema.

paliarush (commented):
GraphQL schema looks good and corresponds to the design https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md

lenaorobei:
@paliarush please confirm interfaces comply with requirements and standards.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,216 @@
...
+                    'Could not add the product with SKU %sku to the shopping cart: %message',
+                    ['sku' => $sku, 'message' => $e->getMessage()]
+                )->render(),
+                $cartItemPosition
+            );
+            return;
+        }
+
+        if (is_string($result)) {
+            $errors = array_unique(explode("\n", $result));
```
lenaorobei, line 168: For complex errors we should be able utilize composite `GraphQlInputException`. Please see example `\Magento\QuoteGraphQl\Model\Cart\AddSimpleProductToCart::execute`.

app/code/Magento/QuoteGraphQl/Model/Resolver/AddProductsToCart.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $cartItems = [];
+        foreach ($cartItemsData as $cartItemData) {
+            $cartItems[] = (new CartItemFactory())->create($cartItemData);
+        }
+
+        /** @var AddProductsToCartOutput $addProductsToCartOutput */
+        $addProductsToCartOutput = $this->addProductsToCart->execute($maskedCartId, $cartItems);
+
+        return [
```
lenaorobei, line 77: With mentioned exception we should be able to return both - cart data and errors.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
rogyar: Hi @lenaorobei. Thank you for your suggestion. In case of GraphQl modules, this approach will work. However, in this particular case, we work with the `Quote` module that has no dependency to GraphQl modules. 
We use a similar approach in the `Reorder` functionality 

https://github.com/magento/magento2/blob/03a9a241ee14b0b9761d1e40db0370e22e999389/app/code/Magento/Sales/Model/Reorder/Reorder.php#L189

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,214 @@
...
+     * Error message codes
+     */
+    private const ERROR_PRODUCT_NOT_FOUND = 'PRODUCT_NOT_FOUND';
+    private const ERROR_INSUFFICIENT_STOCK = 'INSUFFICIENT_STOCK';
+    private const ERROR_NOT_SALABLE = 'NOT_SALABLE';
+    private const ERROR_UNDEFINED = 'UNDEFINED';
+    /**#@-*/
+
+    /**
+     * List of error messages and codes.
```
mslabko, line 36: Please, add a comment that this is an ad-hoc solution to get an error code related to some message. And maybe we can introduce a new class to provide error code based on error message to avoid logic duplicate: \Magento\Sales\Model\Reorder\Reorder::getErrorCode

app/code/Magento/Quote/Model/Cart/Data/SelectedOption.php
```diff
@@ -0,0 +1,34 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Quote\Model\Cart\Data;
+
+class SelectedOption
```
mslabko, line 13: Please add a description

app/code/Magento/Quote/Model/Cart/Data/EnteredOption.php
```diff
@@ -0,0 +1,51 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\Quote\Model\Cart\Data;
+
+class EnteredOption
```
mslabko, line 13: Please add a description

app/code/Magento/Quote/Model/Cart/Data/AddProductsToCartOutput.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @return CartInterface
+     */
+    public function getCart(): CartInterface
+    {
+        return $this->cart;
+    }
+
+    /**
+     * Get errors happened during reorder
```
mslabko: Fix method description

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
mslabko: @lenaorobei this ugly approach comes from \Magento\Quote\Model\Quote::addProduct and due to we need to provide error message(s) we need to "parse" it in that way, as was "designed" in Quote::addProduct
```
       /**
         * Error message
         */
        if (is_string($cartCandidates) || $cartCandidates instanceof \Magento\Framework\Phrase) {
            return (string)$cartCandidates;
        }
```

app/code/Magento/QuoteConfigurableOptions/Model/Cart/BuyRequest/SuperAttributeDataProvider.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Validates the provided options structure
+     *
+     * @param array $optionData
+     * @throws LocalizedException
+     */
+    private function validateInput(array $optionData): void
+    {
+        if (count($optionData) !== 3) {
```
akaplya, line 70: the mentioned structure is an implementation detail that should not be exposed to a client. the only reason we are using base64 because we can not introduce unique ID for all options values in a backward-compatible way right now. but we are going to introduce it eventually.


Anyway, all IDs should be retrieved from the regular option values query.

akaplya (changes requested):

app/code/Magento/QuoteBundleOptions/Model/Cart/BuyRequest/BundleDataProvider.php
```diff
@@ -0,0 +1,76 @@
...
+     * Validates the provided options structure
+     *
+     * @param array $optionData
+     * @throws LocalizedException
+     */
+    private function validateInput(array $optionData): void
+    {
+        if (count($optionData) !== 4) {
+            $errorMessage = __('Wrong format of the entered option data. ' .
+                'Must be "bundle/option_id/option_value_id/option_quantity"');
```
mslabko: We cannot expose implementation details to Customer. "id" should be interpreted as one unit. Current implementation with encoded value just a workaround until UUID will be introduced in Magento (see details in https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md#options)

Maybe we can log it just for debug purpose and throw a generic error message aka "Provided option %1 is not valid product with SKU %2". I think log details is acceptable and will help/save time during debug. @akaplya  what do you think?

mslabko (changes requested):
Please, check minor errors.

Do you plan to implement in this phase "id_v2" field for product schema? Without this information, the Customer will not be able to add the product to the cart without reverse-engineering and relying on the current implementation. Please see details here: https://github.com/magento/architecture/blob/master/design-documents/graph-ql/coverage/add-items-to-cart-single-mutation.md#options

rogyar:
Hi @mslabko. Thank you for your suggestions. We are working on the id_V2 implementation for options within the scope of another PR. I didn't want to overwhelm this huge PR with more details.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
rogyar: Totally agree. I was thinking about that but then came up with the conclusion that we will need this new class on the framework level (since we will need to reuse it in different independent modules). As long as we plan to deliver this PR as a part of 2.4.0 release, I don't want to introduce new interfaces or implementations on the framework level. So, let's leave a comment for now, and then, once this PR is delivered, I will introduce a new class and refactor the existing logic. 

Thank you.

app/code/Magento/QuoteBundleOptions/Model/Cart/BuyRequest/BundleDataProvider.php
akaplya: Sounds reasonable

rogyar:
@magento run all tests

mslabko (approved):

app/code/Magento/Quote/Model/Cart/AddProductsToCartInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+declare(strict_types=1);
+
+namespace Magento\Quote\Model\Cart;
+
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Quote\Model\Cart\Data\AddProductsToCartOutput;
+
+/**
+ * Unified approach to add products to the Shopping Cart.
```
paliarush: Should we introduce storefront API instead (in `Magento/Quote/Api` directory)? Then it should also be responsible for authorization checks. Current approach seems unreliable and can lead to pretty serious security vulnerabilities in case of mistake in the client implementation.

rogyar:
@magento run all tests

rogyar:
Hi @paliarush. Thank you for your review. 
According to our discussion yesterday, I've removed the interface so only the implementation is left

rogyar:
@magento run all tests

rogyar:
@magento run Functional Tests CE

engcom-Bravo (changes requested):
Please, add a message when the _sku_ is incorrect or missing
When I use the following query
```graphql
mutation {
  addProductsToCart(
    cartId: "bkEkcQug6ZqR9xWtakN1YhCzpHeTe6u8"
    cartItems: [
      {
        sku: "no such sku"
        quantity: 0
      }
    ]
  ) {
    cart {
      id
      items{
        quantity
        product{
          name
        }
      }
    }
  }
}
``` 
the response is 
```graphql
{
  "data": {
    "addProductsToCart": {
      "cart": {
        "id": "bkEkcQug6ZqR9xWtakN1YhCzpHeTe6u8",
        "items": [
          {
            "quantity": 1,
            "product": {
              "name": "Sprod"
            }
          }
        ]
      }
    }
  }
}
```
instead of displaying a message like on ```addSimpleProductsToCart``` mutation. 
```graphql
{
  "errors": [
    {
      "message": "Missed \"sku\" in cart item data",
      "extensions": {
        "category": "graphql-input"

``` 
if the sku is empty ( ``` sku: " " ``` ) and 
```graphql {
  "errors": [
    {
      "message": "Could not find a product with SKU \"123\"",
      "extensions": {
        "category": "graphql-no-such-entity"
```
when there is no such _sku_

---------------------------------------------------------------------------------------

Please, cover the behavior when the quantity is not available. The query response might be
```graphql
{
  "errors": [
    {
      "message": "Could not add the product with SKU Sprod to the shopping cart: The most you may purchase is 10000.",
      "extensions": {
        "category": "graphql-input"
```

rogyar:
Hi @engcom-Bravo. Thank you for testing. In this particular solution we use the same approach as for the reorder mutation

https://github.com/magento/magento2/blob/b9e28fd3dc2b32a03984f4f62fc5cf612842932f/app/code/Magento/SalesGraphQl/etc/schema.graphqls#L27

So we have a separate field called `userInputErrors` which shows errors if any occur. 
To see the error, you need to request this field explicitly 

Query example with wrong SKU:
```graphql
mutation {
  addProductsToCart(cartId:"OsVNT94k0SHggNKchkV72n3XriSvomVA", cartItems: [
    {
    	sku:"simple-custom-optionss"
      quantity:1
      selected_options: [
        "Y3VzdG9tLW9wdGlvbi8w"
      ],
      entered_options: [
        {
          id: "Y3VzdG9tLW9wdGlvbi8z"
          value: "Here is the entered value"
        }
      ]
      
    }
  ]) {
    cart {
      id
      items {
        product {
          name
        }
      }
    },
    userInputErrors {
      message
    }
  }
}
```

Response

```graphql
{
  "data": {
    "addProductsToCart": {
      "cart": {
        "id": "OsVNT94k0SHggNKchkV72n3XriSvomVA",
        "items": [
          {
            "product": {
              "name": "Simple With Custom Options"
            }
          },
          {
            "product": {
              "name": "Downloadable"
            }
          },
          {
            "product": {
              "name": "Simple With Custom Options"
            }
          }
        ]
      },
      "userInputErrors": [
        {
          "message": "Could not find a product with SKU \"simple-custom-optionss\""
        }
      ]
    }
  }
}
```

Please, let me know if you need more details.
Thank you.

rogyar:
@magento run Static Tests

rogyar:
Hi @lenaorobei. I've adjusted the schema correspondingly. The failing static tests now point to the parameters that were not introduced in this PR (mergeCarts mutation).

Thank you

nrkapoor:
> Hi @lenaorobei. I've adjusted the schema correspondingly. The failing static tests now point to the parameters that were not introduced in this PR (mergeCarts mutation).
> 
> Thank you

@prabhuram93 @cpartica @avattam06 Please review and see why the tests are failing.

prabhuram93:
As @rogyar mentioned the static tests are failing because of mergeCarts mutation on the same schema file, mergeCarts(source_cart_id: String!, destination_cart_id: String!). @paliarush I am assuming updating the args now will be a breaking change. How to approach this? 
cc: @cpartica @lenaorobei

nrkapoor:
> > As @rogyar mentioned the static tests are failing because of mergeCarts mutation on the same schema file, mergeCarts(source_cart_id: String!, destination_cart_id: String!). @paliarush I am assuming updating the args now will be a breaking change. How to approach this?
> > cc: @cpartica @lenaorobei
> 
> There is a proposed change for mergeCarts. https://github.com/magento/architecture/blob/mergeCarts-optional-destination/design-documents/graph-ql/coverage/mergeCarts-optional-destination.md
> This is not going to be a breaking change but I am not sure if this will resolve the static test failures.

@prabhuram93

paliarush:
@prabhuram93  if static test is failing because of underscored arguments in the `mergeCarts` I would not touch them for now. It should not fail on the mainline after merge.

prabhuram93:
@magento run all tests

rogyar:
@magento run all tests

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,222 @@
...
+     */
+    private const MESSAGE_CODES = [
+        'The required options you selected are not available' => self::ERROR_NOT_SALABLE,
+        'Product that you are trying to add is not available' => self::ERROR_NOT_SALABLE,
+        'This product is out of stock' => self::ERROR_NOT_SALABLE,
```
prabhuram93: Hi @rogyar. Solid work! I have couple of minor comments. shouldn't  'This product is out of stock' point to self::ERROR_INSUFFICIENT_STOCK? 
cc: @keharper

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,222 @@
...
+     * TODO: introduce a separate class for getting error code from a message
+     *
+     * @param string $message
+     * @return string
+     */
+    private function getErrorCode(string $message): string
+    {
+        $code = self::ERROR_UNDEFINED;
+
+        $matchedCodes = array_filter(
```
prabhuram93: array_filter scans through all items. Since we are returning a single error code anyway, simple foreach with an early return should do the job. Figured out substring search is inevitable on this array map, at-least we can save some computational complexity. 

foreach ($haystack as $key => $value) {
        if (false !== stripos($value, $needle)) {
            return $key;
        }
    }

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -315,6 +319,11 @@ type SetGuestEmailOnCartOutput {
     cart: Cart!
 }
 
+input EnteredOptionInput {
+    id: String! @doc(description: "base64 encoded option ID")
```
prabhuram93: field type should be ID! 
Based on this https://github.com/magento/architecture/pull/395/files the description shouldn't mention anywhere  how the id is encoded (should be opaque to users/clients). The clients will possess the id which had already been encoded by the server.

prabhuram93: @keharper need your help for schema descriptions.

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -368,3 +377,22 @@ type Order {
...
+    code: CheckoutUserInputErrorCodes! @doc(description: "Checkout-specific error code")
+}
+
+type AddProductsToCartOutput {
+    cart: Cart!
+    userInputErrors:[CheckoutUserInputError]!
+}
+
+enum CheckoutUserInputErrorCodes {
+    PRODUCT_NOT_FOUND
```
prabhuram93, line 387: @keharper even here. you can find what each code means on this class /Quote/Model/Cart/AddProductsToCart.php

app/code/Magento/QuoteGraphQl/Model/Resolver/AddProductsToCart.php
```diff
@@ -0,0 +1,93 @@
...
+            'cart' => [
+                'model' => $addProductsToCartOutput->getCart(),
+            ],
+            'userInputErrors' => \array_map(
+                function (Error $error) {
+                    return [
+                        'code' => $error->getCode(),
+                        'message' => $error->getMessage(),
+                        'path' => $error->getCartItemPosition()
```
prabhuram93: path is expecting an array of string, int is being returned. This might throw an error. Please check.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,222 @@
...
+     * @param string $message
+     * @param int $cartItemPosition
+     * @param string|null $code
+     * @return void
+     */
+    private function addError(string $message, int $cartItemPosition = 0, string $code = ''): void
+    {
+        $this->errors[] = new Data\Error(
+            $message,
+            $code ?? $this->getErrorCode($message),
```
prabhuram93: when $code = "", this validation wont work. GraphQl will error out. Please change validation to something like this
empty($code) ? $this->getErrorCode($message) : $code,

You can replicate this issue by trying to get an out of stock error.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,222 @@
...
+            $cartItemPosition
```
prabhuram93, line 187: Also I found a related issue. when we keep adding products to the cart, and if we run into errors like out of stock, the cart becomes empty. 
Jul-10-2020 00-15-10

prabhuram93 (changes requested):

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
```diff
@@ -0,0 +1,222 @@
...
+     * @param Data\CartItem[] $cartItems
+     * @return AddProductsToCartOutput
+     * @throws NoSuchEntityException Could not find a Cart with provided $maskedCartId
+     */
+    public function execute(string $maskedCartId, array $cartItems): AddProductsToCartOutput
+    {
+        $cartId = $this->maskedQuoteIdToQuoteId->execute($maskedCartId);
+        $cart = $this->cartRepository->get($cartId);
+
+        foreach ($cartItems as $n => $cartItem) {
```
avattam06: Can we change the variable $n to a more meaningful name?

avattam06 (changes requested):

app/code/Magento/QuoteGraphQl/Model/Resolver/AddProductsToCart.php
dthampy: @rogyar :
Below is a list of few of the use cases that are failing.  Please take a look.
Case 1: add product to cart with a qty> the stock available 
Actual result: 
```
"errors": {
      "debugMessage": "Expected a value of type \"CheckoutUserInputErrorCodes\" but received: (empty string)","message": "Internal server error",
```
Case 2 : Add product with qty=0 and a valid sku, 

> Expected: cart is empty and error message to enter a quantity >0
> Actual: The product is still added to the cart with qty =1

Case 3: Add disabled product to cart
> Expected: Cart is empty and message that "product is not available"
> Actual result : 
```
exception: {
  "errors": [
    {
      "debugMessage": "Expected a value of type \"CheckoutUserInputErrorCodes\" but received: (empty string)",
      "message": "Internal server error",
Now, if we omit "code" from userInputError{} and run the same query, below exception is returned
exception: ""errors": [
    {
      "debugMessage": "User Error: expected iterable, but did not find one for field CheckoutUserInputError.path.","
```
Case 4 : Same result as above when you try to add a product with "Out of Stock" status.

app/code/Magento/QuoteBundleOptions/Model/Cart/BuyRequest/BundleDataProvider.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @throws LocalizedException
+     */
+    public function execute(CartItem $cartItem): array
+    {
+        $bundleOptionsData = [];
+
+        foreach ($cartItem->getSelectedOptions() as $optionData) {
+            // phpcs:ignore Magento2.Functions.DiscouragedFunction
+            $optionData = \explode('/', base64_decode($optionData->getId()));
```
prabhuram93, line 32: Hmm. I don't think we should be doing this. An ID is something that should be opaque and there could be several ways we arrive at an ID (base64 is just one of it). And we should not be decoding it to get optionId,  optionValueId etc. It should be explicit from the clients. And this decode is used on almost every data provider on this PR. @paliarush @DrewML is this ok?

paliarush: As soon as no decoding/encoding is required on the storefront - the approach follows the design.

prabhuram93: Let me try to explain this better. Consider the case for a bundle product with 2 options.

Option 1 has 2 products, with **dropdown** type and **user can enter quantity**
Option 2 has 2 products with **multiselect** type with a fixed quantity

So the way we are encoding the id for each product in the option is base_64```(bundle/<option-id>/<option-value-id>/<option-quantity>)```. It takes in the hypothesis that option_quantity is always fixed, which is not since user can change it.

**Problem 1**
When we add the bundle item to the cart, the design is not respecting the quantity the user is passing, rather its decoding from the encoded id. The user entered quantities will be ignored here.

**Problem 2**
Screen Shot 2020-07-14 at 5 22 48 PM

For a configurable product the clients have the code and value_index. And I am not sure why we are encoding them and also returning the non encoded values at the same time. Might be to support the old mutations? 

It is evident that we are encoding the ids for entities/types with very specific properties of each product type. More than cache just keys it will have business/behavioral implications. If we are going to rely on the ids for decoding and using the properties I think we should have strict interfaces per product type for encoding/decoding and also interfaces for abstracting the algorithm we use for the same. Currently its all over the place.

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -22,6 +22,7 @@ type Mutation {
     setPaymentMethodAndPlaceOrder(input: SetPaymentMethodAndPlaceOrderInput): PlaceOrderOutput @deprecated(reason: "Should use setPaymentMethodOnCart and placeOrder mutations in single request.") @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\SetPaymentAndPlaceOrder")
     mergeCarts(source_cart_id: String!, destination_cart_id: String!): Cart! @doc(description:"Merges the source cart into the destination cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\MergeCarts")
     placeOrder(input: PlaceOrderInput): PlaceOrderOutput @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\PlaceOrder")
+    addProductsToCart(cartId: String!, cartItems: [CartItemInput!]!): AddProductsToCartOutput @doc(description:"Add products to the cart. Provide single interface for all product types ") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\AddProductsToCart")
```
keharper:
```suggestion
    addProductsToCart(cartId: String!, cartItems: [CartItemInput!]!): AddProductsToCartOutput @doc(description:"Add any type of product to the cart") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\AddProductsToCart")
```
```suggestion
    addProductsToCart(cartId: String!, cartItems: [CartItemInput!]!): AddProductsToCartOutput @doc(description:"Add products to the cart. Provide single interface for all product types ") @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\AddProductsToCart")
```

dthampy:
Adding additional use cases that are failing : 
In addition to the ones that are listed here: https://github.com/magento/magento2/pull/27914/#discussion_r453959504
```
Use case 5: 
Create a configurable product with two attribute "color" and "size" each having at least couple values
Also make sure to have a customizable option of type drowndown ( I just chose a selectable option type) with option values
Now, add the above configurable product to cart using the single mutation.
Please the screenshot attached for the query and the response is an internal server error. Please take a look
```
addConfigurableToCartWithCustomOptions

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -51,6 +52,9 @@ input VirtualProductCartItemInput {
 input CartItemInput {
     sku: String!
     quantity: Float!
+    parent_sku: String,
```
keharper:
```suggestion
    parent_sku: String @doc(description: "For child products, the SKU its parent product")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -51,6 +52,9 @@ input VirtualProductCartItemInput {
...
+    selected_options: [String!]
+    entered_options: [EnteredOptionInput!]
```
keharper:
```suggestion
    entered_options: [EnteredOptionInput!] @doc(description: "An array of entered options for the base product, such as personalization text")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
keharper: Note: I deleted the comma that was originally after String

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -51,6 +52,9 @@ input VirtualProductCartItemInput {
 input CartItemInput {
     sku: String!
     quantity: Float!
+    parent_sku: String,
+    selected_options: [String!]
```
keharper:
```suggestion
    selected_options: [String!] @doc(description: "The selected options for the base product, such as color or size")
```

app/code/Magento/QuoteGraphQl/Model/Resolver/AddProductsToCart.php
rogyar: Thanks for mentioning that. This behavior is fixed in the recent commits.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
rogyar: Good point, I've revised this behavior slightly in the recent commits.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
rogyar: That's a good catch. The problem is it's a default system behavior introduced to the core a couple of months ago.

https://github.com/magento/magento2/blob/faced716b53f4d152113d0a0171cfc171ecbefa0/app/code/Magento/Quote/Model/Quote.php#L1684

I'm not sure about the best way to overcome it. @prabhuram93 maybe you have ideas, thank you.

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/AddDownloadableProductToCartSingleMutationTest.php
```diff
@@ -0,0 +1,211 @@
...
+    private function getQuery(
+        string $maskedQuoteId,
+        int $qty,
+        string $sku,
+        string $customizableOptions
+    ): string {
+        return <<<MUTATION
+mutation {
+    addProductsToCart(
+        cart_id: "{$maskedQuoteId}",
```
dthampy: Please change the argument to "cartId"

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/AddDownloadableProductToCartSingleMutationTest.php
```diff
@@ -0,0 +1,211 @@
...
+        cart_items: [
```
dthampy: and here to cartItems

dev/tests/api-functional/testsuite/Magento/GraphQl/Bundle/AddBundleProductToCartSingleMutationTest.php
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * @var ProductRepositoryInterface
+     */
+    private $productRepository;
+
+    /**
+     * @inheritdoc
+     */
+    protected function setUp()
```
dthampy: please add the return type

dev/tests/api-functional/testsuite/Magento/GraphQl/Bundle/AddBundleProductToCartSingleMutationTest.php
```diff
@@ -0,0 +1,225 @@
...
+        $optionId = $option->getId();
+        $selectionId = $selection->getSelectionId();
+
+        $bundleOptionIdV2 = $this->generateBundleOptionIdV2((int) $optionId, (int) $selectionId, 1);
+        $maskedQuoteId = $this->quoteIdToMaskedId->execute((int)$this->quote->getId());
+
+        $query = <<<QUERY
+mutation {
+    addProductsToCart(
+        cart_id: "{$maskedQuoteId}",
```
dthampy: Please check for all occurrences of  "cart_id " and "cart_items" and change them to "cartId" and "cartItems"

rogyar:
Hi @dthampy. Regarding the issue you've mentioned here 

https://github.com/magento/magento2/pull/27914#issuecomment-659573092

I'm able to reproduce the same issue but it's not connected to the current changeset somehow. The problem is in an attempt to retrieve customizable option for a configurable product in the following resolver 

https://github.com/magento/magento2/blob/89ba392798b3e4ea0914e88c3341efb49ba327d3/app/code/Magento/ConfigurableProductGraphQl/Model/Resolver/ConfigurableCartItemOptions.php#L57

The changes we are introducing within the scope of this PR don't modify the way how we retrieve the cart information but the way of input for this information (but not the way how it's stored in DB, it stays the same). So we have the same issue for retrieving customizable dropdown option value even if we use the existing mutation for adding configurable products.

dthampy:
> Hi @dthampy. Regarding the issue you've mentioned here
> 
> [#27914 (comment)](https://github.com/magento/magento2/pull/27914#issuecomment-659573092)
> 
> I'm able to reproduce the same issue but it's not connected to the current changeset somehow. The problem is in an attempt to retrieve customizable option for a configurable product in the following resolver
> 
> https://github.com/magento/magento2/blob/89ba392798b3e4ea0914e88c3341efb49ba327d3/app/code/Magento/ConfigurableProductGraphQl/Model/Resolver/ConfigurableCartItemOptions.php#L57
> 
> The changes we are introducing within the scope of this PR don't modify the way how we retrieve the cart information but the way of input for this information (but not the way how it's stored in DB, it stays the same). So we have the same issue for retrieving customizable dropdown option value even if we use the existing mutation for adding configurable products.

Thank you @rogyar for getting back. 
I will create a bug for the issue and will let the PO to prioritize.
Just a general feedback on the tests:
As part of validating UserInputError message , please include "code" and "path" in addition, to make sure that we cover that.
Also it 'd be great if negative cases like(i) addProduct with quantity=0   (ii) or adding a disabledProduct or (iv)addProduct sku=""  etc.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
prabhuram93: Hey @rogyar. This took me a while to figure it out. Apparently this is happening only on the new mutation, not the existing ones. 

Even though the cart item is being deleted in core, the cart is not saved. But I see in this PR the cart is being saved here even when there are errors. app/code/Magento/Quote/Model/Cart/AddProductsToCart.php:116. I don't see any to reason to do that, please correct me if I am wrong.

So fixing that should fix this issue.

dev/tests/api-functional/testsuite/Magento/GraphQl/ConfigurableProduct/AddConfigurableProductToCartSingleMutationTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+        $option = current($cartItem['configurable_options']);
+        self::assertEquals($attributeId, $option['id']);
+        self::assertEquals($valueIndex, $option['value_id']);
+        self::assertArrayHasKey('option_label', $option);
+        self::assertArrayHasKey('value_label', $option);
+    }
+
+    /**
+     * Generates Id_v2 for super configurable product super attributes
```
prabhuram93: Hi @rogyar . Based on this arch PR https://github.com/magento/architecture/pull/396/files would be nice if you can update the id_v2 references to uid.

prabhuram93:
@rogyar, I just wanted to update the status of the PR. Most of the issues are fixed. The following items are pending

**Pending Items**
comment1: https://github.com/magento/magento2/pull/27914#discussion_r459636002
comment2: https://github.com/magento/magento2/pull/27914#discussion_r459673501

**For Tests**:
Please fix the comments Deepty has put in place
I checked in with @dthampy and additional coverage is needed on the following items

```
Case 1: add a simple product with custom option ( both entered and selected option types)
Case 1. Add the max qty of product +1  to the cart
Case 2: Add the product with sku:""
Case 3: Add qty=0 or  -ve qty 
Case 5: Add a disable product to cart
Case 6: End to end test case ( a general idea below)
   -- Create a product (product with custom options that has both entered and selected options)
   -- Query product and get the uid 
   -- use that uid to add the product to the cart using the new addProductsToCart mutation
```

Once these are complete we will be in a very good position to merge this. 
Thanks for your hard work on this. We really appreciate it!

rogyar:
Hi @prabhuram93. Thanks a lot for summing everything up. I'm working on the pending issues.

app/code/Magento/Quote/Model/Cart/AddProductsToCart.php
rogyar: Hi @prabhuram93. That's a good point. But unfortunately, without explicit cart saving on `Model/Cart/AddProductsToCart.php:116` the quote is not being saved anywhere. You will still get the correct cart output since the corresponding model with the current changes will be used. But you will not get the record saved in the database.

You may test it in the following way: 
- Set stock qty of some product (let's say "simple1") to 10
- Comment out the following line `Model/Cart/AddProductsToCart.php:116`
- Add "simple1" to the shopping cart with QTY 3
- Add "simple1" using the same mutation (QTY 3)
- Check the result, you will have total QTY 3 (not 6) because the result of the very first mutation is not being saved anywhere. 

So I'm still thinking that the approach of removing item if it has an error
https://github.com/magento/magento2/blob/faced716b53f4d152113d0a0171cfc171ecbefa0/app/code/Magento/Quote/Model/Quote.php#L1684

Does the dirty thing. 
I'm going to spend more time to overcome this weird behavior but any other idea will be very appreciated. 

Thank you!

prabhuram93: Hey @rogyar. I understand that cart is not being saved anywhere else. But I don't see any reason to save the cart when there are errors. So in the scenario you mentioned I wouldn't just comment out the line in step 2, rather we can handle it to save the cart when only there are no errors. That way when there are errors just the error objects will be updated for that particular mutation and the cart will not be saved. When there are no errors cart will be saved. I managed to try it out and it works for both the cases you mentioned. Let me know if this is still unclear. Thanks!

rogyar: Got it, thank you. We may adjust it. The only thing that will be lost in this case, described in the following scenario.

- Add 3 different items to the cart
- There's no issue with first 2 but 3rd is problematic
- We will have no items in the cart instead of having 2 of them

... but, it's another story :)

prabhuram93: Hey @rogyar, I don't see a use case where we add 3 different items to the cart at the same point. We might do it one after the other, but not together. So we are good on that front.

prabhuram93:
@magento run all tests

prabhuram93:
@magento run all tests

rogyar:
@magento run all tests

avattam06:
**Case 1:** Internal server error: On adding a `simple product` to cart with the `single mutation query` with only the `required selected options` is given(no entered options).

OnlySelectedAdminOptions

OnlySelectedOptionsOnSP

OnlySelectedOptionsOnSPStorefront

**Case 2:** Internal server error: On adding a simple product to cart with both the `required and non required entered options`(no selected options).
OnlyEnteredOptionsOnSPSF

OnlyEnteredOptionsOnSPSFAddedtoCart


**Case 3:** Internal server error: On adding a `simple product` with both the `required selected and entered options`.
BothSelected EnteredAdmin

BothSelected EnteredOptiononSP

BothSelected EntredQueryError


On all the above cases, we are able to add products on the storefront to the cart. Please also check if the selected or entered options are not required cases for the above scenarios.

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -315,6 +319,11 @@ type SetGuestEmailOnCartOutput {
     cart: Cart!
 }
 
+input EnteredOptionInput {
+    id: ID! @doc(description: "base64 encoded option ID")
```
keharper:
```suggestion
    id: ID! @doc(description: "An encoded option ID")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -315,6 +319,11 @@ type SetGuestEmailOnCartOutput {
...
+    value: String!
```
keharper:
```suggestion
    value: String!  @doc(description: "The value of the entered option")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -368,3 +377,22 @@ type Order {
     order_number: String!
     order_id: String @deprecated(reason: "The order_id field is deprecated, use order_number instead.")
 }
+
+type CheckoutUserInputError @doc(description:"An error encountered while adding an item the the cart.") {
```
keharper:
```suggestion
type CheckoutUserInputError @doc(description:"An error encountered while adding an item to the cart.") {
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -368,3 +377,22 @@ type Order {
...
+    message: String! @doc(description: "Localized error message")
```
keharper:
```suggestion
    message: String! @doc(description: "A localized error message")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -368,3 +377,22 @@ type Order {
...
+    path: [String]! @doc(description: "Path to the input field that caused an error. See the GraphQL specification about path errors for details: http://spec.graphql.org/draft/#sec-Errors")
+    code: CheckoutUserInputErrorCodes! @doc(description: "Checkout-specific error code")
+}
+
+type AddProductsToCartOutput {
+    cart: Cart!
```
keharper:
```suggestion
    cart: Cart!  @doc(description: "The cart after products have been added")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -368,3 +377,22 @@ type Order {
...
+    userInputErrors:[CheckoutUserInputError]!
```
keharper:
```suggestion
    userInputErrors:[CheckoutUserInputError]! @doc(description: "An error encountered while adding an item to the cart.")
```

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
keharper: I can't say with certainty.

keharper (commented):
My comments on the attribute descriptions

prabhuram93:
@magento run all tests

rogyar:
@magento run WebAPI Tests

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -368,3 +372,22 @@ type Order {
...
```
prabhuram93: @rogyar using this userInputErrors:[CheckoutUserInputError], will have an impact on the reorder schema. I am not sure if this should be named with "Checkout" here. 
cc: @paliarush @nrkapoor

paliarush: This violates modularity because there is no dependency of Quote on Sales.

prabhuram93: @paliarush Exactly that was my concern too. I have updated the schema.

prabhuram93:
@magento run all tests

prabhuram93:
@magento run all tests

prabhuram93 (approved):

prabhuram93:
@magento run all tests

prabhuram93:
@magento run all tests

PR #27933, opened 2020-04-22
Update ColumnResolver.php

### LINKED ISSUE #27883: Duplicated keys for Weight
### Preconditions (*)
Resource Models use `const` to avoid typos and represent reusable element across the application.

https://github.com/magento/magento2/blob/eab4eba4809e866c55527fd2cbe63ab9b488a6b7/app/code/Magento/OfflineShipping/Model/ResourceModel/Carrier/Tablerate/CSV/ColumnResolver.php#L11-L28

You can notice that 2 `const`s are using the same value, which is developer's mistake.

The result of such mistake is fact, that just after - in `nameToPositionMap` - one position is overridden by another :facepalm: 
image

But when you scan the Module, you notice that there is actually field for Package Weight (destination):

https://github.com/magento/magento2/blob/95c176af90e7e9775cf01414f1fe3b2cdba1990b/app/code/Magento/OfflineShipping/Model/Carrier/Tablerate.php#L229-L240

### Steps to reproduce (*)

1.
2.

### Expected result (*)

1. [Screenshots, logs or description]
2.

### Actual result (*)

1. [Screenshots, logs or description]
2.

manishthummar111:
### Description (*)
COLUMN_WEIGHT and COLUMN_WEIGHT_DESTINATION column both are same and not need COLUMN_WEIGHT_DESTINATION so removed it and tested

### Related Pull Requests
Duplicated keys for Weight #27883

1. Fixes magento/magento2#27883 : Issue title

1. ...
2. ...

COLUMN_WEIGHT and COLUMN_WEIGHT_DESTINATION column both are same and not need COLUMN_WEIGHT_DESTINATION so removed it and tested

### DIFF
--- a/app/code/Magento/OfflineShipping/Model/ResourceModel/Carrier/Tablerate/CSV/ColumnResolver.php
+++ b/app/code/Magento/OfflineShipping/Model/ResourceModel/Carrier/Tablerate/CSV/ColumnResolver.php
@@ -12,7 +12,6 @@ class ColumnResolver
     const COLUMN_REGION = 'Region/State';
     const COLUMN_ZIP = 'Zip/Postal Code';
     const COLUMN_WEIGHT = 'Weight (and above)';
-    const COLUMN_WEIGHT_DESTINATION = 'Weight (and above)';
     const COLUMN_PRICE = 'Shipping Price';
 
     /**
@@ -23,7 +22,6 @@ class ColumnResolver
         self::COLUMN_REGION => 1,
         self::COLUMN_ZIP => 2,
         self::COLUMN_WEIGHT => 3,
-        self::COLUMN_WEIGHT_DESTINATION => 3,
         self::COLUMN_PRICE => 4,
     ];
 


orlangur:
@manishthummar111 `COLUMN_WEIGHT_DESTINATION` seems to be a different thing, what is it for?

rogyar:
I believe there was an idea of this column in 2016. But since it's not being used for 4+ years (the array index is overridden) we may remove it. The system behavior will not be touched anyhow by this change. 

@manishthummar111 could you look at the failing static tests, please?

orlangur:
@rogyar are we allowed to remove constants right away or should deprecate it first?

rogyar:
@orlangur that's a good point. @manishthummar111 please, mark the constant as deprecated instead of removing (you may still remove the corresponding array element)

VladimirZaets (changes requested):
Hi @manishthummar111, thanks for collaboration. Due to [Magento backward-compatible guide](http://devdocs.magento.com/guides/v2.2/contributor-guide/backward-compatible-development/) we can't remove constants. Firstly we should deprecate it. Please use `@deprecate` tag to deprecate one of these constant

engcom-Charlie:
Hi @manishthummar111, I'm closing this PR now due to inactivity.

PR #27938, opened 2020-04-22
catch Throwable instead of Exception

rvitaliy:
### Description (*)
Provides possibility to catch `\Throwable` instead of `\Exception`.
This Interface is introduced in php 7.0
This change allow us to catch errors like `\Error`, `\TypeError`, etc...

### Related Pull Requests
none

### Fixed Issues
none

generate an exception `\Error`  in any point of related code and verify that catch is executed


orlangur (changes requested):
Nice catch @rvitaliy!

Please enforce such change with a static test though.

orlangur:
@rvitaliy, hm, and after reading https://stackoverflow.com/questions/2274102/difference-between-using-throwable-and-exception-in-a-try-catch I don't think it's a good idea to catch Throwable in any case. Please provide some reasoning.

rvitaliy:
Yes, i agree, catch \Throwable instead of \Exception in any case is not a good idea. But in our case it is, at 99% of cases we try to catch something and log it as described on StackOverflow.
Probably is better to analize all this points of code to undestand if is better to catch \Throwable or \Exception.
Since it is very difficult to do this in a single task, I suggest to keep magento developers in mind to start using catch \Throwable instead of \Exception from now on and slowly migrate in a more targeted way.
Good luck!

orlangur:
@rvitaliy yeah, totally makes sense, thank you!

Hope we already have them in main critical points, like "API call failed due to Fatal Error".

PR #27964, opened 2020-04-24
Deprecate AbstractAction and it's public methods

lbajsarowicz:
### Description (*)
[Abstract Action and Action class should not be used](https://community.magento.com/t5/Magento-DevBlog/Decomposition-of-Magento-Controllers/ba-p/430883)

### Related Pull Requests
N/A

N/A

N/A

### Questions or comments
CC: @lenaorobei 

### DIFF
--- a/lib/internal/Magento/Framework/App/Action/AbstractAction.php
+++ b/lib/internal/Magento/Framework/App/Action/AbstractAction.php
@@ -1,7 +1,5 @@
 <?php
 /**
- * Abstract redirect/forward action class
- *
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
@@ -11,10 +9,12 @@ use Magento\Framework\App\RequestInterface;
 use Magento\Framework\App\ResponseInterface;
 
 /**
+ * Abstract redirect/forward action class
+ *
  * @deprecated Use \Magento\Framework\App\ActionInterface
  * @see https://community.magento.com/t5/Magento-DevBlog/Decomposition-of-Magento-Controllers/ba-p/430883
  */
-abstract class AbstractAction implements ActionInterface
+abstract class AbstractAction implements \Magento\Framework\App\ActionInterface
 {
     /**
      * @var \Magento\Framework\App\RequestInterface
--- a/lib/internal/Magento/Framework/App/Action/AbstractAction.php
+++ b/lib/internal/Magento/Framework/App/Action/AbstractAction.php
@@ -10,7 +10,11 @@ namespace Magento\Framework\App\Action;
 use Magento\Framework\App\RequestInterface;
 use Magento\Framework\App\ResponseInterface;
 
-abstract class AbstractAction implements \Magento\Framework\App\ActionInterface
+/**
+ * @deprecated Use \Magento\Framework\App\ActionInterface
+ * @see https://community.magento.com/t5/Magento-DevBlog/Decomposition-of-Magento-Controllers/ba-p/430883
+ */
+abstract class AbstractAction implements ActionInterface
 {
     /**
      * @var \Magento\Framework\App\RequestInterface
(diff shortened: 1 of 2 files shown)


lib/internal/Magento/Framework/App/Action/AbstractAction.php
```diff
@@ -7,28 +7,35 @@
  */
 namespace Magento\Framework\App\Action;
 
+use Magento\Framework\App\ActionInterface;
 use Magento\Framework\App\RequestInterface;
 use Magento\Framework\App\ResponseInterface;
+use Magento\Framework\Controller\Result\RedirectFactory;
+use Magento\Framework\Controller\ResultFactory;
```
orlangur: Please avoid such changes in future pull requests, it needs to be a separate effort rather than noise in regular diffs.

lbajsarowicz: I'm sorry about that.

lbajsarowicz: Removed noise for you.

lbajsarowicz:
@orlangur Reverted the "noise" and rebased to hide the traces.

lib/internal/Magento/Framework/App/Action/AbstractAction.php
```diff
@@ -56,6 +60,7 @@ public function __construct(
      * Retrieve request object
      *
      * @return \Magento\Framework\App\RequestInterface
+     * @deprecated This method should not be used anymore. Inject `RequestInterface` into constructor instead
```
lenaorobei: We do not deprecate methods, but classes.

lib/internal/Magento/Framework/App/Action/AbstractAction.php
```diff
@@ -10,7 +10,11 @@
 use Magento\Framework\App\RequestInterface;
 use Magento\Framework\App\ResponseInterface;
 
-abstract class AbstractAction implements \Magento\Framework\App\ActionInterface
+/**
+ * @deprecated Use \Magento\Framework\App\ActionInterface
+ * @see https://community.magento.com/t5/Magento-DevBlog/Decomposition-of-Magento-Controllers/ba-p/430883
```
lenaorobei: Please change to:
```
/**
 * @deprecated Inheritance was eliminated, composition should be used instead.
 * @see \Magento\Framework\App\ActionInterface
 */
```

`@see` tag will allow navigate class in IDE.

lib/internal/Magento/Framework/App/Action/Action.php
```diff
@@ -24,6 +24,7 @@
  * Action classes that do not extend from this class will lose this behavior and might not function correctly
  *
  * @deprecated Use \Magento\Framework\App\ActionInterface
+ * @see https://community.magento.com/t5/Magento-DevBlog/Decomposition-of-Magento-Controllers/ba-p/430883
```
lenaorobei: Please change to 
```
/**
 * @deprecated Inheritance was eliminated, composition should be used instead.
 * @see \Magento\Framework\App\ActionInterface
 */
```

ihor-sviziev: I would keep both links, just to have more info

lbajsarowicz: It's forbidden to add links to code

ihor-sviziev: @lenaorobei  @lbajsarowicz  Could you explain where it declared and why?
From my perspective add link that could describe more detailed why class is deprecated and what to use instead is really good option!

I even checked documentation for phpdoc - `@see` tag could contains URL to documentation https://docs.phpdoc.org/latest/references/phpdoc/tags/see.html

lenaorobei: @ihor-sviziev we can put links to official documentation only.

lenaorobei (approved):

lib/internal/Magento/Framework/App/Action/Action.php
ihor-sviziev: Got your point. Thx

PR #27972, opened 2020-04-24
ISSUE-27954: Forgot password save user only one column

### LINKED ISSUE #27954: Can't create/reset user password if user address is not valid
### Preconditions (*)

1. Magento 2.4-develop
2. Open `Create New Account Options` and enable `Require Emails Confirmation`
3. Create user with address (no special validation) via admin

### Steps to reproduce (*)

1. Add some kind of validation to address fields (limit length for city name)

> UPDATE customer_eav_attribute SET validate_rules='{"input_validation":1,"min_text_length":0,"max_text_length":2}'
    WHERE attribute_id=(SELECT attribute_id FROM eav_attribute WHERE entity_type_id=2 AND attribute_code='city');

OR for example you could create a plugin `beforeSave` on `\Magento\Customer\Api\AddressRepositoryInterface` that will add custom validation of address data.
2. Go to frontend
3. Click `Sign In`
4. Click `Forgot Your Password?`
5. Fill fields and click `Reset My Password`
6. Open email and click `Set a New Password`

### Expected result (*)

1. Forgot password functionality should allow the user to reset password
2. Address data shouldn't be used in reset password process

### Actual result (*)

1. You will be redirected to `Forgot Your Password?` page

Problem is in class `Magento\Customer\Model\ForgotPasswordToken\ConfirmCustomerByToken` in method `execute` which runs save on customer object.

We should disable address validation in this process or make an error message more clear for user.

sheepfy:
Fixes https://github.com/magento/magento2/issues/27954


aleron75:
Hello @sheepfy thanks for your contribution, it's very appreciated.

Due to [Magento Definition of Done](https://devdocs.magento.com/guides/v2.4/contributor-guide/contributing_dod.html), all code must be covered by tests. 

For this specific case, this fix should be covered by automated tests with the scenario which leads to an issue. 
Tests should fail on the mainline and pass with this fix.

To answer the question "which kind of tests should we write", I would go with a lightweight unit test.

ihor-sviziev (changes requested):
Hi @sheepfy,
Could you update your solution as it was requested here? https://github.com/magento/magento2/pull/27972#issuecomment-683285733

sheepfy:
> Hi @sheepfy,
> Could you update your solution as it was requested here? [#27972 (comment)](https://github.com/magento/magento2/pull/27972#issuecomment-683285733)


Hi, so, to be clear for me, you want me to implement unit tests for "ProductResource:: updateColumn" function ? 
If that so, i can do it untill the end of this week

engcom-Charlie:
Hi, @sheepfy could you please fix static and unit tests?
Thank you.

sheepfy:
Hi, i’ll do my best do finish the test untill the end of the week, if not , i’ll let you know

engcom-Charlie:
Hi, @sheepfy do you have any updates?
Thank you.

engcom-Charlie:
Hi, @sheepfy please fix static and unit tests. Otherwise, we can't proceed with your PR.
Thank you.

engcom-Charlie:
Hi, @sheepfy I'll continue with your PR.

engcom-Charlie:
@magento run all tests

app/code/Magento/Customer/Model/ResourceModel/Customer.php
```diff
@@ -403,4 +403,20 @@ public function changeResetPasswordLinkToken(\Magento\Customer\Model\Customer $c
         }
         return $this;
     }
+
+    /**
+     * @param int $customerId
+     * @param string $column
+     * @param string $value
+     */
+    public function updateColumn($customerId, $column, $value)
```
ihor-sviziev: For newly added methods we should add argument & return types, like this
```suggestion
    public function updateColumn(int $customerId, string $column, string $value)
```

app/code/Magento/Customer/Model/ResourceModel/Customer.php
```diff
@@ -403,4 +403,20 @@ public function changeResetPasswordLinkToken(\Magento\Customer\Model\Customer $c
         }
         return $this;
     }
+
+    /**
+     * @param int $customerId
+     * @param string $column
+     * @param string $value
+     */
```
ihor-sviziev:
```suggestion
    /**
     * Update one column in customer_entity table
     *
     * @param int $customerId
     * @param string $column
     * @param string $value
     * @return $this
     */
```

ihor-sviziev (changes requested):
Hi @sheepfy @engcom-Charlie,
I just reviewed this PR again and see that class where updateColumn method was adde - it is marked as `@api`, so adding there new methods is not allowed and that's why SVC check currently fails.

Also the customer model stored in EAV, so it will be more logical to make "updateAttribute" method that will save the only needed attribute, it doesn't matter if it stored in the main table `customer_entity` or in some of EAV tables.

I do understand that it will be huge refactoring , so instead of going to this direction - I could suggest two options:
1. move the `updateColumn` method to separate model `ClearAccountConfirmation`, similar to `ConfirmCustomerByToken`. It would just update confirmation field to null using resource model. The only one concern I have here - earlier `$customer->setConfirmation(null)` was executed and it basically changed the object, while now it's not changed.

2. In the account management model I see quite similar case:
https://github.com/magento/magento2/blob/68770fcefd93b8f2806fd60356efc318f91d2598/app/code/Magento/Customer/Model/AccountManagement.php#L579-L582
https://github.com/magento/magento2/blob/68770fcefd93b8f2806fd60356efc318f91d2598/app/code/Magento/Customer/Model/AccountManagement.php#L1617-L1626
Seems like this flag allow to skip the address validation process, and basically it will fix original issue.

What to you thing about those two suggestions? I'm ok with implementing both of them

engcom-Charlie:
@ihor-sviziev totally agree with you.
As for me, the second suggestion about the flag would be better.
@sheepfy what do you think?

engcom-Charlie:
@sheepfy I'll continue with your PR.

engcom-Charlie:
@magento run all tests

ihor-sviziev:
@engcom-Charlie thank you! now looks much better!

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-8388](https://jira.corp.magento.com/browse/ENGCOM-8388) has been created to process this Pull Request

magento-engcom-team:
@sheepfy thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

sheepfy:
Hello, thanks for inviting me into the Community Contributors team.
And sorry for the late reply and not continuing with the PR. Since first
time a created and till now, things changed and i didn't have time to go on.

engcom-Bravo:
This PR fixes an issue for Magento Enterprise Edition 

:heavy_check_mark: QA Passed

### Manual testing scenario:
1. Go to Store->Configuration->Customer->Customer Configuration
2. Enable 'Require Emails Confirmation'
3. On Admin go to Customers->All customers and Add New Customer with address
4. Add some kind of validation to address fields in the Database (limit length for city name)
`UPDATE customer_eav_attribute SET validate_rules='{"input_validation":1,"min_text_length":0,"max_text_length":2}'
WHERE attribute_id=(SELECT attribute_id FROM eav_attribute WHERE entity_type_id=2 AND attribute_code='city');`
5. Go to frontend
6. Click Sign In
7. Click Forgot Your Password?
8. Fill fields and click Reset My Password
9. Open the email and click Set a New Password

**Before** applying changes provided in this PR, the Customer has been redirected to _Forgot Your Password_
Before PR

**After** switching to the PR, the Customer is redirected to _Set a New Password_
After PR

PR #28127, opened 2020-05-06
Implement missing sortOrder for Plugins in PageCache module

lbajsarowicz:
### Description (*)
During work with #27270 I noticed that there is no way to inject plugin optimizing the Layout output, by default when there is no `sortOrder` for PageCache plugins - it's assumed to be `0`, so you need to use negative value for your plugin (eg. `-10`) optimizing output.

1. ...
2. ...

### Questions or comments
This change **is not backwards incompatible**, so that does not need additional architects' approval.
image

### DIFF
--- a/app/code/Magento/PageCache/etc/frontend/di.xml
+++ b/app/code/Magento/PageCache/etc/frontend/di.xml
@@ -7,22 +7,21 @@
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
     <type name="Magento\Framework\App\FrontControllerInterface">
-        <plugin name="front-controller-builtin-cache" type="Magento\PageCache\Model\App\FrontController\BuiltinPlugin"/>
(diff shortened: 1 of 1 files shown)


slavvka:
@magento run all tests

ihor-sviziev:
Looks like something was wrong during running the tests, just restarted

ihor-sviziev:
Hi @lbajsarowicz,
I think would be better to add these changes to https://github.com/magento/magento2/pull/27270 and remove negative sort order there. What  do you think?

lbajsarowicz:
@ihor-sviziev I'm afraid this is going to freeze the original PR either way.

ihor-sviziev:
Wow. more than 160 failed mftf tests. Looks like false positive. But to be sure let's re-run tests again 
@magento run Functional Tests B2B, Functional Tests CE, Functional Tests EE

ihor-sviziev (changes requested):
Hi @lbajsarowicz,
Please review failing functional tests. Looks like your change introduced quite big amount of test failures.

lbajsarowicz:
OK, let's give up with that for a moment, that is by-design the issue of Magento that has to be addressed across the platform.

PR #28147, opened 2020-05-08
Introduce separate BlockByIdentifier class to get Layout Block based on CMS Block Identifier

### LINKED ISSUE #28309: [Issue] Introduce separate BlockByIdentifier class to get Layout Block based on CMS Block Identifier
This issue is automatically created based on existing pull request: magento/magento2#28147: Introduce separate BlockByIdentifier class to get Layout Block based on CMS Block Identifier

---------
### Preconditions: (*)

The existing class is anti-pattern:
1. Uses `\Magento\Framework\Model\AbstractModel::load` to load Model
1. If `block_id` is identifier, the `getIdentities` method returns same identity whatever the Scope is (`return [\Magento\Cms\Model\Block::CACHE_TAG . '_' . $this->getBlockId()];`), which is painful when you use the same `identifier` for multiple per-store CMS blocks.

Why can't we fix the original Block? Some of developers might have used:
```xml
        <referenceContainer name="page.top">
            <block class="Magento\Cms\Block\Block" name="header.info" after="navigation.sections">
                <arguments>
                    <argument name="block_id" xsi:type="string">666</argument>
                </arguments>
            </block>
        </referenceContainer>
```

That is why we keep old Block backwards-compatible and introducing new one that can be used explicitly:
```xml
        <referenceContainer name="page.top">
            <block class="Magento\Cms\Block\BlockByIdentifier" name="header.info" after="navigation.sections">
                <arguments>
                    <argument name="identifier" xsi:type="string">header-info</argument>
                </arguments>
            </block>
        </referenceContainer>
```

### Expected Result (*)

1. ...
2. ...

lbajsarowicz:
### Description (*)

The existing class is anti-pattern:
1. Uses `\Magento\Framework\Model\AbstractModel::load` to load Model
1. If `block_id` is identifier, the `getIdentities` method returns same identity whatever the Scope is (`return [\Magento\Cms\Model\Block::CACHE_TAG . '_' . $this->getBlockId()];`), which is painful when you use the same `identifier` for multiple per-store CMS blocks.

Why can't we fix the original Block? Some of developers might have used:
```xml
        <referenceContainer name="page.top">
            <block class="Magento\Cms\Block\Block" name="header.info" after="navigation.sections">
                <arguments>
                    <argument name="block_id" xsi:type="string">666</argument>
                </arguments>
            </block>
        </referenceContainer>
```

That is why we keep old Block backwards-compatible and introducing new one that can be used explicitly:
```xml
        <referenceContainer name="page.top">
            <block class="Magento\Cms\Block\BlockByIdentifier" name="header.info" after="navigation.sections">
                <arguments>
                    <argument name="identifier" xsi:type="string">header-info</argument>
                </arguments>
            </block>
        </referenceContainer>
```

1. ...
2. ...

### Resolved issues:
1. [x] resolves magento/magento2#28309: Introduce separate BlockByIdentifier class to get Layout Block based on CMS Block Identifier


lbajsarowicz:
After further investigation, I realized that additional Identities are not necessary, as these are used only as tags, not as cache keys. The only thing that was missing before and is missing now is to tag block with CMS Block identities, so that when CMS Block is requested to be removed from cache, the Layout Block cache should be cleaned too.

I'm going to adjust that now.

ihor-sviziev (changes requested):
Hi @lbajsarowicz,
Could you fix unit tests failures ?

lbajsarowicz:
I can't find any failing report for Static Tests, neither for Unit Tests.

ihor-sviziev:
Hi @slavvka could you add more info why unit and static tests fails? We don’t have anything in reports

ihor-sviziev:
@magento run all tests

ihor-sviziev (approved):

magento-engcom-team:
[ENGCOM-7560](https://jira.corp.magento.com/browse/ENGCOM-7560) has been created to process this Pull Request

ihor-sviziev:
MHI test fails due to used protected method, that was overridden from parent class, there is no way to fix it. I would say this is false positive that shouldn't require approval.

slavvka:
@magento create issue

app/code/Magento/Cms/Block/BlockByIdentifier.php
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Loads the CMS block by `identifier` provided as an argument
+     *
+     * @return BlockInterface|BlockModel
+     * @throws NoSuchEntityException
+     */
+    private function getCmsBlock(): BlockInterface
+    {
+        if (!$this->getIdentifier()) {
+            throw new NoSuchEntityException(
```
Serializator: A `NoSuchEntityException` seems a bit unrelated to the actual cause of the exception, it wasn't that the entity was't found by the provided identifier, the identifier was simply not there.

Might there be a more suitable exception that could be thrown here?

app/code/Magento/Cms/Block/BlockByIdentifier.php
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * @inheritDoc
+     */
+    protected function _toHtml(): string
+    {
+        try {
+            return $this->filterOutput(
+                $this->getCmsBlock()->getContent()
+            );
+        } catch (NoSuchEntityException $e) {
```
Serializator, line 76: If there is a block in the layout using the `Magento\Cms\Block\BlockByIdentifier` class with an invalid / unknown identifier it'd probably raise some eyebrows if this exception isn't logged somewhere.

It will most likely get trampled on by all the other logs but at the same time it might be a valuable clue for someone in the future.

app/code/Magento/Cms/Block/BlockByIdentifier.php
lbajsarowicz: Nice catch, this should actually be `InvalidArgumentException`

app/code/Magento/Cms/Block/BlockByIdentifier.php
```diff
@@ -0,0 +1,170 @@
...
+            ->filter($content);
+    }
+
+    /**
+     * Loads the CMS block by `identifier` provided as an argument
+     *
+     * @return BlockInterface|BlockModel
+     * @throws NoSuchEntityException
+     */
+    private function getCmsBlock(): BlockInterface
```
Serializator, line 112: `private` prevents extensibility in a big way, rather use `protected` so that (for whatever reason) this class can be extended upon at least through a preference if really necessary. It's unfortunate that plugins can't be used on protected methods.

^ I won't comment on every private keyword in here (was thinking about it...) but I think you know what I'm grasping at!

app/code/Magento/Cms/Block/BlockByIdentifier.php
Serializator: I see what you did there! 😜

app/code/Magento/Cms/Block/BlockByIdentifier.php
lbajsarowicz: That was also actually something I was thinking about, as some of Administrators would like to have the block displayed conditionally - eg. for specific store, but don't have an error when the block by identifier is missing.

Also - logging a warning / notice for every single display of a page does not sound like good approach.

app/code/Magento/Cms/Block/BlockByIdentifier.php
lbajsarowicz: Using protected methods is discouraged and is going to fail Static code checks / Magento health check.

Also please consider reviewing these resources:
- https://devdocs.magento.com/guides/v2.3/ext-best-practices/extension-coding/common-programming-bp.html#composition-over-inheritance
- https://github.com/magento/magento2/issues/1019#issuecomment-72757989

So I am not going to change this into `protected` method as it's against what I consider as best practices.

app/code/Magento/Cms/Block/BlockByIdentifier.php
Serializator: I agree, it's a good point and didn't think about it that way 👍

app/code/Magento/Cms/Block/BlockByIdentifier.php
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if (null === $this->cmsBlock) {
+            $this->cmsBlock = $this->blockByIdentifier->execute(
+                (string)$this->getIdentifier(),
+                $this->getCurrentStoreId()
+            );
+        }
+
+        return $this->cmsBlock;
```
Serializator, line 131: What happens if the CMS block is disabled?

lbajsarowicz: That is actually interesting catch, as

https://github.com/magento/magento2/blob/c74256b9c637ac9808f8b0877d895b4a9b44f90f/app/code/Magento/Cms/Model/GetBlockByIdentifier.php#L43-L54

does not care about Block status. I'm going to adjust the logic.

ihor-sviziev:
@magento run all tests

lbajsarowicz:
@magento run Unit Tests

ihor-sviziev:
@magento run all tests

ihor-sviziev:
@slavvka could you explain why DB compare test fails?

ihor-sviziev (approved):
✔ Approved!

Failed functional test isn't related to changes from this PR, failed MHI looks like false positive.

lbajsarowicz:
@magento run all tests

I'd like to verify if there are no issues with current mainline.

engcom-Echo:
:heavy_check_mark: QA Passed
A simple block was created, using class = "Magento \ Cms \ Block \ BlockByIdentifier". getIdentities() return different result for different store. Before, the function returned the same result.

ihor-sviziev:
@engcom-Alfa @engcom-Echo is there any reasons why this PR isn't moving forward?

engcom-Foxtrot:
@magento create issue

engcom-Foxtrot:
Magento Health Index failure should be approved.

ihor-sviziev:
@engcom-Foxtrot MHI test failure is false positive! 
It fails on the method https://github.com/magento/magento2/pull/28147/files#diff-014e853ca19a6acdd6c6c6a51aa6fb7eR70
```php
protected function _toHtml(): string
``
which is overrides protected method from parent.

**Comment by sidolov** [comment]
Hi @ihor-sviziev! All failures should be approved, even if they are false-positive to make sure we will not introduce breaking changes. We will proceed with the approval ASAP

**Comment by lbajsarowicz** [comment]
@sidolov Ping.

**Comment by gabrieldagama** [comment]
Internal approval has been created and being processed (MC-37066).

**Comment by lbajsarowicz** [comment]
@gabrieldagama Where we are regarding that PR?

**Review by sidolov — APPROVED** [review]

**Comment by magento-engcom-team** [comment]
[ENGCOM-7560](https://jira.corp.magento.com/browse/ENGCOM-7560) has been created to process this Pull Request

File: app/code/Magento/Cms/Block/Block.php [hunk 5]
```diff
@@ -10,6 +10,8 @@
 
 /**
  * Cms block content block
+ * @deprecated This class introduces caching issues and should no longer be used
```
vovayatsyuk, line 13: What issues?

vovayatsyuk: Must be this one, that was fixed? https://github.com/magento/magento2/commit/5c84c1ad6aecbf6209d04bcaa160c49f356344d4

app/code/Magento/Cms/Block/BlockByIdentifier.php
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Returns the value of `identifier` injected in `<block>` definition
+     *
+     * @return string|null
+     */
+    private function getIdentifier(): ?string
+    {
+        return $this->getdata('identifier') ?: null;
```
vovayatsyuk: Must be getData

vovayatsyuk (changes requested):

app/code/Magento/Cms/Block/BlockByIdentifier.php
ihor-sviziev:
```suggestion
        return $this->getData('identifier') ?: null;
```

vovayatsyuk:
I don't understand why do Magento needs this new block. It seems that everything from this PR could be done in the existing block without creating a new entity and without BC.

Something like this:

```php
if (is_numeric($identifier)) {
  // loadById to be compatible with custom code that uses id.
} else {
 // new logic
}
```

app/code/Magento/Cms/Block/Block.php
lbajsarowicz: Not really. Except the fact the previous one uses both `id` and `identifier` for cache key, there was something more (I'll try to find out what was the other issue)

lbajsarowicz:
@vovayatsyuk If we want to follow SOLID principles - the arguments have to explicit, without type juggling. It's exactly the same like Page by Identifier interface.

vovayatsyuk:
But we already have a block to show cms blocks. I can only agree with this change if we'll remove another block in the future. But this leads to the next thought - why create a new block instead of refactoring the existing one and remove deprecated logic in the next major release?

lbajsarowicz:
@vovayatsyuk The only issue for introducing that Interface was the requirement for backwards compatibility, as this PR was created **before 2.4 was released**. If I knew that's going to hang for a year, I'd just modify the original implementation with happy "do it all" (against the SOLID principles) way.

ihor-sviziev:
@vovayatsyuk in my opinion - we can go with both options. in the current implementation of cms/block/block we have following issues:
1. not enforced to use CMS block ID or Identified, and both currently supported and widely used by extension developers
2. different cache tags in case if block ID or identified was used
3. deprecated method of loading the block (using $model->load())
4. In case if the block with such identifier / id is missing / disabled  / not assigned to current store view - we don't have any errors/warnings/etc.

Also as @lbajsarowicz said - addressing all of these issues will cause the not following SOLID principles and probably will be backward incompatible.
Technically this block isn't marked as `@api` and we might add backward incompatible changes, but if we'll change the logic - it definitely will cause issues during magento upgrade.

ihor-sviziev:
@magento run all tests

ihor-sviziev:
@sidolov is there any reason why this PR isn't getting merged?

sidolov:
@ihor-sviziev the pull request in the delivery queue right now and will be delivered soon. There are no reasons to postpone the merging process.

lbajsarowicz:
@sidolov , @ihor-sviziev @vovayatsyuk - Thanks for your support. Glad to have this merged.

PR #28192, opened 2020-05-11
Allows JSON unserializing to be associative or object

jajajaime:
### Description (*)

Allows a developer to pass an optional boolean argument when unserializing a JSON string, so it can be returned as object or associative. Keeps associative as default for backwards compatibility.

1. Create code that unserializes json string while passing the extra argument of false

### DIFF
--- a/lib/internal/Magento/Framework/Serialize/Serializer/Json.php
+++ b/lib/internal/Magento/Framework/Serialize/Serializer/Json.php
@@ -32,9 +32,9 @@ class Json implements SerializerInterface
      * @inheritDoc
      * @since 100.2.0
      */
-    public function unserialize($string)
+    public function unserialize($string, $assoc = true)
     {
-        $result = json_decode($string, true);
+        $result = json_decode($string, $assoc);
         if (json_last_error() !== JSON_ERROR_NONE) {
             throw new \InvalidArgumentException("Unable to unserialize value. Error: " . json_last_error_msg());
         }


ihor-sviziev:
I remember we had some architecture discussions in https://github.com/magento/architecture/pull/254, https://github.com/magento/magento-coding-standard/pull/91 and 
 previously in https://github.com/magento/architecture/pull/161.

AFAIK no real decision was done yet. @lenaorobei do we have any updates on that?

lib/internal/Magento/Framework/Serialize/Serializer/Json.php
```diff
@@ -32,9 +32,9 @@ public function serialize($data)
      * @inheritDoc
      * @since 100.2.0
      */
-    public function unserialize($string)
+    public function unserialize($string, $assoc = true)
```
orlangur, line 35: Such change is not allowed: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/

Let's first find out whether we need such functionality at all.

orlangur (changes requested):

ihor-sviziev:
@orlangur please explain why you moved this PR to "on hold" status.

lenaorobei:
Unfortunately, we cannot accept this PR since it violates backward compatibility policy.

There were some discussions regarding introduction new JSON contract, but there is no resolution yet. Please see details here https://github.com/magento/architecture/pull/254

jajajaime:
@lenaorobei is this because of "Adding parameters in public methods"? If so, I think it should be on a case by case basis, as creating a new class or interface would create more work for devs to switch to the new one in this case.

orlangur:
@jajajaime backward compatibility policy is mandatory, no exception.

> when unserializing a JSON string, so it can be returned as object or associative

Please create a proposal in architecture repo first to describe why it needs to be returned as object.

PR #28216, opened 2020-05-13
Replace hard-coded list of category attributes

### LINKED ISSUE #13440: Adding custom attribute to category doesn't show store specific value
Created a custom short_description attribute following the exact ways to create as shown in module-catalog/Setup/CategorySetup.php and module-catalog/view/adminhtml/ui_component/category_form.xml

### Preconditions
1. Magento 2.4-develop

### Steps to reproduce
1. Create new module with Setup InstallData.php
2. Install new category attribute following Magento's current category attribute array.
[magento-bug2.txt](https://github.com/magento/magento2/files/1681035/magento-bug2.txt)

3. Add view/adminhtml/ui_component/category_form.xml for above attribute
[magento-bug.txt](https://github.com/magento/magento2/files/1681033/magento-bug.txt)

4. I am also getting this for a custom text and image attribute.
5. php bin/magento setup:upgrade
6. Set different values per scope level using store switcher

### Expected result
1. Should see [Store View] label under attribute label.
2. Should be able to see different values per scope level (global, website, store)

### Actual result
1. Can't see [Store View] label under attribute label.
2. Can't see different values per scope level (global, website, store)
3. Can see different values in database

fredden:
### Description
`Magento\Catalog\Model\Category\DataProvider::getFieldsMap()` had a hard-coded list of category attributes. This pull request replaces this with a list generated from cache. The cache is built from reading `view/adminhtml/ui_component/category_form.xml` files, so this dynamically generated list includes all category attributes, not only those in the hard-coded list.

### Fixed Issues
1. Fixes #13440 - Adding custom attribute to category doesn't show store specific value

### Related Pull Requests
https://github.com/magento/partners-magento2ee/pull/267

### Manual testing scenarios
See issue #13440 for reproduction steps, expected result, and actual result.

### Questions or comments
- I have not yet updated any tests, however I have tested this locally.
- [`magento/module-catalog-staging`](https://github.com/magento/partners-magento2ee/blob/2.4-develop/app/code/Magento/CatalogStaging/Model/Category/DataProvider.php) will need to also be updated for Commerce as a similar hard-coded list exists there and that class extends this. - See https://github.com/magento/partners-magento2ee/pull/267


FredericMartinez:
It works in `2.3.5-p1`
Thanks @fredden

marvinhinz:
@magento run all tests

thampe:
This Looks good, I also tested it in `2.35-p1`.

marvinhinz:
@magento-engcom-team any idea why there are no build reports?

marvinhinz:
@magento run all tests

fredden:
@magento run all tests

fredden:
@magento run all tests

lenaorobei:
@magento run all tests

app/code/Magento/Catalog/Model/Category/DataProvider.php
```diff
@@ -233,6 +241,7 @@ public function __construct(
         $this->storeManager = $storeManager;
         $this->request = $request;
         $this->categoryFactory = $categoryFactory;
+        $this->uiConfigFactory = $uiConfigFactory;
```
lenaorobei: To keep backward compatibility let's add nullable argument.

app/code/Magento/Catalog/Model/Category/DataProvider.php
```diff
@@ -645,56 +654,43 @@ public function getDefaultMetaData($result)
...
+                    foreach ($childNode['children'] as $grandchildName => $grandchildNode) {
+                        $fieldsMap[$group][] = $grandchildName;
+                    }
+                } else {
+                    $fieldsMap[$group][] = $childName;
+                }
+            }
+
+            // Remove empty groups
+            if (empty($fieldsMap[$group])) {
```
lenaorobei: I believe we can remove this along with assignment in the line 674.

lenaorobei (changes requested):
Please see review comments. 

Looks like the fix should be applied to commerce repo. You can easily open commerce PR and link this this one through PR description.

app/code/Magento/Catalog/Model/Category/DataProvider.php
fredden: Technically we can remove this and the assignment on 674. I'd prefer to keep these however, as I feel like this method is more explicit in intent: showing clearly that empty groups are omitted on purpose.

The PHP manual also suggests that implicit array creation is discouraged. (Emphasis is mine.)

https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.modifying

> If $arr doesn't exist yet, it will be created, so this is also an alternative way to create an array. **This practice is however discouraged** because if $arr already contains some value (e.g. string from request variable) then this value will stay in the place and [] may actually stand for string access operator. **It is always better to initialize a variable by a direct assignment.**

app/code/Magento/Catalog/Model/Category/DataProvider.php
fredden: As far as I know, Magento's stance on extension is to prefer [composition over inheritance](https://devdocs.magento.com/guides/v2.4/ext-best-practices/extension-coding/common-programming-bp.html#composition-over-inheritance). If a third party has extended this model via a preference (which is documented as "[a method of last resort](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/build/di-xml-file.html#override-a-method-using-preference-nodes)"), then keeping their extension up to date with changes here is a necessary cost to using such a technique. I do not see any benefit in using a nullable argument here.

My objections aside, I do not know how to create this object if not via dependency injection. (I have no experience with the Object Manager as "[Magento prohibits the direct use of the `ObjectManager`](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/object-manager.html)" and any existing usages "[are not tacit endorsements of using the `ObjectManager` directly.](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/object-manager.html#usage-rules)" I understand this exists as an internal implementation detail of Magento's dependency injection system.)
How should I create this `virtualType` if not supplied via dependency injection?
https://github.com/magento/magento2/blob/8b7d949bf1cb67d7b98379fb30b6721ec7b92b37/app/code/Magento/Ui/etc/di.xml#L339-L343

lenaorobei: You are absolutely right about composition over inheritance principle. However, constructor modification is considered as a braking change. As a workaround, we recommend adding new optional parameter to the constructor at the end of the arguments list.

Please refer to the [Backward compatible development guide](https://devdocs.magento.com/contributor-guide/backward-compatible-development/) and section about `Adding a constructor parameter`.

app/code/Magento/Catalog/Model/Category/DataProvider.php
lenaorobei: In this case I'd prefer creating array variable to collect needed data and if it is not empty, write this data to `$fieldsMap[$group]`.

fredden:
Can I please get some help writing suitable tests for this change?

fredden:
@magento run all tests

fredden:
@magento run all tests

app/code/Magento/Catalog/Model/Category/DataProvider.php
```diff
@@ -645,56 +657,42 @@ public function getDefaultMetaData($result)
...
+                    // <container/> nodes need special handling
+                    foreach (array_keys($childNode['children']) as $grandchildName) {
+                        $fields[] = $grandchildName;
+                    }
+                } else {
+                    $fields[] = $childName;
+                }
+            }
+
+            if (count($fields)) {
```
lenaorobei:
```suggestion
            if (!empty($fields)) {
```

lenaorobei:
I recommend looking at `Magento\Catalog\Model\Category\DataProviderTest` as an example.

Please note also that Integration build failed and needs to be addressed https://public-results-storage-prod.magento-testing-service.engineering/reports/magento/magento2/pull/28216/24348333bfd24d15a02ce17ab5aabbc7/Integration/allure-report-ce/index.html#suites/1a47ad2d0b8dde3b8bc5178a8df7e3fb/8fa7f03a7c7e6e9c/

The same for Static build https://public-results-storage-prod.magento-testing-service.engineering/reports/magento/magento2/pull/28216/2370b22c2e6f31410ec42677e0b5a155/Statics/allure-report-b2b/index.html#suites/51d3ab4d5d77a0bc6a7ea344e9e7e6be/e670dd9849cb50d8/

engcom-Foxtrot:
@magento run all tests

engcom-Foxtrot:
@magento run all tests

engcom-Foxtrot:
@magento run all tests

lenaorobei (approved):

engcom-Alfa:
:heavy_check_mark: QA Passed

**Manual testing scenario:**

1. Create new module with Setup InstallData.php
2. Install new category attribute following Magento's current category attribute array.
[magento-bug2.txt](https://github.com/magento/magento2/files/1681035/magento-bug2.txt)
3. Add view/adminhtml/ui_component/category_form.xml for above attribute
[magento-bug.txt](https://github.com/magento/magento2/files/1681033/magento-bug.txt)
4. php bin/magento setup:upgrade
5. Set different values per scope level using store switcher

**Before:**
:heavy_multiplication_x:  Can't see [Store View] label under attribute label.
:heavy_multiplication_x:  Can't see different values per scope level (global, website, store)



**After:** 
:heavy_check_mark: [Store View] label under attribute label.
:heavy_check_mark: Able to see different values per scope level (global, website, store)


_On Storefront:_
Peek 2020-10-01 11-33

_On Admin panel:_ 
Peek 2020-10-01 12-08

engcom-Foxtrot:
@magento run all tests

engcom-Foxtrot:
@magento run all tests

Issue #28068, opened 2020-05-01
Configurable product not showing price on category and product page


hemendraajmera:
### Preconditions (*)

1. using Magento 2.3.5-p1
2. added new configurable product with two product

### Steps to reproduce (*)

1. after adding configurable products with two products and provide stock and status as in stock. on the category page, it showing price and on product detail page it showing option and price
![image](https://user-images.githubusercontent.com/2478037/80807318-7a58e900-8bda-11ea-934a-708f96cec5c3.png)
![image](https://user-images.githubusercontent.com/2478037/80807360-a5433d00-8bda-11ea-96dd-f01bdbef69a8.png)

2. chaining status to out of stock for it not showing product price on the category page and on product detail page also not showing price and option 
![image](https://user-images.githubusercontent.com/2478037/80807208-36fe7a80-8bda-11ea-9e17-286c9a602af9.png)
![image](https://user-images.githubusercontent.com/2478037/80807389-b3915900-8bda-11ea-98b6-2b47ed445dd6.png)

### Expected result (*)

1. show the price in whatever status of the product
![image](https://user-images.githubusercontent.com/2478037/80807318-7a58e900-8bda-11ea-934a-708f96cec5c3.png)
![image](https://user-images.githubusercontent.com/2478037/80807360-a5433d00-8bda-11ea-96dd-f01bdbef69a8.png)

### Actual result (*)

1. not showing price in out of stock 
![image](https://user-images.githubusercontent.com/2478037/80807208-36fe7a80-8bda-11ea-9e17-286c9a602af9.png)
![image](https://user-images.githubusercontent.com/2478037/80807389-b3915900-8bda-11ea-98b6-2b47ed445dd6.png)

shikhamis11:
hi @hemendraajmera 
I also believe that product price should display for out of stock configurable product just like simple products.

magento-engcom-team:
:white_check_mark: Confirmed by @shikhamis11 
 Thank you for verifying the issue. Based on the provided information internal tickets `MC-33935` were created

**Issue Available**: @shikhamis11, _You will be automatically unassigned. Contributors/Maintainers can claim this issue to continue. To reclaim and continue work, reassign the ticket to yourself._

hemendraajmera:
@magento give me 2.5-p1-develop instance

magento-engcom-team:
Hi @hemendraajmera. Thank you for your request. I'm working on Magento 2.5-p1-develop instance for you

[...]

PR #28252, opened 2020-05-18
Configurable product not showing price on category and product page #…

madhurajawat:
…28068

Configurable product not showing price on category and product page when all the associated products are out of stock.

### Description (*)

Configurable product not showing price on category and product page when all the associated products are out of stock.

1. magento/magento2#<issue_number>: Configurable product not showing price on category and product page #28068

1. Create a configurable product, all simple products are out of stock
2.  Go to -> detail page of the configurable product
3. The product price does not show

### DIFF
--- a/app/code/Magento/Catalog/Pricing/Render/FinalPriceBox.php
+++ b/app/code/Magento/Catalog/Pricing/Render/FinalPriceBox.php
@@ -64,26 +64,36 @@ class FinalPriceBox extends BasePriceBox
      */
     protected function _toHtml()
     {
-        if (!$this->salableResolver->isSalable($this->getSaleableItem())) {
-            return '';
-        }
-
         $result = parent::_toHtml();
-        //Renders MSRP in case it is enabled
-        if ($this->isMsrpPriceApplicable()) {
-            /** @var BasePriceBox $msrpBlock */
-            $msrpBlock = $this->rendererPool->createPriceRender(
-                MsrpPrice::PRICE_CODE,
-                $this->getSaleableItem(),
-                [
-                    'real_price_html' => $result,
-                    'zone' => $this->getZone(),
-                ]
-            );
-            $result = $msrpBlock->toHtml();
-        }
 
-        return $this->wrapResult($result);
+        if(!$result) {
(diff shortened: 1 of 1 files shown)


app/code/Magento/Catalog/Pricing/Render/FinalPriceBox.php
```diff
@@ -64,26 +64,36 @@ public function __construct(
...
-                    'real_price_html' => $result,
-                    'zone' => $this->getZone(),
-                ]
-            );
-            $result = $msrpBlock->toHtml();
-        }
 
-        return $this->wrapResult($result);
+        if(!$result) {
+            $result = BasePriceBox::_toHtml();
```
Serializator, line 70: `Magento\Catalog\Pricing\Render\FinalPriceBox` extends from `Magento\Framework\Pricing\Render\PriceBox` (`BasePriceBox`).

If I'm not mistaken, with the above in the mind the parent class is actually called twice, once through `parent::_toHtml` and once through `BasePriceBox::_toHtml`, is there a reason for this and thus on purpose or not?

app/code/Magento/Catalog/Pricing/Render/FinalPriceBox.php
```diff
@@ -64,26 +64,36 @@ public function __construct(
...
+            try {
+                /** @var MsrpPrice $msrpPriceType */
+                $msrpPriceType = $this->getSaleableItem()->getPriceInfo()->getPrice('msrp_price');
+            } catch (\InvalidArgumentException $e) {
+                $this->_logger->critical($e);
+                return $this->wrapResult($result);
```
Serializator, line 76: I'd refactor everything related to `$msrpPriceType` into the try-catch in which `$msrpPriceType` is defined, this way the return logic (`$this->wrapResult($result)`) is not duplicated.

I can understand if this was done to prevent too much indentation by using an early return. In this case I'd argue that it might be better to refactor it into a separate method responsible for returning either the HTML for the MSRP price rendering or an empty string in case of the exception for example, minimising the overal indentation.

pradeepkanaka:
Is this a valid fix?

PR #28284, opened 2020-05-19
Cannot Create Credit Memos with ZeroGrandTotal via Order/Invoice Service

amenk:
When creating credit memos via the orderService, allowZeroGrandTotal is ignored

Steps to reproduce:

1. Have 0 Subtotal order
2. try to cancel it via Services, for example:

        \Magento\Sales\Api\RefundOrderInterface $orderRefunder,
        \Magento\Sales\Api\RefundInvoiceInterface $invoiceRefunder

                $invoiceRefunder->execute(
                    $invoice->getId(),
                    [],
                    false,
                    true
                );

   or

       $this->orderRefunder->execute(
                $id,
                [],
                true
            );

Leads to errors like:

      The credit memo's total must be positive.

See also https://github.com/magento/magento2/pull/16005

### DIFF
--- a/app/code/Magento/Sales/Model/Order/Creditmemo/Validation/QuantityValidator.php
+++ b/app/code/Magento/Sales/Model/Order/Creditmemo/Validation/QuantityValidator.php
@@ -94,7 +94,7 @@ class QuantityValidator implements ValidatorInterface
             }
         }
 
-        if ($entity->getGrandTotal() <= 0) {
+        if (!$entity->isValidGrandTotal()) {
             $messages[] = __('The credit memo\'s total must be positive.');
         } elseif ($totalQuantity <= 0 && !$this->canRefundShipping($order)) {
             $messages[] = __('You can\'t create a creditmemo without products.');


amenk:
Still it is not possible to use

            // orderRefunder =  \Magento\Sales\Api\RefundOrderInterface
            $this->orderRefunder->execute(
                $id,
                [],
                true
            );


Because of \Magento\Sales\Model\Order\Validation\CanRefund::isTotalPaidEnoughForRefund

amenk:
There is a lot of logic duplication between CreditMemos and Order, not sure how to refactor this properly, help appreciated

amenk:
We probably have a workaround: Instead of creating 0-value credit memos we can use $order->cancel(), still it is strange that via backend we can create a 0-value credit memo but not via the service

edenduong (changes requested):
Hi @amenk ,

Thank you for your contribution. I see the logic is correct. Could you please cover your changes with Unit Test ? Thank you very much.

amenk:
This seems to need some more work. Would be nice if someone else could jump in

I will not follow this bug anymore because we decided for another approach without invoices and credit memos for 0 totals

edenduong:
@amenk: Thank you. I will close this Pull Request. If someone continue to work, he/she can create new Pull request.

amenk:
Does anybody know if this problem still appears?

PR #28293, opened 2020-05-20
Allow to get currect redirect url of Result/Redirect

### LINKED ISSUE #28954: [Issue] Allow to get currect redirect url of Result/Redirect
This issue is automatically created based on existing pull request: magento/magento2#28293: Allow to get currect redirect url of Result/Redirect

---------
### Summary (*)

During my development I created some plugins to change the redirects of the Controllers:

- `Magento\Customer\Controller\Account\EditPost`
- `Magento\Customer\Controller\Address\FormPost`
- `Magento\Newsletter\Controller\Manage`

The redirects should be changed only in some cases. In an Around- or After-Plugin it was impossible to get the defined redirect url because `\Magento\Framework\Controller\Result\Redirect::$url` is protected and there is no public getter.

### Examples (*)

1. Create a custom module `MyVendor\MyModule`
2. Add di.xml
3. Write a plugin preference f.e.
```
    <type name="Magento\Customer\Controller\Account\EditPost">
        <plugin
            name="CustomEditPost"
            type="MyVendor\MyModule\Plugin\Controller\Customer\Account\EditPostPlugin" />
    </type>
```
4. In the Plugin `MyVendor\MyModule\Plugin\Controller\Customer\Account\EditPostPlugin` it should be possible now to do:
```
    public function aroundExecute(EditPost $subject, \Closure $proceed)
    {
        $resultRedirect = $proceed();
        $url = $resultRedirect->getUrl();
```

borriglione:
### Description

During my development I created some plugins to change the redirects of the Controllers:

- `Magento\Customer\Controller\Account\EditPost`
- `Magento\Customer\Controller\Address\FormPost`
- `Magento\Newsletter\Controller\Manage`

The redirects should be changed only in some cases. In an Around- or After-Plugin it was impossible to get the defined redirect url because `\Magento\Framework\Controller\Result\Redirect::$url` is protected and there is no public getter.

1. Create a custom module `MyVendor\MyModule`
2. Add di.xml
3. Write a plugin preference f.e.
```
    <type name="Magento\Customer\Controller\Account\EditPost">
        <plugin
            name="CustomEditPost"
            type="MyVendor\MyModule\Plugin\Controller\Customer\Account\EditPostPlugin" />
    </type>
```
4. In the Plugin `MyVendor\MyModule\Plugin\Controller\Customer\Account\EditPostPlugin` it should be possible now to do:
```
    public function aroundExecute(EditPost $subject, \Closure $proceed)
    {
        $resultRedirect = $proceed();
        $url = $resultRedirect->getUrl();
```

### Resolved issues:
1. [x] resolves magento/magento2#28954: Allow to get currect redirect url of Result/Redirect


rogyar:
Hi, @borriglione. Thank you for your collaboration. That's an interesting use case but it makes sense.
In some other cases, I would recommend invoking the `::render` method and get the location from response headers. However, this operation has side effects for the redirect itself. So, not a sort of recommended approach either.

lib/internal/Magento/Framework/Controller/Result/Redirect.php
```diff
@@ -83,6 +83,15 @@ public function setUrl($url)
         return $this;
     }
 
+    /**
+     * URL Getter
+     * @return string
```
rogyar:
```suggestion
     * @return string|null
```

rogyar (changes requested):
According to the [definition of done](https://devdocs.magento.com/guides/v2.3/contributor-guide/contributing_dod.html) all changes need to be covered by automated tests. 
Could I kindly ask you to extend the [existing unit test](https://github.com/magento/magento2/blob/05c9e0ea61a2c218a363334a3fef7dac34c38093/lib/internal/Magento/Framework/Controller/Test/Unit/Result/RedirectTest.php#L15) for checking the getter. 

Thank you

engcom-Echo:
@magento run all tests

engcom-Echo:
@magento run all tests

rogyar:
@magento run all tests

rogyar (dismissed):
I believe the failing tests are not related to the current PR. The same happens for other PRs

engcom-Delta:
:heavy_check_mark: QA passed
Was checked case from [Manual testing scenarios](https://github.com/magento/magento2/pull/28293#issue-420587913) . 
#28293PR

engcom-Delta:
@mageno create issue

engcom-Delta:
@magento create issue

engcom-Delta:
@magento create issue

sidolov:
@magento run all tests

lib/internal/Magento/Framework/Controller/Result/Redirect.php
```diff
@@ -83,6 +84,16 @@ public function setUrl($url)
         return $this;
     }
 
+    /**
+     * URL Getter
+     *
+     * @return string|null
+     */
+    public function getUrl()
```
sidolov, line 92: Unfortunately, adding new public methods to classes is backward incompatible according to our [Backward Compatible Development Guide](https://devdocs.magento.com/guides/v2.4/contributor-guide/backward-compatible-development/)

sidolov (changes requested):

sidolov:
@borriglione  I am closing this PR now due to inactivity.

lib/internal/Magento/Framework/Controller/Result/Redirect.php
borriglione: @sidolov 
Ok, but what would be the best/correct approach for the described use case: https://github.com/magento/magento2/pull/28293#issue-420587913 ?

schmengler: @sidolov Please point out where in the BC development guide this is stated. I don't find it.

schmengler:
Thank you, kind bot, but what now? I just commented and don't even have rights to reopen the PR :grinning:

lib/internal/Magento/Framework/Controller/Result/Redirect.php
sidolov: @schmengler you may find the information in the section: **Introduction of a method to a class or interface**

of https://devdocs.magento.com/contributor-guide/backward-compatible-development/

schmengler: But in this case it isn't productive to introduce a new class. The `setUrl()` method isn't even part of an interface, neither the newly introduced `getUrl()`.

Let's think about the reasoning behind the rule: Adding a method to an interface breaks alternative implementations. Adding a method to a class that does not have an interface doesn't because if one would have replaced the implementation, it needs to extend the original class and will inherit the new methods

adamlavery:
@schmengler I've got to agree with you, this process is nonsense. The second PR on this has been mishandled by bots too. What does it take to get these things implemented? @sidolov put the spanner in the works but had nothing useful to contribute on how to get this very simple request, for something that should have been there from the start, addressed.

There are plenty of use-cases for needing to know where we're going next.

@sidolov, as you put a halt to this, can you suggest how else this can be solved? As @borriglione pointed out, creating a new class does not solve the issue because that would mean refactoring Magento core to pass the new class around - no? Let's apply some common sense to this.

